Issue #6, opened 2020-07-12
Implement layout algorithm


GabrielMajeri:
Once the layout is parsed into memory, we have to compute the offsets and sizes (width/height) of the objects to be rendered.

## Layout algorithm

To calculate dimensions (width/height):
- elements with **fixed** sizes (e.g. `123px`) are laid out, in the order they are encountered.
- then we handle elements with `inner`, which indicates that the size of the component depends on the size of the child.
- then **percentages** (such as `10%`) are computed, based on the total size of the parent container.
- then `@` rules (e.g. `100@`) are applied, which divide up the remaining free space.

When the `width`/`height` attributes are not specified, they each default to `inner`.

To calculate positions (x/y):
- this only makes sense when there is some free space left (otherwise, all the elements fit tightly together and are positioned one after another).
- `x-align`/`y-align` take a percentage, which indicates where it should be along the respective axis.
  `0%` would mean completely to the left/to the top, `100%` would mean completely to the right/to the bottom, and `50%` would be halfway between.

If there's not enough space in the parent container to lay out all children, we need to handle the overflow (for instance, through _scrolling_).

Keavon:
This card refers to this issue: https://github.com/Keavon/Graphite/projects/1#card-41708195

Keavon:
Shelved on account of switching to a web app frontend for the short term.

PR #8, opened 2020-07-13
Start implementing layout engine

GabrielMajeri:
Starts work towards #6.

I've opened this to get some feedback.

Currently, it only computes sizes (not offsets), and it only works for the top-level window.

### DIFF
--- a/gui/README.md
+++ b/gui/README.md
@@ -15,6 +15,23 @@ The layout engine does a top-down pass through the component tree in order to de
 
 Layout is controlled using predefined attributes, such as `width`, `height`, `x-align`, `y-align`, `gap` or `padding`.
 
+### Layout algorithm
+
+To calculate dimensions (width/height):
+- elements with **fixed** sizes (e.g. `123px`) are laid out, in the order they are encountered.
+- then we handle elements with `inner`, which indicates that the size of the component depends on the size of the child. Their sizes have to be recursively computed. Afterwards we add their padding and spacing.
+- then we handle elements with **percentage** sizes (such as `10%`), based on the computed total size of the parent container.
+- then `@` rules (e.g. `100@`) are applied, which divide up the remaining free space in the container.
+
+When the `width`/`height` attributes are not specified, they each default to `inner`.
+
+If there's not enough space in the parent container to lay out all children, the container overflows (this can be handled, for example, through _scrolling_).
+
+To calculate positions (x/y):
+- this only makes sense when there is some free space left (otherwise, all the elements fit tightly together and are positioned one after another).
+- `x-align`/`y-align` take a percentage, which indicates where it should be along the respective axis.
+  `0%` would mean completely to the left/to the top, `100%` would mean completely to the right/to the bottom, and `50%` would be halfway between.
+
 ## Component lifetime
 
 The children of a component are passed to it as a `content` attribute. For example, looking at the row component:
--- a/src/layout_abstract_syntax.rs
+++ b/src/layout_abstract_syntax.rs
@@ -63,6 +63,16 @@ impl LayoutComponentNode {
 		Self::Text(text)
 	}
 
+	pub fn layout_attributes(&self) -> LayoutAttributes {
+		match self {
+			Self::Tag(tag) => tag.layout_arguments.clone(),
+			Self::Text(text) => LayoutAttributes {
+				width: Dimension::AbsolutePx((text.len() as f64) * 10.0),
+				..Default::default()
+			},
+		}
+	}
+
 	/// Print the component node (for debugging)
 	#[allow(dead_code)]
 	pub fn debug_print(&self) {
@@ -252,6 +262,27 @@ impl AttributeArg {
 
 // ====================================================================================================
 
+/// An axis along which layout is performed
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+pub enum Axis {
+	/// Horizontal axis, which is used to lay out the children of a row.
+	Horizontal,
+	/// Vertical axis, which is used to lay out the children of a column.
+	Vertical,
+}
+
+impl Axis {
+	/// Returns the axis perpendicular to this one.
+	pub fn perpendicular(&self) -> Axis {
+		match self {
+			Axis::Horizontal => Axis::Vertical,
+			Axis::Vertical => Axis::Horizontal,
+		}
+	}
+}
+
+// ====================================================================================================
+
 /// Attributes used by the layout engine to calculate sizing and placement
 #[derive(Clone, Debug, PartialEq)]
 pub struct LayoutAttributes {
@@ -261,6 +292,16 @@ pub struct LayoutAttributes {
 	pub y_align: f64,
 	pub gap: BoxDimensions,
 	pub padding: BoxDimensions,
+}
+
+impl LayoutAttributes {
+	/// Retrieves the size of this element along the specified axis.
+	pub fn size_along(&self, axis: Axis) -> Dimension {
+		match axis {
+			Axis::Horizontal => self.width,
+			Axis::Vertical => self.height,
+		}
+	}
 }
 
 impl Default for LayoutAttributes {
--- a/src/layout_engine.rs
+++ b/src/layout_engine.rs
@@ -1,0 +1,135 @@
+//! The layout engine converts the layout tree to rectangles of fixed sizes on the screen,
+//! in order to know how to draw the various components.
+
+use crate::layout_abstract_syntax::*;
+use crate::layout_abstract_types::*;
+
+/// Unit for offsets and sizes, expressed in pixels.
+pub struct Pixel;
+
+/// A 2D rectangular extent of pixels.
+pub type Extent = euclid::Size2D<f64, Pixel>;
+
+/// Constructs a new extent from values along an axis.
+fn extent_from_axis_values(main_axis_size: f64, cross_axis_size: f64, axis: Axis) -> Extent {
+	match axis {
+		Axis::Horizontal => Extent::new(main_axis_size, cross_axis_size),
+		Axis::Vertical => Extent::new(cross_axis_size, main_axis_size),
+	}
+}
+
+/// Returns the length of the rectangle's sides along the main and the cross axis.
+fn extent_sizes_along(extent: Extent, main_axis: Axis) -> (f64, f64) {
+	match main_axis {
+		Axis::Horizontal => (extent.width, extent.height),
+		Axis::Vertical => (extent.height, extent.width),
+	}
+}
+
+// ====================================================================================================
+
+/// Recurisvely computes the layout of a node.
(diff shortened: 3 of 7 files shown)


src/layout_abstract_syntax.rs
```diff
@@ -35,6 +35,16 @@ impl LayoutAbstractTag {
...
+
+	/// Retrieves the `width` attribute of this node, if any.
+	pub fn width(&self) -> Option<&AttributeValue> {
+		self.attributes.iter().find(|attribute| attribute.name == "width").map(|attribute| &attribute.value)
+	}
+
+	/// Retrieves the `height` attribute of this node, if any.
+	pub fn height(&self) -> Option<&AttributeValue> {
+		self.attributes.iter().find(|attribute| attribute.name == "height").map(|attributes| &attributes.value)
+	}
```
GabrielMajeri, line 47: I've created these helper functions, but it makes sense to eventually just store the layout attributes as a dedicated structure in the node (because these are some built-in attributes).

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+			Axis::Vertical => Axis::Horizontal,
+		}
+	}
+}
+/// Built-in component attributes used during layout.
+#[derive(Debug)]
+struct LayoutAttributes {
+	pub width: Dimension,
+	pub height: Dimension,
+}
```
GabrielMajeri, line 30: Eventually this will also store alignments and padding and such.

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+
+fn compute_linear_layout(node: &LayoutNode, tag: &LayoutAbstractTag) {
+	// Extract the layout-specific attributes of this node
+	let mut attribs = node.borrow().layout_attributes();
+
+	// For the sake of example, we'll assume the main window has a fixed size
+	if tag.namespace == "window" && tag.name == "main" {
+		attribs.width = Dimension::AbsolutePx(1920.0);
+		attribs.height = Dimension::AbsolutePx(1080.0);
+	}
```
GabrielMajeri, line 137: This is also temporary. The `main.xml` file said the window would be called with a specific width and height, so for now I've taken some example values.

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+
+	// At the end, we lay out the elements which use the `@` specifier.
+	// They split up the remaining free space.
+	let free_space = {
+		if let Dimension::AbsolutePx(size) = attribs.size_along(main_axis) {
+			size - min_main_axis_size
+		}
+		else {
+			panic!("child layout uses free space percentage but parent doesn't have a well-defined size");
+		}
```
GabrielMajeri, line 191: This is a problematic scenario. What if the child wants to be 100@ of the free space of the parent, but the parent has an unbound 100% size?

The solution would be to always make sure the root of the hierarchy has a fixed size (for example, the window).

gui/README.md
```diff
@@ -15,6 +15,23 @@ The layout engine does a top-down pass through the component tree in order to de
 
 Layout is controlled using predefined attributes, such as `width`, `height`, `x-align`, `y-align`, `spacing` or `padding`.
 
+### Layout algorithm
+
+To calculate dimensions (width/height):
+- elements with **fixed** sizes (e.g. `123px`) are laid out, in the order they are encountered.
+- then **percentages** (such as `10%`) are computed, based on the total size of the parent container.
+- then we handle elements with `inner`, which indicates that the size of the component depends on the size of the child.
+- then `@` rules (e.g. `100@`) are applied, which divide up the remaining free space.
```
Keavon, line 24: `inner` elements will actually need to be calculated before `%` elements.

Think of `px` and `inner` as elements with outwards-pushing sizes. Their demanded size is set within (either the literal `px` value or by recursing into the `inner` element and solving for a pixel value). Then padding and spacing are added outside of them (padding grows the actual size, spacing just pushes the x/y alignment within its container and takes up remaining space).

After these two are calculated (`px` then `inner`), the container has been pushed to expand by these types of children. Its size is now known, so it can go push inwards on `%` and `@` elements.

Think of `%` and `@` as elements with sizes that are pushed inwards by their container. The container's size is now known, and these two types of elements have their dimensions based on that known size. `%` elements are calculated and that means all the dimensions are known except `@` elements. Add up all the known sizes and subtract that sum from the container's dimension to get the remaining unfilled pixels. Now the `@` elements are finally given a size equal to their percentage of the remainder. (`50@` = 50% of the remainder, so a `100px` remainder becomes an absolute `50px` size for that element.)

So in summary, the order goes like this:
1. Immediately count all the `px` values, these are done.
2. Recursively calculate `inner` sizes to get their pixel values, now all outwards-pushing elements are done.
3. If the container was `inner`, it now has a known size: `max(outwardsPusher + itsPadding + itsSpacing, ...)` pixels
4. Now `%` values are `percentage / 100 * containerSize` pixels
5. Finally `@` values are `atSignPercentage / 100 * (containerSize - (knownElementSizes + theirPadding + theirSpacing))` pixels

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
+//! The layout engine converts the layout tree to rectangles of fixed sizes on the screen,
+//! in order to know how to draw the various components.
+
+use crate::layout_abstract_syntax::*;
+use crate::layout_abstract_types::*;
+
+/// An axis along which layout is performed
+#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+enum Axis {
+	/// Horizontal axis, which is normally used by rows.
```
Keavon, line 10: > normally used **within** rows

(rows themselves are stacked on top of one another if there are multiple sibling rows, and sibling `<row>`s and `<column>`s are an error by the way)

Not trying to nitpick but this is a distinction that could lead to confusion (I had to think about it for a few minutes).

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+	Width,
+	/// Size relativ to the height of this component.
+	Height,
+}
+
+impl Dimension {
+	/// Converts an `AttributeValue` to a `Dimension`.
+	fn from_attribute_value(value: &AttributeValue) -> Self {
+		if let AttributeValue::TypeValue(values) = value {
+			assert_eq!(values.len(), 1, "Unexpected number of values");
```
Keavon, line 65: It might be more useful for this function to return a vector of `Dimension`s. It would make sense to treat everything as multi-value attributes (for example `padding="1px 2px 3px 4px"`), and single-value attributes are multi-value attributes of length 1.

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+				// For now, we assume the text only takes up one row,
+				// and we have a monospaced font where each character is 20px wide.
+				let width = (content.len() * 20) as f32;
+				let height = 20.0;
+
+				let width = Dimension::AbsolutePx(width);
+				let height = Dimension::AbsolutePx(height);
+
+				LayoutAttributes { width, height }
+			},
```
Keavon, line 110: We won't even need to bother with this branch, text nodes aren't handled by the layout engine. They exist so you can pass text to a particular node's `content` attribute. The `<text>` primitive node will be in charge of rendering text and since it is `inner` by default, it will usually report its width when the text renderer reports that. The text (if you bind it, like `content="SOME_VAR"`) can also be used in the JavaScript(?) scripts associated with each layout component. For example, the `<dropdown>` element will need a list of selectable dropdown items so the JS will be `SOME_VAR.split("\n").map(s => s.trim())`.

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+				let dimension = match *value {
+					TypeValue::AbsolutePx(pixels) => Dimension::AbsolutePx(pixels),
+					TypeValue::Percent(percent) => Dimension::Percent(percent),
+					TypeValue::PercentRemainder(percent) => Dimension::PercentRemainder(percent),
+					TypeValue::Inner => Dimension::Inner,
+					TypeValue::Width => Dimension::Width,
+					TypeValue::Height => Dimension::Height,
+					_ => panic!("unsupported type in layout attribute"),
+				};
+				dimension
```
Keavon, line 77: It might make sense to break this out into its own conversion function since it will be useful to call it in other places.

src/layout_engine.rs
```diff
@@ -0,0 +1,206 @@
...
+	// Extract the layout-specific attributes of this node
+	let mut attribs = node.borrow().layout_attributes();
+
+	// For the sake of example, we'll assume the main window has a fixed size
+	if tag.namespace == "window" && tag.name == "main" {
+		attribs.width = Dimension::AbsolutePx(1920.0);
+		attribs.height = Dimension::AbsolutePx(1080.0);
+	}
+
+	let main_axis = if tag.namespace == "" && tag.name == "row" { Axis::Horizontal } else { Axis::Vertical };
```
Keavon, line 139: Okay, let's default to `Horizontal` (so things flow like they are inside a `<row>` if neither `<row>` nor `<col>` is specified. (`window/main.xml` is now being wrapped in `<col>`.)

src/layout_engine.rs
Keavon: The window will indeed be always called with absolute pixel values, like shown on line 1 of `window/main.xml`, equal to the dimensions of the window frame.

GabrielMajeri:
I've updated the PR. It is now able to recurse through the layout for `window:main` without crashing.

It seems to compute main axis sizes (i.e. heights) correctly. The cross axis sizes aren't really implemented yet.

src/layout_engine.rs
GabrielMajeri: Makes sense, horizontal is also the default for the web.

src/layout_engine.rs
GabrielMajeri: I've let it return width/height 100% for now so it doesn't crash my WIP cross-axis calculation code.

src/layout_engine.rs
GabrielMajeri: I'll leave that as a to-do for the future

src/layout_engine.rs
GabrielMajeri: I've reworded it, hopefully it makes more sense now.

gui/README.md
GabrielMajeri: Thanks for the explanation, I've updated the `README.md` for GUI with a summary.

GabrielMajeri:
I've updated this earlier, but it seems GitHub didn't display the changes :thinking: 

The changes are as follows:
- rebased on top of the latest `master`, fixing what changed.
- switched to using the `euclid` crate for 2D geometry. This is the same one used by the Servo browser's layout engine, so it's right on target.

GabrielMajeri:
I've rebased the PR, and incorporated the latest changes from `master`. The code now compiles and runs, and produces mostly the same output as before.

The code is still messy in my opinion and I'll look into breaking it up into multiple PRs or refactors over the following days.

Keavon:
Shelved on account of switching to a web app frontend for the short term.

PR #10, opened 2020-07-15
Separately store layout attributes

### LINKED ISSUE #9: Extract layout attributes into a separate structure
Right now, the layout attributes (such as `width` or `padding`) are stored together with other, user-defined attributes in the `attributes` vector:
https://github.com/Keavon/Graphite/blob/c41710ab7696799d6d81974eee3768d01c58c60c/src/layout_abstract_syntax.rs#L23
To store these more efficiently, and to simplify the downstream code, the parser should extract these attributes and store them into a Rust structure. Something similar to this:
https://github.com/Keavon/Graphite/blob/c41710ab7696799d6d81974eee3768d01c58c60c/src/layout_dom_node.rs#L8-L12
But instead of `f32`s, width and height would store a `Dimension`.

If the parser notices a node has a value given for a layout attribute, for example `width = "10px"`, it will store it as is (`Dimension::AbsolutePx(10.0)`). Otherwise it will use the default value (in this example, `Dimension::Inner` for unspecified width).

GabrielMajeri:
This PR is a smaller, independent part broken off of #8.

It extracts the attributes used by the layout system into a separate structure, to make it easier to access them.

If a node does not specify a certain value for one of its layout attributes, the default is used. So, for example, nodes with no alignment attribute will be, by default, left and top aligned.

Fixes #9.


src/layout_abstract_syntax.rs
```diff
@@ -28,12 +29,27 @@ impl LayoutAbstractTag {
...
+			"width" => self.layout_attributes.width = attribute.dimension(),
+			"height" => self.layout_attributes.height = attribute.dimension(),
+			"x-align" => self.layout_attributes.x_align = attribute.percent(),
+			"y-align" => self.layout_attributes.y_align = attribute.percent(),
+			"x-padding" => self.layout_attributes.padding.set_horizontal(attribute.dimension()),
+			"y-padding" => self.layout_attributes.padding.set_vertical(attribute.dimension()),
+			"padding" => self.layout_attributes.padding = attribute.box_dimensions(),
+			"x-spacing" => self.layout_attributes.spacing.set_horizontal(attribute.dimension()),
+			"y-spacing" => self.layout_attributes.spacing.set_vertical(attribute.dimension()),
+			"spacing" => self.layout_attributes.spacing = attribute.box_dimensions(),
```
Keavon, line 49: Can we move this match to another function and call that function when there is no `:` prefix? That will be cleaner when it comes time to add `on:`-prefixed event handler attributes (and any other future attribute categories).

Keavon: Are also might want to add the `scroll` attribute for later implementation.

src/layout_abstract_syntax.rs
```diff
@@ -20,6 +20,7 @@ impl LayoutAbstractNode {
 pub struct LayoutAbstractTag {
 	pub namespace: String,
 	pub name: String,
+	pub layout_attributes: LayoutAttributes,
 	pub attributes: Vec<Attribute>,
```
Keavon, line 24: Let's rename this `user_attributes`.

src/layout_abstract_syntax.rs
GabrielMajeri: Good idea, I've refactored that function.

src/layout_abstract_syntax.rs
GabrielMajeri: Did you mean the field below this one?

src/layout_abstract_syntax.rs
```diff
@@ -20,6 +20,11 @@ impl LayoutAbstractNode {
 pub struct LayoutAbstractTag {
 	pub namespace: String,
 	pub name: String,
+	/// Layout attributes, which are used by the layout engine.
+	pub layout_attributes: LayoutAttributes,
+	/// The special content attribute, representing the inner elements of this tag.
+	pub content: Option<AttributeValue>,
```
GabrielMajeri, line 26: I've added a `content` field where we can store the node content. I think it's better this way, instead of searching for it linearly.

Keavon: Now that we're breaking attributes into `layout_attributes`, `user_attributes`, (and `event_attributes` in the future for `on:` prefixes) I quite like this change!

PR #18, opened 2020-08-14
Switch to the Bevy engine

GabrielMajeri:
Replaces the current WIP rendering system with Bevy. The `main.rs` file now contains all the relevant application code.

**TODO**: I've unsafely derived `Send` and `Sync` for `LayoutSystem`, because I just wanted to see if it works. So far, there haven't been any issues. But we should try to review our thread safety.


GabrielMajeri:
@Keavon I've updated the PR to showcase how we could render the direct children of the main window.

This is how it looks like now:
bevy

(it's flipped upside-down because Bevy's Y axis goes from bottom-to-top :sweat_smile:)

Keavon:
Shelved on account of switching to a web app frontend for the short term.

GabrielMajeri:
I've deleted the branch since it hasn't been relevant for quite a while.

Keavon:
> I've deleted the branch since it hasn't been relevant for quite a while.

Ah, ancient history :) Thanks for your contributions early on, even though we did end up shifting to a new codebase and wiping out all your work (sorry about that! but it was ultimately the right strategy to make progress). Let me know if you're interested at all in contributing again at some point in the future, there's lots of exciting progress being made now with the node system just now being integrated.

PR #31, opened 2021-03-15
Define the js wasm-editor interface

TrueDoctor:
### Goal
Define functions to allow interfacing with the wasm editor state from the javascript frontend code

### Issues/TODO
* Explore the use of wasm-bindgen to pass structs as parameters
* Extend and de-duplicate functionality
* Develop concept for handling panels and other ui elements


packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+/// Modify the currently selected tool in the document state store
+#[wasm_bindgen]
+pub fn select_tool(tool_id: u32) {
+	todo!()
+}
+
```
Keavon, line 15: Let's wrap things like the `u32` of `document_id` as a `DocumentId`.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+/// Returns a unique document identifier
+#[wasm_bindgen]
+pub fn load_document(raw_data: &[u8]) -> u32 {
+	todo!()
+}
+
+/// Mouse path drawn on the canvas if the mouse is clicked for a longer period of time
+/// (every mouse position between the mouse down and up event)
+#[wasm_bindgen]
+pub fn canvas_mouse_path(mousepath: &[u64], modkeys: u8, ui_element_id: u32) -> u32 {
```
Keavon, line 33: Best way to do this is probably letting JS accumulate mouse pixel movements in a buffer. Then when it gets the chance to call this function with its buffer, the buffer gets purged and it can keep recording afterwards.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+
+/// Notify the editor that the mouse hovers above a panel
+#[wasm_bindgen]
+pub fn panel_enter_hover(panel_id: u32) {
+	todo!()
+}
+
+/// Notify the editor that the mouse leaves a panel
+#[wasm_bindgen]
+pub fn panel_leave_hover(panel_id: u32) {
```
Keavon, line 45: Might not need this because we can change it to 0 when you're hovering over the main window.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+	todo!()
+}
+
+/// Notify the editor that a UI-element is clicked
+// TODO: should this function return UI changes?
+#[wasm_bindgen]
+pub fn click_element(ui_element_id: u32, modkeys: u8) {
```
Keavon, line 52: `modkeys` probably unnecessary

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
 
+/// Modify the currently selected tool in the document state store
+#[wasm_bindgen]
+pub fn select_tool(tool_id: u32) {
```
Keavon, line 12: String probably better.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+/// Notify the editor that a UI-element is clicked
+// TODO: should this function return UI changes?
+#[wasm_bindgen]
+pub fn click_element(ui_element_id: u32, modkeys: u8) {
+	todo!()
+}
+
+/// Update working colors
+#[wasm_bindgen]
+pub fn update_colors(primary_color: u32, secondary_color: u32) {
```
Keavon, line 58: Colors will be defined as:

struct Color {
    r: f32,
    g: f32,
    b: f32,
    a: f32,
}

protheory8:
```
struct Color {
    r: f32,
    g: f32,
    b: f32,
    a: f32,
}
```

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+	todo!()
+}
+
+/// Query the name of a specific document
+#[wasm_bindgen]
+pub fn get_document_name(document_id: u32) {
```
Keavon, line 64: Returns a string

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
 #[global_allocator]
 static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
Keavon, line 9: We should break up these functions into separate files based on what panel they are a part of. *Window* is the name of the window shell "panel". It would also be helpful if we can make the JS side aware of the "namespacing" of these functions to avoid collisions and make it clear when code is calling the window, the viewport panel, the properties panel, the layers panel, the graph panel, etc. One way to do this is by prefixing everything, however a more proper way of "namespacing" would be preferable.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+// TODO: should it be possible to open multiple documents at once?
+#[wasm_bindgen]
+pub fn select_document(document_id: u32) {
+	todo!()
+}
+
+/// Load a new .gdd file into the editor
+/// Returns a unique document identifier
+#[wasm_bindgen]
+pub fn load_document(raw_data: &[u8]) -> u32 {
```
Keavon, line 26: Graphite 0.1 has no GDD file, let's remove this for now.

packages/wasm-bindings/src/lib.rs
```diff
@@ -7,6 +7,82 @@ use wasm_bindgen::prelude::*;
...
+/// Modify the currently selected tool in the document state store
+#[wasm_bindgen]
+pub fn select_tool(tool_id: u32) {
+	todo!()
+}
+
+/// Modify the currently selected Document in the editor state store
+// TODO: should it be possible to open multiple documents at once?
+#[wasm_bindgen]
+pub fn select_document(document_id: u32) {
```
Keavon, line 19: The Editor Library should be document-agnostic. You might have two different document viewports open side-by-side. I believe the Event Dispatcher can infer which document the event is referring to based on the mouse hover status (`panel_enter_hover(...)`). Then the Event Dispatcher sends either an Action or Operation. Operations all include the Document ID. And Actions that pertain to the Open Documents State Store also include the Document ID.

web-frontend/wasm-wrapper/src/.exrc
```diff
@@ -0,0 +1,213 @@
...
+set statusline=%!UnifiedGensl()
+set suffixes=.bak,~,.o,.h,.info,.swp,.obj,.lock
+set tabstop=4
+set timeoutlen=500
+set undodir=~/.vim_runtime/temp_dirs/undodir
+set undofile
+set wildignore=*.o,*~,*.pyc,*/.git/*,*/.hg/*,*/.svn/*,*/.DS_Store,*/target/*
+set wildmode=longest,list
+set window=41
+" vim: set ft=vim :
```
bjorn3, line 213: Should this file be added to `.gitignore`?

PR #33, opened 2021-03-19
Add init function to wasm-bindings crate and update dependencies

protheory8:


Keavon:
Let's also remove wee_alloc per our discussion on Discord.

PR #42, opened 2021-03-24
Translate tool names to rust enum values

TrueDoctor:
- Add default colors
- Translate Strings to ToolTypes

### DIFF
--- a/client/web/wasm/src/viewport.rs
+++ b/client/web/wasm/src/viewport.rs
@@ -1,10 +1,18 @@
-use crate::wrappers::Color;
+use crate::wrappers::{translate_tool, Color};
+use graphite_editor_core::tools::ToolState;
 use wasm_bindgen::prelude::*;
+
+pub static mut TOOL_STATE: ToolState = ToolState::default();
 
 /// Modify the currently selected tool in the document state store
 #[wasm_bindgen]
-pub fn select_tool(tool: String) {
-	todo!()
+pub fn select_tool(tool: String) -> Result<(), JsValue> {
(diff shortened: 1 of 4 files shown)


client/web/wasm/src/viewport.rs
```diff
@@ -1,10 +1,18 @@
...
 use wasm_bindgen::prelude::*;
 
+pub static mut TOOL_STATE: ToolState = ToolState::default();
+
 /// Modify the currently selected tool in the document state store
 #[wasm_bindgen]
-pub fn select_tool(tool: String) {
-	todo!()
+pub fn select_tool(tool: String) -> Result<(), JsValue> {
+	let tool_state = unsafe { &mut TOOL_STATE };
```
Keavon, line 10: Why do we need `unsafe` code? I'm going to suggest here again that we switch from using the inflexible index-based `ToolType` enum, that we instead use a HashMap based on the string name of the tool, as this will help support plugin-provided tools down the road. The performance will be entirely negligible.

TrueDoctor: We need the unsafe code because we use a static mutable variable, that is just due to the necessity of storing a state that persists across multiple invocations

TrueDoctor: I'm not sure the plugin argument is entirely valid though, because the tool state is defined in the editor-core and the plugin would never touch that. With the current design I don't see how plugins could add tools

TrueDoctor: But I do understand It from a code maintainability standpoint

Keavon (approved):

PR #44, opened 2021-03-27
Add camelcase rule, import Error type from JS and rename get_inner_color to inner.

protheory8:


client/web/wasm/src/shims.rs
```diff
@@ -0,0 +1,10 @@
+use wasm_bindgen::prelude::*;
+
+#[wasm_bindgen]
+extern "C" {
+	#[derive(Clone, Debug)]
+	pub type Error;
+
+	#[wasm_bindgen(constructor)]
+	pub fn new(msg: &str) -> Error;
+}
```
TrueDoctor, line 10: Why the new shims file?
would Color fit in there as well?

Keavon (approved):

PR #45, opened 2021-03-27
Add metadata

protheory8:


client/web/wasm/Cargo.toml
```diff
@@ -12,14 +12,25 @@ crate-type = ["cdylib", "rlib"]
...
-# code size when deploying.
 console_error_panic_hook = { version = "0.1.6", optional = true }
+graphite-editor-core = { path = "../../../core/editor" }
+wasm-bindgen = "0.2.72"
 
 [dev-dependencies]
 wasm-bindgen-test = "0.3.22"
+
+[package.metadata.wasm-pack.profile.dev]
+wasm-opt = false
```
TrueDoctor, line 23: I think we maybe should enable wasm-opt, it adds about 1-2 s of compile time, but might make an noticeable runtime difference
or is that to allow for debugging?

protheory8: I don't know, I think `wasm-opt` is disabled for dev builds by default anyway.

TrueDoctor: Alright we can change it if we notice a significant difference between release and debug builds

TrueDoctor (approved):

Keavon (approved):

PR #46, opened 2021-03-27
Add .gitattributes file

protheory8:
I also ran `git add --renormalize .` to apply the settings, not sure if everything works correctly now though.


TrueDoctor (approved):
Look good, The only <CR> I find seems to be in the node_modules folder.
(`grep -rIl -m 1 $'\r' .`)

PR #51, opened 2021-03-28
Record mouse movement

RustyNixieTube:


client/web/wasm/src/viewport.rs
```diff
@@ -16,22 +16,35 @@ pub fn select_tool(tool: String) -> Result<(), JsValue> {
...
 pub fn on_mouse_move(x: u32, y: u32) {
-	// SAFETY: This is safe because the code can only ever run in a single thread
-	unsafe {
-		POS = (x, y);
-	}
+	EDITOR_STATE.with(|editor| {
+		let mut editor = editor.borrow_mut();
+		if editor.tools.mouse_is_clicked {
+			editor.tools.trace.append_point(x, y)
+		}
```
TrueDoctor, line 26: I think it might be useful to save the state regardless of whether the button was clicked, because we store that information in the mouse state. Butl well talk about that in a few mins

core/editor/src/tools/mod.rs
```diff
@@ -1,7 +1,9 @@
-use crate::Color;
+use crate::{events::Trace, Color};
 use std::collections::HashMap;
 
 pub struct ToolState {
+	pub mouse_is_clicked: bool,
```
TrueDoctor, line 5: Could you just save either a MouseState struct, or a MousKeys to allow to track all mouse buttons?

core/editor/src/dispatcher/events.rs
```diff
@@ -18,6 +18,34 @@ pub enum Response {
...
+	pub fn new() -> Self {
+		Self(vec![])
+	}
+	pub fn first_point(&self) -> Option<&MouseState> {
+		if !self.0.is_empty() {
+			Some(&self.0[0])
+		} else {
+			None
+		}
+	}
```
TrueDoctor, line 32: You can just use `self.0.get(0)` instead

core/editor/src/dispatcher/events.rs
```diff
@@ -18,6 +18,34 @@ pub enum Response {
...
+	pub fn last_point(&self) -> Option<&MouseState> {
+		let trace = &self.0;
+		let trace_length = trace.len();
+		if trace_length > 0 {
+			Some(&trace[trace_length - 1])
+		} else {
+			None
+		}
```
TrueDoctor, line 40: see above

TrueDoctor (changes requested):

client/web/wasm/src/viewport.rs
```diff
@@ -16,22 +16,35 @@ pub fn select_tool(tool: String) -> Result<(), JsValue> {
...
 #[wasm_bindgen]
 pub fn on_mouse_click(x: u32, y: u32) -> Result<(), JsValue> {
 	let ev = events::Event::Click(events::MouseState::from_pos(x, y));
-	EDITOR_STATE.with(|editor| editor.borrow_mut().handle_event(ev)).map_err(|err| Error::new(&err.to_string()).into())
+	EDITOR_STATE
+		.with(|editor| {
+			let mut editor = editor.borrow_mut();
+			editor.tools.mouse_is_clicked = true;
+			editor.tools.trace.clear();
+			editor.handle_event(ev)
```
T0mstone, line 39: Maybe we should rename the event since it sounds kind of weird that each point in a trace will send a `Click` event

core/editor/src/dispatcher/events.rs
```diff
@@ -18,6 +18,34 @@ pub enum Response {
 
 #[derive(Debug, Clone)]
 pub struct Trace(Vec<MouseState>);
+
+impl Trace {
+	pub fn new() -> Self {
```
T0mstone, line 23: This should implement `Default`

core/editor/src/dispatcher/events.rs
T0mstone: also `Vec::last`

core/editor/src/dispatcher/events.rs
T0mstone: even better: `Vec::first`

core/editor/src/dispatcher/events.rs
```diff
@@ -18,6 +18,29 @@ pub enum Response {
...
+		Self(vec![])
+	}
+	pub fn first_point(&self) -> Option<&MouseState> {
+		self.0.get(0)
```
T0mstone, line 27: even better would be `self.0.first()`

core/editor/src/dispatcher/events.rs
```diff
@@ -18,6 +18,29 @@ pub enum Response {
...
+	}
+	pub fn last_point(&self) -> Option<&MouseState> {
+		// avoid underflow
+		if let Some(idx) = self.0.len().checked_sub(1) {
```
T0mstone, line 31: even better would be `self.0.last()`

TrueDoctor (approved):

PR #53, opened 2021-03-28
Implement basic circle stamping

T0mstone:


core/document/src/lib.rs
```diff
@@ -1,7 +1,30 @@
...
+#[derive(Default, Debug, Clone, PartialEq)]
+pub struct Document {
+	pub svg: Vec<SvgElement>,
+}
+
+impl Document {
+	pub fn render(&self) -> String {
+		self.svg.iter()
+			.map(|element| element.render())
+			.collect::<Vec<_>>() // multiline
```
Keavon, line 27: Would you mind if we do actually make this all one line? It's not actually that long.

core/editor/src/dispatcher/mod.rs
```diff
@@ -9,60 +9,68 @@ pub struct Dispatcher {
...
 			}
 			Event::MouseDown(mouse_state) => {
-				tool_state.mouse_state = mouse_state;
+				state.tools.mouse_state = mouse_state;
 				// the state has changed so we add a trace point
-				tool_state.record_trace_point();
+				state.tools.record_trace_point();
 
-				self.emit_response(Response::UpdateCanvas);
+				// self.emit_response(Response::UpdateCanvas { document: state.document.render() });
```
Keavon, line 40: For the record, this will be removed later, and then the dispatcher will send events based on the context (like which tool is active) to the tool router, then it's sent to the tool and that tool is in charge of emitting Responses.

Keavon (approved):

PR #58, opened 2021-03-31
Add Active Tool Router

### LINKED ISSUE #57: Active Tool Router and stub for Tool State Machine
Inputs events (let's start with mouse inputs and modifier keys for now) should be directed from the Dispatcher to the Active Tool Router, then from there to the corresponding tool's state machine definition run by the Tool State Machine system.

The Active Tool Router is basically just a switch. It redirects events it receives to the FSM corresponding to the `active_tool`.

This issue should not add any functionality for the user to run, but it should extend the stubbed out tool `Response` to a click on the canvas into the Active Tool Router and to a stub for the FSM. This is followed up with https://github.com/GraphiteEditor/Graphite/issues/54 for the actual FSM implementation.

protheory8:
Fixes #57
There is a FSM stub for `Select` tool.

### DIFF
--- a/core/document/src/operation.rs
+++ b/core/document/src/operation.rs
@@ -1,0 +1,3 @@
+pub enum Operation {
+	AddCircle((f64, f64), f64),
+}
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -13,38 +13,38 @@ impl Dispatcher {
 		log::trace!("{:?}", event);
 		match event {
 			Event::SelectTool(tool_type) => {
-				state.tools.active_tool = tool_type;
+				state.tool_state.active_tool_type = tool_type;
 				Ok(())
 			}
 			Event::SelectPrimaryColor(color) => {
-				state.tools.primary_color = color;
+				state.tool_state.primary_color = color;
 				Ok(())
 			}
 			Event::SelectSecondaryColor(color) => {
-				state.tools.secondary_color = color;
+				state.tool_state.secondary_color = color;
 				Ok(())
 			}
 			Event::SwapColors => {
-				std::mem::swap(&mut state.tools.primary_color, &mut state.tools.secondary_color);
+				std::mem::swap(&mut state.tool_state.primary_color, &mut state.tool_state.secondary_color);
 				Ok(())
 			}
 			Event::ResetColors => {
-				state.tools.primary_color = Color::BLACK;
-				state.tools.secondary_color = Color::WHITE;
+				state.tool_state.primary_color = Color::BLACK;
+				state.tool_state.secondary_color = Color::WHITE;
 				Ok(())
 			}
 			Event::MouseDown(mouse_state) => {
-				state.tools.mouse_state = mouse_state;
-				// the state has changed so we add a trace point
-				state.tools.record_trace_point();
+				state.tool_state.mouse_state = mouse_state;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
 
-				// self.emit_response(Response::UpdateCanvas { document: state.document.render() });
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::MouseUp(mouse_state) => {
-				state.tools.mouse_state = mouse_state;
-				// the state has changed so we add a trace point
-				state.tools.record_trace_point();
+				state.tool_state.mouse_state = mouse_state;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
 
 				state.document.svg.push(SvgElement::Circle(Circle {
 					center: Point {
@@ -55,32 +55,39 @@ impl Dispatcher {
 				}));
 
 				self.emit_response(Response::UpdateCanvas { document: state.document.render() });
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::MouseMovement(pos) => {
-				state.tools.mouse_state.position = pos;
-				state.tools.record_trace_point();
+				state.tool_state.mouse_state.position = pos;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::ModifierKeyDown(mod_keys) => {
-				state.tools.mod_keys = mod_keys;
-				// the state has changed so we add a trace point
-				state.tools.record_trace_point();
+				state.tool_state.mod_keys = mod_keys;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::ModifierKeyUp(mod_keys) => {
-				state.tools.mod_keys = mod_keys;
-				// the state has changed so we add a trace point
-				state.tools.record_trace_point();
+				state.tool_state.mod_keys = mod_keys;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::KeyPress(key) => todo!(),
 		}
 	}
+
 	pub fn emit_response(&self, response: Response) {
 		let func = &self.callback;
 		func(response)
 	}
+
 	pub fn new(callback: Callback) -> Dispatcher {
 		Dispatcher { callback }
 	}
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -11,40 +11,42 @@ pub struct Dispatcher {
 impl Dispatcher {
 	pub fn handle_event(&self, state: &mut EditorState, event: Event) -> Result<(), EditorError> {
 		log::trace!("{:?}", event);
+
 		match event {
 			Event::SelectTool(tool_type) => {
 				state.tool_state.active_tool_type = tool_type;
+
 				Ok(())
 			}
 			Event::SelectPrimaryColor(color) => {
 				state.tool_state.primary_color = color;
+
 				Ok(())
 			}
 			Event::SelectSecondaryColor(color) => {
 				state.tool_state.secondary_color = color;
+
 				Ok(())
 			}
 			Event::SwapColors => {
 				std::mem::swap(&mut state.tool_state.primary_color, &mut state.tool_state.secondary_color);
+
 				Ok(())
 			}
 			Event::ResetColors => {
 				state.tool_state.primary_color = Color::BLACK;
 				state.tool_state.secondary_color = Color::WHITE;
+
 				Ok(())
 			}
 			Event::MouseDown(mouse_state) => {
 				state.tool_state.mouse_state = mouse_state;
-				// The state has changed so we add a trace point.
-				state.tool_state.record_trace_point();
+				state.tool_state.active_tool()?.handle_input(event);
 
-				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::MouseUp(mouse_state) => {
 				state.tool_state.mouse_state = mouse_state;
-				// The state has changed so we add a trace point.
-				state.tool_state.record_trace_point();
 
 				state.document.svg.push(SvgElement::Circle(Circle {
 					center: Point {
@@ -53,30 +55,28 @@ impl Dispatcher {
 					},
 					radius: 10.0,
 				}));
-
 				self.emit_response(Response::UpdateCanvas { document: state.document.render() });
-				state.tool_state.active_tool().handle_input(event);
+
+				state.tool_state.active_tool()?.handle_input(event);
+
 				Ok(())
 			}
 			Event::MouseMovement(pos) => {
 				state.tool_state.mouse_state.position = pos;
-				// The state has changed so we add a trace point.
-				state.tool_state.record_trace_point();
-				state.tool_state.active_tool().handle_input(event);
+				state.tool_state.active_tool()?.handle_input(event);
+
 				Ok(())
 			}
 			Event::ModifierKeyDown(mod_keys) => {
 				state.tool_state.mod_keys = mod_keys;
-				// The state has changed so we add a trace point.
-				state.tool_state.record_trace_point();
-				state.tool_state.active_tool().handle_input(event);
+				state.tool_state.active_tool()?.handle_input(event);
+
 				Ok(())
 			}
 			Event::ModifierKeyUp(mod_keys) => {
 				state.tool_state.mod_keys = mod_keys;
-				// The state has changed so we add a trace point.
-				state.tool_state.record_trace_point();
-				state.tool_state.active_tool().handle_input(event);
+				state.tool_state.active_tool()?.handle_input(event);
+
 				Ok(())
 			}
 			Event::KeyPress(key) => todo!(),
--- a/core/editor/src/error.rs
+++ b/core/editor/src/error.rs
@@ -3,13 +3,14 @@ use crate::Color;
 use std::error::Error;
 use std::fmt::{self, Display};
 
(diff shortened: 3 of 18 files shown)


core/editor/src/tools/mod.rs
```diff
@@ -37,6 +69,10 @@ impl ToolState {
 			mod_keys: self.mod_keys,
 		})
 	}
+
+	pub fn active_tool(&mut self) -> &mut Box<dyn Tool> {
+		self.tools.get_mut(&self.active_tool_type).unwrap()
```
TrueDoctor, line 74: Could you either avoid the unwrap by  returning an option or add a comment on why it is safe?

core/editor/src/tools/select.rs
```diff
@@ -0,0 +1,42 @@
...
+use crate::events::MouseKeys;
+use crate::tools::Tool;
+
+pub struct Select(Fsm);
+
+impl Tool for Select {
+	fn handle_input(&mut self, event: Event) {
+		match event {
+			Event::MouseDown(state) => {
+				if state.mouse_keys == MouseKeys::LEFT {
```
TrueDoctor, line 11: I'm not to familiar with the bit vector crate but that seems like it would compare all bits but that might not be what we want

core/editor/src/tools/mod.rs
```diff
@@ -1,14 +1,30 @@
...
+mod sample;
+mod select;
+mod shape;
+
+use crate::events::{Event, ModKeys, MouseState, Trace, TracePoint};
+use crate::Color;
 use std::collections::HashMap;
 
+pub trait Tool {
+	fn handle_input(&mut self, event: Event);
```
TrueDoctor, line 17: Didn't you want to rename that function to `handle`at some point?

TrueDoctor, line 17: I think the function should also return an `Operation`
The Operation would need to be defined in the `document-core` and would just be an enum with values like AddCircle(center, radius, color, …)
But we could also put that into a seperate PR if it is too much work

core/editor/src/tools/ellipse.rs
```diff
@@ -0,0 +1,16 @@
...
+	fn handle_input(&mut self, event: Event) {
+		todo!();
+	}
+}
+
+impl Default for Ellipse {
+	fn default() -> Self {
+		Self
+	}
+}
```
TrueDoctor, line 16: I suppose you could just #[derive(Default)] it for now

core/editor/src/dispatcher/mod.rs
```diff
@@ -55,32 +55,39 @@ impl Dispatcher {
...
 				self.emit_response(Response::UpdateCanvas { document: state.document.render() });
+				state.tool_state.active_tool().handle_input(event);
 				Ok(())
 			}
 			Event::MouseMovement(pos) => {
-				state.tools.mouse_state.position = pos;
-				state.tools.record_trace_point();
+				state.tool_state.mouse_state.position = pos;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
```
TrueDoctor, line 64: You can delete every reference to Trace that is currently in the `ToolState`.
The `Trace` will be handled by the different tool.
I just wanted you to keep the modifications to the mod_keys/mouse_keys to avoid weird issues with holding a modifier key while changing the tool.

core/editor/src/dispatcher/mod.rs
```diff
@@ -13,38 +13,38 @@ impl Dispatcher {
...
 			}
 			Event::MouseUp(mouse_state) => {
-				state.tools.mouse_state = mouse_state;
-				// the state has changed so we add a trace point
-				state.tools.record_trace_point();
+				state.tool_state.mouse_state = mouse_state;
+				// The state has changed so we add a trace point.
+				state.tool_state.record_trace_point();
 
 				state.document.svg.push(SvgElement::Circle(Circle {
```
TrueDoctor, line 49: Could you if you decide that it is feasible to Include the Operations in this PR, move this to the Circle Tool? 
That would be an awesome end to end demonstration and show how the information flow is intended to take place

core/editor/src/tools/mod.rs
```diff
@@ -1,14 +1,30 @@
...
+}
+
 pub struct ToolState {
 	pub mouse_state: MouseState,
 	pub mod_keys: ModKeys,
 	pub trace: Trace,
```
TrueDoctor, line 23: This can be removed, see comment above

core/editor/src/tools/mod.rs
protheory8: I expected that it will be guaranteed that the `tools` HashMap will have every tool in it, so I used `unwrap` here. I will change it to `Result` return type I guess.

core/editor/src/tools/select.rs
protheory8: Not sure how to fix this one.

core/editor/src/tools/mod.rs
protheory8: No, I actually wanted to rename it to `handle_input` and I did (you suggested `handle`).

protheory8: I will do that, I'm wondering if it should return `Option<Operation>` instead though.

core/editor/src/tools/ellipse.rs
protheory8: I forgot that `Default` is derivable, I will fix that.

core/editor/src/dispatcher/mod.rs
protheory8: Operations that actually work probably not, I can do that in the next PR though.
I can make an enum though.

core/editor/src/tools/mod.rs
TrueDoctor: I think it might be better to use an option and do the error handling one leve above

core/editor/src/tools/select.rs
TrueDoctor: `if state.mouse_keys && MousKeys::LEFT == 1`

core/editor/src/tools/mod.rs
TrueDoctor: Oh, sorry I got confused, I originally used the name `handle_input` and just abbreviated it from there on, so we are on the same page

TrueDoctor: Either that or add a none operation but I think an option would be cleaner

core/editor/src/tools/mod.rs
TrueDoctor: What your saying is true, but might cause some headache when adding tools and forgetting to initialize them. This is not as bad because we have stack traces but it is cleaner that way

protheory8: I meant how to handle the error on the level above, just unwrap it?

protheory8: This is a core library, we can't throw a JS exception here.

core/editor/src/dispatcher/mod.rs
TrueDoctor: If you hold down the shift key, then switch to a different tool, release the shift key and switch back to the first tool, the first tool never got the notification that the key was released

core/editor/src/tools/mod.rs
TrueDoctor: That is true it would be nice though if our core libraries could not panic, you could just add an new error type and propagate it through

TrueDoctor (approved):

core/editor/src/tools/mod.rs
```diff
@@ -20,7 +38,23 @@ impl Default for ToolState {
...
+				hash_map.insert(ToolType::Select, Box::new(select::Select::default()));
+				hash_map.insert(ToolType::Crop, Box::new(crop::Crop::default()));
+				hash_map.insert(ToolType::Navigate, Box::new(navigate::Navigate::default()));
+				hash_map.insert(ToolType::Sample, Box::new(sample::Sample::default()));
+				hash_map.insert(ToolType::Path, Box::new(path::Path::default()));
+				hash_map.insert(ToolType::Pen, Box::new(pen::Pen::default()));
+				hash_map.insert(ToolType::Line, Box::new(line::Line::default()));
+				hash_map.insert(ToolType::Rectangle, Box::new(rectangle::Rectangle::default()));
+				hash_map.insert(ToolType::Ellipse, Box::new(ellipse::Ellipse::default()));
+				hash_map.insert(ToolType::Shape, Box::new(shape::Shape::default()));
```
T0mstone, line 54: maybe use the `maplit` crate (a small dependency) to make this code easier to maintain?

core/editor/src/tools/select.rs
```diff
@@ -0,0 +1,40 @@
...
+use document_core::Operation;
+
+#[derive(Default)]
+pub struct Select(Fsm);
+
+impl Tool for Select {
+	fn handle_input(&mut self, event: Event) -> Option<Operation> {
+		match event {
+			Event::MouseDown(state) => {
+				if !(state.mouse_keys & MouseKeys::LEFT).is_empty() {
```
T0mstone, line 13: use `!state.mouse_keys.contains(MouseKays::LEFT)` instead.

core/document/src/operation.rs
```diff
@@ -0,0 +1,3 @@
+pub enum Operation {
+	AddCircle((f64, f64), f64),
```
T0mstone, line 2: Maybe this should use a `struct Point(f64, f64)` or something similar

core/editor/src/error.rs
```diff
@@ -3,13 +3,14 @@ use crate::Color;
...
 
-/// The error type used by the graphite editor.
+/// The error type used by the Graphite editor.
 #[derive(Clone, Debug)]
 pub enum EditorError {
 	InvalidOperation(String),
 	InvalidEvent(String),
 	Misc(String),
 	Color(String),
+	ToolNotInitialized,
```
T0mstone, line 13: The problem may be on my end, but why do we need this?

core/editor/src/tools/select.rs
T0mstone: Also I think there can only ever be a single mouse key pressed per event, so that could even be `state.mouse_keys != MouseKeys::LEFT` with a comment stating what I just said.

core/editor/src/tools/select.rs
```diff
@@ -0,0 +1,40 @@
...
+					self.0 = Fsm::LmbDown;
+				}
+			}
+			Event::MouseUp(state) => {
+				if self.0 == Fsm::LmbDown && state.mouse_keys == MouseKeys::LEFT {
```
T0mstone, line 18: This can be left untouched if my second comment from just above is implemented. Otherwise replace this with `state.mouse_keys.contains(MouseKeys::LEFT)`

T0mstone (changes requested):

T0mstone:
Also how do we do immediate FSM transitions? If I recall correctly, the FSM should directly jump from `SelectedObject` to `Ready`

core/editor/src/tools/mod.rs
protheory8: That would be pretty nice, should I make a similar macro to the one from `maplit` or should I just use the crate?

T0mstone: I think using a similar macro would be better since

1. It doesn't add a dependency
2. We can also include `Box::new` and `ToolType::` in the macro, so we only have to provide the enum variants when invoking the macro

be sure to initialize the hashmap with `HashMap::with_capacity` though

core/editor/src/tools/select.rs
T0mstone: oh right

core/document/src/operation.rs
T0mstone: I agree. That was meant as a strong 'Maybe' anyway

core/editor/src/tools/mod.rs
```diff
@@ -10,6 +10,7 @@ mod select;
 mod shape;
 
 use crate::events::{Event, ModKeys, MouseState, Trace, TracePoint};
+// use crate::gen_tools_hashmap;
```
T0mstone: You forgot(?) to remove this

T0mstone (changes requested):

core/editor/src/tools/mod.rs
protheory8: I already did

core/editor/src/macros.rs
```diff
@@ -0,0 +1,15 @@
+macro_rules! count_args {
```
Keavon, line 1: Could you please provide a detailed comment describing exactly what each of these do? (Feel free to use multiple lines of comments if needed.) They are nontrivial to read.

core/editor/src/tools/crop.rs
```diff
@@ -0,0 +1,12 @@
+use crate::events::Event;
+use crate::tools::Tool;
+use document_core::Operation;
+
+#[derive(Default)]
+pub struct Crop;
+
+impl Tool for Crop {
+	fn handle_input(&mut self, event: Event) -> Option<Operation> {
```
Keavon, line 9: If I understand this correctly, tools won't respond with an Operation. They will end up needing direct access to make calls to the Dispatcher with zero, one, or many events (defined in the lambda function in each state of the FSM).

core/editor/src/macros.rs
protheory8: T0mstone told me to use `count_args`, I can't understand it very well either.

gen_tools_hash_map generates
```rust
{
    let hash_map = HashMap::with_capacity(/* Number of args passed in, counted using `count_args` macro */);

    hash_map.insert(ToolType::Select, select::Select::default());
    hash_map.insert(ToolType::Crop, crop::Crop::default());

    hash_map
}
```
from
```rust
gen_tools_hash_map! {
    Select => select::Select,
    Crop => crop::Crop,
}
```

core/editor/src/tools/crop.rs
protheory8: What do you mean?

core/editor/src/tools/mod.rs
```diff
@@ -1,14 +1,32 @@
...
+
 pub struct ToolState {
 	pub mouse_state: MouseState,
 	pub mod_keys: ModKeys,
 	pub trace: Trace,
 	pub primary_color: Color,
 	pub secondary_color: Color,
-	pub active_tool: ToolType,
+	pub active_tool_type: ToolType,
+	pub tools: HashMap<ToolType, Box<dyn Tool>>,
```
Keavon, line 29: Does it make sense to store this on `ToolState`? It doesn't seem like state to me.

core/editor/src/tools/mod.rs
```diff
@@ -1,14 +1,32 @@
...
+mod shape;
+
+use crate::events::{Event, ModKeys, MouseState, Trace, TracePoint};
+use crate::Color;
+use crate::EditorError;
+use document_core::Operation;
 use std::collections::HashMap;
 
+pub trait Tool {
+	fn handle_input(&mut self, event: Event) -> Option<Operation>;
```
Keavon, line 19: We'll want to replace `Option` with `Vec` so a queue of `Operation`s can be returned.

core/editor/src/tools/crop.rs
Keavon: We discussed this further in Discord and concluded that we should indeed stick with returning an Operation, although a Vec of Operations instead of just 1 or 0. See https://github.com/GraphiteEditor/Graphite/pull/58#discussion_r605509501

Keavon (approved):
Please fix comments before merging.

core/editor/src/macros.rs
Keavon: Oh to clarify, I meant as comments in the code 😊

core/editor/src/tools/mod.rs
TrueDoctor: Yes it does, pleas @ me for details

T0mstone (approved):

PR #59, opened 2021-03-31
Implement svg layer structure

TrueDoctor:


core/document/src/lib.rs
```diff
@@ -2,12 +2,16 @@ pub use kurbo::{Circle, Point};
 
 #[derive(Debug, Clone, PartialEq)]
 pub enum SvgElement {
+    Folder(Folder),
```
Keavon, line 5: These are space indents

core/document/src/lib.rs
```diff
@@ -16,12 +20,32 @@ impl SvgElement {
 }
 
 #[derive(Default, Debug, Clone, PartialEq)]
+pub struct Folder {
+    elements: Vec<SvgElement>,
```
Keavon, line 24: Would a single Vec of tuples be better?

Keavon (approved):

core/document/src/lib.rs
TrueDoctor: Cache efficiency Data oriented programming sorts of stuff

core/document/src/lib.rs
```diff
@@ -22,31 +26,148 @@ impl SvgElement {
...
+pub enum DocumentError {
+	ElementNotFound,
+	NotAShape,
+	ElementAlreadyExists,
+	InvalidPath,
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct Folder {
+	elements: HashMap<String, SvgElement>,
```
Keavon, line 39: I'm thinking we want to use a vector not a hashmap. Names should be a property on every `SvgElement`, not something intrinsic (and strictly unique) to folders. Most shapes won't be named, unless you go through the effort to rename them, and in that case there should be not requirement about name uniqueness. Using a vector instead would also help avoid the requirement of naming folders arbitrarily, like where you concatenate in the `fold()` function above.

core/document/src/lib.rs
```diff
@@ -22,31 +26,148 @@ impl SvgElement {
...
+impl Default for Folder {
+	fn default() -> Self {
+		Self { elements: HashMap::new() }
+	}
+}
+
+#[derive(Debug, Clone, PartialEq)]
 pub struct Document {
-	pub svg: Vec<SvgElement>,
+	pub svg: SvgElement,
```
Keavon, line 79: Should the root actually be a folder, so it can contain N number of `SvgElements` inside?

core/document/src/lib.rs
```diff
@@ -22,31 +26,148 @@ impl SvgElement {
...
 impl Document {
 	pub fn render(&self) -> String {
-		self.svg.iter().map(|element| element.render()).collect::<Vec<_>>().join("\n")
+		self.svg.render()
+	}
+
+	pub fn open(&self, path: &str) -> Result<&SvgElement, DocumentError> {
+		assert!(matches!(self.svg, SvgElement::Folder(_)), "SVG root has to be of type Folder");
+		let mut root = &self.svg;
+		for s in path.split('/') {
```
Keavon, line 98: Throughout the whole PR, could you please change the use of single-letter variable names into more descriptive names? I still need to find the time to write a code style guide. But basically `s` -> `segment`, `f` -> `folder`, `c` -> `color`, `d` -> `document`, etc. However `i` in a for-loop is fine and `e` in an error are fine because those are well known "common" and "standard" variable names in their own right.

core/document/src/lib.rs
```diff
@@ -22,31 +26,148 @@ impl SvgElement {
...
+	}
 }
 
 impl Document {
 	pub fn render(&self) -> String {
-		self.svg.iter().map(|element| element.render()).collect::<Vec<_>>().join("\n")
+		self.svg.render()
+	}
+
+	pub fn open(&self, path: &str) -> Result<&SvgElement, DocumentError> {
```
Keavon, line 95: I'm not really sure I understand the purpose of reading and writing in a path-like format. What capability does that provide us for our vector editor (during 0.1 and 0.2)?

core/document/src/lib.rs
Keavon, line 98: I'm thinking a vector (or slice) of strings would be safer than a slash-delimited path string. In case you want a slash in your path, for example, which could add complexity around escaping and room for mistakes.

core/document/src/lib.rs
```diff
@@ -1,17 +1,21 @@
...
+	Folder(Folder),
 	Circle(Circle),
 	Rect(Rect),
 }
 
 impl SvgElement {
 	pub fn render(&self) -> String {
 		match self {
+			Self::Folder(f) => f.elements.values().map(|e| e.render()).fold(String::with_capacity(f.elements.len() * 30), |s, e| s + &e),
 			Self::Circle(c) => {
```
Keavon, line 19: Throughout the whole PR, could you please change the use of single-letter variable names into more descriptive names? I still need to find the time to write a code style guide. But basically `s` -> `segment`, `f` -> `folder`, `c` -> `circle`/`color`, `d` -> `document`, etc. However `i` in a for-loop is fine and `e` in an error are fine because those are well known "common" and "standard" variable names in their own right.

core/document/src/lib.rs
```diff
@@ -1,17 +1,21 @@
...
 pub enum SvgElement {
+	Folder(Folder),
 	Circle(Circle),
 	Rect(Rect),
 }
 
 impl SvgElement {
 	pub fn render(&self) -> String {
 		match self {
+			Self::Folder(f) => f.elements.values().map(|e| e.render()).fold(String::with_capacity(f.elements.len() * 30), |s, e| s + &e),
```
Keavon, line 18: Throughout the whole PR, could you please change the use of single-letter variable names into more descriptive names? I still need to find the time to write a code style guide. But basically `s` -> `segment`, `f` -> `folder`, `c` -> `color`, `d` -> `document`, etc. However `i` in a for-loop is fine and `e` in an error are fine because those are well known "common" and "standard" variable names in their own right.

core/document/src/operation.rs
```diff
@@ -1,4 +1,6 @@
 pub enum Operation {
-	AddCircle { cx: f64, cy: f64, r: f64 },
-	AddRect { x0: f64, y0: f64, x1: f64, y1: f64 },
+	AddCircle { path: String, cx: f64, cy: f64, r: f64 },
+	AddRect { path: String, x0: f64, y0: f64, x1: f64, y1: f64 },
+	DeleteElement { path: String },
+	AddFolder { path: String },
```
Keavon, line 5: We'll also need an index for where (top, bottom, somewhere in the middle) to insert the circle/rect/folder within the directory specified by the `path`.

core/document/src/lib.rs
TrueDoctor: I'm not sure what you mean, the root is a folder

core/editor/src/tools/ellipse.rs
```diff
@@ -34,6 +34,7 @@ impl Default for EllipseToolFsmState {
 #[derive(Clone, Debug, Default)]
 struct EllipseToolData {
 	drag_start: ViewportPosition,
+	index: u64,
```
Keavon, line 37: Is `index` just a hack to deal with unique naming? This should be able to go away after switching from a hashmap to vector approach.

core/editor/src/tools/ellipse.rs
```diff
@@ -50,7 +51,13 @@ impl Fsm for EllipseToolFsmState {
 			(EllipseToolFsmState::LmbDown, Event::MouseUp(mouse_state)) => {
 				let r = data.drag_start.distance(&mouse_state.position);
 				log::info!("draw ellipse with radius: {:.2}", r);
+				let name = format!("ellipses/ellipse-{}", data.index);
+				if data.index == 0 {
+					operations.push(Operation::AddFolder { path: "ellipses".to_string() });
+				}
+				data.index += 1;
 				operations.push(Operation::AddCircle {
+					path: name,
```
Keavon, line 60: Name will just be a default ("Circle") until the user renames it.

core/document/src/lib.rs
Keavon: I mean changing this line to say `pub svg: Folder,`

core/document/src/lib.rs
```diff
@@ -22,31 +26,152 @@ impl SvgElement {
...
 }
 
 impl Document {
 	pub fn render(&self) -> String {
-		self.svg.iter().map(|element| element.render()).collect::<Vec<_>>().join("\n")
+		self.svg.render()
+	}
+
+	pub fn open(&self, path: &str) -> Result<&SvgElement, DocumentError> {
+		assert!(matches!(self.svg, SvgElement::Folder(_)), "SVG root has to be of type Folder");
```
T0mstone, line 96: This assert could also be removed if we make the root just a folder

Keavon (approved):
Awesome work! This is a great architecture for both 0.1 as well as 0.2 and onwards.

PR #60, opened 2021-04-01
Microtransactions (April Fools)

T0mstone:
This PR implements microtransactions, a feature essential to every piece of modern software.
The implementation is only a stub for now, as we still have to figure out how to process payments

### DIFF
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -16,6 +16,10 @@ impl Dispatcher {
 			Event::SelectTool(tool_type) => {
 				state.tool_state.active_tool_type = tool_type;
 
+				if !state.tool_state.can_use_tool(&tool_type) {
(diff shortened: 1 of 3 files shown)


TrueDoctor:
We might need to implement some sort of cryptographic verification to stop users from just modifying our memory from the js frontend

Keavon (approved):
Glad we are getting an early start on these important features!

TrueDoctor:
And what do you think about a limited stroke length? That way users could still use all tools, just not for any longer than 5 cm or some amount of time

Keavon:
We may also want to charge for either the red, green, or blue color primary. That way people can get a taste for Graphite's capabilities with 2/3 of the hue spectrum, but if they want to unlock the full capabilities of color and stop making art with stylized color palettes, a microtransaction would be required.

T0mstone:
April 1st should be over in every time zone now

PR #62, opened 2021-04-04
Add FSM and related tool plumbing (Fix #54)

### LINKED ISSUE #54: Implement the Tool State Machine
This is a finite state machine with various states and transitions, see the diagram below for an example. Each transition also represents a "hint" that shows up in the status bar footer.

It is important to think of the tool menu as, conceptually, a mode selector for the behavior of editing within the viewport. You perform operations on the viewport using your mouse and keyboard, and those operations change the node graph for you. Depending on which mode you're in, different operations are applied. That mode is chosen by selecting the active tool in the toolbar, or you could think of it as the "tool mode" in the "tool mode selector bar".

Each of these "modes" will be given its own FSM definition designed to model the exact behavior for the user experience of the tool.

The state machines will probably be pretty gnarly, it is way easier to work with them in a format resembling the Nomnoml diagram code (see below) than it is to figure out all that in Rust code. So the FSM for each tool will be defined as a text file. This probably makes extensions easier to support in the future also. We will need to design a format grammar and write a parser to construct the internal FSM data structure, which is implemented as:

- Each tool has a HashMap to store all the `ToolState`s (orange and gold states in the graph).
- A `ToolState` stores: `name: String; isIntermediate: boolean; transitions: Vec<Transition>; action: a closure or a string name to look up the closure in a HashMap;`
- A `Transition` stores: `condition: TransitionCondition; destination: string` (each `Transition` is owned by its outgoing state within the `Vec`)
- The "destination" string is the name of another `ToolState`. A `TransitionCondition` is an enum with all the possible input events and stateful details about the tool options, editor, and document (examples: mouse move event, mouse button down and up event for each mouse button, key down and up events for every key), plus `ε` (an instant, automatic transition). Perhaps we need a concept of event (which gets triggered) and a persistent condition (like a tool setting or some stateful characteristic about the document or editor). Or instead, perhaps transitions are only inputs or `ε`, but they can be gated by persistent conditions. The array of `TransitionCondition`s will probably set the precedence.
- The state machine is hard coded to begin with the `Ready` state. There is no end state, the machine never halts. 

Example subset of the FSM for the Select tool:

[Nomnoml diagram link](https://nomnoml.com/#view/%23spacing%3A%20120%0A%23arrowSize%3A%200.5%0A%23leading%3A%201.5%0A%23gravity%3A%201%0A%23direction%3A%20down%0A%23bendSize%3A%200.2%0A%0A%23.state%3A%20fill%3Dorange%20stroke%3Dblack%20visual%3Droundrect%0A%23.intermediatestate%3A%20fill%3Dgold%20stroke%3Dblack%20visual%3Droundrect%0A%23.input%3A%20fill%3Dturquoise%20stroke%3Dblack%20visual%3Dinput%0A%0A%0A%5B%3Cstate%3E%20Ready%5D%0A%5B%3Cstate%3E%20Transform%20Selected%5D%0A%5B%3Cstate%3E%20Grab%20Selected%5D%0A%5B%3Cstate%3E%20Rotate%20Selected%5D%0A%5B%3Cstate%3E%20Scale%20Selected%5D%0A%5B%3Cstate%3E%20Select%20Object%5D%0A%5B%3Cstate%3E%20Select%20Object%20Grow%2FShrink%20Selection%5D%0A%5B%3Cstate%3E%20Select%20Area%5D%0A%5B%3Cstate%3E%20Select%20Area%20Grow%2FShrink%20Selection%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Up%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Right%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Down%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Left%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Up%20Big%20Increment%20Nudge%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Right%20Big%20Increment%20Nudge%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Down%20Big%20Increment%20Nudge%5D%0A%5B%3Cstate%3E%20Nudge%20Selected%20Left%20Big%20Increment%20Nudge%5D%0A%5B%3Cstate%3E%20Move%20Duplicate%5D%0A%5B%3Cstate%3E%20Duplicate%5D%0A%0A%5B%3Cstart%3E%20Activate%20tool%5D%20-%3E%20%5BReady%5D%0A%5BReady%5D%20-%20%5B%3Cinput%3E%20LMB%20Down.%5D%0A%5BLMB%20Down.%5D%20-%3E%20%5B%3Cintermediatestate%3E%20LMB%20Down%2C%20Pending.%5D%0A%20%20%20%20%5BLMB%20Down%2C%20Pending.%5D%20-%20%5B%3Cinput%3E%20LMB%20Up.%5D%0A%20%20%20%20%5BLMB%20Up.%5D%20-%3E%20%5BSelect%20Object%5D%0A%20%20%20%20%20%20%5BSelect%20Object%5D%20-%20%5B%3Cinput%3E%20%CE%B5.%5D%0A%20%20%20%20%20%20%5B%CE%B5.%5D%20-%3E%20%5BReady%5D%0A%20%20%20%20%5BLMB%20Down%2C%20Pending.%5D%20-%20%5B%3Cinput%3E%20MouseMove.%5D%0A%20%20%20%20%5BMouseMove.%5D%20-%3E%20%5BTransform%20Selected%5D%0A%20%20%20%20%09%5BTransform%20Selected%5D%20-%20%5B%3Cinput%3E%20LMB%20Up..%5D%0A%20%20%20%20%20%20%20%20%5BLMB%20Up..%5D%20-%3E%20%5BReady%5D%0A%20%20%20%20%20%20%20%20%5BTransform%20Selected%5D%20-%20%5B%3Cinput%3E%20MouseMove..%5D%0A%20%20%20%20%20%20%20%20%5BMouseMove..%5D%20-%3E%20%5BTransform%20Selected%5D)

Orange is a main state that shows up in the footer of the Graphite editor as an input hint. Yellow/gold is a "intermediate" state that doesn't show up in the footer, but instead the result of the state it leads to shows up. (You can get to Transform Selected through a LMB Down combined with a MouseMove, for example.)

Footer

geom3trik:
- Added FSM trait
- Added FSM to select and ellipse tools
- Moved circle stamping code to ellipse tool

### DIFF
--- a/core/document/src/operation.rs
+++ b/core/document/src/operation.rs
@@ -1,3 +1,3 @@
 pub enum Operation {
-	AddCircle((f64, f64), f64),
+	AddCircle { cx: f64, cy: f64, r: f64 },
 }
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -38,13 +38,8 @@ impl Dispatcher {
 			Event::MouseMove(pos) => {
 				editor_state.tool_state.mouse_state.position = *pos;
 			}
-			Event::ModifierKeyDown(mod_keys) => {
-				editor_state.tool_state.mod_keys = *mod_keys;
-			}
-			Event::ModifierKeyUp(mod_keys) => {
-				editor_state.tool_state.mod_keys = *mod_keys;
-			}
-			Event::KeyPress(key) => todo!(),
+			Event::KeyUp(key) => todo!(),
+			Event::KeyDown(key) => todo!(),
 		}
 
 		let (responses, operations) = editor_state.tool_state.active_tool()?.handle_input(event, &editor_state.document);
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -1,5 +1,5 @@
 pub mod events;
-use crate::{Color, Document, EditorError, ToolState};
+use crate::{Color, Document, EditorError, EditorState};
 use document_core::Operation;
 use events::{Event, Response};
 
@@ -9,47 +9,47 @@ pub struct Dispatcher {
 }
 
 impl Dispatcher {
-	pub fn handle_event(&self, tool_state: &mut ToolState, document: &mut Document, event: &Event) -> Result<(), EditorError> {
+	pub fn handle_event(&self, editor_state: &mut EditorState, event: &Event) -> Result<(), EditorError> {
 		log::trace!("{:?}", event);
 
 		match event {
 			Event::SelectTool(tool_type) => {
-				tool_state.active_tool_type = *tool_type;
+				editor_state.tool_state.active_tool_type = *tool_type;
 			}
 			Event::SelectPrimaryColor(color) => {
-				tool_state.primary_color = *color;
+				editor_state.tool_state.primary_color = *color;
 			}
 			Event::SelectSecondaryColor(color) => {
-				tool_state.secondary_color = *color;
+				editor_state.tool_state.secondary_color = *color;
 			}
 			Event::SwapColors => {
-				std::mem::swap(&mut tool_state.primary_color, &mut tool_state.secondary_color);
+				std::mem::swap(&mut editor_state.tool_state.primary_color, &mut editor_state.tool_state.secondary_color);
 			}
 			Event::ResetColors => {
-				tool_state.primary_color = Color::BLACK;
-				tool_state.secondary_color = Color::WHITE;
+				editor_state.tool_state.primary_color = Color::BLACK;
+				editor_state.tool_state.secondary_color = Color::WHITE;
 			}
 			Event::MouseDown(mouse_state) => {
-				tool_state.mouse_state = *mouse_state;
+				editor_state.tool_state.mouse_state = *mouse_state;
 			}
 			Event::MouseUp(mouse_state) => {
-				tool_state.mouse_state = *mouse_state;
+				editor_state.tool_state.mouse_state = *mouse_state;
 			}
-			Event::MouseMovement(pos) => {
-				tool_state.mouse_state.position = *pos;
+			Event::MouseMove(pos) => {
+				editor_state.tool_state.mouse_state.position = *pos;
 			}
 			Event::ModifierKeyDown(mod_keys) => {
-				tool_state.mod_keys = *mod_keys;
+				editor_state.tool_state.mod_keys = *mod_keys;
 			}
 			Event::ModifierKeyUp(mod_keys) => {
-				tool_state.mod_keys = *mod_keys;
+				editor_state.tool_state.mod_keys = *mod_keys;
 			}
 			Event::KeyPress(key) => todo!(),
 		}
 
-		let (responses, operations) = tool_state.active_tool()?.handle_input(event, document);
+		let (responses, operations) = editor_state.tool_state.active_tool()?.handle_input(event, &editor_state.document);
 
-		self.dispatch_operations(document, &operations);
+		self.dispatch_operations(&mut editor_state.document, &operations);
 		// TODO - Dispatch Responses
 
 		Ok(())
@@ -63,20 +63,21 @@ impl Dispatcher {
 
 	fn dispatch_operation(&self, document: &mut Document, operation: &Operation) {
 		document.handle_operation(operation, |svg: String| {
-			self.dispatch_response(&Response::UpdateCanvas { document: svg });
+			self.dispatch_response(Response::UpdateCanvas { document: svg });
 		});
 	}
 
 	pub fn dispatch_responses(&self, responses: &[Response]) {
 		for response in responses {
-			self.dispatch_response(response);
+			// TODO - Remove clone when Response is Copy
+			self.dispatch_response(response.clone());
 		}
 	}
 
-	pub fn dispatch_response(&self, response: &Response) {
+	pub fn dispatch_response(&self, response: Response) {
 		let func = &self.callback;
 		// TODO - Remove clone if possible
-		func(response.clone())
+		func(response)
 	}
 
 	pub fn new(callback: Callback) -> Dispatcher {
--- a/core/editor/src/dispatcher/mod.rs
+++ b/core/editor/src/dispatcher/mod.rs
@@ -1,6 +1,6 @@
 pub mod events;
-use crate::{Color, EditorError, EditorState};
-use document_core::{Circle, Point, SvgElement};
+use crate::{Color, Document, EditorError, ToolState};
+use document_core::Operation;
 use events::{Event, Response};
 
 pub type Callback = Box<dyn Fn(Response)>;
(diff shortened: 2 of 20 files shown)


core/editor/src/tools/ellipse.rs
```diff
@@ -1,12 +1,54 @@
...
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum EllipseToolState {
+	Ready,
+	LmbDown,
+	TransformSelected,
+}
+
+impl Default for EllipseToolState {
```
Keavon, line 29: Can we make this happen automatically for every tool by deriving a shared trait? `Ready` should always be the default trait, preferably without having to specify it on each tool. Or maybe we can pass a function or macro with a list of states and it adds the "Ready" state automatically.

core/editor/src/tools/ellipse.rs
```diff
@@ -1,12 +1,54 @@
...
-pub struct Ellipse;
+pub struct Ellipse {
+	state: EllipseToolState,
+}
 
 impl Tool for Ellipse {
-	fn handle_input(&mut self, event: Event) -> Vec<Operation> {
-		todo!();
+	fn handle_input(&mut self, event: &Event, document: &Document) -> (Vec<Response>, Vec<Operation>) {
+		let mut responses = Vec::new();
```
Keavon, line 14: I don't love the idea of returning data to the caller of `.transition()` by passing in mutable `Vec`s. Feels way too C-like. It's probably better that we just return a 3-tuple from `.transition()` with the two `Vec`s (transferring ownership to the caller) plus the current return value. We can destructure them in the caller.

core/editor/src/tools/mod.rs
```diff
@@ -9,14 +9,19 @@ mod sample;
...
 use document_core::Operation;
 use std::collections::HashMap;
 
 pub trait Tool {
-	fn handle_input(&mut self, event: Event) -> Vec<Operation>;
+	fn handle_input(&mut self, event: &Event, document: &Document) -> (Vec<Response>, Vec<Operation>);
+}
+
+pub trait Fsm {
+	fn transition(self, event: &Event, document: &Document, responses: &mut Vec<Response>, operations: &mut Vec<Operation>) -> Self;
```
Keavon, line 24: I don't love the idea of returning data to the caller of `.transition()` by passing in mutable `Vec`s. Feels way too C-like. It's probably better that we just return a 3-tuple from `.transition()` with the two `Vec`s (transferring ownership to the caller) plus the current return value. We can destructure them in the caller.

core/editor/src/tools/select.rs
```diff
@@ -1,40 +1,51 @@
...
+
+			(SelectToolState::LmbDown, Event::MouseUp(mouse_state)) if mouse_state.mouse_keys.contains(MouseKeys::LEFT) => SelectToolState::Ready,
+
+			(SelectToolState::LmbDown, Event::MouseMovement(mouse_state)) => SelectToolState::TransformSelected,
+
+			(SelectToolState::TransformSelected, Event::MouseMovement(mouse_state)) => SelectToolState::TransformSelected,
+
+			(SelectToolState::TransformSelected, Event::MouseUp(mouse_state)) if mouse_state.mouse_keys.contains(MouseKeys::LEFT) => SelectToolState::Ready,
+
+			(state, _) => state,
```
Keavon, line 48: It's probably nicer to use `(_, _) => self,`

core/editor/src/tools/select.rs
```diff
@@ -1,40 +1,51 @@
...
+}
+
+impl Fsm for SelectToolState {
+	fn transition(self, event: &Event, document: &Document, responses: &mut Vec<Response>, operations: &mut Vec<Operation>) -> Self {
+		match (self, event) {
+			(SelectToolState::Ready, Event::MouseDown(mouse_state)) if mouse_state.mouse_keys.contains(MouseKeys::LEFT) => SelectToolState::LmbDown,
+
+			(SelectToolState::LmbDown, Event::MouseUp(mouse_state)) if mouse_state.mouse_keys.contains(MouseKeys::LEFT) => SelectToolState::Ready,
+
+			(SelectToolState::LmbDown, Event::MouseMovement(mouse_state)) => SelectToolState::TransformSelected,
```
Keavon, line 42: Let's make this line return `self` to make it clearer on reading it that it loops back to itself.

core/editor/src/tools/ellipse.rs
```diff
@@ -1,12 +1,54 @@
...
+				operations.push(Operation::AddCircle {
+					cx: mouse_state.position.x as f64,
+					cy: mouse_state.position.y as f64,
+					r: 10.0,
+				});
+
+				EllipseToolState::Ready
+			}
+
+			(state, _) => state,
```
Keavon, line 51: It's probably nicer to use `(_, _) => self,`

core/editor/src/dispatcher/mod.rs
```diff
@@ -9,83 +9,74 @@ pub struct Dispatcher {
...
-					radius: 10.0,
-				}));
-				self.emit_response(Response::UpdateCanvas { document: state.document.render() });
-
-				state.tool_state.active_tool()?.handle_input(event);
-
-				Ok(())
+				tool_state.mouse_state = *mouse_state;
 			}
 			Event::MouseMovement(pos) => {
```
Keavon, line 38: Let's rename this `MouseMove`.

core/editor/src/dispatcher/mod.rs
```diff
@@ -9,83 +9,74 @@ pub struct Dispatcher {
...
+				tool_state.mod_keys = *mod_keys;
 			}
 			Event::ModifierKeyUp(mod_keys) => {
-				state.tool_state.mod_keys = mod_keys;
-				state.tool_state.active_tool()?.handle_input(event);
-
-				Ok(())
+				tool_state.mod_keys = *mod_keys;
 			}
 			Event::KeyPress(key) => todo!(),
```
Keavon, line 47: Let's turn this into `KeyDown` and `KeyUp`. I'm also thinking we should get rid of explicit `ModifierKeyDown`/`Up` events but still allow them to be queryable as the current input state. We can address this in a followup PR though.

core/editor/src/tools/mod.rs
TrueDoctor: This might be a minor thing, but reusing vectors would yield better performance as you don't have to do heap allocations for every call because the maximum number of responses/operations for one tool should not fluctuate too much (I'd suspect something in the range of 1-10) so reusing a vec might be a good idea.

Keavon: Does this get reused? I thought it's called once per input event.

core/editor/src/tools/select.rs
Keavon: Or just `_ => self,`

geom3trik: I'm not sure that will work because the match statement consumes self, so it will throw an error about trying to use self after it's been moved. I'll give it a go thought and see.

core/document/src/operation.rs
```diff
@@ -1,3 +1,3 @@
 pub enum Operation {
-	AddCircle((f64, f64), f64),
+	AddCircle { cx: f64, cy: f64, r: f64 },
```
T0mstone, line 2: Is this a dependency thing or why can't we just use `AddCircle(kurbo::Circle)`?

core/editor/src/dispatcher/mod.rs
```diff
@@ -9,83 +9,74 @@ pub struct Dispatcher {
 }
 
 impl Dispatcher {
-	pub fn handle_event(&self, state: &mut EditorState, event: Event) -> Result<(), EditorError> {
+	pub fn handle_event(&self, tool_state: &mut ToolState, document: &mut Document, event: &Event) -> Result<(), EditorError> {
```
T0mstone, line 12: why is this necessary? passing the whole editor state seems more natural to me and is also more extensible.

core/editor/src/dispatcher/mod.rs
```diff
@@ -9,83 +9,74 @@ pub struct Dispatcher {
...
+			self.dispatch_response(response);
+		}
 	}
 
-	pub fn emit_response(&self, response: Response) {
+	pub fn dispatch_response(&self, response: &Response) {
 		let func = &self.callback;
-		func(response)
+		// TODO - Remove clone if possible
+		func(response.clone())
```
T0mstone, line 79: This is bad API design. Instead, you want to keep the function as taking a `Response` and let the caller of this function clone that if needed.

core/editor/src/tools/ellipse.rs
T0mstone: That's not as easy as it sounds since the `Ready` states are variants of different enums and the only thing they share is their name

core/editor/src/tools/ellipse.rs
T0mstone: In that case you'd use `_ => self`

core/editor/src/tools/ellipse.rs
```diff
@@ -1,12 +1,54 @@
...
+		EllipseToolState::Ready
+	}
+}
+
+impl Fsm for EllipseToolState {
+	fn transition(self, event: &Event, document: &Document, responses: &mut Vec<Response>, operations: &mut Vec<Operation>) -> Self {
+		match (self, event) {
+			(EllipseToolState::Ready, Event::MouseDown(mouse_state)) if mouse_state.mouse_keys.contains(MouseKeys::LEFT) => EllipseToolState::LmbDown,
+
+			// TODO - Check for left mouse button
```
T0mstone, line 40: don't postpone a simple thing like this, just add the check analogous to the previous match arm

core/editor/src/tools/select.rs
T0mstone: That isn't the case since `_` doesn't capture the argument

T0mstone (dismissed):

core/editor/src/tools/mod.rs
TrueDoctor: You could store a reusable Vec in the tool definition which we currently don't do. At the moment it does not make afaik any difference, but passing the vec by ref could in the fufure provide better performance, just thought I'd mention that while we are defining traits to avoid having to change things down the line

core/editor/src/tools/ellipse.rs
TrueDoctor: That would not work, because the mouse state contains the currently pressed buttons and upon releasing the button, the button in question is no longer pressed. We would have to compute the difference between this and the last mouse state to get the pressed button which would likely happen in the dispatcher. We could also add individual events for LMBDown, LMBUP, etc.

T0mstone: Ok, I didn't think about that. I think the latter makes the most sense but that can also be done in a later PR

core/document/src/operation.rs
TrueDoctor: We could debate on whether the use of kurbo should be an implementation detail, because we might replace kurbo with a more fully featured svg lib

core/editor/src/tools/ellipse.rs
Keavon: Ah, fair point. Perhaps we can optimize the cleanliness of this in the future but this should be a reasonable approach for now.

core/editor/src/dispatcher/mod.rs
Keavon: @geom3trik was this resolved yet? Feel free to hit "Resolve Conversation" if the button exists for you.

Keavon (approved):

PR #65, opened 2021-04-07
Add key handling

TrueDoctor:


Keavon (approved):

maroider:
Due to edge cases like keyboard-layout-changing keyboard shortcuts and [dead keys](https://www.w3.org/TR/uievents/#keys-dead), key presses and releases should (probably) be associated based on the `code` rather than the `key`.

TrueDoctor:
> Due to edge cases like keyboard-layout-changing keyboard shortcuts and [dead keys](https://www.w3.org/TR/uievents/#keys-dead), key presses and releases should (probably) be associated based on the `code` rather than the `key`.

For now we wll just use the logical keys to avoid having to handle different keyboard layouts

PR #66, opened 2021-04-08
Populate tool shelf and tool options bar

Keavon:
The "active" state is currently statically set to the Select tool. You can use the keyboard shortcuts <kbd>V</kbd> (for the Select Tool, which currently has no function), <kbd>M</kbd> (for the Rectangle Tool, which draws rectangles), and <kbd>E</kbd> (for the Ellipse Tool, which currently draws circles). However the active tool won't show up in the tool shelf.

Screenshot


client/web/src/components/widgets/IconButton.vue
```diff
@@ -0,0 +1,64 @@
...
+<script lang="ts">
+import { defineComponent } from "vue";
+
+export default defineComponent({
+	props: {
+		size: { type: Number, required: true },
+		gapAfter: { type: Boolean, default: false },
+	},
+});
+</script>
```
TrueDoctor, line 64: how hard would it be to add some functionality to the button?
basically calling the switch tool function with the tool name and maybe even changing the highlighted tool

TrueDoctor (approved):
But as far as I can tell all of the tools look good and the css look good as well. We can do everything else in a separate pr

T0mstone:
I don't have anything to add.

client/web/src/components/widgets/IconButton.vue
Keavon: Slightly hard enough that I'm saving it for another commit. You're more than welcome to tackle it, or I can try and do it this weekend.

PR #67, opened 2021-04-10
Polish a few things

T0mstone:
This PR contains two minor changes that don't change any behaviour.

Regarding the added dependency, `thiserror` is written by [dtolnay](https://github.com/dtolnay) and adds no transitive dependencies, so it should be fine.


TrueDoctor (approved):
Looks good! even though it ironically made the code twice as long xD

Issue #89, opened 2021-05-01
Change Ellipse Tool to work like a standard ellipse tool (not just circles)


Keavon:
This should be relatively easy for a newcomer or someone who has a few extra minutes. The Ellipse Tool currently only draws circles from the center. Instead, it should draw ellipses from the corner by default. If you want to constrain it to a 1:1 aspect ratio (a circle), you hold the <kbd>Shift</kbd> modifier key. If you want to have the circle originate from the center instead of the corner, you hold the <kbd>Alt</kbd> modifier key.

These modifier keys can also be extended to the other shape drawing tools.

PR #98, opened 2021-05-03
Improve Ellipse Tool

pkupper:
Implement Ellipse tool as specified in #89.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/98)

### DIFF
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -193,7 +193,21 @@ impl Document {
 		self.work_operations.push(operation.clone());
 		let responses = match operation {
 			Operation::AddCircle { path, insert_index, cx, cy, r, style } => {
-				self.add_layer(&path, Layer::new(LayerDataTypes::Circle(layers::Circle::new(kurbo::Point::new(cx, cy), r, style))), insert_index)?;
+				self.add_layer(&path, Layer::new(LayerDataTypes::Circle(layers::Circle::new((cx, cy), r, style))), insert_index)?;
+
+				Some(vec![DocumentResponse::DocumentChanged])
+			}
+			Operation::AddEllipse {
+				path,
+				insert_index,
+				cx,
+				cy,
+				rx,
+				ry,
+				rot,
+				style,
+			} => {
+				self.add_layer(&path, Layer::new(LayerDataTypes::Ellipse(layers::Ellipse::new((cx, cy), (rx, ry), rot, style))), insert_index)?;
 
 				Some(vec![DocumentResponse::DocumentChanged])
 			}
@@ -206,11 +220,7 @@ impl Document {
 				y1,
 				style,
 			} => {
-				self.add_layer(
-					&path,
-					Layer::new(LayerDataTypes::Rect(Rect::new(kurbo::Point::new(x0, y0), kurbo::Point::new(x1, y1), style))),
-					insert_index,
-				)?;
+				self.add_layer(&path, Layer::new(LayerDataTypes::Rect(Rect::new((x0, y0), (x1, y1), style))), insert_index)?;
 
 				Some(vec![DocumentResponse::DocumentChanged])
 			}
@@ -223,11 +233,7 @@ impl Document {
 				y1,
 				style,
 			} => {
-				self.add_layer(
-					&path,
-					Layer::new(LayerDataTypes::Line(Line::new(kurbo::Point::new(x0, y0), kurbo::Point::new(x1, y1), style))),
-					insert_index,
-				)?;
+				self.add_layer(&path, Layer::new(LayerDataTypes::Line(Line::new((x0, y0), (x1, y1), style))), insert_index)?;
 
 				Some(vec![DocumentResponse::DocumentChanged])
 			}
@@ -247,7 +253,7 @@ impl Document {
 				sides,
 				style,
 			} => {
-				let s = Shape::new(kurbo::Point::new(x0, y0), kurbo::Vec2 { x: x0 - x1, y: y0 - y1 }, sides, style);
+				let s = Shape::new((x0, y0), (x0 - x1, y0 - y1), sides, style);
 				self.add_layer(&path, Layer::new(LayerDataTypes::Shape(s)), insert_index)?;
 
 				Some(vec![DocumentResponse::DocumentChanged])
--- a/core/document/src/layers/ellipse.rs
+++ b/core/document/src/layers/ellipse.rs
@@ -1,0 +1,37 @@
+use super::style;
+use super::LayerData;
+
+use std::fmt::Write;
+
+#[derive(Debug, Clone, Copy, PartialEq, Default)]
+pub struct Ellipse {
+	shape: kurbo::Ellipse,
+	style: style::PathStyle,
+}
+
+impl Ellipse {
+	pub fn new(center: impl Into<kurbo::Point>, radii: impl Into<kurbo::Vec2>, rotation: f64, style: style::PathStyle) -> Ellipse {
+		Ellipse {
+			shape: kurbo::Ellipse::new(center, radii, rotation),
+			style,
+		}
+	}
+}
+
+impl LayerData for Ellipse {
+	fn render(&mut self, svg: &mut String) {
+		let kurbo::Vec2 { x: rx, y: ry } = self.shape.radii();
+		let kurbo::Point { x: cx, y: cy } = self.shape.center();
+
+		let _ = write!(
+			svg,
+			r#"<ellipse cx="0" cy="0" rx="{}" ry="{}" transform="translate({} {}) rotate({})" {} />"#,
+			rx,
+			ry,
+			cx,
+			cy,
+			self.shape.rotation().to_degrees(),
+			self.style.render(),
+		);
+	}
+}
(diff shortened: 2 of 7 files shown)


core/document/src/layers/ellipse.rs
```diff
@@ -0,0 +1,34 @@
...
+	fn render(&mut self, svg: &mut String) {
+		let _ = write!(
+			svg,
+			r#"<ellipse cx="0" cy="0" rx="{}" ry="{}" transform="translate({} {}) rotate({})" {} />"#,
+			self.shape.radii().x,
+			self.shape.radii().y,
+			self.shape.center().x,
+			self.shape.center().y,
+			self.shape.rotation().to_degrees(),
+			self.style.render(),
```
T0mstone, line 31: This would be a lot more readable with named format arguments:
```
… rx="{rx}" ry="{ry}" transform="translate({cx} {cy}) rotate({rot})" {}/>"#, rx=self.shape.radii().x, ry=self.shape.radii().y, …
```
also the multiple calls to `radii` and `center` should be factored out. You can even do it like this:
```
let kurbo::Vec2 { x: rx, y: ry } = self.shape.radii();
let kurbo::Point { x: cx, y: cy } = self.shape.center();
```
note that format args don't have shorthand syntax afaik, so you still have to write `cx=cx` etc.

core/document/src/document.rs
```diff
@@ -197,6 +197,24 @@ impl Document {
...
+				cx,
+				cy,
+				rx,
+				ry,
+				rot,
+				style,
+			} => {
+				self.add_layer(
+					&path,
+					Layer::new(LayerDataTypes::Ellipse(layers::Ellipse::new(kurbo::Point::new(cx, cy), kurbo::Vec2::new(rx, ry), rot, style))),
```
T0mstone, line 212: since `layers::Ellipse::new` uses `impl Into<kurbo::{Point, Vec2}>`, this code could be made simpler by omitting those `new` calls in favor of `(f64, f64)`: 
```
layers::Ellipse::new((cx, cy), (rx, ry), …)
```
This change should then also be made in the other match arms.
If not, we should remove the `impl Into` from those `new` function's arguments.
What do you think, @Keavon?

core/editor/src/tools/ellipse.rs
```diff
@@ -58,37 +62,112 @@ impl Fsm for EllipseToolFsmState {
...
 	}
 }
+
+fn make_operation(data: &EllipseToolData, tool_data: &DocumentToolData) -> Operation {
+	let x0 = data.drag_start.x as f64;
+	let y0 = data.drag_start.y as f64;
+	let x1 = data.drag_current.x as f64;
+	let y1 = data.drag_current.y as f64;
+
+	let (cx, cy) = if data.center_around_cursor { (x0, y0) } else { ((x0 + x1) * 0.5, (y0 + y1) * 0.5) };
```
T0mstone, line 138: Is there a reason this uses `* 0.5` instead of the conceptually simpler `/ 2.0`?

core/editor/src/tools/ellipse.rs
```diff
@@ -58,37 +62,112 @@ impl Fsm for EllipseToolFsmState {
...
+	let (mut rx, mut ry) = if data.constrain_to_circle {
+		let r = f64::max((x1 - x0).abs(), (y1 - y0).abs());
+		(r, r)
+	} else {
+		((x1 - x0).abs(), (y1 - y0).abs())
+	};
+
+	if !data.center_around_cursor {
+		rx *= 0.5;
+		ry *= 0.5;
```
T0mstone, line 149: same as above: why not use `/= 2.0`?

core/editor/src/tools/ellipse.rs
```diff
@@ -58,37 +62,112 @@ impl Fsm for EllipseToolFsmState {
...
+fn make_operation(data: &EllipseToolData, tool_data: &DocumentToolData) -> Operation {
+	let x0 = data.drag_start.x as f64;
+	let y0 = data.drag_start.y as f64;
+	let x1 = data.drag_current.x as f64;
+	let y1 = data.drag_current.y as f64;
+
+	let (cx, cy) = if data.center_around_cursor { (x0, y0) } else { ((x0 + x1) * 0.5, (y0 + y1) * 0.5) };
+
+	let (mut rx, mut ry) = if data.constrain_to_circle {
+		let r = f64::max((x1 - x0).abs(), (y1 - y0).abs());
```
T0mstone, line 141: maybe this should use the min instead?

core/editor/src/tools/ellipse.rs
```diff
@@ -58,37 +62,112 @@ impl Fsm for EllipseToolFsmState {
...
+			path: vec![],
+			insert_index: -1,
+			cx,
+			cy,
+			rx,
+			ry,
+			rot: 0.0,
+			style: style::PathStyle::new(None, Some(style::Fill::new(tool_data.primary_color))),
+		}
+	}
```
T0mstone, line 172: checking `data.constrain_to_circle` twice seems kind of clunky to me.
Maybe this `if` should include the radius calculating code...

core/editor/src/tools/rectangle.rs
```diff
@@ -72,8 +72,6 @@ impl Fsm for RectangleToolFsmState {
 				RectangleToolFsmState::LmbDown
 			}
 			(RectangleToolFsmState::LmbDown, Event::LmbUp(mouse_state)) => {
-				let r = data.drag_start.distance(&mouse_state.position);
-				log::info!("draw rectangle with radius: {:.2}", r);
```
T0mstone, line 76: why is this included in this PR? it has nothing to do with it

T0mstone:
oh and sorry for sounding a bit harsh at times

core/document/src/document.rs
Keavon: The tuple form makes it shorter and easier to read.

core/editor/src/tools/ellipse.rs
Keavon: Do they get optimized into the same assembly? I've always understood that multiplication is faster than division. To my eyes, they look equally readable. I wouldn't bother changing it.

core/editor/src/tools/ellipse.rs
Keavon: Author's choice, see my comment above.

core/editor/src/tools/ellipse.rs
T0mstone: Yes they do: https://godbolt.org/z/1rso37Tqv

core/editor/src/tools/ellipse.rs
T0mstone: It's not equivalent though.`max` means the radius will be the larger of the two side lengths,
`min` means the shorter. I would think the shorter to be better since then the circle fits into the drawn rectangle and also since affinity does it like that iirc

core/editor/src/tools/ellipse.rs
T0mstone: It is [different assermbly](https://godbolt.org/z/1rso37Tqv) but both only a single instruction and I'm not qualified to talk about the performance difference

core/editor/src/tools/ellipse.rs
T0mstone: and I think you meant to reply to the previous (up by one) comment

Keavon: Sorry, somehow this comment ended up in the wrong thread. I was replying in reference to `/= 2.0;`

core/editor/src/tools/ellipse.rs
Keavon: Author's choice, see my comment above.

Keavon (approved):

core/editor/src/tools/ellipse.rs
Keavon: Food for thought: https://stackoverflow.com/questions/4125033/floating-point-division-vs-floating-point-multiplication

TrueDoctor: why not just use
```rust
#[repr(C)]
union AwesomeFloat {
    u: u32,
    f: f32,
}
let fl = AwesomeFloat { f: 3.141};
unsafe { fl.u += 1<<24};
unsafe { fl.u -= 1<<24};
```
while we're already at it?

core/document/src/layers/ellipse.rs
pkupper: It looks like format arguments requires the unstable feature `format_args_capture`. I'll factor out the `radii`and `center` calls but leave the arguments for now.

core/editor/src/tools/ellipse.rs
pkupper: GIMP uses max for their ellipse tool. That way, the starting and ending cursor positions are both on the bounding square of the circle. This will feel more intuitive when overlays are in place. I would however be fine with both.

core/editor/src/tools/ellipse.rs
TrueDoctor: The performance of / 2.0 is most likely worse, but the assembly does not differ when compiled with optimizations.
I'd say the slight difference in debug mode is most likely completely negligible

PR #103, opened 2021-05-04
Change circle radius calculation

T0mstone:
I "reverse engineered" (basically just looked at) how Affinity Photo handles this calculation and implemented it here.

IMO this feels really nice and we should at least make it an option

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/103)


T0mstone:
Moved to #104

PR #105, opened 2021-05-04
Fix ellipse tool behavior

pkupper:
Previously, the corner of the Circle's bounding box was not fixed to the starting location, when holding Shift. This PR changes this to the expected behavior. In addition the Circle now expands to the cursor location as expected when holding both Shift and Alt at the same time.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/105)

### DIFF
--- a/core/editor/src/tools/ellipse.rs
+++ b/core/editor/src/tools/ellipse.rs
@@ -135,10 +135,14 @@ fn make_operation(data: &EllipseToolData, tool_data: &DocumentToolData) -> Opera
 	let x1 = data.drag_current.x as f64;
 	let y1 = data.drag_current.y as f64;
 
-	let (cx, cy, r_scale) = if data.center_around_cursor { (x0, y0, 1.0) } else { ((x0 + x1) * 0.5, (y0 + y1) * 0.5, 0.5) };
-
 	if data.constrain_to_circle {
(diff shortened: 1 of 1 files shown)


core/editor/src/tools/ellipse.rs
```diff
@@ -135,10 +135,14 @@ fn make_operation(data: &EllipseToolData, tool_data: &DocumentToolData) -> Opera
 	let x1 = data.drag_current.x as f64;
 	let y1 = data.drag_current.y as f64;
 
-	let (cx, cy, r_scale) = if data.center_around_cursor { (x0, y0, 1.0) } else { ((x0 + x1) * 0.5, (y0 + y1) * 0.5, 0.5) };
-
 	if data.constrain_to_circle {
-		let r = f64::max((x1 - x0).abs(), (y1 - y0).abs()) * r_scale;
+		let (cx, cy, r) = if data.center_around_cursor {
+			(x0, y0, ((x1 - x0) * (x1 - x0) + (y1 - y0) * (y1 - y0)).sqrt())
```
T0mstone, line 140: please rewrite this using [`f64::hypot`](https://doc.rust-lang.org/std/primitive.f64.html#method.hypot)

pkupper: Done! Good to know this function exists.

T0mstone (approved):

PR #115, opened 2021-05-07
Fix "P" keybind

Keavon:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/115)


TrueDoctor (approved):
trival fix, looks good

Issue #118, opened 2021-05-08
Make the show/hide layer button work


Keavon:
Start with `core\editor\src\dispatcher\mod.rs:40` and plumb the state updates through the document layer system's visibility state. It needs to end up in the state encoded in a `LayerPanelEntry` in `core\document\src\response.rs:9` which gets sent through the WASM back to the frontend.

akshay1992kalbhor:
Hey, can I work on this?

Keavon:
Yes, this should be available. Feel free to ping me on Discord and I can answer some questions if you need any help, or through GitHub in this discussion if you don't use Discord. Thanks for volunteering!

TrueDoctor:
ou basically only have to call the already exposed wasm functions from vue, add an operation for it and handle to operation in the document.
This also has be added to the dispatcher

Keavon:
To clarify, this is actually already piped from the Vue through the WASM to the code in `core\editor\src\dispatcher\mod.rs:40`.

akshay1992kalbhor:
Thanks guys. I am still reading the code to understand what's going on. Your comments have also been helpful!

akshay1992kalbhor:
Should probably close this one. @Keavon

Keavon:
For future reference, you can write "closes #118" in your commit message and it'll close the associated issue for you.

PR #127, opened 2021-05-16
Make the show/hide layer buttons work (#118)

akshay1992kalbhor:
@Keavon @TrueDoctor Any suggestions/changes?

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/127)


core/document/src/document.rs
```diff
@@ -300,6 +300,14 @@ impl Document {
 				let children = self.layer_panel(path.as_slice())?;
 				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::ExpandFolder { path, children }])
 			}
+			Operation::ToggleVisibility { path } => {
+				let _ = self.layer_mut(&path).map(|x| {
```
Keavon, line 304: Please rename `x` to `layer`.

core/document/src/document.rs
```diff
@@ -300,6 +300,14 @@ impl Document {
 				let children = self.layer_panel(path.as_slice())?;
 				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::ExpandFolder { path, children }])
 			}
+			Operation::ToggleVisibility { path } => {
```
Keavon, line 303: @TrueDoctor did we decide we generally want to do `set` or `toggle`?

TrueDoctor: For now toggle

akshay1992kalbhor:
I have made the necessary changes. Is there a way to update this PR with the changes? Or do I need to make a new one?

Keavon (approved):

core/document/src/document.rs
```diff
@@ -300,6 +300,14 @@ impl Document {
...
+				let _ = self.layer_mut(&path).map(|layer| {
+					layer.visible = !layer.visible;
+					layer.cache_dirty = true;
+				});
+				let children = self.layer_panel(&[])?;
```
TrueDoctor, line 308: you should either update the folder the current layer resides (for that you can use `children = self.layer_panel(path.as_slice()[..path.len()-1])?`)
or you need to add a separate response for the vis toggle

akshay1992kalbhor: Should `children = self.layer_panel(path.as_slice()[..path.len()-1])?` replace line 308?

TrueDoctor: you still need the `let` but yeah, you can try and see if it works

Keavon:
Closes #118

PR #129, opened 2021-05-18
Popovers now close when clicked out or mouse strays 100px outside bounds

Keavon:
---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)

### DIFF
--- a/client/web/src/components/widgets/PopoverMount.vue
+++ b/client/web/src/components/widgets/PopoverMount.vue
@@ -76,8 +76,11 @@ import { defineComponent } from "vue";
 
 export default defineComponent({
 	components: {},
-	props: {
-		open: { type: Boolean, default: false },
+	data() {
+		return {
+			open: false,
+			mouseStillDown: false,
+		};
 	},
 	updated() {
 		const popoverContent = this.$refs.popoverContent as HTMLElement;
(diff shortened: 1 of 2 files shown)


client/web/src/components/widgets/PopoverMount.vue
```diff
@@ -91,5 +94,88 @@ export default defineComponent({
...
+		setClosed() {
+			this.open = false;
+		},
+		mouseMoveHandler(e: MouseEvent) {
+			const MOUSE_STRAY_DISTANCE = 100;
+
+			// Close the popover if the mouse has strayed far enough from its bounds
+			if (this.isMouseEventOutsidePopover(e, MOUSE_STRAY_DISTANCE)) {
+				this.setClosed();
+			}
```
bjorn3, line 110: I am not quite convinced that it should automatically close when you move the cursor out of the bounds of the popover.

Keavon: This is Blender's behavior, and I quite like it. It is on my radar as a topic for doing proper UX studies instead of riding with the assumption that Blender's choice was objectively correct. It does perhaps take some getting used to popovers (and dropdown menus) disappearing when your mouse leaves the vicinity, but it also provides a very convenient method of dismissing menus you aren't interacting with. We can also adjust the `MOUSE_STRAY_DISTANCE` from 100px to something different (Blender seems to use about 54 pixels per my counts, but I felt that was a little on the low side). We can also consider leaving more important popovers like the color picker up until a click or the Escape key dismisses them. Let's keep the code for now and study the UX question once we have users.

TrueDoctor: Should we maybe add a specific kind of comment/tag for ux preference stuff?
Basically that should be configurable in the future/impacts ux can be tagged to later easily search for all such occurrences. (Might be good to use a unified approach for both rust and vue)

TrueDoctor (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)* status: all files reviewed, 2 unresolved discussions (waiting on @bjorn3 and @Keavon)

___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2Wzv_3gwV8bTt_j_y:b-168dae) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
> ```
> 				// Track if the left mouse button is now down so its later click event can be canceled
> 				const eventIsForLmb = e.button === 0;
> ```
This line does not match the comment above, is that intentional?

Keavon (commented):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)* status: all files reviewed, 2 unresolved discussions (waiting on @bjorn3 and @TrueDoctor)

___
*[client/web/src/components/widgets/PopoverMount.vue, line 110 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-M_yHW9XK_80z5z4tHB2-r1-110:-Ma2ZfD04FCKfQ7roO1Z:b-z6rl5f) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L110)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

Should we maybe add a specific kind of comment/tag for ux preference stuff?
Basically that should be configurable in the future/impacts ux can be tagged to later easily search for all such occurrences. (Might be good to use a unified approach for both rust and vue)
</blockquote></details>

I'll think about that and act on it when it becomes more of a concern. I'm not worried about forgetting this particular variable in the mean time.
___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2_DnU4gob3SaDHEv5:bfungv) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

> ```
> 				// Track if the left mouse button is now down so its later click event can be canceled
> 				const eventIsForLmb = e.button === 0;
> ```
This line does not match the comment above, is that intentional?
</blockquote></details>

The comment is for the subsequent pair of lines, that's why they are part of the same contiguous block. Or did you mean that the comment is confusing and it was not initially obvious how it described that pair of lines? If so, I should reword it.

TrueDoctor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)* status: all files reviewed, 2 unresolved discussions (waiting on @bjorn3 and @Keavon)

___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2a-Ie7oIpuB35s2kR:budxwp7) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
<details><summary><i>Previously, Keavon (Keavon Chambers) wrote…</i></summary><blockquote>

The comment is for the subsequent pair of lines, that's why they are part of the same contiguous block. Or did you mean that the comment is confusing and it was not initially obvious how it described that pair of lines? If so, I should reword it.
</blockquote></details>

Comparing to 0 checks if no buttons are pressed and not if the lmb was not pressed

Keavon:
___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2aS5C86sJoyuht35-:b1mf2p5) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

Comparing to 0 checks if no buttons are pressed and not if the lmb was not pressed
</blockquote></details>

I see your confusion. I'll update the comment to clarify that the `.button` API defines 0 as LMB, which is different from the `.buttons` API which uses bit flags where 1 is LMB.

TrueDoctor (approved):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)* status: all files reviewed, 2 unresolved discussions (waiting on @bjorn3 and @Keavon)

___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2bMZy6c8LzkbBPrHR:b-ud2wdw) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
<details><summary><i>Previously, Keavon (Keavon Chambers) wrote…</i></summary><blockquote>

I see your confusion. I'll update the comment to clarify that the `.button` API defines 0 as LMB, which is different from the `.buttons` API which uses bit flags where 1 is LMB.
</blockquote></details>

ah, I see, my bad maybe some constants for the mouse buttons would be helpful but that is optional

TrueDoctor (approved):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/129)* status: all files reviewed, 1 unresolved discussion (waiting on @bjorn3)

Keavon:
___
*[client/web/src/components/widgets/PopoverMount.vue, line 128 at r1](https://reviewable.io/reviews/graphiteeditor/graphite/129#-Ma2Wzv_3gwV8bTt_j_x:-Ma2c5VD3hwMkfpR9Yvk:b-2bwdss) ([raw file](https://github.com/graphiteeditor/graphite/blob/d5a57761545d47f5cad06b1689dfe1ab9f3683a9/client/web/src/components/widgets/PopoverMount.vue#L128)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

ah, I see, my bad maybe some constants for the mouse buttons would be helpful but that is optional
</blockquote></details>

I'll make some clarifications on that in my upcoming PR involving further popover improvements, it's more expedient to merge this now.

Issue #100, opened 2021-05-04
Document scaling


Keavon:
We need a system of document units within our (currently) infinite viewport. Then you can use the scale input area or zoom buttons to adjust that zoom.

![](https://files.keavon.com/-/EntireViciousFairyfly/capture.png)

Tool actions need to be projected from viewport space to canvas space. We actually probably will want some standard functions specifically for projecting between coording spaces that can be used by lots of code. In the future, you'll also be able to rotate the viewport's display of the canvas.

And then the canvas space needs to be shown at the correct pan and zoom (and rotation) in the viewport.

PR #156, opened 2021-05-23
Add Canvas transform

0HyperCube:
- [x] Canvas position represents a point on the infinite canvas
- [x] Canvas transform represents the location of the infinite canvas relative to the viewport (i.e. the svg).
- [x] All tools convert the viewport positions to canvas positions and produce an operation with the positions in canvas space.
- [x] Operations produce an svg element with the correct location scale and rotation
- [x] Svg is rendered with the correct transform from the Rust code

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)


Keavon:
Associated issue: https://github.com/GraphiteEditor/Graphite/issues/100

client/web/wasm/src/document.rs
```diff
@@ -21,11 +21,19 @@ pub fn select_tool(tool: String) -> Result<(), JsValue> {
...
+	let ev = DocumentMessage::DocumentResize(ViewportPosition { x: new_width, y: new_height });
+	EDITOR_STATE.with(|editor| editor.borrow_mut().handle_message(ev)).map_err(convert_error)
+}
+
 // TODO: When a mouse button is down that started in the viewport, this should trigger even when the mouse is outside the viewport (or even the browser window if the browser supports it)
 /// Mouse movement within the screenspace bounds of the viewport
 #[wasm_bindgen]
 pub fn on_mouse_move(x: u32, y: u32) -> Result<(), JsValue> {
-	// TODO: Convert these screenspace viewport coordinates to canvas coordinates based on the current zoom and pan
+	// TODO: Convert these screenspace viewport coordinates to document coordinates based on the current zoom and pan
```
TrueDoctor, line 36: This comment can be removed

core/document/src/layers/circle.rs
```diff
@@ -6,13 +6,15 @@ use std::fmt::Write;
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
 pub struct Circle {
 	shape: kurbo::Circle,
+	rotation: f64,
 	style: style::PathStyle,
 }
 
 impl Circle {
-	pub fn new(center: impl Into<kurbo::Point>, radius: f64, style: style::PathStyle) -> Circle {
+	pub fn new(center: impl Into<kurbo::Point>, radius: f64, rotation: f64, style: style::PathStyle) -> Circle {
```
TrueDoctor, line 14: does a circle need a rotation?

core/editor/src/document/document_message_handler.rs
```diff
@@ -41,6 +42,9 @@ impl DocumentMessageHandler {
 	pub fn active_document(&self) -> &Document {
 		&self.documents[self.active_document]
 	}
+	pub fn mut_active_document(&mut self) -> &mut Document {
+		&mut self.documents[self.active_document]
+	}
 	pub fn active_document_mut(&mut self) -> &mut Document {
 		&mut self.documents[self.active_document]
 	}
```
TrueDoctor, line 50: duplicate fuction

core/editor/src/frontend/frontend_message_handler.rs
```diff
@@ -11,6 +11,7 @@ pub enum FrontendMessage {
 	ExpandFolder { path: Vec<LayerId>, children: Vec<LayerPanelEntry> },
 	SetActiveTool { tool_name: String },
 	UpdateCanvas { document: String },
+	UpdateDocumentTransform { transform: String },
```
TrueDoctor, line 14: What was the reasoning for  returning a string vs 4 floats? returning floats seems cleaner, but I might very well be wrong on that one.

core/editor/src/input/mouse.rs
```diff
@@ -1,5 +1,79 @@
...
+			x: self.x * cosine - self.y * sine,
+			y: self.x * sine + self.y * cosine,
+		}
+	}
+	pub fn add(&self, x: f64, y: f64) -> Self {
+		Self { x: self.x + x, y: self.y + y }
+	}
+	pub fn multiply(&self, x: f64) -> Self {
+		Self { x: self.x * x, y: self.y * x }
+	}
```
TrueDoctor, line 29: You should probably just implement the Add and Mul traits for DocumentPosition.

core/editor/src/input/mouse.rs
```diff
@@ -1,5 +1,79 @@
...
+}
+
+impl DocumentTransform {
+	pub fn transform_string(&self) -> String {
+		let inverse_scale = 1. / self.scale;
+		let size = DocumentPosition {
+			x: self.size.x as f64 / 2.,
+			y: self.size.y as f64 / 2.,
+		};
```
TrueDoctor, line 65: after adding the trait impls, this should become
```rust
let size = self.size * 0.5;
```

core/editor/src/tool/tools/ellipse.rs
```diff
@@ -58,7 +59,7 @@ struct EllipseToolData {
 impl Fsm for EllipseToolFsmState {
 	type ToolData = EllipseToolData;
 
-	fn transition(self, event: ToolMessage, _document: &SvgDocument, tool_data: &DocumentToolData, data: &mut Self::ToolData, input: &InputPreprocessor, responses: &mut VecDeque<Message>) -> Self {
+	fn transition(self, event: ToolMessage, document: &Document, tool_data: &DocumentToolData, data: &mut Self::ToolData, input: &InputPreprocessor, responses: &mut VecDeque<Message>) -> Self {
```
TrueDoctor, line 62: we could consider doing the conversion to doc coordinates in the EllipseMessageHandler and pass that along to reduce boilerplate code

core/editor/src/frontend/frontend_message_handler.rs
pkupper: I second this. Let's keep frontend messages frontend-agnostic. This string seems very SVG specific.

core/editor/src/input/mouse.rs
```diff
@@ -1,5 +1,79 @@
...
+			degrees: 45.,
+			size: ViewportPosition::default(),
+		}
+	}
+}
+
+impl DocumentTransform {
+	pub fn transform_string(&self) -> String {
+		let inverse_scale = 1. / self.scale;
+		let size = DocumentPosition {
```
pkupper, line 72: If this represents the center of the viewport in viewport-coordinates it should be a ViewportPositon.

core/editor/src/input/mouse.rs
```diff
@@ -1,5 +1,79 @@
...
+			x: self.size.x as f64 / 2.,
+			y: self.size.y as f64 / 2.,
+		};
+		let translation = self.location.multiply(-inverse_scale).rotate(-self.radians);
+		format!(
+			"translate({},{}) scale({}) rotate({})",
+			translation.x + size.x,
+			translation.y + size.y,
+			inverse_scale,
+			self.radians * -57.295779513, // 180 / pi
```
pkupper, line 82: Use `self.radians.to_degrees()`

core/editor/src/input/mouse.rs
```diff
@@ -13,6 +87,28 @@ impl ViewportPosition {
 		let y_diff = other.y as i64 - self.y as i64;
 		f64::sqrt((x_diff * x_diff + y_diff * y_diff) as f64)
 	}
+	pub fn to_document_position(&self, document_transform: &DocumentTransform, apply_rotation: bool) -> DocumentPosition {
```
pkupper, line 100: What's difference between using apply_rotation and not? Intuitively there should only be a single way to transform viewport coordinates to document coordinates. Both cases seem to do a rotation here.

0HyperCube (commented):
> Dismissed @TrueDoctor from a discussion.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 0 of 23 files reviewed, 6 unresolved discussions (waiting on @geom3trik, @pkupper, @T0mstone, and @TrueDoctor)

___
*[client/web/wasm/src/document.rs, line 36 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvg-r8-36:-MarDGGf-dcoX3Eo1WIv:b-896fix) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/client/web/wasm/src/document.rs#L36)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

This comment can be removed
</blockquote></details>

Done.
___
*[core/document/src/layers/circle.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Maj2iAoMqoeckuV3RA6-r8-14:-MarCo5CDUkJSaNiXrQQ:btsx2d8) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/document/src/layers/circle.rs#L14)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

does a circle need a rotation?
</blockquote></details>

No, I just put it in there so the transition function of `core/editor/src/tool/tools/ellipse.rs` would not have to handle both unrotated and rotated coordinates (that got very messy when I tried to do it, but you are welcome to try)
___
*[core/editor/src/document/document_message_handler.rs, line 50 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvi-r8-50:-MarCdi61sULiG2jhIsF:b-896fix) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/document/document_message_handler.rs#L50)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

duplicate fuction
</blockquote></details>

Done.
___
*[core/editor/src/frontend/frontend_message_handler.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvj-r8-14:-Mar4LanEPQaCzVh7Tr9:btfh11g) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/frontend/frontend_message_handler.rs#L14)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

What was the reasoning for  returning a string vs 4 floats? returning floats seems cleaner, but I might very well be wrong on that one.
</blockquote></details>

I did the converting to strings in rust because I'm not great at vue, but you could do it either way.
___
*[core/editor/src/input/mouse.rs, line 65 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-65:-MarC4HcDKoXB8ypDa-H:b-896fix) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L65)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

after adding the trait impls, this should become
```rust
let size = self.size * 0.5;
```
</blockquote></details>

Done.
___
*[core/editor/src/tool/tools/ellipse.rs, line 62 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvm-r8-62:-MarFrf0CCkhinlC_CT8:b-pzj98a) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/tool/tools/ellipse.rs#L62)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

we could consider doing the conversion to doc coordinates in the EllipseMessageHandler and pass that along to reduce boilerplate code
</blockquote></details>

I tried to do that but I'm very confused about how the viewport position gets into self.EllipseToolData

pkupper:
Some basic (temporary) keybindings to move/rotate the canvas would be nice to see that it works, as opposed to having a static transformation.

core/editor/src/input/mouse.rs
```diff
@@ -1,5 +1,79 @@
...
+		(item.x, item.y)
+	}
+}
+
+// The location of the viewport (or anything else) in the canvas
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct DocumentTransform {
+	pub location: DocumentPosition,
+	pub radians: f64,
+	pub degrees: f64,
```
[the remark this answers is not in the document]
Keavon: I'm inclined to prefer degrees throughout the application since degrees are probably more useful in 2D graphics than radians. I'm probably neglecting to consider something so please push back if you disagree.

core/document/src/document.rs
```diff
@@ -197,10 +205,14 @@ impl Document {
 				cy,
 				rx,
 				ry,
-				rot,
+				rotation,
```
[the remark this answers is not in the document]
Keavon: Right side up in the viewport, tilted relative to the canvas.

TrueDoctor (commented):
> Reviewed 4 of 8 files at r7, 5 of 12 files at r8, 2 of 3 files at r9.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 11 of 23 files reviewed, 9 unresolved discussions (waiting on @0HyperCube, @geom3trik, @pkupper, @T0mstone, and @TrueDoctor)

___
*[core/editor/src/input/mouse.rs, line 42 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-42:-MarJX1dASv7t8d0Pugz:bw9ks6v) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L42)):*
<details><summary><i>Previously, Keavon (Keavon Chambers) wrote…</i></summary><blockquote>

I'm inclined to prefer degrees throughout the application since degrees are probably more useful in 2D graphics than radians. I'm probably neglecting to consider something so please push back if you disagree.
</blockquote></details>

Usually all the math is done in radians and only user input/output is done in degrees

core/editor/src/input/mouse.rs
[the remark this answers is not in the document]
pkupper: Agree with @TrueDoctor s comment. For trigonometric functions and some other calculations radians usually are more natural  (lead to simpler equations), hence why sin() and cos() use radians. We can make sure however, that users only ever see degrees.

TrueDoctor (changes requested):
> Reviewed 1 of 14 files at r4.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 12 of 23 files reviewed, 12 unresolved discussions (waiting on @0HyperCube, @geom3trik, @Keavon, @pkupper, @T0mstone, and @TrueDoctor)

___
*[client/web/wasm/src/document.rs, line 28 at r10](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MarbjVk5hKfnrYt1NPN:-MarbjVk5hKfnrYt1NPO:b-sujk4h) ([raw file](https://github.com/graphiteeditor/graphite/blob/5f7f3da83bee64e978b27aba528d827aa432fa4c/client/web/wasm/src/document.rs#L28)):*
> ```XML
> #[wasm_bindgen]
> pub fn on_viewport_resize(new_width: u32, new_height: u32) -> Result<(), JsValue> {
> 	let ev = DocumentMessage::DocumentResize(ViewportPosition { x: new_width, y: new_height });
> ```

I still dislike the idea of document coming in contact with the viewport dimensions directly, I feel like we should send this to the ipp first (sorry if my opinion on that matter is still a bit half baked)
___
*[core/document/src/layers/circle.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Maj2iAoMqoeckuV3RA6-r8-14:-MarcGHV7uuisPJoXwF7:b2pvn2z) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/document/src/layers/circle.rs#L14)):*
<details><summary><i>Previously, 0HyperCube wrote…</i></summary><blockquote>

No, I just put it in there so the transition function of `core/editor/src/tool/tools/ellipse.rs` would not have to handle both unrotated and rotated coordinates (that got very messy when I tried to do it, but you are welcome to try)
</blockquote></details>

I'll try it out
___
*[core/editor/src/input/mouse.rs, line 100 at r10](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Marju3_Aj2i0cx_Y7m4:-Marju3a5lBi3XU_iNp2:bzcelqh) ([raw file](https://github.com/graphiteeditor/graphite/blob/5f7f3da83bee64e978b27aba528d827aa432fa4c/core/editor/src/input/mouse.rs#L100)):*
> ```XML
> 		f64::sqrt((x_diff * x_diff + y_diff * y_diff) as f64)
> 	}
> 	pub fn to_document_position(&self, document_transform: &DocumentTransform, apply_rotation: bool) -> DocumentPosition {
> ```

I think it would be cleaner to do the reverse implementation and provide a from viewport position for documentPosition as a function on document. having to pass the transform as a parameter seems a bit cumbersome 
___
*[core/editor/src/input/mouse.rs, line 103 at r10](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MarjG5BEUqD7iUdU2ib:-MarjG5BEUqD7iUdU2ic:bx6teua) ([raw file](https://github.com/graphiteeditor/graphite/blob/5f7f3da83bee64e978b27aba528d827aa432fa4c/core/editor/src/input/mouse.rs#L103)):*
> ```XML
> 		if apply_rotation {
> 			DocumentPosition { x: self.x as f64, y: self.y as f64 }
> 				.add((document_transform.size.x as f64 * -0.5, document_transform.size.y as f64 * -0.5))
> ```

These could be replaced with + for readability

TrueDoctor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 12 of 23 files reviewed, 13 unresolved discussions (waiting on @0HyperCube, @geom3trik, @Keavon, @pkupper, @T0mstone, and @TrueDoctor)

___
*[core/document/src/layers/circle.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Maj2iAoMqoeckuV3RA6-r8-14:-MarpDchAP03OHIanDEo:b-gfhgpe) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/document/src/layers/circle.rs#L14)):*
<details><summary><i>Previously, TrueDoctor wrote…</i></summary><blockquote>

I'll try it out
</blockquote></details>

why can't you just leave out the rotation parameter when constructing the circle operation?
___
*[core/editor/src/input/mouse.rs, line 62 at r10](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MarnRtYChm9ymI93t7P:-MarnRtYChm9ymI93t7Q:b-in8hy9) ([raw file](https://github.com/graphiteeditor/graphite/blob/5f7f3da83bee64e978b27aba528d827aa432fa4c/core/editor/src/input/mouse.rs#L62)):*
> ```XML
> 			location: DocumentPosition { x: 800., y: 800. },
> 			scale: 2.4,
> 			radians: 0.785398163, // 45 degrees in radians
> ```

This is probably just for testing and will get removed, but consider using `f64::consts::FRAC_PI_4` instead

core/document/src/layers/circle.rs
0HyperCube: Applying the coordinates got messy

0HyperCube (commented):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 10 of 23 files reviewed, 13 unresolved discussions (waiting on @0HyperCube, @geom3trik, @pkupper, @T0mstone, and @TrueDoctor)

___
*[core/document/src/document.rs, line 208 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Maj2iAoMqoeckuV3RA5-r8-208:-MarKsrTDx42L9dLfilS:b2cw2nf) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/document/src/document.rs#L208)):*
<details><summary><i>Previously, Keavon (Keavon Chambers) wrote…</i></summary><blockquote>

Right side up in the viewport, tilted relative to the canvas.
</blockquote></details>

So basically working on a rotated canvas should look the same as working on a non rotated canvas.
___
*[core/editor/src/frontend/frontend_message_handler.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvj-r8-14:-MarLXDe2al-gQ184qxL:b-2ewvwq) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/frontend/frontend_message_handler.rs#L14)):*
<details><summary><i>Previously, 0HyperCube wrote…</i></summary><blockquote>

I did the converting to strings in rust because I'm not great at vue, but you could do it either way.
</blockquote></details>

We are already sending an SVG to the front end.
___
*[core/editor/src/input/mouse.rs, line 42 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-42:-MbLxRZw-RX3xrBIcpWd:b9m38jm) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L42)):*
<details><summary><i>Previously, pkupper (Paul Kupper) wrote…</i></summary><blockquote>

Agree with @TrueDoctor s comment. For trigonometric functions and some other calculations radians usually are more natural  (lead to simpler equations), hence why sin() and cos() use radians. We can make sure however, that users only ever see degrees.
</blockquote></details>

It is needed in both radians and degrees. Storing it both ways reduces expensive floating point multiplications
___
*[core/editor/src/input/mouse.rs, line 72 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-72:-MarJlAbDI8wU9EjTmul:b-896fix) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L72)):*
<details><summary><i>Previously, pkupper (Paul Kupper) wrote…</i></summary><blockquote>

Use `self.radians.to_degrees()`
</blockquote></details>

Done.
___
*[core/editor/src/input/mouse.rs, line 90 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-90:-MarHCYACIKKdImYFnso:b2zoxcl) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L90)):*
<details><summary><i>Previously, pkupper (Paul Kupper) wrote…</i></summary><blockquote>

What's the difference between using apply_rotation and not? Intuitively there should only be a single way to transform viewport coordinates to document coordinates. Both cases seem to do a rotation here.
</blockquote></details>

This is a less than ideal fix because some tools have their rotations applied in the canvas transform. If not apply rotation, the rotation is not applied to the actual point, only the offset (it seems to work)

TrueDoctor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/156)* status: 10 of 23 files reviewed, 13 unresolved discussions (waiting on @0HyperCube, @geom3trik, @pkupper, @T0mstone, and @TrueDoctor)

___
*[core/document/src/document.rs, line 208 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-Maj2iAoMqoeckuV3RA5-r8-208:-MbMcc4mD8DRCR3RcPvk:b-gvf3at) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/document/src/document.rs#L208)):*
<details><summary><i>Previously, 0HyperCube wrote…</i></summary><blockquote>

So basically working on a rotated canvas should look the same as working on a non rotated canvas.
</blockquote></details>

Yes I think it is easiest to just think of rotating the canvas as a transform on the document root
___
*[core/editor/src/frontend/frontend_message_handler.rs, line 14 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvj-r8-14:-MbMe6ru4oUjKWqcEx5w:bk5uchw) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/frontend/frontend_message_handler.rs#L14)):*
<details><summary><i>Previously, 0HyperCube wrote…</i></summary><blockquote>

We are already sending an SVG to the front end.
</blockquote></details>

This might be more of a symptom than the cause. I was arguing to return floats purely for frontend  ui info and not to rotate the canvas. That should already happen in the svg. And in that case returning numbers seems like the better approach
___
*[core/editor/src/input/mouse.rs, line 42 at r8](https://reviewable.io/reviews/graphiteeditor/graphite/156#-MaPi7_C-3tAeCkWcnvl-r8-42:-MbMdPLM6l3uqCt6Z22d:b80iiuv) ([raw file](https://github.com/graphiteeditor/graphite/blob/ddcc22e6219434fde5889f1602a7a1335e517398/core/editor/src/input/mouse.rs#L42)):*
<details><summary><i>Previously, 0HyperCube wrote…</i></summary><blockquote>

It is needed in both radians and degrees. Storing it both ways reduces expensive floating point multiplications
</blockquote></details>

having a cache miss can cost you up to 100 cpu cycles or so meanwhile doing the conversion will take less than 20 cycles. (Just some ballpark numbers)
reducing memory consumption is also a plus

PR #160, opened 2021-05-24
Export current document as SVG when pressing Ctrl+Shift+S

TrueDoctor:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/160)


client/web/src/components/panels/Document.vue
```diff
@@ -223,12 +223,29 @@ export default defineComponent({
...
+		download(filename: string, svgData: string) {
+			const svgBlob = new Blob([svgData], { type: "image/svg+xml;charset=utf-8" });
+			const svgUrl = URL.createObjectURL(svgBlob);
+			const element = document.createElement("a");
+
+			element.href = svgUrl;
+			element.setAttribute("download", filename);
+			element.style.display = "none";
+
+			document.body.appendChild(element);
```
Keavon, line 235: I believe we can get rid of the body.appendChild and body.removeChild parts.

client/web/wasm/src/wrappers.rs
```diff
@@ -85,8 +85,11 @@ pub fn translate_key(name: &str) -> Key {
 		"r" => KeyR,
 		"m" => KeyM,
 		"x" => KeyX,
```
Keavon, line 87: Should we just call a function to lower-case all the strings before they are matched against? This might also be as good a time as ever to add all the letters in the alphabet.

core/editor/src/document/document_message_handler.rs
```diff
@@ -69,6 +70,12 @@ impl MessageHandler<DocumentMessage, ()> for DocumentMessageHandler {
 				assert!(id < self.documents.len(), "Tried to select a document that was not initialized");
 				self.active_document = id;
 			}
+			ExportDocument => responses.push_back(
+				FrontendMessage::ExportDocument {
+					document: format!("<svg>{}</svg>", self.active_document_mut().document.render_root()),
```
Keavon, line 75: `<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1920 1080">{}</svg>`

Keavon:
Needs color :)

Keavon (approved):
Great work! Thanks for working through my nitpicks.

PR #163, opened 2021-05-31
Implement menu bar floating list menus

Keavon:
Demo

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/163)

### DIFF
--- a/client/web/src/components/widgets/inputs/MenuBarInput.vue
+++ b/client/web/src/components/widgets/inputs/MenuBarInput.vue
@@ -1,11 +1,12 @@
 <template>
 	<div class="menu-bar-input">
-		<div class="entry"><Icon :icon="'GraphiteLogo'" /></div>
-		<div class="entry"><span>File</span></div>
-		<div class="entry"><span>Edit</span></div>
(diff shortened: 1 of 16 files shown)


client/web/src/components/widgets/inputs/MenuBarInput.vue
```diff
@@ -42,13 +49,108 @@
...
+							{ label: "Another Recent File.gdd" },
+							{ label: "An Older File.gdd" },
+							{ label: "Some Other Older File.gdd" },
+							{ label: "Yet Another Older File.gdd" },
+						],
+					],
+				},
+			],
+			[
+				{ label: "Close", shortcut: ["Ctrl", "W"] },
```
TrueDoctor, line 86: Wont this just close the browser window?

TrueDoctor (approved):

client/web/src/components/widgets/inputs/MenuBarInput.vue
Keavon: Yes, a lot of these aren't supported because the browser doesn't let you override them. We might have to use the keyboard lock API.

PR #178, opened 2021-06-10
Implement viewport selection

### LINKED ISSUE #140: Select Tool: Viewport layer click selection
Click a layer in the viewport to select it. We'll probably want to add a `data-` attribute to each SVG element to identify its path and just use `addEventHandler` to handle the click event. ~~Depends on #137.~~

This pertains to the Select Tool.
Select Tool

Complexity: 1
Involves: Frontend (JS)

### LINKED ISSUE #145: Select Tool: Viewport box selection
Extending upon #140, we need to drag a box around a region in the viewport and select all the layers inside (which would be selected if you went and selected every single pixel inside that rectangular region). We don't want to use the bounding box of each layer, because your box selection might include shapes which don't have geometry located in the rectangular region. This will possibly be tricky to solve and might need to use some of Kurbo's algorithms. It would be nice if we can do it purely in the frontend with JS though. ~~Depends on #137.~~

Complexity: 3
Involves: Frontend (JS), Editor (Rust)

pkupper:
Closes #140 and closes #145

- [x] Ellipse layer intersection
- [x] Line layer intersection
- [x] Polyline layer intersection
- [x] Rectangle layer intersection
- [x] Shape layer intersection
- [x] Select tool drag box
- [x] Select tool point selection
- [x] Select tool box selection
- [x] Update layer panel when deselecting
- [x] Consistent tolerance when selecting lines

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/178)


Keavon:
I'm going to merge this now @TrueDoctor but then we can go back and make changes after you have time to code review it. I'd really like to get this submitted.

PR #209, opened 2021-06-13
Show warning message if BigInt64Array is not available

### LINKED ISSUE #183: Create a warning that Safari is unsupported
Create a way to tell the user that their browser is unsupported. Provide justification to avoid pissing people off. The reason is Safari's lack of support for the [BigInt64Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array) API, because Safari is basically today's Internet Explorer.

It would probably be nice to still let the user view the web page even if functionality is broken. Let's do something that hides the page until the user reads and acknowledges the message. Try to implement this entirely in `App.vue` to avoid polluting other parts of the frontend codebase.

tillarnold:
Closes #183. @Keavon is this roughly what you imagined? I'm a bit of a vuejs novice so I'm sorry if I approached this wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/209)


Keavon:
This looks great! Let me make a few adjustments to the text if you don't mind.

```
Your browser currently doesn't support Graphite
```

```
Unfortunately, some features won't work properly in your browser. Please use a modern browser other than Safari, such as Firefox, Chrome, or Edge. Rest assured, Safari compatibility is planned.

Your browser is missing support for the <a href="https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt64Array#browser_compatibility" target="_blank"><code>BigInt64Array</code></a> JavaScript API which is required for using the editor. You can still explore the user interface.
```

```
I understand, let's just see the interface
```

(you'll want to style the link to use the accent color)

Keavon:
Tiny nitpicks in the code:

* Let's use a 4px border-radius instead of 5px
* Instead of 100vw and 100vh, it might be safer to use right: 0 and bottom: 0 (disregard if this doesn't work, I haven't tested it)
* For the padding, we don't use the rem unit anywhere else in Graphite, but we try to use multiples of 4px everywhere
* Use two <p> tags for the two paragraphs in the altered text above
* Remove the units from your 0-valued CSS properties like 0px and 0rem, so it just says 0 instead for those

Keavon (approved):

Issue #189, opened 2021-06-12
Duplicate the selected layers


Keavon:
Ctrl+D to duplicate the selected layers.

This should be a relatively simple task purely touching the Rust codebase. If you're interested in this issue, we can answer questions and write up a guide of steps to implement this feature.

tillarnold:
I'd be interested in taking this task.

Keavon:
Perfect, I assigned it to you @tillarnold. @TrueDoctor, would you be able to describe what this task will involve? We don't need layer offset for the duplicated or pasted layer.

TrueDoctor:
I'm nor sure how this can be implemented cleanly, the easiest way would be to store a clipboard in the backend and manipulate it through operations but this is far from ideal. The other place to store these would be the editor, but until we have a history system setup we would need to resort to sending layers directly wrapped as operations which is not really idiomatic either.
I don't expect you to understand half of what I just said, It is also mean as a question for @Keavon

Keavon:
Let's just do the duplicate part for now, then we can evaluate what is involved in the copy/paste part. @TrueDoctor

TrueDoctor:
The easiest way to implement duplicate functionality is to add an Operation `Operation::DuplicateLayer{path: Vec<LayerId>}`.
That operation then just clones the layer to the same folder it resides in

Keavon:
@tillarnold are you on the Discord server for Graphite yet? That's typically the way we onboard new contributors and provide an introduction to the code, either through a chat or voice call format. If that would be beneficial for your understanding, since I realize this may be confusing or vague without any context in the backend codebase. Or if you're the kind of person who's great at just reading through code files to orient yourself, that's also great. Discord is most commonly the easiest way to help you get up to speed but if that's not your thing, we want to help remove friction and confusion whatever it exists on any platform. Please let me or @TrueDoctor what would be best to help us help you succeed.

Issue #124, opened 2021-05-11
Change how Vue sends input to the backend


Keavon:
We are currently binding keys in Vue. This was set up as a temporary measure. We'll want to properly send input events to the WASM code without involving Vue. Some considerations:

* [x] All keyboard input should be sent to the WASM
* [x] Only mouse inputs targeting the viewport should be sent, not those involving clicks on GUI elements
* [x] Mouse that has dragged outside the viewport should continue to be sent until (and including) release
* [x] Properly handle (to the best of its abilities) events when the application has been removed from focus in the OS
* [x] Use input to outside-the-viewport items to help reconstruct the status of inputs that may have been interrupted while mouse button was down in a viewport interaction and then focus lost
* [ ] Support multiple panels at once
* [ ] Generic approach so it can be used in other panel types, like the Layer Graph
* [ ] Consideration should be put into pen inputs, and other types of input devices

TrueDoctor:
Alt + Shift keys are combined into a `meta` key press.
![image](https://user-images.githubusercontent.com/21245806/119231311-833b9600-bb20-11eb-8b51-cdda6cc8bd55.png)

One way of fixing this would be to send the current modifier state (+ pressed mouse key)
as additional information in the events to the input preprocessor. This gives the input preprocessor more information to work with and reduces the impact of missed key events.

TrueDoctor:
+ Add scroll wheel handling to the js and the input preprocessor.
https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent

Keavon:
I'm moving this back to Available, but it will require some experience in webdev to properly implement this in a way that doesn't conflict with Vue. If you take this issue, please work closely with me to make sure all considerations are dealt with.

Keavon:
We should standardize on the names of each key defined in https://www.w3.org/TR/uievents-code/#key-alphanumeric-writing-system

PR #214, opened 2021-06-16
Add keyboard shortcut to duplicate layers

tillarnold:
Implements part of #189. @TrueDoctor is this roughly how you imagined the implementation? As you described I added `Operation::DuplicateLayer`. I then also added `DocumentMessage::DuplicateSelectedLayers` which mirrors how deleting the selected layers works. I added the `e.preventDefault();` calls in `Document.vue` because control-D (at least in Firefox) is the shortcut for adding a new bookmark. I'm not sure if this is the way you want to handle this or if we only want to preventDefault on some specific key presses.

Also about the 
```rust
// TODO: Replace with drain_filter https://github.com/rust-lang/rust/issues/59618
let paths: Vec<Vec<LayerId>> = self.active_document().layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())).collect();
```

codeblock. It is now duplicated in the match arms of `DeleteSelectedLayers` and `DuplicateSelectedLayers`. Should that be extracted to a `selected_layers` function?

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/214)

### DIFF
--- a/core/editor/src/document/document_message_handler.rs
+++ b/core/editor/src/document/document_message_handler.rs
@@ -162,9 +162,7 @@ impl MessageHandler<DocumentMessage, ()> for DocumentMessageHandler {
 				}
 			}
 			DuplicateSelectedLayers => {
-				// TODO: Replace with drain_filter https://github.com/rust-lang/rust/issues/59618
-				let paths: Vec<Vec<LayerId>> = self.active_document().layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())).collect();
-				for path in paths {
+				for path in self.active_document().layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())) {
 					responses.push_back(DocumentOperation::DuplicateLayer { path }.into())
 				}
 			}
--- a/core/editor/src/document/document_message_handler.rs
+++ b/core/editor/src/document/document_message_handler.rs
@@ -11,6 +11,7 @@ pub enum DocumentMessage {
 	SelectLayers(Vec<Vec<LayerId>>),
 	DeleteLayer(Vec<LayerId>),
 	DeleteSelectedLayers,
+	DuplicateSelectedLayers,
 	AddFolder(Vec<LayerId>),
 	RenameLayer(Vec<LayerId>, String),
 	ToggleLayerVisibility(Vec<LayerId>),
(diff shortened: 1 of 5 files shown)


core/editor/src/document/document_message_handler.rs
```diff
@@ -160,6 +161,13 @@ impl MessageHandler<DocumentMessage, ()> for DocumentMessageHandler {
 					responses.push_back(DocumentOperation::DeleteLayer { path }.into())
 				}
 			}
+			DuplicateSelectedLayers => {
+				// TODO: Replace with drain_filter https://github.com/rust-lang/rust/issues/59618
+				let paths: Vec<Vec<LayerId>> = self.active_document().layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())).collect();
+				for path in paths {
+					responses.push_back(DocumentOperation::DuplicateLayer { path }.into())
+				}
```
TrueDoctor, line 169: You could avoid the collection into a vec to reduce heap allocations

TrueDoctor: We have to collect it to a vec for deleting layers, because we can't iterate through a collection and simultaneously remove things from  it (hence the drain_filter comment). But in this case that is not necessary, as we only iterate over a non mutable reference

TrueDoctor:
Overall excellent work, if you fix the one thing, I think we are good to merge

Keavon:
@TrueDoctor has a better working knowledge of the backend as the owner of that system. But I'll pose the question, which might be baseless because of my lack of context: should `Operation::DuplicateLayer` be plural?

Keavon:
> I added the e.preventDefault(); calls in Document.vue because control-D (at least in Firefox) is the shortcut for adding a new bookmark. I'm not sure if this is the way you want to handle this or if we only want to preventDefault on some specific key presses.

That's great. There are also a number of other shortcuts like Ctrl+E, Shift+Tab, and maybe more I'm not thinking of right now which trigger default browser behavior. #124 is the comprehensive issue involving that aspect of input transfer to the backend wherever it can be done without conflicting with Vue input. I have had that issue claimed for a while with the intention of starting out but other pressing (and perhaps slightly more fun) tasks got in the way. After we land this PR, would you be interesting in taking over #124 if you think it's within your range of capabilities? In the mean time, that preventDefault is a great measure.

TrueDoctor (approved):

Issue #218, opened 2021-06-16
Copy and paste layers


Keavon:
Copy the selected layers and paste copies of them. The MVP for this feature is basically the same as duplication (#189), but not instant (you can copy, then paste later). If you copy layers then delete some of what you had copied, the paste will still recreate those deleted layers. The paste operation should preserve their order from top to bottom of the stack, but they should all be placed at the top of the current folder (which is presently just the root until we get folders working in the frontend). A copy buffer will need to be introduced, @TrueDoctor thinks in the editor backend is probably cleanest. This will later evolve or be replaced to support actual nodes (layers) in the node (layer) graph. Once #100 is merged, a helpful addition beyond the MVP for this issue will involve checking if the pasted layer is currently visible in the viewport. If so, paste in the same place where the layers were copied from. If not, paste in the center of the viewport.

Keavon:
@tillarnold would you like to follow up #189 with this? Hopefully the description provides come clarity for you. It shouldn't be blocked, except the last two sentences which is blocked on #100.

tillarnold:
Yeah I'll do this one.

Keavon:
@tillarnold would you like any help getting started or continuing with your progress on this feature? I'd be more than happy to lend a hand.

Issue #219, opened 2021-06-17
Ctrl+A to select all layers


Keavon:

AyeTbk:
I'm interested in taking this issue.

Issue #201, opened 2021-06-13
Nudge the selected layers around the canvas with the arrow keys


Keavon:
The arrow keys nudge (move) the selected layer(s) by one canvas pixel. When the <kbd>Shift</kbd> modifier key is held down, the nudges go in increments of 10 canvas pixels.

This is probably a good first issue to get acquainted with some concepts in the Rust backend. Please reach out to @TrueDoctor here or on [Discord](https://discord.graphite.design) for in-depth instructions to help you get a feel for the code that this touches.

henryksloan:
I've started working on this.

Keavon:
This is blocked on the inability to transform layers directly, rather than just folders.

Some discussion from Discord:

> @henryksloan: Operation::TransformLayer seems to act on Folder layers; is that the wrong way to move an individual layer?

> @TrueDoctor: That is a shortcoming of the current design. We already have a fix, it just needs to get implemented.

Keavon:
@henryksloan you should be unblocked now courtesy of the swift work of @0HyperCube.

Issue #149, opened 2021-05-23
Add folders in the Layer Tree panel of the frontend


Keavon:
Documents (and layer paths) involve a nested layer structure. But currently everything is added flatly to the root because the frontend does not support folders. We need to add folders to the Layer Tree panel which can represent the nested layer structure and display the information provided by the Editor backend. After that, we'll be able to rearrange layers into the folders (#141) and select other layers that will be the hierarchical target for new layers to be drawn as siblings of (or a folder for new layers to be children of).

Still left to do:
- [ ] Fix frontend layer rendering JS (@Keavon)
- [ ] Make it so we can copy and paste hierarchy, not just flat layers. This also includes the ability to select multiple layers at various levels in parts of the hierarchy and it copies each of them, to eventually be pasted as new siblings (descendants of selected folders are de-duplicated). (@TrueDoctor)
- [ ] When a layer (shape or folder) is selected and being grouped, we copy and delete its deepest common ancestor folder, put a new folder exactly in its place (including its original index among its siblings), then paste back into that new folder. In the special case of root (because we can't delete the root), we select everything in root, delete everything, create a new folder, then paste everything in that single folder. (@TrueDoctor)

Complexity: 3
Involves: Frontend (JS)

azeemba:
I was considering working on this.

I looked at the `LayerPanelEntry` interface at https://github.com/GraphiteEditor/Graphite/blob/master/client/web/src/utilities/response-handler.ts#L204-L210
It does have the `layer_type` flag which identifies something as a `Folder`. It's not clear to me though how to determine what other layers belong to this folder. Is the goal of this issue to update this interface? Or is that structure encapsulated in the "path" that needs to be decoded?

Keavon:
This is a frontend task that involves two major things. One: building the HTML/CSS for the actual hierarchical layer structure. Two: retrieving `ExpandFolder` and `CollapseFolder` messages and maintaining a JavaScript object in the frontend which is mutated by those two messages. Any time a folder is collapsed, we empty that part of the hierarchy and any children. If anything is updated in a folder, a whole new status is sent for that folder to replace it (but not replace its child folder contents).

Keavon:
[...]

PR #220, opened 2021-06-17
Copy and paste layers MVP

tillarnold:
corresponding issue: #218

@TrueDoctor  I'm not sure if this is what you imagined here so I created this PR in a bit of an unfinished state to ask for some feedback. Since the copy buffer is in the editor I think I need to send Layers wrapped as operations and since operations are serializable I needed to make Layers serializable too. And in turn I needed to enable to `serde` feature of kurbo. I’m not sure if this is desired.

Also I think there is a problem with the keybindings: I had to add the control-V shortcut to the beginning of the `mapping!` because otherwise the `SelectTool(ToolType::Select)` message bound to V did take precendence.

Furthermore the way I select the active layers in `CopySelectedLayers` (which I think is the same way we do it in `DuplicateSelectedLayers` and `DeleteSelectedLayers`) is flawed as it appears to also take into account some sort of temporary layers. This is why I added the warn log in `process_action` for `CopySelectedLayers`. To reproduce: create a rect and then press control-C which correctly copies the rect and logs a debug message to the console. Now just click the canvas with the rect tool which will not create a new layer and then press control-C. You’ll now see the warning message in the console indicating that `layer_data` contained a selected layer which does not actually exist in the real document layers.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/220)


TrueDoctor:
> @TrueDoctor I'm not sure if this is what you imagined here so I created this PR in a bit of an unfinished state to ask for some feedback. Since the copy buffer is in the editor I think I need to send Layers wrapped as operations and since operations are serializable I needed to make Layers serializable too. And in turn I needed to enable to `serde` feature of kurbo. I’m not sure if this is desired.

Yes this is the messiness I was talking about, it is not ideal. One option would be not turning on serialization and just storing the layer as a byte array but that gets messy as soon as we have heap allocations (not sure if kurbo uses any)

> 
> Also I think there is a problem with the keybindings: I had to add the control-V shortcut to the beginning of the `mapping!` because otherwise the `SelectTool(ToolType::Select)` message bound to V did take precendence.
> 
This is atm by design

> Furthermore the way I select the active layers in `CopySelectedLayers` (which I think is the same way we do it in `DuplicateSelectedLayers` and `DeleteSelectedLayers`) is flawed as it appears to also take into account some sort of temporary layers. This is why I added the warn log in `process_action` for `CopySelectedLayers`. To reproduce: create a rect and then press control-C which correctly copies the rect and logs a debug message to the console. Now just click the canvas with the rect tool which will not create a new layer and then press control-C. You’ll now see the warning message in the console indicating that `layer_data` contained a selected layer which does not actually exist in the real document layers.

Will have to look into that

tillarnold:
Question: Is there already code that receives the selected layers in order/sorts layers by their order in the document or should I implement such a thing as part of this PR?

TrueDoctor:
> Question: Is there already code that receives the selected layers in order/sorts layers by their order in the document or should I implement such a thing as part of this PR?

such does currently not exist. This also bears the question of whether we should honor the folder structure when copy pasting (paste the layer into a folder structure that resembles the location they were copied from)

Keavon:
> such does currently not exist. This also bears the question of whether we should honor the folder structure when copy pasting (paste the layer into a folder structure that resembles the location they were copied from)

No, except when the layers that are selected *are* the groups. Let's think of groups as layers which automatically include their children. If you select various layers (some shapes, some groups with their children included), then that set of selected layers will be pasted in a flat structure ordered from top to bottom as they would have been read by scrolling through the layer panel (i.e. a depth first traversal, I think?). Regardless of the folder topology that they are copied *from*, they end up in a flat list where you paste them, but when groups are selected then those are also pasted flatly as groups with their children included. If a group and some of its children are selected, I think it's the same story and you end up with multiple copies of the children, flat and within the folder.

TrueDoctor:
> > such does currently not exist. This also bears the question of whether we should honor the folder structure when copy pasting (paste the layer into a folder structure that resembles the location they were copied from)
> 
> No, except when the layers that are selected _are_ the groups. Let's think of groups as layers which automatically include their children. If you select various layers (some shapes, some groups with their children included), then that set of selected layers will be pasted in a flat structure ordered from top to bottom as they would have been read by scrolling through the layer panel (i.e. a depth first traversal, I think?). Regardless of the folder topology that they are copied _from_, they end up in a flat list where you paste them, but when groups are selected then those are also pasted flatly as groups with their children included. If a group and some of its children are selected, I think it's the same story and you end up with multiple copies of the children, flat and within the folder.

Just keep in mind that this might create issues with blend modes. In that case layers might look differently when pasting them which might be confusing

Keavon:
Blend modes really screw up most approaches, don't they? That said, I think that is fine— there have to be compromises somewhere if you are effectively reordering things in a random way that removes them from their hierarchy.

tillarnold:
Am I correct in my assumption that currently folders and groups cannot be created from the GUI? Should i try to implement this and then write some tests (since manual testing with the GUI is unavailable)?

Keavon:
That is correct. And automated tests would always be excellent additions!

If you are so inclined, adding folders to the frontend (#149) is an unclaimed issue that might be a good next issue once your current pipeline is free. The core of the issue is a moderately simple algorithm but its interplay with other features will take some care. For example it'll make rearranging layers more difficult, once that is implemented.

TrueDoctor:
That is correct and is currently tracked here: https://github.com/GraphiteEditor/Graphite/issues/149
It would be awesome if you could implement that and also make testing easier.
It should actually not be too hard to do so. you have to change how the frontend stores layers to insert the expand messages into a  tree structure and add some left padding for indented layers

TrueDoctor:
wow same brain xD

tillarnold:
So I added some tests. I had to add them to `dispatcher.rs` instead of making them integration tests because I needed access to the `document_message_handler` field to make asserts about the document. Is this the style of tests you imagined? If yes I will add some more test as I haven't yet added any tests for  layers in folders.

Also I added some initial code to order the layers correctly. My code is horrendous but the overall approach I use is to take a Layer path (`Vec<LayerId>`) and map it to a path of indices into the tree of layers (`Vec<usize>`). This is implemented in `indices_for_path`. I then use that in `selected_layers_sorted` to sort the layers. Most of the ugliness of `selected_layers_sorted` comes from me trying to handle the before described case of `active_document().layer_data` containing layers paths which later cannot be found in the actual `Document`. If that wasn't the case  `selected_layers_sorted` could probably be implemented in terms of just a `sort_by_cached_key`. Do you think this sorting approach is overall OK?

Also what I don't fully understand yet is the difference between a Folder and a Group and where Groups are implemented in the code.

Keavon:
I'll have to defer to @TrueDoctor for his expertise on the other questions.

> Also what I don't fully understand yet is the difference between a Folder and a Group and where Groups are implemented in the code.

Folders are Groups should be synonymous. I believe our preference is to call them Folders. This should ideally be standardized to avoid intermixed terminology. Conceptually, a folder is equivalent to a document (the open document is the root folder, and subfolders could also be opened in new tabs) and additionally equivalent to a subgraph (in the node/layer graph panel, you can view folders as child graphs by jumping into the breadcrumb trail of the folder subgraph). In the three different concepts for a document/folder/graph (subdocument/subfolder/subgraph), they can either be edited in situ, in isolation, or independently:

Document:
- In situ: drag things around the parent document viewport and change layer properties, even if they're grouped into a subfolder
- In isolation: a specific element can be isolated in the viewport, dimming the parents/siblings in the background, so it can be edited without the distraction of the (sibling and parent) content
- Independently: a subdocument can be opened and edited in an entirely new tab of the Graphite editor, saved changes show up in the parent unless saved as a separate document

Folder:
- In situ: folders can be modified in the Layer Tree panel (rearranged, properties modified, etc.)
- In isolation: a breadcrumb trail in the Layer Tree panel is shown, and only the isolated subfolder contents are listed in the tree (this is how we see the folder containing your mask while you're creating the selection mask in Mask Mode)
- Independently: Again, open the document in a separate tab

Graph:
- In situ: expand the layer (node) graph for a folder to see its children
- In isolation: jump into the folder's graph in isolation mode, dimming the parent graph in the background, causing a breadcrumb trail to show at the top of the graph viewport
- Independently: Again, open the document in a separate tab

TrueDoctor:
I'll try to look at the code soon

Keavon:
@tillarnold to follow up on that explanation with a few more details that you haven't seen yet, here's the design for a document. A Graphite document contains several blocks, including a set of layers and an embedded storage block. The layers form a set that maps an ID (always incremented within the document, never reused), the name and other metadata for the layer, and data of the layer like its various properties and any pointers to referenced data. Those pointers to referenced data are strings that can point to a path on your local file system, a URL to some resource on the internet, a path on the asset store, a path on your local network (useful for studios), and embedded data which is stored in binary format within the Graphite document (.GDD) file in the embedded storage block. That block is just a bunch of data with pointers to it. Either it has a name given by the user or it has a UUID provided automatically for subgraphs/subdocuments. This embedded data section is where actual folders get stored, since they are actually embedded Graphite documents that live within the file. Unless you take one of your folders, save it to its own GDD file on disk, and convert it from embedded to linked, then by default it will be stored internally within the embedded storage block. This works recursively, since an embedded document also has a set of layers block and an embedded storage block.

TrueDoctor:
I think I know the reason for some of the issues, when using the working folder, the paths are added to the current selection, but don't get removed upon ClearWorkingFolder. We probably need a document response that removes elements from the selection cache which we can then call during the handling of ClearWorkingFolder

tillarnold:
I tried to merge master into this branch but it appears that some changes make the approach of serializing layer impossible/more difficult because gleam::DAffine2 does not appear to implement `Serializable`?

0HyperCube:
@tillarnold Use the serde feature flag on glam https://docs.rs/glam/0.17.0/glam/index.html#feature-gates to get the serde derives

tillarnold:
I already added the serde feature flag and `glam::DVec2` appears to implement Serializable (at least I'm not getting any errors for it). I only get errors for `DAffine2`.

tillarnold:
@Keavon just as a status update for you on this issue: Once the code compiles again (when the above problem about `DAffine2` not being `Serializable` is solved) and I get the OK from @TrueDoctor about the overall approach (and the tests) I should be able to implement your desired behavior for Folders and finish writing the test and then hopefully this PR is in a shape where it can be merged.

0HyperCube:
@tillarnold Sorry for late reply, I've never really used serde before. What you can do is add:
```rust
#[derive(Serialize, Deserialize)]
#[serde(remote = "glam::DAffine2")]
struct DAffine2Ref {
    pub matrix2: DMat2,
    pub translation: DVec2,
}
```
above the Layer struct and then use:
```rust
#[serde(with = "DAffine2Ref")]
pub transform: glam::DAffine2,
```
when declaring the layer

TrueDoctor:
> […] and I get the OK from @TrueDoctor about the overall approach (and the tests) […].

I've looked though the code for a bit and the approach looks good

Keavon:
> @Keavon just as a status update for you on this issue: Once the code compiles again (when the above problem about `DAffine2` not being `Serializable` is solved) and I get the OK from @TrueDoctor about the overall approach (and the tests) I should be able to implement your desired behavior for Folders and finish writing the test and then hopefully this PR is in a shape where it can be merged.

I greatly appreciate the status update, and it sounds like both of the blockers have been solved. Wonderful job so far and I'm excited to have this merged soon! 🚀

tillarnold:
@TrueDoctor should the issue with ClearWorkingFolder be solved in this PR or will there be a separate one?

core/editor/src/misc/test_utils.rs
```diff
@@ -19,9 +19,8 @@ pub trait EditorTestUtils {
 	fn drag_tool(&mut self, typ: ToolType, x1: u32, y1: u32, x2: u32, y2: u32);
 	fn move_mouse(&mut self, x: u32, y: u32);
 	fn mousedown(&mut self, state: MouseState);
-	fn mouseup(&mut self, state: MouseState);
+	fn mouseup(&mut self, x: u32, y: u32);
```
Keavon, line 22: Question: is the inconsistency between `x: u32, y: u32` for `mouseup` and `state: MouseState` for `mousedown` intentional?

Keavon:
> @TrueDoctor should the issue with ClearWorkingFolder be solved in this PR or will there be a separate one?
> 
> also the CI for my latest commit failed in a strange way. Is this a known issue?

The CI can occasionally break before it actually runs the code. In those cases, it's best to just rerun it. If you don't have permission to trigger a rerun, you can either ping me or True Doctor to rerun it, or make some trivial changes (hopefully improvements, maybe add some useful comments?) and push those to cause it to rerun.

TrueDoctor:
> @TrueDoctor should the issue with ClearWorkingFolder be solved in this PR or will there be a separate one?
> 
> also the CI for my latest commit failed in a strange way. Is this a known issue?

Up to you

core/editor/src/misc/test_utils.rs
tillarnold: There might be an issue with how I named the functions but my idea was that a for `mousedown` one wants to provide which `MouseKeys` are pressed while `mouseup` releases all `MouseKeys`. I did this because if we have separate `left_mouseup`and `right_mouseup`and `center_mouseup` functions we'd need to keep track of which mouse buttons are currently pressed to be able to only realase the correct buttons.

tillarnold:
@TrueDoctor the code in this PR currently handles the additional elements in the selection cache so do you think we could first get  this PR ready to merge and tackle the ClearWorkingFolder problem later? If yes I think this is ready for review now.

TrueDoctor:
Technically it would be cleaner to remove that special case handling but I don't ha a strong opinion on that issue

core/editor/src/input/input_mapper.rs
```diff
@@ -103,6 +103,7 @@ macro_rules! mapping {
 impl Default for Mapping {
 	fn default() -> Self {
 		let (up, down, pointer_move) = mapping![
+			entry! {action=DocumentMessage::PasteLayers, key_down=KeyV, modifiers=[KeyControl]},
```
Keavon, line 106: Shouldn't this one be down near the bottom with Duplicate and Copy?

core/editor/src/misc/test_utils.rs
Keavon: I see. But what about, say, if both LMB and RMB are down but then only RMB is released (mouseup)?

Also, the rest of the code tends to use the terminology LMB, RMB, and MMB instead of Left, Right, Middle. This is just for tests, but it's up to you if you want to rename it accordingly or keep it as is.

Keavon:
By the way @tillarnold, I'm happy to merge this once you've had a chance to look at my recent comments. Please let me know if there's anything else I can do to help you!

core/editor/src/input/input_mapper.rs
tillarnold: I had to add the control-V shortcut to the beginning of the `mapping!`  because otherwise the `SelectTool(ToolType::Select)` message bound to V did take precedence.

core/editor/src/misc/test_utils.rs
tillarnold: Do you think it would be better to have `fn mouseup(&mut self, state: MouseState);` to make this more clear and consistent?

tillarnold:
@TrueDoctor @0HyperCube  one small issue: #230 removes `PartialEq` for `Layer` but since `Operation` is `PartialEq` and `PasteLayer` contains a `Layer`  this breaks this PR. Should I readd `PartialEq` for `Layer`? Or is there a better solution?

core/editor/src/misc/test_utils.rs
Keavon: It would make it more consistent, but I'll let you decide if it would make it more clear. I don't require a change, but I bring it up as a potentially confusing/inconsistent part of the code.

core/editor/src/input/input_mapper.rs
Keavon: Ah, makes sense. Hopefully we can implement the precedence system sooner than later to deal with that more cleanly, but currently art features are a higher priority. Thanks for the explanation!

0HyperCube:
> one small issue: #230 removes `PartialEq` for `Layer` but since `Operation` is `PartialEq` and `PasteLayer` contains a `Layer` this breaks this PR. Should I readd `PartialEq` for `Layer`? Or is there a better solution?

@tillarnold Solved by removing `PartialEq` for operation in #232

Keavon:
@tillarnold #232 was merged so you should be unblocked now. If not, please let me know ASAP and I'll do what I can to make sure your needs are met. So I'm up-to-date with the status, how much work is left to do on this before we're ready to merge?

tillarnold:
@Keavon if I didn't miss anything i think this should be ready to merge?

Keavon (approved):

Keavon (approved):
> Reviewed 8 of 18 files at r5, 1 of 1 files at r6, 2 of 5 files at r7, 3 of 5 files at r8, 1 of 2 files at r10, 1 of 1 files at r11, 1 of 1 files at r13, 3 of 3 files at r14.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/220)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tillarnold)

Keavon:
Alrighty, merged! Thank you @tillarnold. And for future reference, since I recently added you to the repository, you should be able to delete your GitHub user's fork and work directly on branches in this official repository now. That will also mean that Cloudflare Pages does deployment runs on pushes to your branches in this repo which is helpful during code review.

If you have time this week (before the end of the sprint this Saturday), there are more items (a couple of them are pretty quick: #201 and #219) in the [Tasks Board](https://github.com/GraphiteEditor/Graphite/projects/6) under the Available category.

PR #228, opened 2021-06-27
Don't use eslint formatter for rust files in vscode

pkupper:
Currently "dbaeumer.vscode-eslint" is globally set as the default formatter for the entire workspace. This does not work for rust files and therefore stops formatOnSave from working in VSCode. I propose removing this definition as this should be a user setting and not a workspace setting. Alternatively "dbaeumer.vscode-eslint" could be set as the default formatter for javascript/typescript/vue files only.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/228)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,7 +24,6 @@
 	],
 	"files.eol": "\n",
 	"html.format.wrapLineLength": 200,
-	"editor.defaultFormatter": "dbaeumer.vscode-eslint",
 	"eslint.format.enable": true,
 	"eslint.workingDirectories": ["./client/web"],
 	"eslint.validate": [
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,29 +3,20 @@
 		"editor.formatOnSave": true,
 		"editor.formatOnPaste": true
 	},
-	"[vue]": {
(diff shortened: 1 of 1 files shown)


Keavon:
Thanks for the heads up, that should target only the web-related files. It should not be removed though.

Keavon:
@pkupper do you want to submit a patch with the following contents of `.vscode/settings.json`?

```json
{
	"[rust]": {
		"editor.formatOnSave": true,
		"editor.formatOnPaste": true
	},
	"[typescript, javascript, json, vue]": {
		"editor.codeActionsOnSave": {
			"source.fixAll.eslint": true
		},
		"editor.formatOnSave": true,
		"editor.defaultFormatter": "dbaeumer.vscode-eslint",
	},
	"rust-analyzer.diagnostics.disabled": [
		"missing-unsafe" // Remove when rust-analyzer bug fixes unsafe code on WASM JavaScript https://github.com/rust-analyzer/rust-analyzer/issues/5412
	],
	"files.eol": "\n",
	"html.format.wrapLineLength": 200,
	"eslint.format.enable": true,
	"eslint.workingDirectories": [
		"./client/web"
	],
	"eslint.validate": [
		"javascript",
		"typescript",
	],
	"vetur.format.enable": false,
}
```

Keavon (approved):

PR #232, opened 2021-07-02
Re-add PartialEq removed in #230

0HyperCube:
~~Remove more partialeq derives from the messaging system.~~ (Edit: see last comment.) Unblocks #220

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/232)

### DIFF
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -5,7 +5,7 @@ use crate::{
 	DocumentError, DocumentResponse, LayerId, Operation,
 };
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Document {
 	pub root: Layer,
 	pub work: Layer,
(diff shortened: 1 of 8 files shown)


Keavon (approved):

TrueDoctor:
wait a sec

TrueDoctor:
We should evaluate it there is a good reason to require partial eq before just removing it

Keavon:
Ok. I gave my approval but let's hold off until we can determine if this is the best course of action.

TrueDoctor:
We will need an eq implementation for message deduplication which is why I initially put it there

TrueDoctor:
We can have a discussion on if we can avoid requiring partial eq for that but it is a discussion we should have

0HyperCube:
When would messages be duplicated?

TrueDoctor:
> When would messages be duplicated?

For example if you have multiple requests to update the canvas for example, you obviously only need to process the last one

Keavon:
Summary from the discussion on Discord:

From @TrueDoctor:
> I would favor reverting most of #230 but keeping the deprecation of our fork

Issue #143, opened 2021-05-23
Select Tool: Grab/drag selected layers around viewport


Keavon:
If the user has no layers selected, this will select the layer under the cursor (see #140) and then drag (transform) it. If multiple are selected, and the drag occurs on one of the selected, it will drag (transform) all of them. If one or more are selected but the drag occurs on another layer, it deselects the current selection and selects the new layer and drags it (like in the first sentence).

After this is implemented, we can then add support for snapping: #200.

Complexity: 3
Involves: Editor (Rust), Document (Rust)

Keavon:
@pkupper do you know if this task is blocked, or soft-blocked in some capacity, by your work on #178?

pkupper:
This is does require #178. I'll see how much work this is and perhaps include it in my PR. We haven't really had the need to modify existing Layer yet so it will require some new Document Operations

Keavon:
Blocked on #243.

Keavon:
@AyeTbk you might like this one.

henryksloan:
I'd like to take this one, now that Fill is merged.

Keavon:
Some additional features for this one: <kbd>Shift</kbd> hotkey constraints to a 15° angle, and adding the <kbd>Ctrl</kbd> hotkey locks to that particular constraint angle no matter where you move your mouse after. But let's merge the basic feature before the modifier keys.

PR #244, opened 2021-07-11
Allow transforming layers

### LINKED ISSUE #243: Allow layers, not only folders, to have transforms

0HyperCube:
Closes #243 and unblocks #201 and #143.

I accidentally used the wrong function to get the layer. I can't find one that gets the root with an empty path and the layer / folder if that is where the path is pointing (unless I'm missing something @TrueDoctor) so I made a new one.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/244)

### DIFF
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -140,6 +140,18 @@ impl Document {
 		Ok(root)
 	}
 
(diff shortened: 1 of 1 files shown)


TrueDoctor (approved):
I have not verified the transformation change, but this looks good to merge

PR #245, opened 2021-07-11
Implement nudging selected layers with arrow keys

### LINKED ISSUE #201: Nudge the selected layers around the canvas with the arrow keys

henryksloan:
Fixes #201

- Added arrow keys to `Key` enum and WASM key translator
- Added `NudgeSelectedLayers` Document message with x and y components
- Handled the new message, transforming the selected layers by the given components
- Added input entries for shift and non-shift variants of the command

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/245)


0HyperCube:
@henryksloan  One of the web things seems to not like `2f64.sqrt()` and instead wants `Math.SQRT2` saying `calls in constants are limited to constant functions`. This should be an easy fix.

henryksloan:
@0HyperCube Oh yeah, that const slipped my mind. Fixed.

core/editor/src/input/input_mapper.rs
```diff
@@ -6,6 +6,11 @@ use super::{
 	InputPreprocessor,
 };
 
+const NUDGE_AMOUNT: f64 = 1.;
+const NUDGE_AMOUNT_DIAG: f64 = NUDGE_AMOUNT * std::f64::consts::FRAC_1_SQRT_2;
```
Keavon, line 10: Sorry to contradict the earlier discussion from the team on Discord but a slower diagonal nudge distance is not actually a desirable feature because the user explicitly expects integer movements as multiples of 1 or 10 (while holding shift), so going 1.414 times faster diagonally is not really a problem compared to resulting in unpredictable, non-integer values for the final transform position. Even supporting simultaneous diagonal movement is an improvement over Photoshop's behavior, which just goes side-to-side or up-or-down depending on whether you hit the left/right/up/down key most recently while holding a diagonal pair.

Keavon (approved):

PR #254, opened 2021-07-13
Implement fill tool

### LINKED ISSUE #152: New Tool: Fill Tool
The Fill Tool changes the fill color property for the layer which gets clicked on by setting it to the primary working color. Although it will eventually be a raster tool for flood filling a thresholded range of colors in an area of the raster image, we can duplicate its icon and turn it blue and move it to the bottom section of tools to use it as a way to change the color of shapes that click on it.

henryksloan:
Closes #152

- Enabled the Fill tool icon
- Added a `set_fill` method to `Fill`
    - Should all getters and setters be added to `Fill` and `Stroke` for consistency?
- Added a new operation that replaces a layer's `fill` with a given color
- Added an input entry mapping `MouseDown` + `Lmb` to the Fill tool
- Added Fill as a `#[child]` to the `tool_message_handler`
- Implemented the `Fill` click handler, generating a `FillLayer` operation with the primary color, targeting a clicked layer (if any)

The biggest caveat at the moment is the ad hoc click detection. I'm using `intersects_quad_root` with a small rectangle centered on the mouse. This works (though it gives an extra pixel of padding to the click), but I imagine `intersects_point` and `intersects_point_root` functions should be implemented in the future (I can open an issue for that if we agree it would be preferred).

Also, when multiple layers are returned by `intersects_quad_root`, I am using the first of them. This seems to correspond to the visually lowest layer; should it instead use the last item returned?

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/254)


core/document/src/layers/style/mod.rs
```diff
@@ -56,6 +56,9 @@ impl PathStyle {
 	pub fn new(stroke: Option<Stroke>, fill: Option<Fill>) -> Self {
 		Self { stroke, fill }
 	}
+	pub fn set_fill(&mut self, fill: Option<Fill>) {
```
TrueDoctor, line 59: I would suggest set_fill should take fill as an argument and wrap it in Some before assigning.
The value is wrapped in an option because you don't have to have a fill, but if you set the fill, thät is always the case

core/document/src/document.rs
```diff
@@ -354,6 +354,14 @@ impl Document {
 				let path = path.as_slice()[..path.len() - 1].to_vec();
 				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::FolderChanged { path }])
 			}
+			Operation::FillLayer { path, color } => {
+				if let Ok(layer) = self.layer_mut(path) {
+					layer.style.set_fill(Some(layers::style::Fill::new(*color)));
+					layer.cache_dirty = true;
+				}
+				self.root.cache_dirty = true;
```
TrueDoctor, line 362: Could you while your at it maybe include this snippet of code?
```rust
fn mark_as_dirty(&mut self, path: &[PathId]) {
   let mut root = self.root
   for id in path {
       root.dirty = true;
       root = root.as_folder().layer(id) // skip in last iteration
  }
}
```
That probably won't compile but that should not be to hard to fix if you look at the other functions, would just be nice to do the dirtification properly

core/editor/src/tool/tools/fill.rs
```diff
@@ -0,0 +1,34 @@
...
+#[impl_message(Message, ToolMessage, Fill)]
+#[derive(PartialEq, Clone, Debug)]
+pub enum FillMessage {
+	MouseDown,
+}
+
+impl<'a> MessageHandler<ToolMessage, ToolActionHandlerData<'a>> for Fill {
+	fn process_action(&mut self, action: ToolMessage, data: ToolActionHandlerData<'a>, responses: &mut VecDeque<Message>) {
+		let mouse_pos = data.2.mouse.position;
+		let quad = [
```
TrueDoctor, line 17: You could take a look at `src/tool/tools/select.rs` for how to genericise selection tolerances.
It would be great if you could also put the selection tolerance const into the `src/consts.rs` module
oh wait, nevermind that does not exist yet but you could add the constant with a todo comment

Keavon:
> Should all getters and setters be added to `Fill` and `Stroke` for consistency?

Might as well!

core/document/src/document.rs
0HyperCube, line 362: The root doesn't need to be set as dirty or have a `DocumentChanged` operation if the layer does not exist

core/document/src/layers/style/mod.rs
0HyperCube: We may want a way to clear the fill on for example a rectangle with both an outline and a fill, but it should probably be with a `clear_fill` method

core/editor/src/tool/tools/fill.rs
```diff
@@ -0,0 +1,34 @@
...
+			glam::DVec2::new(mouse_pos.x as f64 - 1., mouse_pos.y as f64 - 1.),
+			glam::DVec2::new(mouse_pos.x as f64 + 1., mouse_pos.y as f64 - 1.),
+			glam::DVec2::new(mouse_pos.x as f64 - 1., mouse_pos.y as f64 + 1.),
+			glam::DVec2::new(mouse_pos.x as f64 + 1., mouse_pos.y as f64 + 1.),
+		];
+		if let Some(path) = data.0.intersects_quad_root(quad).get(0) {
```
0HyperCube, line 23: I assume you want to fill the top most layer, so `.get(0)` should be `.last()`

0HyperCube (dismissed):

core/document/src/layers/style/mod.rs
henryksloan: Good points. I had chosen this method to ensure the value could be reset to None. I just implemented @0HyperCube's recommendation, as it's the best of both worlds.

core/editor/src/tool/tools/fill.rs
henryksloan: Agreed, done.

core/document/src/document.rs
henryksloan: You're right, I have it this way because making both changes in the same scope requires taking `self` as mut twice. I'll rewrite it slightly to avoid that. By the way, this operation is unconditional, I was using `if let` because it separated the scopes in which self was borrowed.

henryksloan: I reordered the operations and replaced the `if let` with an `unwrap`. While most operations that receive paths `unwrap` them, `ToggleVisibility` uses Result::map and discards the result (which I think is an antipattern). I imagine the operations should be consistent about this, and it may be a good idea to move away from `unwrap` before it becomes a problem for concurrency or something.

henryksloan: I was able to get that function compiled, but I couldn't get it to actually properly dirty the cache. I tried some variations with no luck; have you been able to get this function working?

TrueDoctor: did you try some debug printing? I would imagine it to be relatively easy to debug

henryksloan: Done, I had to change the order of the `root.dirty = true` statements to ensure that the last layer in the path was dirtied.

core/document/src/layers/style/mod.rs
```diff
@@ -56,6 +56,24 @@ impl PathStyle {
 	pub fn new(stroke: Option<Stroke>, fill: Option<Fill>) -> Self {
 		Self { stroke, fill }
 	}
+	pub fn get_fill(&self) -> Option<Fill> {
```
TrueDoctor, line 59: please remove the get wherever you see it: https://rust-lang.github.io/api-guidelines/naming.html
our codebase is not yet consistent in that regard but it is something we want to improve on

TrueDoctor (approved):

TrueDoctor:
Alright, feel free to merge once the ci is done

PR #258, opened 2021-07-14
Implement collaborative editing

TrueDoctor:
~~does not work at all~~
does work after all

Notes from early testing:
* don't broadcast the download document message
* each user has his own working folder
* only operations and their responses are exchanged -> server could store all past operations for new users that join
* should operations explicitly target one document? (should users be allowed to work on different documents simultaneously)

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/258)


Keavon:
Let's put the server into a folder called `nanotube`.

Keavon:
Closing this for the time being.

PR #264, opened 2021-07-14
Transform nudge coordinates to viewport space

### LINKED ISSUE #259: Nudge should move in viewport space, not canvas space
Draw something, rotate the document to ~45° and nudge the shape with the up/down/left/right keys. The shape will move along the diagonal.

henryksloan:
Fixes #259

I transformed the given nudge vector by rotating it by the negative of the root layer rotation.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/264)


core/editor/src/document/document_message_handler.rs
```diff
@@ -501,10 +501,16 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 			}
 			NudgeSelectedLayers(x, y) => {
 				let paths: Vec<Vec<LayerId>> = self.selected_layers_sorted();
+
+				let delta = {
+					let neg_angle = -self.layerdata_mut(&vec![]).rotation;
+					let (cos, sin) = (neg_angle.cos(), neg_angle.sin());
+					DVec2::new(x * cos - y * sin, x * sin + y * cos)
+				};
```
TrueDoctor, line 509: I think you can just use 
```rust
let delta = self.layerdata_mut(&vec![]).inverse((x,y).into());
```

henryksloan: The `inverse` function seems to be a member of `Transform`, but I'm getting the rotation from LayerData::rotation. Do you mean I should create a transform with rotation equal to `self.layerdata_mut(&vec![]).rotation`?

TrueDoctor:
This will still not work for subfolders.
I would still suggest to use active_document().document.root.transform.inverse

Keavon (approved):

PR #266, opened 2021-07-15
Add a fullscreen button and the keyboard lock API

### LINKED ISSUE #249: Add a fullscreen button to take over browser-reserved keybinds
The [keyboard lock](https://web.dev/keyboard-lock/) API (implemented only on Chromium-based browsers at present) lets us take control of shortcuts like <kbd>Ctrl</kbd><kbd>N</kbd>,<kbd>Ctrl</kbd><kbd>W</kbd>, and <kbd>Shift</kbd><kbd>Tab</kbd> when in fullscreen mode. Replace the Windows-style Minimize/Maximize/Close window buttons, which are useless in a browser, with a button to go fullscreen and enable keyboard lock. Show the fullscreen button but not the label about enabling shortcuts in browsers that don't support the API. Disable the inaccessible keyboard shortcuts printed in menus when keyboard lock is not active.

Keavon:
Closes #249

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/266)


client/web/src/App.vue
```diff
@@ -196,10 +196,12 @@ img {
 
 <script lang="ts">
 import { defineComponent } from "vue";
+import fullscreen from "@/utilities/fullscreen";
 import MainWindow from "./components/window/MainWindow.vue";
```
henryksloan, line 200: Probably avoid mixing relative with absolute paths.

client/web/src/components/widgets/floating-menus/MenuList.vue
```diff
@@ -15,7 +15,16 @@
...
 				<span class="entry-label">{{ entry.label }}</span>
-				<UserInputLabel v-if="entry.shortcut && entry.shortcut.length" :inputKeys="[entry.shortcut]" />
+				<Icon
+					:icon="'Info'"
+					v-if="entry.shortcutRequiresLock && !fullscreen.keyboardLocked"
+					:title="
+						keyboardLockApiSupported
+							? 'This hotkey is reserved by the browser, but becomes available in fullscreen mode'
+							: 'This hotkey is reserved by the browser, but becomes available in Chrome, Edge, and Opera which support the Keyboard.lock() API'
+					"
```
henryksloan, line 25: This is very verbose, maybe use a computed property. Alternatively, since `keyboardLockApiSupported` is only set once, just put this text in member data on component init.

client/web/src/components/widgets/floating-menus/MenuList.vue
```diff
@@ -15,7 +15,16 @@
...
+				/>
+				<UserInputLabel v-else-if="entry.shortcut && entry.shortcut.length" :inputKeys="[entry.shortcut]" />
```
henryksloan, line 27: If typescript is allowed in these attributes, you may be able to do something cleaner like `entry.shortcut?.length`

client/web/src/components/widgets/floating-menus/MenuList.vue
```diff
@@ -242,6 +254,7 @@ const MenuList = defineComponent({
 	data() {
 		return {
 			currentEntry: this.activeEntry,
+			keyboardLockApiSupported: keyboardLockApiSupported(),
```
henryksloan, line 257: Consider changing the function name to `isKeyboardLockApiSupported` to differentiate it from the data identifier

client/web/src/utilities/fullscreen.ts
```diff
@@ -0,0 +1,37 @@
...
+		state.keyboardLocked = true;
+	}
+}
+
+export async function exitFullscreen() {
+	await document.exitFullscreen();
+}
+
+export async function toggleFullscreen() {
+	if (state.windowFullscreen) await exitFullscreen();
```
henryksloan, line 33: An if/else pair with no braces is a bit strange to look at IMO

client/web/src/utilities/input.ts
```diff
@@ -6,7 +8,11 @@ export function shouldRedirectKeyboardEventToBackend(e: KeyboardEvent): boolean
 	if (target.nodeName === "INPUT" || target.nodeName === "TEXTAREA" || target.isContentEditable) return false;
 
 	// Don't redirect a fullscreen request
-	if (e.key.toLowerCase() === "f11") return false;
+	if (e.key.toLowerCase() === "f11" && e.type === "keydown" && !e.repeat) {
```
henryksloan, line 11: Would it be a good idea to lowercase `e.type`? I'm sure it's consistent in practice, but I figure the same is also true of `e.key`, so it seems reasonable to lowercase either both or neither

henryksloan (approved):
LGTM besides a bunch of minor stylistic and clarity recommendations

client/web/src/components/widgets/floating-menus/MenuList.vue
Keavon: Infuriatingly, WebPack doesn't let us use the optional chaining operator.

client/web/src/components/widgets/floating-menus/MenuList.vue
Keavon: Excellent idea

client/web/src/App.vue
Keavon: Will be fixed in an upcoming PR which replaces all paths in the project with absolute ones.

client/web/src/utilities/input.ts
Keavon: The JS spec guarantees that capitalization of its event names. I'm less certain if that's the case for keyboard input names.

PR #271, opened 2021-07-15
Support rearranging layers with hotkeys

### LINKED ISSUE #238: Rearranging layers with hotkeys
<kbd>Ctrl</kbd><kbd>[</kbd> lowers the selected layer(s) in the stack, <kbd>Ctrl</kbd><kbd>]</kbd> raises, and the <kbd>Shift</kbd> modifier raises the selected layer(s) all the way to the top or bottom. Raising and lowering is constrained to the folder for now. Leave a `// TODO` for handling the case where layers from multiple separate folders or depths in the folder hierarchy are selected when trying to raise/lower.

ajweeks:
Closes #238

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/271)


core/editor/src/document/document_message_handler.rs
```diff
@@ -161,6 +162,32 @@ impl DocumentMessageHandler {
 		layers_with_indices.sort_by_key(|(_, indices)| indices.clone());
 		return layers_with_indices.into_iter().map(|(path, _)| path).collect();
 	}
+
+	/// Returns the paths to all layers in order
+	fn all_layers_sorted(&self) -> Vec<Vec<LayerId>> {
```
TrueDoctor, line 167: This function should be combined with the selected_layers_sorted function to reduce code duplication

core/editor/src/input/input_mapper.rs
```diff
@@ -235,6 +235,10 @@ impl Default for Mapping {
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, -NUDGE_AMOUNT), key_down=KeyArrowRight, modifiers=[KeyArrowUp]},
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, NUDGE_AMOUNT), key_down=KeyArrowRight, modifiers=[KeyArrowDown]},
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, 0.), key_down=KeyArrowRight},
+			entry! {action=DocumentMessage::MoveSelectedLayer(-1), key_down=KeyLeftBracket, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::MoveSelectedLayer(1), key_down=KeyRightBracket, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::MoveSelectedLayer(-999999), key_down=KeyLeftBracket, modifiers=[KeyControl, KeyShift]},
```
ajweeks, line 240: These key binds (ctrl+shift+[ & ]) aren't working, I'm not fully sure why, any ideas?

TrueDoctor: Can you rebase onto the current master?
It is caused by the specified precedence (Ctrl+[) also matches and it is the first match.
But I added a change that should also sort them by the number of modifiers used, this would be a good test to see if my code is working.

client/web/src/components/widgets/inputs/MenuBarInput.vue
```diff
@@ -126,6 +126,10 @@ const menuEntries: MenuListEntries = [
 			[
 				{ label: "Select All", shortcut: ["Ctrl", "A"], action: async () => (await wasm).select_all_layers() },
 				{ label: "Deselect All", shortcut: ["Ctrl", "Alt", "A"], action: async () => (await wasm).deselect_all_layers() },
+				{ label: "Move Layer Up", shortcut: ["Ctrl", "]"], action: async () => (await wasm).move_selected_layer(1) },
+				{ label: "Move Layer Down", shortcut: ["Ctrl", "["], action: async () => (await wasm).move_selected_layer(-1) },
+				{ label: "Move Layer To Top", shortcut: ["Ctrl", "Shift", "]"], action: async () => (await wasm).move_selected_layer(999999) },
+				{ label: "Move Layer To Bottom", shortcut: ["Ctrl", "Shift", "["], action: async () => (await wasm).move_selected_layer(-999999) },
```
Keavon, line 132: Let's put these in a flyout menu category called "Arrange" and use the following naming:
"Raise"

"Lower"

core/editor/src/document/document_message_handler.rs
```diff
@@ -54,6 +54,7 @@ pub enum DocumentMessage {
 	WheelCanvasZoom,
 	SetCanvasRotation(f64),
 	NudgeSelectedLayers(f64, f64),
+	MoveSelectedLayer(i32),
```
Keavon, line 57: "Move" sounds like it means *translate* to me. Would you please rename variables/functions/types like "move" or "position" to "rearrange" or "reorder"?

core/editor/src/input/input_mapper.rs
```diff
@@ -235,6 +235,10 @@ impl Default for Mapping {
...
+			entry! {action=DocumentMessage::MoveSelectedLayer(999999), key_down=KeyRightBracket, modifiers=[KeyControl, KeyShift]},
```
Keavon, line 241: Please use `i32::MAX` and `i32::MIN` instead.

client/web/wasm/src/document.rs
```diff
@@ -163,14 +163,22 @@ pub fn select_all_layers() -> Result<(), JsValue> {
 	EDITOR_STATE.with(|editor| editor.borrow_mut().handle_message(DocumentMessage::SelectAllLayers)).map_err(convert_error)
 }
 
-/// Select all layers
+/// Deselect all layers
```
Keavon, line 166: Nice catch! Thanks.

core/editor/src/input/input_mapper.rs
ajweeks: If you're referring to 5b39776, I already had it in my branch history. I updated to latest anyway and the issue wass still occurring. Even when manually reordering the entries in code, or removing all other actions it still wouldn't work.

I found however that I was able to get it working by adding another two keybinds, specifically for `{` & `}`, which are the shift-equivalents of `[` & `]` on most keyboards.

TrueDoctor: Ah, that makes a lot of sense, glad you got it working. For these issues it is also helpful to pres `3` to get debug output in the console.
But this will still cause issues, on my keyboard that would be the `λ` and `χ` keys… that will potentially be resolved when we switch to hardware key presses. Should we switch to something else in the meantime?

TrueDoctor:
Oh, and in general, feel free to add new commits instead of just amending them, everything will be squashed in the end anyway, it just makes it easier for us to keep track of the changes. (even tough I fully empathize with not wanting to write commit messages)

core/editor/src/input/input_mapper.rs
ajweeks: Ah that's handy to know about - I've just been littering the code with debug logs up until now 😄

Hm, I see. I'm not overly familiar with handling inputs in web apps so I'm not sure what an alternative solutions there are, besides perhaps trying to guess what kind of keyboard the user has or supporting key remapping.

client/web/src/components/widgets/inputs/MenuBarInput.vue
ajweeks: Sure thing. Could you clarify exactly what you mean by "flyout menu" though? Is there precedence for this type of feature elsewhere in the codebase? My skills lie much more in backend than frontend so it might make more sense for someone else to hook it up if it's not trivial.

core/editor/src/document/document_message_handler.rs
ajweeks: Agreed, I think "reorder" is a good term for it.

core/editor/src/input/input_mapper.rs
TrueDoctor: Yeah this kinda belongs to #124 and has yet to be addressed, it has already caused issues in the past

core/editor/src/input/input_mapper.rs
ajweeks: That should cause the addition in `DocumentMessageHandler.process_action` to overflow... but I just tested it and it doesn't somehow? I don't like it when code works but I don't know why :thinking:

This is the behaviour I expected: (panic with 'attempt to add with overflow') https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=91a719d050d320467beaf56c7a2dcda5

We can also special case these two values if that's preferable to using arbitrarily large ints here.

TrueDoctor: You should either add a special case or cast through i64. Casting is probably better because the user could still give weird input that overflows otherwise

ajweeks:
This is why I force pushed before, I think I messed up the rebase 🤦‍♂️

README.md
```diff
@@ -1,29 +1,23 @@
 ![Graphite Logo](graphite_splash.png)
```
Keavon, line 1: Discard changes to this file (merge error).

client/web/vue.config.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
 				(Plugin) =>
 					new Plugin({
 						crateDirectory: path.resolve(__dirname, "wasm"),
+						watchDirectories: [path.resolve(__dirname, "../../core")],
```
Keavon, line 16: Discard changes to this file (merge error).

core/document/src/document.rs
```diff
@@ -227,13 +227,28 @@ impl Document {
...
+			self.layer_local_bounding_box(path)
+		} else {
+			let layer = self.document_layer(path)?;
+			Ok(layer.bounding_box(self.root.transform * layer.transform, layer.style))
+		}
 	}
 
-	pub fn layer_local_bounding_box(&self, path: &[LayerId]) -> Result<[DVec2; 2], DocumentError> {
-		let layer = self.layer(path)?;
+	pub fn layer_local_bounding_box(&self, path: &[LayerId]) -> Result<Option<[DVec2; 2]>, DocumentError> {
```
Keavon, line 249: I believe this function change is a merge error.

.vscode/settings.json
```diff
@@ -12,6 +12,7 @@
 		"editor.defaultFormatter": "dbaeumer.vscode-eslint",
 	},
 	"rust-analyzer.experimental.procAttrMacros": true,
+	"rust-analyzer.cargo.target": "wasm32-unknown-unknown",
```
Keavon, line 15: Discard changes to this file (merge error).

core/document/src/layers/folder.rs
```diff
@@ -101,6 +119,35 @@ impl Folder {
 			_ => None,
 		}
 	}
+
+	pub fn bounding_box(&self, transform: glam::DAffine2) -> Option<[DVec2; 2]> {
```
Keavon, line 123: I believe this function is also a merge error.

core/document/src/layers/mod.rs
```diff
@@ -175,8 +175,12 @@ impl Layer {
 		self.data.to_kurbo_path(self.transform, self.style)
 	}
 
-	pub fn bounding_box(&self, transform: glam::DAffine2, style: style::PathStyle) -> [DVec2; 2] {
-		self.data.bounding_box(transform, style)
+	pub fn bounding_box(&self, transform: glam::DAffine2, style: style::PathStyle) -> Option<[DVec2; 2]> {
```
Keavon, line 178: Discard changes to this file (merge error).

core/editor/src/document/document_message_handler.rs
```diff
@@ -136,21 +137,22 @@ impl DocumentMessageHandler {
 		);
 	}
 
-	/// Returns the paths to the selected layers in order
-	fn selected_layers_sorted(&self) -> Vec<Vec<LayerId>> {
+	/// Returns the paths to all layers in order, optionally including only selected layers
+	fn get_sorted_layers(&self, only_selected: bool) -> Vec<Vec<LayerId>> {
```
Keavon, line 141: We don't use the `get` prefix in Rust code, per Rust style guidelines.

core/editor/src/document/document_message_handler.rs
```diff
@@ -161,6 +163,16 @@ impl DocumentMessageHandler {
...
+
+	/// Returns the paths to all layers in order
+	fn get_all_layers_sorted(&self) -> Vec<Vec<LayerId>> {
+		self.get_sorted_layers(false)
+	}
+
+	/// Returns the paths to all selected layers in order
+	fn get_selected_layers_sorted(&self) -> Vec<Vec<LayerId>> {
+		self.get_sorted_layers(true)
+	}
```
Keavon, line 175: `get`

client/web/src/components/widgets/inputs/MenuBarInput.vue
Keavon: A sub-menu. See "Open Recent" a few lines above for an example.

ajweeks:
Thanks for the merge help @TrueDoctor

ajweeks:
Yeah, that code will change imminently though since I'm adding support for moving multiple layers now anyway

core/editor/src/document/document_message_handler.rs
```diff
@@ -159,7 +161,17 @@ impl DocumentMessageHandler {
...
+
+	/// Returns the paths to all layers in order
+	fn get_all_layers_sorted(&self) -> Vec<Vec<LayerId>> {
+		self.get_sorted_layers(false)
+	}
+
+	/// Returns the paths to all selected layers in order
+	fn get_selected_layers_sorted(&self) -> Vec<Vec<LayerId>> {
+		self.get_sorted_layers(true)
```
Keavon, line 174: We don't use the `get` prefix in Rust code, per Rust style guidelines.

client/web/src/components/widgets/inputs/MenuBarInput.vue
```diff
@@ -126,6 +126,10 @@ const menuEntries: MenuListEntries = [
 			[
 				{ label: "Select All", shortcut: ["Ctrl", "A"], action: async () => (await wasm).select_all_layers() },
 				{ label: "Deselect All", shortcut: ["Ctrl", "Alt", "A"], action: async () => (await wasm).deselect_all_layers() },
+				{ label: "Raise To Front", shortcut: ["Ctrl", "]"], action: async () => (await wasm).reorder_selected_layer(2147483647) },
+				{ label: "Raise", shortcut: ["Ctrl", "["], action: async () => (await wasm).reorder_selected_layer(1) },
+				{ label: "Lower", shortcut: ["Ctrl", "Shift", "]"], action: async () => (await wasm).reorder_selected_layer(-1) },
+				{ label: "Lower to Back", shortcut: ["Ctrl", "Shift", "["], action: async () => (await wasm).reorder_selected_layer(-2147483648) },
```
Keavon, line 132: The displayed keyboard shortcuts are out of order.

core/editor/src/input/input_mapper.rs
```diff
@@ -235,6 +235,10 @@ impl Default for Mapping {
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, -NUDGE_AMOUNT), key_down=KeyArrowRight, modifiers=[KeyArrowUp]},
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, NUDGE_AMOUNT), key_down=KeyArrowRight, modifiers=[KeyArrowDown]},
 			entry! {action=DocumentMessage::NudgeSelectedLayers(NUDGE_AMOUNT, 0.), key_down=KeyArrowRight},
+			entry! {action=DocumentMessage::ReorderSelectedLayer(-1), key_down=KeyLeftBracket, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::ReorderSelectedLayer(1), key_down=KeyRightBracket, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::ReorderSelectedLayer(i32::MIN), key_down=KeyLeftCurlyBracket, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::ReorderSelectedLayer(i32::MAX), key_down=KeyRightCurlyBracket, modifiers=[KeyControl]},
```
Keavon, line 241: These also need the <kbd>Shift</kbd> modifier key.

client/web/src/components/widgets/inputs/MenuBarInput.vue
ajweeks: I've set this up but the menu items now no longer their actions, I think there's a bug somewhere in `MenuList.vue`. I'd guess grandchildren's actions are not being registered, but I'm not familiar enough with Vue to find a solution.

core/editor/src/input/input_mapper.rs
ajweeks: This was discussed [above](https://github.com/GraphiteEditor/Graphite/pull/271#discussion_r670810447), I could only get this to work by listening for the curly bracket key, rather than shift + square bracket. Once #124 is resolved this can perhaps be changed. Should I leave a comment saying so?

Keavon: Ah, gotcha. Yes, please definitely leave a comment explaining the situation to anyone unfamiliar with the context for the strange reasoning behind this so it doesn't look like an error, and also so it is clear when this can be cleaned up someday once the input system is more feature-complete.

client/web/src/components/widgets/inputs/MenuBarInput.vue
Keavon: Working on fixing this now, it appears that the click handler for submenu entries isn't being called for some reason. You are fine to commit this but my fix should solve the functionality aspect.

Keavon:
Looks good, thanks for making the changes. Is this ready for a final review or do you need to handle more edge cases or add features or tests or something? Let me know if you need any help completing it.

core/document/src/layers/folder.rs
```diff
@@ -65,6 +65,84 @@ impl Folder {
...
+				.concat();
+			}
+
+			rearrange(&mut self.layers, source_pos, source_pos_end, target_pos);
+			rearrange(&mut self.layer_ids, source_pos, source_pos_end, target_pos);
+
+			let min_index = source_pos_end.min(target_pos);
+			let max_index = source_pos_end.max(target_pos);
+			for layer_index in min_index..max_index {
+				self.layers[layer_index].cache_dirty = true;
```
ajweeks, line 110: I'm not fully sure this is necessary, there seems to be quite a few ways to mark things dirty in the codebase. In the current things do update correctly, I may be doing something non-optimally though.

Keavon: @TrueDoctor

ajweeks:
I hadn't considered testing, that's not a bad idea though, I've added a few. Let me know if they're sufficient.

I think it's ready for a final review now, it's working as I'd expect it to.

TrueDoctor:
> I hadn't considered testing, that's not a bad idea though, I've added a few. Let me know if they're sufficient.
> 
> I think it's ready for a final review now, it's working as I'd expect it to.
That's nice, will get right to reviewing

core/document/src/document.rs
```diff
@@ -226,6 +226,15 @@ impl Document {
 		Ok(())
 	}
 
+	pub fn reorder_layers(&mut self, source_paths: &[Vec<LayerId>], target_path: &[LayerId]) -> Result<(), DocumentError> {
+		// TODO: Detect when moving between folders and handle properly
+
+		let source_layer_ids = source_paths.iter().map(|x| *x.last().unwrap()).collect();
```
TrueDoctor, line 232: You could throw document Errors here instead of just panicking because this is a public function

Keavon:
Looks great, thanks for adding the tests! (This is the kind of thing that can be trickier to get right so testing is indeed helpful). My only comment about the tests is just a nitpick about the comments: it took me about a minute of understand what you meant by `''` until I realized that's the "same as above" indicator, my mind was thinking along the lines of single-quoted strings or anything else. If you get the chance, could you consider moving the comment to the line above the one it refers to (typically it's better stylistic practice for descriptive comments, rather than TODOs, to be on its own line) so the comment refers to the block of following lines, and the `''` mark is therefore unnecessary? I really don't intend to be a stickler over comments or style like that, so please feel more than free to ignore this comment, it's truly a nitpick I'm totally happy to ignore.

core/document/src/layers/folder.rs
```diff
@@ -65,6 +65,84 @@ impl Folder {
 		Ok(())
 	}
 
+	pub fn reorder_layers(&mut self, source_ids: Vec<LayerId>, target_id: LayerId) -> Result<(), DocumentError> {
+		let source_pos = self.layer_ids.iter().position(|x| *x == source_ids[0]).ok_or(DocumentError::LayerNotFound)?;
+		let source_pos_end = source_pos + source_ids.len() - 1;
+		let target_pos = self.layer_ids.iter().position(|x| *x == target_id).ok_or(DocumentError::LayerNotFound)?;
+
+		let mut last_pos = source_pos;
+		for layer_id in &source_ids[1..source_ids.len()] {
```
TrueDoctor, line 74: can you do something like `source_ids.take(source_ids.len())`? might be a bit more ideomatic

core/document/src/layers/folder.rs
```diff
@@ -65,6 +65,84 @@ impl Folder {
 		Ok(())
 	}
 
+	pub fn reorder_layers(&mut self, source_ids: Vec<LayerId>, target_id: LayerId) -> Result<(), DocumentError> {
+		let source_pos = self.layer_ids.iter().position(|x| *x == source_ids[0]).ok_or(DocumentError::LayerNotFound)?;
```
TrueDoctor, line 69: you could put that into a local function (write either a fn in the fn or bind it to a lambda)

core/document/src/layers/folder.rs
```diff
@@ -65,6 +65,84 @@ impl Folder {
...
+		let source_pos_end = source_pos + source_ids.len() - 1;
+		let target_pos = self.layer_ids.iter().position(|x| *x == target_id).ok_or(DocumentError::LayerNotFound)?;
+
+		let mut last_pos = source_pos;
+		for layer_id in &source_ids[1..source_ids.len()] {
+			let layer_pos = self.layer_ids.iter().position(|x| *x == *layer_id).ok_or(DocumentError::LayerNotFound)?;
+			if (layer_pos as i32 - last_pos as i32).abs() > 1 {
+				// Selection is not contiguous
+				return Err(DocumentError::InvalidPath);
```
TrueDoctor, line 78: It's probably best to add a separate error code for that

core/document/src/layers/folder.rs
TrueDoctor: no, this would not do anything, the cache of the folder contaning the updats has to be updated

core/document/src/layers/folder.rs
```diff
@@ -65,6 +65,84 @@ impl Folder {
...
+			}
+		} else {
+			// Dragging down
+
+			// Prevent shifting past end
+			if source_pos == 0 {
+				return Err(DocumentError::InvalidPath);
+			}
+
+			fn rearrange<T>(arr: &mut Vec<T>, source_pos: usize, source_pos_end: usize, target_pos: usize)
```
TrueDoctor, line 120: yeah we can at least reduce the code length of this function by half to make it more readable

core/editor/src/document/document_message_handler.rs
```diff
@@ -136,21 +137,23 @@ impl DocumentMessageHandler {
...
+			// 'path.len() > 0' filters out root layer since it has no indices
+			.filter_map(|(path, data)| (!path.is_empty() && !only_selected || data.selected).then(|| path.clone()))
 			.filter_map(|path| {
 				// Currently it is possible that layer_data contains layers that are don't actually exist
 				// and thus indices_for_path can return an error. We currently skip these layers and log a warning.
 				// Once this problem is solved this code can be simplified
 				match self.active_document().document.indices_for_path(&path) {
 					Err(err) => {
-						warn!("selected_layers_sorted: Could not get indices for the layer {:?}: {:?}", path, err);
+						warn!("layers_sorted: Could not get indices for the layer {:?}: {:?}", path, err);
```
TrueDoctor, line 156: Oh and just fyi you should be able to use clippy now

TrueDoctor (changes requested):

core/document/src/layers/folder.rs
ajweeks: That isn't equivalent code, the slice excludes the first element intentionally. If there's a better way of doing that I'm all ears.

core/document/src/layers/folder.rs
ajweeks: Good point

core/document/src/layers/folder.rs
ajweeks: Cool, I wasn't sure but thought the same.

core/document/src/layers/folder.rs
ajweeks: I'm not quite sure what you mean?

core/editor/src/document/document_message_handler.rs
ajweeks: Yep, thanks! I haven't seen any warnings appearing on these changes.

ajweeks:
> Looks great, thanks for adding the tests! (This is the kind of thing that can be trickier to get right so testing is indeed helpful). My only comment about the tests is just a nitpick about the comments: it took me about a minute of understand what you meant by `''` until I realized that's the "same as above" indicator, my mind was thinking along the lines of single-quoted strings or anything else. If you get the chance, could you consider moving the comment to the line above the one it refers to (typically it's better stylistic practice for descriptive comments, rather than TODOs, to be on its own line) so the comment refers to the block of following lines, and the `''` mark is therefore unnecessary? I really don't intend to be a stickler over comments or style like that, so please feel more than free to ignore this comment, it's truly a nitpick I'm totally happy to ignore.

Yep, that's totally fair, I agree it wasn't the clearest.

core/document/src/layers/folder.rs
TrueDoctor: I suppose you could call `skip(1)` but at that point it might get longer than your current code.
Oh, wait, you are indexing source ids with source ids, in that case it should either be `source_ids.skip(1)` or &source_ids[1..]

TrueDoctor: The nice thing about the skip is, that it does not panic if the list is empty but that depends on what you want the code to do

core/document/src/document.rs
```diff
@@ -229,8 +229,16 @@ impl Document {
...
-		self.root.as_folder_mut()?.reorder_layers(source_layer_ids, *target_path.last().unwrap())?;
+		let mut source_layer_ids = Vec::with_capacity(source_paths.len());
+		for path in source_paths {
+			let result = path.last();
+			match result {
+				Some(result) => source_layer_ids.push(*result),
+				None => return Err(DocumentError::LayerNotFound),
+			}
+		}
+
```
TrueDoctor, line 240: You can actually do something fancy here: 
```rust
let source_layer_ids = source_paths.iter().map(|x| *x.last().ok_or(DocumentError::LayerNotFound)).collect()?;
```

TrueDoctor: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect

core/document/src/layers/folder.rs
```diff
@@ -59,36 +59,36 @@ impl Folder {
 	}
 
 	pub fn remove_layer(&mut self, id: LayerId) -> Result<(), DocumentError> {
-		let pos = self.layer_ids.iter().position(|x| *x == id).ok_or(DocumentError::LayerNotFound)?;
+		let pos = self.position_of_layer(id).ok_or(DocumentError::LayerNotFound)?;
```
TrueDoctor, line 62: I would probably let the function return an error directly but that is probably personal preference

core/document/src/layers/folder.rs
```diff
@@ -153,15 +141,19 @@ impl Folder {
 	}
 
 	pub fn layer(&self, id: LayerId) -> Option<&Layer> {
-		let pos = self.layer_ids.iter().position(|x| *x == id)?;
+		let pos = self.position_of_layer(id)?;
```
TrueDoctor, line 144: I would probably use `ok()` here and return a result to reduce code duplication

core/document/src/document.rs
ajweeks: I tried doing something similar, I couldn't get it to work though. The code you supplied doesn't compile. I tried a few approaches again just now but couldn't find a solution, perhaps because I'm rather new to rust 😄 If you can find a more idiomatic solution I'd take it. 🙂

core/document/src/layers/folder.rs
ajweeks: @TrueDoctor if you thought rearrange was the same in both blocks, they're actually slightly different. I've renamed them to make that more clear. Or is there something else I'm missing?

core/document/src/layers/folder.rs
ajweeks: Ah I see. The caller definitely shouldn't be passing in an empty list (moving a list of zero layers makes no sense).

I think `&source_ids[1..]` is probably the best solution here.

core/document/src/layers/folder.rs
ajweeks: Yep, that works.

core/document/src/layers/folder.rs
ajweeks: That's what I had at first, but I made the function more generic so it could be used in more places. I'll change it back though if you think it's better.

PR #273, opened 2021-07-16
Add a document method that returns a layer's bounding box

### LINKED ISSUE #151: Unify bounding box for shapes during creation
Provide a mechanism to calculate and update the bounding box of a shape while drawing it.
This will honour things like the Center or LockAspectRatio modifiers.

If done before #148, there will be more to refactor in that issue. However this is higher priority.

Complexity: 2
Involves: Editor (Rust)

AyeTbk:
Closes #151

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/273)

### DIFF
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -227,6 +227,11 @@ impl Document {
 		Ok(())
 	}
 
+	pub fn layer_bounding_box(&self, path: &[LayerId]) -> Result<[DVec2; 2], DocumentError> {
+		let layer = self.layer(path)?;
+		Ok(layer.bounding_box(self.root.transform * layer.transform, layer.style))
+	}
+
 	fn mark_as_dirty(&mut self, path: &[LayerId]) -> Result<(), DocumentError> {
 		let mut root = &mut self.root;
 		root.cache_dirty = true;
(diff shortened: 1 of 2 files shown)


core/document/src/layers/mod.rs
```diff
@@ -100,6 +100,13 @@ impl LayerDataTypes {
 			}
 		}
 	}
+
+	pub fn bounding_box(&self, transform: glam::DAffine2, style: style::PathStyle) -> [DVec2; 2] {
+		use kurbo::Shape;
```
henryksloan, line 105: Move `use` to the top of the file.

henryksloan (approved):
LGTM

core/document/src/layers/mod.rs
AyeTbk: done

core/document/src/document.rs
```diff
@@ -227,6 +227,11 @@ impl Document {
 		Ok(())
 	}
 
+	pub fn layer_bounding_box(&self, path: &[LayerId]) -> Result<[DVec2; 2], DocumentError> {
```
Keavon, line 230: Please rename `layer_axis_aligned_bounding_box` and create another function called `layer_local_bounding_box` which returns the bounding box of the layer's own rotation.

AyeTbk: Done

Keavon (approved):

PR #278, opened 2021-07-17
Apply clippy lints

TrueDoctor:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/278)

### DIFF
--- a/core/document/src/layers/line.rs
+++ b/core/document/src/layers/line.rs
@@ -11,12 +11,12 @@ use super::LayerData;
 use serde::{Deserialize, Serialize};
 use std::fmt::Write;
 
-#[derive(Debug, Clone, Copy, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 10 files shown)


core/document/src/layers/line.rs
```diff
@@ -11,12 +11,12 @@ use super::LayerData;
 use serde::{Deserialize, Serialize};
 use std::fmt::Write;
 
-#[derive(Debug, Clone, Copy, PartialEq, Deserialize, Serialize)]
-pub struct Line {}
+#[derive(Debug, Default, Clone, Copy, PartialEq, Deserialize, Serialize)]
+pub struct Line;
 
 impl Line {
 	pub fn new() -> Line {
```
0HyperCube, line 18: We don't need new() as we have default

0HyperCube (approved):
Looks good

PR #287, opened 2021-07-21
Add documentation

RustyNixieTube:
PS: it was unrelated to documentation but I have modified the `message_prelude` because using `super` when `super` is the root of the crate is a bit odd

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/287)

### DIFF
--- a/core/document/src/color.rs
+++ b/core/document/src/color.rs
@@ -20,7 +20,7 @@ impl Color {
 	pub const GREEN: Color = Color::from_unsafe(0., 1., 0.);
 	pub const BLUE: Color = Color::from_unsafe(0., 0., 1.);
 
-	/// Return Some(Color) if `red`, `green`, `blue` and `alpha` have a valid value. Negative number (including `-0.0`), `f32::NAN` and infinity are not valid value and return `None`.
+	/// Returns `Some(Color)` if `red`, `green`, `blue` and `alpha` have a valid value. Negatives numbers (including `-0.0`), NaN and infinity are not valid values and return `None`.
 	/// Values greater than `1.0` for alpha are not valid.
 	/// # Examples
 	/// ```
--- a/core/document/src/color.rs
+++ b/core/document/src/color.rs
@@ -1,10 +1,9 @@
 use serde::{Deserialize, Serialize};
 
-/// Structure that represent a color\
-/// internally alpha is stored as f32 that go from 0.0 (transparent) to 1.0
-/// the other components (rgb) are stored as f32 that go from 0.0 up to f32::MAX,
-/// the value are proportinal to the number of photons hitting the camera sensor
-/// if we have two different cameras with two different photon measuring capabilities, the reference point is the highest photon measuring capabilities
+/// Structure that represent a color.
+/// Internally alpha is stored as `f32` that range from `0.0` (transparent) to 1.0 (opaque).
+/// The other components (RGB) are stored as `f32` that range from `0.0` up to `f32::MAX`,
+/// the values encode the brightness of each channel proportional to the light intensity in cd/m² (nits) in HDR, and 0.0 (black) to 1.0 (white) in SDR color.
 #[repr(C)]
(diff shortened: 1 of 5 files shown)


core/document/src/color.rs
```diff
@@ -1,5 +1,10 @@
 use serde::{Deserialize, Serialize};
 
+/// Structure that represent a color\
```
Keavon, line 3: Replace the trailing `\` with a period.

core/document/src/color.rs
```diff
@@ -1,5 +1,10 @@
...
+/// internally alpha is stored as f32 that go from 0.0 (transparent) to 1.0
```
Keavon, line 4: Capitalize the start of the new sentence. Replace "go" with "range" and add "(opaque)." after "1.0".

core/document/src/color.rs
```diff
@@ -1,5 +1,10 @@
...
+/// the other components (rgb) are stored as f32 that go from 0.0 up to f32::MAX,
```
Keavon, line 5: Capitalize the start of the new sentence (I'll stop mentioning this now, but please end each sentence with a period and start the next one with an uppercase letter, unless the entire comment is only one sentence then it doesn't need a period but should still start capitalized).

Uppercase "rgb" -> "RGB". Probably also a good idea to put all the `f32`, `0.0`, `f32::MAX`, etc. values in backticks to make those inline code.

core/document/src/color.rs
```diff
@@ -1,5 +1,10 @@
...
+/// the value are proportinal to the number of photons hitting the camera sensor
```
Keavon, line 6: For better clarity about the concept, would you mind switching this sentence to the following?:

> the values encode the brightness of each channel proportional to the light intensity in cd/m² (nits) in HDR, and `0.0` (black) to `1.0` (white) in SDR color.

core/document/src/color.rs
```diff
@@ -16,21 +21,51 @@ impl Color {
 	pub const GREEN: Color = Color::from_unsafe(0., 1., 0.);
 	pub const BLUE: Color = Color::from_unsafe(0., 0., 1.);
 
+	/// return Some(Color) if `red`, `green`, `blue` and `alpha` have a valid value. Negative number (including -0.0), Nan and infinity are not valid value and return None.
+	/// Value greater than 1.0 for alpha are not valid
```
Keavon, line 25: Value -> Values

core/document/src/color.rs
```diff
@@ -16,21 +21,51 @@ impl Color {
 	pub const GREEN: Color = Color::from_unsafe(0., 1., 0.);
 	pub const BLUE: Color = Color::from_unsafe(0., 0., 1.);
 
+	/// return Some(Color) if `red`, `green`, `blue` and `alpha` have a valid value. Negative number (including -0.0), Nan and infinity are not valid value and return None.
```
Keavon, line 24: "Return" -> "Returns". Put `Some(Color)` in backticks. "Negative number" -> "Negative numbers". Put `-0.0` in backticks. Should "Nan" be "`f32::NAN`"? Otherwise "Nan" -> "NaN". "not valid value" -> "not valid values". Backtick "None".

core/renderer/src/lib.rs
```diff
@@ -1,3 +1,9 @@
+//! Graphite Renderer Core Library: `/core/renderer/`  
+//! A stateless library (with the help of in-memory and/or on-disk caches for performance) for rendering Graphite's render graph (GRD) files.
+//! The official Graphite CLI and Document Core Library are the primary users,
+//! but this library is intended to be useful to any application that wants to link the library for the purpose of rendering Graphite's render graphs.
+//! For example, games can link the library and render procedural textures with customizable parametric input values.
+
```
TrueDoctor, line 6: we will probably remove the renderer as it was planned because its functionality will be integrated into the nodegraph

TrueDoctor (approved):

PR #288, opened 2021-07-22
Implement tool messaging and shape flipping

### LINKED ISSUE #199: Flip shapes operator
When a layer is selected, the flip operators should allow them to be flipped horizontally and vertically using these buttons in the Select Tool: This can be achieved by multiplying the scale by -1.

henryksloan:
Closes #199

- Added a WASM-bound `send_tool_message` function that takes a `ToolType` and a `&JsValue`
    - It (safely) attempts to deserialize the `JsValue` into a `ToolMessage` associated with the given `ToolType`
- Added an optional `onClick` prop to `IconButton`
- Added an optional `message` field to `IconButtonOption`
    - It dispatches `send_tool_message` with the active tool
- Added a `FlipLayer(path, flip_horizontal, flip_vertical)` to `DocumentMessage `
- Added `FlipHorizontal` and `FlipVertical` messages to `SelectMessage`
    - Adding to a tool's message enum is the chosen approach for tool messages
    - Handling messages in the `transition` function allows for state-dependent message handling
- Fixed typos and wording for `set_tool_options`

## Caveats
The basic scale-matrix implementation of flipping means that flips are with respect to the shape's origin (generally the initial point it's drawn from). This is arguably desired behavior, but users may expect the operations to depend on the shape's "center of mass" or something similar.

Any suggestions to clean up the following? I tried to use an or-pattern, but the borrow checker prevents us from subsequently checking the variant within the block.
```rust
(_, FlipHorizontal) => {
	let selected_layers = document.layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone()));
	for path in selected_layers {
		responses.push_back(DocumentMessage::FlipLayer(path, true, false).into());
	}

	self
}
(_, FlipVertical) => {
	let selected_layers = document.layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone()));
	for path in selected_layers {
		responses.push_back(DocumentMessage::FlipLayer(path, false, true).into());
	}

	self
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/288)

### DIFF
--- a/client/web/wasm/src/document.rs
+++ b/client/web/wasm/src/document.rs
@@ -4,7 +4,7 @@ use crate::EDITOR_STATE;
 use editor_core::input::input_preprocessor::ModifierKeys;
 use editor_core::input::mouse::ScrollDelta;
 use editor_core::message_prelude::*;
-use editor_core::tool::tool_options::ToolOptions;
+use editor_core::tool::{tool_options::ToolOptions, tools, ToolType};
 use editor_core::{
 	input::mouse::{MouseState, ViewportPosition},
 	LayerId,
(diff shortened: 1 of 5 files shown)


Keavon (approved):

Keavon:
Approved for the web part. Hoping for @TrueDoctor to give his approval for the Rust part.

client/web/wasm/src/document.rs
```diff
@@ -30,12 +30,31 @@ pub fn set_tool_options(tool: String, options: &JsValue) -> Result<(), JsValue>
...
+			},
+			_ => Err(Error::new(&format!("Tool message sending not implemented for {}", tool)).into()),
+		},
+		None => Err(Error::new(&format!("Couldn't send message for {} because it was not recognized as a valid tool", tool)).into()),
+	};
+	EDITOR_STATE.with(|editor| match tool_message {
+		Ok(tool_message) => editor.borrow_mut().handle_message(tool_message).map_err(convert_error),
+		Err(err) => Err(err),
+	})
+}
```
TrueDoctor, line 56: What are the cons of just adding a  `dispatch_message(message: &JsValue)` ? (or `dispatch_tool_message`)

TrueDoctor (approved):

client/web/wasm/src/document.rs
henryksloan: I chose this approach because it allows the frontend to simply send a message to the active tool, without having to wrap it in `ToolMessage::Select { SelectMessage { ... } }`. For the specific use case of messaging, I think it makes sense to allows the frontend to just specify the name of the message, and the backend to handle wrapping and error handling as granularly as necessary. I feel that most cases of TS -> Rust messaging can and should be specific, otherwise constructing messages on the backend.

I'm not sure what you mean by `dispatch_tool_message`... do you mean I could rename this function, or change it somehow?

TrueDoctor: > I'm not sure what you mean by dispatch_tool_message... do you mean I could rename this function, or change it somehow?
removing one layer of wrapping basically, was just worried about the scalability of that approach but after all thee wasm-wrapper was never intended to be pretty so that should be fine.
Feel free to merge then

PR #292, opened 2021-07-22
Change tool option kinds to PascalCase

henryksloan:
As suggested by @Keavon

This naming scheme is more sensible, as each option "(Kind)Option" maps directly to "(Kind)".

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/292)


Keavon:
And additionally, these are equal to the names of the Vue components and their files, which removes unclarity.

Keavon (approved):

PR #293, opened 2021-07-22
Add dragging of layers with the select tool

### LINKED ISSUE #143: Select Tool: Grab/drag selected layers around viewport

henryksloan:
Closes #143

- Added a `DragLayer` message to `DocumentMessage` that accepts a path and an offset
- Modified the select tool's FSM to allow:
    - Dragging all selected layers
    - Dragging a single previously unselected layer
    - Drawing a quad and selecting everything within it (by starting the drag on the background)
    - Deselecting everything by clicking the background

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/293)


TrueDoctor (commented):
Could you add comments where you used the root transform? e.g.
// TODO: Replace with functions of the transform api
this will at least make it easier to fix it afterwards

Keavon:
This looks good to me at a cursory read-through, but the backend code is not my domain of expertise so I defer to True Doctor. Looks like this needs to be synced with master and then you have my blessing to merge if True Doctor is also ok with it.

Keavon (approved):

PR #295, opened 2021-07-24
Rename align operations and icons

henryksloan:
Just renaming the align operations to reduce verbosity and redundancy. The left/right/top/bottom aligns don't need to be labelled as vertical or horizontal. Additionally, I changed "Horizontal/Vertical Align Center" to "Align Horizontal/Vertical Center" so that each option's name starts with "Align", and so "horizontal center" and "vertical center" can be read as directions just like left/right/top/bottom.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/295)


Keavon (approved):
This looks like the best naming given the constraints.

PR #296, opened 2021-07-24
Add alignment of selected layers

### LINKED ISSUE #198: Alignment operations between shapes
When multiple layers are selected, the alignment operators should allow them to be moved to a common left/center/right/top/middle/bottom using these buttons in the Select Tool:

henryksloan:
Closes #198

- Added messages to alignment icon buttons within select tool
- Added a document message `SetLayerCoordinates` that takes a path and two `Option<f64>`
    - It translates the path by a given x (if `Some`) and y (if `Some`) offset
- Added and handled alignment messages to `SelectMessage`
- Added `align_selected` helper to `select.rs`
    - Accepts two new enums, `AlignDimension` and `AlignAggregate`
    - Aggregates a particular coordinate of the selected layers with min, max, or average
    - Uses `SetLayerCoordinates` to update all the selected layers, offsetting them to account for when bounding box point != origin

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/296)

### DIFF
--- a/client/web/src/components/widgets/options/ToolOptions.vue
+++ b/client/web/src/components/widgets/options/ToolOptions.vue
@@ -87,15 +87,15 @@ export default defineComponent({
 			[
 				"Select",
 				[
-					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left" },
-					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center" },
-					{ kind: "IconButton", icon: "AlignRight", title: "Align Right" },
+					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left", message: "AlignLeft" },
+					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: "AlignHorizontalCenter" },
+					{ kind: "IconButton", icon: "AlignRight", title: "Align Right", message: "AlignRight" },
 
 					{ kind: "Separator", type: SeparatorType.Unrelated },
 
-					{ kind: "IconButton", icon: "AlignTop", title: "Align Top" },
-					{ kind: "IconButton", icon: "AlignVerticalCenter", title: "Align Vertical Center" },
-					{ kind: "IconButton", icon: "AlignBottom", title: "Align Bottom" },
+					{ kind: "IconButton", icon: "AlignTop", title: "Align Top", message: "AlignTop" },
+					{ kind: "IconButton", icon: "AlignVerticalCenter", title: "Align Vertical Center", message: "AlignVerticalCenter" },
+					{ kind: "IconButton", icon: "AlignBottom", title: "Align Bottom", message: "AlignBottom" },
 
 					{ kind: "Separator", type: SeparatorType.Related },
 
--- a/client/web/src/components/widgets/options/ToolOptions.vue
+++ b/client/web/src/components/widgets/options/ToolOptions.vue
@@ -38,7 +38,7 @@ interface IconButtonOption {
 	kind: "IconButton";
 	icon: string;
 	title: string;
-	message?: string;
+	message?: string | object;
 }
 
 interface SeparatorOption {
@@ -87,15 +87,15 @@ export default defineComponent({
 			[
 				"Select",
 				[
-					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left", message: "AlignLeft" },
-					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: "AlignHorizontalCenter" },
-					{ kind: "IconButton", icon: "AlignRight", title: "Align Right", message: "AlignRight" },
+					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left", message: { Align: ["X", "Min"] } },
+					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: { Align: ["X", "Average"] } },
+					{ kind: "IconButton", icon: "AlignRight", title: "Align Right", message: { Align: ["X", "Max"] } },
 
 					{ kind: "Separator", type: SeparatorType.Unrelated },
 
-					{ kind: "IconButton", icon: "AlignTop", title: "Align Top", message: "AlignTop" },
-					{ kind: "IconButton", icon: "AlignVerticalCenter", title: "Align Vertical Center", message: "AlignVerticalCenter" },
-					{ kind: "IconButton", icon: "AlignBottom", title: "Align Bottom", message: "AlignBottom" },
+					{ kind: "IconButton", icon: "AlignTop", title: "Align Top", message: { Align: ["Y", "Min"] } },
+					{ kind: "IconButton", icon: "AlignVerticalCenter", title: "Align Vertical Center", message: { Align: ["Y", "Average"] } },
+					{ kind: "IconButton", icon: "AlignBottom", title: "Align Bottom", message: { Align: ["Y", "Max"] } },
 
 					{ kind: "Separator", type: SeparatorType.Related },
 
--- a/client/web/src/components/widgets/options/ToolOptions.vue
+++ b/client/web/src/components/widgets/options/ToolOptions.vue
@@ -88,13 +88,13 @@ export default defineComponent({
 				"Select",
 				[
 					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left", message: { Align: ["X", "Min"] } },
-					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: { Align: ["X", "Average"] } },
+					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: { Align: ["X", "Center"] } },
 					{ kind: "IconButton", icon: "AlignRight", title: "Align Right", message: { Align: ["X", "Max"] } },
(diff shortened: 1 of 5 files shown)


core/editor/src/document/document_message_handler.rs
```diff
@@ -662,6 +663,24 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					responses.push_back(DocumentOperation::SetLayerTransform { path, transform }.into());
 				}
 			}
+			SetLayerCoordinates(path, x_option, y_option) => {
```
Keavon, line 666: Do you think "Position" or "Location" would be less ambiguous or more standardized compared to "Coordinates"?

core/editor/src/tool/tools/select.rs
```diff
@@ -162,6 +168,36 @@ impl Fsm for SelectToolFsmState {
...
+				(_, AlignVerticalCenter) => {
+					align_selected(document, responses, AlignDimension::Y, AlignAggregate::Average);
+
+					self
+				}
+				(_, AlignBottom) => {
+					align_selected(document, responses, AlignDimension::Y, AlignAggregate::Max);
+
+					self
+				}
```
Keavon, line 200: I'm surprised we would need to add those button actions to the select tool's FSM. Why doesn't the button just send a message directly to the document? I'd be concerned that this approach of involving the tool's FSM could cause additional complication if it isn't necessary. But I am also not the most familiar with this part of the code, so if you think it's necessary, I'd be delighted to learn why to help me understand this code better.

core/editor/src/tool/tools/select.rs
```diff
@@ -186,6 +222,60 @@ impl Fsm for SelectToolFsmState {
...
+	Y,
+}
+
+enum AlignAggregate {
+	Min,
+	Max,
+	Average,
+}
+
+fn align_selected(document: &Document, responses: &mut VecDeque<Message>, dimension: AlignDimension, aggregate: AlignAggregate) {
```
Keavon, line 236: It definitely seems like this should be a document function, because we should be able to call it from other parts of the codebase (from the file menu, from a future command search panel like in Blender, from a future console, etc.) without the Select Tool active.

core/editor/src/tool/tools/select.rs
```diff
@@ -186,6 +222,60 @@ impl Fsm for SelectToolFsmState {
...
+	let selected_paths: Vec<Vec<LayerId>> = document.layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())).collect();
+	let n_selected = selected_paths.len();
+	if n_selected == 0 {
+		return;
+	}
```
Keavon, line 241: Opinionated style nitpick: why not just skip assigning this to a variable since it's only used here and once again down below? `selected_paths.len()` is nearly the same length as `n_selected`. Ignore this feedback if you have to make a special trip to edit this file.

core/editor/src/document/document_message_handler.rs
henryksloan: Good point, going with "Translation" as it matches the transform terms

core/editor/src/tool/tools/select.rs
henryksloan: Agreed, done

core/editor/src/tool/tools/select.rs
henryksloan: Fair enough, changed

core/editor/src/tool/tools/select.rs
henryksloan: This is the approach that @TrueDoctor and I converged on. A big advantage of it is that button message handling can depend on state (in a semantically clear way), which we think may be a desirable use case. Similarly, button messages could easily act on state, e.g. aborting a drag action if a certain action is done (maybe by a keyboard shortcut). It also keeps all of a tool's message handling in one place, and keeps every message within the FSM metaphor (button presses generally just being self-transitions).

I did substantially cut down on the code by making this a tuple variant, though.

core/editor/src/tool/tools/select.rs
```diff
@@ -24,10 +24,24 @@ pub enum SelectMessage {
 	MouseMove,
 	Abort,
 
+	Align(AlignDimension, AlignAggregate),
 	FlipHorizontal,
 	FlipVertical,
 }
 
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub enum AlignDimension {
```
Keavon, line 33: I think "Axis" would be a more accurate word.

core/editor/src/document/document_message_handler.rs
```diff
@@ -638,6 +641,47 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+				}
+
+				let selected_layers = selected_paths.iter().map(|path| {
+					let layer = self.active_document().document.layer(path).unwrap();
+					let point = {
+						let bounding_box = layer.bounding_box(layer.transform, layer.style).unwrap();
+						match aggregate {
+							AlignAggregate::Min => bounding_box[0],
+							AlignAggregate::Max => bounding_box[1],
+							AlignAggregate::Average => bounding_box[0].lerp(bounding_box[1], 0.5),
```
Keavon, line 657: Average means it's adding up all the center coordinates of all the shapes then taking the mean value and setting the centers of each to that mean value, which is useful but is not the normal idea of "Center" in graphics software. "Average" will be another button [accessible in popover](https://files.keavon.com/-/BulkyAltruisticMonarch/capture.png) to the right of these alignment buttons. So please keep the `Average` case but add a `Center` case above it which centers everything in the center of the union bounding box of all shape bounding boxes. So you do what you've done here with the `Min` and `Max` case by using `reduce()` to find the min and max of all the bounding boxes, then you take the center (sum divided by two) of the min and max values.

core/editor/src/document/document_message_handler.rs
```diff
@@ -638,6 +641,47 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+
+				let bounding_box_coords = selected_layers.clone().map(|(_, bounding_box_coord, _)| bounding_box_coord);
+				let aggregated_coord = match aggregate {
+					AlignAggregate::Min => bounding_box_coords.reduce(|a, b| a.min(b)).unwrap(),
+					AlignAggregate::Max => bounding_box_coords.reduce(|a, b| a.max(b)).unwrap(),
+					AlignAggregate::Average => bounding_box_coords.sum::<f64>() / selected_paths.len() as f64,
+				};
+				for (path, bounding_box_coord, translation_coord) in selected_layers {
+					let new_coord = match aggregate {
+						AlignAggregate::Min => aggregated_coord - (bounding_box_coord - translation_coord),
```
Keavon, line 675: Maybe break the triple-repeated `bounding_box_coord - translation_coord)` out into a variable?

client/web/src/components/widgets/options/ToolOptions.vue
```diff
@@ -87,15 +87,15 @@ export default defineComponent({
 			[
 				"Select",
 				[
-					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left" },
-					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center" },
-					{ kind: "IconButton", icon: "AlignRight", title: "Align Right" },
+					{ kind: "IconButton", icon: "AlignLeft", title: "Align Left", message: { Align: ["X", "Min"] } },
+					{ kind: "IconButton", icon: "AlignHorizontalCenter", title: "Align Horizontal Center", message: { Align: ["X", "Average"] } },
```
Keavon, line 91: Please change `"Average"` to `"Center"` (see below comment regarding this).

core/editor/src/document/document_message_handler.rs
henryksloan: Ah, you made me realize that all this math is redundant haha.

core/editor/src/document/document_message_handler.rs
```diff
@@ -638,6 +641,69 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+							.unwrap();
+						let max = bounding_boxes
+							.clone()
+							.map(|bbox| match axis {
+								AlignAxis::X => bbox[1].x,
+								AlignAxis::Y => bbox[1].y,
+							})
+							.reduce(|a, b| a.max(b))
+							.unwrap();
+						(min + max) / 2.
```
TrueDoctor, line 709: [(0, max),(1,min)].iter().map().…..sum() / 2;

core/editor/src/document/document_message_handler.rs
```diff
@@ -652,6 +718,24 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					responses.push_back(DocumentOperation::SetLayerTransform { path, transform }.into());
 				}
 			}
+			SetLayerTranslation(path, x_option, y_option) => {
+				if let Ok(layer) = self.active_document_mut().document.layer_mut(&path) {
+					let mut transform = layer.transform;
```
TrueDoctor, line 737: you could probably do something like 
```rust
transform.translation = (x_option.unwrap_or(translate.x), y_option.unwrap_or(translate_y));
```

TrueDoctor: not sure if just codegolfing at this point

Keavon (approved):

PR #300, opened 2021-07-24
MVP eyedropper tool for fill colors

### LINKED ISSUE #154: Full Eyedropper Tool implementation based on pixel sampling
Clicking somewhere samples the color of the pixel under the cursor and sets the primary working color to that. Left click to sample for the primary color, right click to sample for the secondary color.

Since it isn't possible to sample colors from a web page using JavaScript, the simplest approach for this is to probably use the browser's collision detection for the SVG shapes to determine which SVG element is being clicked on, then take either its fill or stroke color. However this approach breaks down for shapes with both a fill and a stroke, and shapes with a gradient or non-solid color. Further investigation is required for solutions that support those edge cases, but one possibility would be switching from rendering via the SVG element to rendering with a 2D canvas using the [Path2D](https://developer.mozilla.org/en-US/docs/Web/API/Path2D/Path2D) API for drawing SVG syntax to the canvas. Canvases can have their color sampled from.

A simpler MVP solution is to just check the color or stroke property of the shape clicked in the viewport, similar to #178. That MVP is filed under #303 and was merged in #300.

### LINKED ISSUE #303: Eyedropper Tool MVP
Let's sample the fill color from the location where clicked. This approach does not work with stroke, or shapes with both stroke and fill. The full feature will be implemented in #154 based on pixel sampling.

henryksloan:
Closes #303.

This was marked for Sprint 7, but I realized this MVP would be a quick and useful addition.

- Enabled eyedropper tool on the tool shelf
- Added getters to `Fill` and `Stroke`
- Added `Lmb => MouseDown` input mapping for the eyedropper tool
- Implemented an MVP for the eyedropper tool
    - Just checks for the layer fill of the topmost layer intersecting the click
    - Updates the primary color

Caveat: The color update is implemented using `FrontendMessage::UpdateWorkingColors` as opposed to `SelectPrimaryColor`. That is because using the former doesn't update colors on the frontend. Adding `UpdateWorkingColors` to the handler for `SelectPrimaryColor` causes an unresponsive infinite-loop on the frontend. This is because `SwatchPairInput` registers a response handler for `ResponseType.UpdateWorkingColors` **but also** `watch`es `primaryColor` and `secondaryColor`, mapping changes to `SwatchPairInput.update{Primary | Secondary}Color`. Those two functions each the respective WASM functions, which in turn dispatch the color selection messages, causing an infinite loop.

In the future, I think this infinite loop should be avoided by removing the `watch`es and instead add callbacks to `ColorPicker`. We should only be dispatching the `Select...Color` messages when the frontend (i.e. the user) updates the colors. In fact, as it currently is, the response handler for `UpdateWorkingColors` always dispatches redundant calls to `SelectPrimaryColor` and `SelectSecondaryColor`.

I could implement that fix, but I think we should leave it for another PR to work out details.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/300)

### DIFF
--- a/core/editor/src/input/input_mapper.rs
+++ b/core/editor/src/input/input_mapper.rs
@@ -183,6 +183,7 @@ impl Default for Mapping {
 			entry! {action=ToolMessage::SelectTool(ToolType::Line), key_down=KeyL},
 			entry! {action=ToolMessage::SelectTool(ToolType::Pen), key_down=KeyP},
 			entry! {action=ToolMessage::SelectTool(ToolType::Shape), key_down=KeyY},
+			entry! {action=ToolMessage::SelectTool(ToolType::Eyedropper), key_down=KeyI},
 			entry! {action=ToolMessage::ResetColors, key_down=KeyX, modifiers=[KeyShift, KeyControl]},
 			entry! {action=ToolMessage::SwapColors, key_down=KeyX, modifiers=[KeyShift]},
 			// Document Actions
--- a/core/editor/src/input/input_mapper.rs
+++ b/core/editor/src/input/input_mapper.rs
@@ -122,6 +122,8 @@ impl Default for Mapping {
 			entry! {action=SelectMessage::DragStop, key_up=Lmb},
 			entry! {action=SelectMessage::Abort, key_down=Rmb},
 			entry! {action=SelectMessage::Abort, key_down=KeyEscape},
+			// Eyedropper
+			entry! {action=EyedropperMessage::MouseDown, key_down=Lmb},
 			// Rectangle
 			entry! {action=RectangleMessage::Center, key_down=KeyAlt},
 			entry! {action=RectangleMessage::UnCenter, key_up=KeyAlt},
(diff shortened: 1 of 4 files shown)


Keavon:
> This was marked for Sprint 7, but I realized this MVP would be a quick and useful addition.

Tasks from upcoming sprints are always on the table! I was also debating with myself if we should even include it in 6 or 7. Thanks for grabbing it.

TrueDoctor (approved):
LGTM

core/editor/src/input/input_mapper.rs
```diff
@@ -122,6 +122,8 @@ impl Default for Mapping {
 			entry! {action=SelectMessage::DragStop, key_up=Lmb},
 			entry! {action=SelectMessage::Abort, key_down=Rmb},
 			entry! {action=SelectMessage::Abort, key_down=KeyEscape},
+			// Eyedropper
+			entry! {action=EyedropperMessage::MouseDown, key_down=Lmb},
```
Keavon, line 126: We also need to add one for the <kbd>I</kbd> key to select the tool.

Keavon (approved):

Keavon:
Can you please add right click mapping to the secondary color? So left click samples into the primary color while right click samples into the secondary color.

PR #305, opened 2021-07-25
Avoid looping behavior with color selection

### LINKED ISSUE #304: Frontend color selection response is prone to infinite loops
From #300:

The [eyedropper tool] color update is implemented using FrontendMessage::UpdateWorkingColors as opposed to SelectPrimaryColor. That is because using the former doesn't update colors on the frontend. Adding UpdateWorkingColors to the handler for SelectPrimaryColor causes an unresponsive infinite-loop on the frontend. This is because SwatchPairInput registers a response handler for ResponseType.UpdateWorkingColors but also watches primaryColor and secondaryColor, mapping changes to SwatchPairInput.update{Primary | Secondary}Color. Those two functions each the respective WASM functions, which in turn dispatch the color selection messages, causing an infinite loop.
In the future, I think this infinite loop should be avoided by removing the watches and instead add callbacks to ColorPicker. We should only be dispatching the Select...Color messages when the frontend (i.e. the user) updates the colors. In fact, as it currently is, the response handler for UpdateWorkingColors always dispatches redundant calls to SelectPrimaryColor and SelectSecondaryColor.

Once the looping tendency is fixed, these select messages should also update the frontend colors.

henryksloan:
Fixes #304

- Removed `watch` declarations for swatch colors
- Allow response handler to update frontend colors
    - This was previously handled by the `update` methods through `watch`, causing loopy pitfalls
- Refactor frontend color updates to a `update_working_colors` helper function
- Update frontend color updates when `SelectPrimaryColor` or `SelectSecondaryColor` is called

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/305)

### DIFF
--- a/client/web/src/components/widgets/inputs/SwatchPairInput.vue
+++ b/client/web/src/components/widgets/inputs/SwatchPairInput.vue
@@ -3,13 +3,13 @@
 		<div class="secondary swatch">
 			<button @click="clickSecondarySwatch" ref="secondaryButton" data-hover-menu-spawner></button>
 			<FloatingMenu :type="MenuType.Popover" :direction="MenuDirection.Right" horizontal ref="secondarySwatchFloatingMenu">
-				<ColorPicker v-model:color="secondaryColor" />
+				<ColorPicker @update:color="secondaryColorChanged" :color="secondaryColor" />
 			</FloatingMenu>
 		</div>
 		<div class="primary swatch">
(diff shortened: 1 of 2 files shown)


Keavon (approved):
Excellent! I'm going to try and go through most of our input widgets and update them to use the `@update:foo="handlerFunction"` approach instead of using watchers, which is clearly a dangerous approach (it gave me hours of troubles with the dropdown menu widget also). Thanks for fixing this.

PR #312, opened 2021-07-27
Layer opacity

### LINKED ISSUE #187: Hook up Opacity and Blend Mode layer settings
Take a look at [mix-blend-mode](https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode) and [opacity](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/opacity). Changing the blend mode and opacity widgets should send a message to the Rust backend requesting to set the blend mode or opacity to the new value for the selected layers. As a followup feature, it should also display the value for the selected layer(s) (with a dash where there is a mismatch between multiple).

- [x] Blend modes (#252)
- [ ] Opacity

Keavon:
Closes #187

Should be reviewed *after* #311 is merged, otherwise you will be reviewing the changes for both PRs at once.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/312)

### DIFF
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -418,6 +418,14 @@ impl Document {
 
 				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::FolderChanged { path }])
 			}
+			Operation::SetLayerOpacity { path, opacity } => {
+				self.mark_as_dirty(path)?;
+				self.layer_mut(&path).unwrap().opacity = *opacity;
+
+				let path = path.as_slice()[..path.len() - 1].to_vec();
+
+				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::FolderChanged { path }])
+			}
 			Operation::FillLayer { path, color } => {
 				let layer = self.layer_mut(path).unwrap();
 				layer.style.set_fill(layers::style::Fill::new(*color));
--- a/core/document/src/document.rs
+++ b/core/document/src/document.rs
@@ -412,7 +412,7 @@ impl Document {
 			}
 			Operation::SetLayerBlendMode { path, blend_mode } => {
 				self.mark_as_dirty(path)?;
-				self.layer_mut(&path).unwrap().blend_mode = *blend_mode;
+				self.layer_mut(path)?.blend_mode = *blend_mode;
 
 				let path = path.as_slice()[..path.len() - 1].to_vec();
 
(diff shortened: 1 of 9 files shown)


henryksloan (approved):
LGTM

core/editor/src/document/document_message_handler.rs
```diff
@@ -365,6 +366,14 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					responses.push_back(DocumentOperation::SetLayerBlendMode { path: path.clone(), blend_mode }.into());
 				}
 			}
+			SetOpacityForSelectedLayers(opacity) => {
+				let opacity = opacity.clamp(0., 1.);
+				let active_document = self.active_document();
+
+				for path in active_document.layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path)) {
```
henryksloan, line 373: It would be a little nicer to do `data.selected.then(|| path.clone)` then later `DocumentOperation::SetLayerOpacity { path, opacity }`... if that would work with the borrow checker

henryksloan (approved):
LGTM

core/document/src/document.rs
```diff
@@ -418,6 +418,14 @@ impl Document {
 
 				Some(vec![DocumentResponse::DocumentChanged, DocumentResponse::FolderChanged { path }])
 			}
+			Operation::SetLayerOpacity { path, opacity } => {
+				self.mark_as_dirty(path)?;
+				self.layer_mut(&path).unwrap().opacity = *opacity;
```
TrueDoctor, line 423: don't use `unwrap()` if there is a realistic chance the given path is not valid, use `?` instead

core/editor/src/document/document_message_handler.rs
TrueDoctor: I think you would then have to chain it with a drain(..). I don't mind the current approach

Keavon: Happy to change it, but I'm not the best person to judge which is best.

Keavon: It compiles as suggested, without requiring `drain`.

henryksloan: Could you post the now-working code so @TrueDoctor can give his thoughts?

Keavon:
```rust
for path in active_document.layer_data.iter().filter_map(|(path, data)| data.selected.then(|| path.clone())) {
	responses.push_back(DocumentOperation::SetLayerOpacity { path, opacity }.into());
}
```

TrueDoctor: > I think you would then have to chain it with a drain(..). I don't mind the current approach

Ah nevermind, it might just work. when using the clone in the iterator you reduce the code length since you then can get rid of the `path: `

PR #282, opened 2021-07-20
Add a bounding box around selected layers in the Select tool

AyeTbk:
Closes #207

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/282)

PR #314, opened 2021-07-28
Add a selected_layers() function and refactor code to use it

Keavon:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/314)

### DIFF
--- a/core/editor/src/document/document_message_handler.rs
+++ b/core/editor/src/document/document_message_handler.rs
@@ -707,13 +707,12 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 			}
 			AlignSelectedLayers(axis, aggregate) => {
 				// TODO: Handle folder nested transforms with the transforms API
-				let selected_layers = self.selected_layers().cloned().peekable();
-				if selected_layers.peek().is_none() {
+				if self.selected_layers().next().is_none() {
 					return;
 				}
 
-				let selected_layers = selected_layers.filter_map(|path| {
-					let layer = self.active_document().document.layer(&path).unwrap();
+				let selected_layers = self.selected_layers().cloned().filter_map(|path| {
+					let layer = self.active_document().document.layer(&path).ok()?;
 					let point = {
 						let bounding_box = layer.bounding_box(layer.transform, layer.style)?;
 						match aggregate {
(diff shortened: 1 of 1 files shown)


Keavon:
I believe we also wanted, in this PR, to change `selected_layers_sorted` to not create a new Vec. @TrueDoctor

TrueDoctor:
> I believe we also wanted, in this PR, to change `selected_layers_sorted` to not create a new Vec. @TrueDoctor

We actually can't because, as the name suggests, we need to sort the layers which is not possible without collecting them first

AyeTbk:
Should these methods (`selected_layers_sorted` and `selected_layers`) be in document_file.rs, on the `Document` type, instead of in document_message_handler.rs? To render the selection's bounding box in #282, the selected layers are needed (I currently compute them on my own, but I would like to reuse a prexisting method, if there is one). It also makes more sense to me for the `Document` to tell us what its selected layers are, instead of having code externalto it having to toy with its innards to determine them...

TrueDoctor:
That sounds like a good idea

Keavon:
> That sounds like a good idea

@TrueDoctor Is that a change you could make then merge this? I only just realized we hadn't merged this yet. I am not sure I understand exactly what change to make per AyeTbk's suggestions.

TrueDoctor:
> > That sounds like a good idea
> 
> @TrueDoctor Is that a change you could make then merge this? I only just realized we hadn't merged this yet. I am not sure I understand exactly what change to make per AyeTbk's suggestions.

Oh lol, I tought rhis code had already been merged xD

TrueDoctor:
I dont think this pr is conceptually the right place to make that change.
The pr does what it says on the label and that is a good thing

TrueDoctor:
> Should this method (`selected_layers`) be in document_file.rs, on the `Document` type, instead of in document_message_handler.rs? To render the selection's bounding box in #282, the selected layers are needed (I currently compute them on my own, but I would like to reuse a prexisting method, if there is one). It also makes more sense to me for the `Document` to tell us what its selected layers are, instead of having code external to it having to toy with its innards to determine them...
> (this goes for a lot of the methods in document_message_handler, I guess)

I think we should do a seperate pr in which we move lines 131-202 into the document_file but it makes sense to move them as a bunch

TrueDoctor:
I will address the movement to the document file + further refactoring in the upcoming transforms api

PR #315, opened 2021-07-29
Add folders to frontend and folder creation to backend

### LINKED ISSUE #149: Add folders in the Layer Tree panel of the frontend

Keavon:
Progress on #149

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/315)


TrueDoctor:
yeah, that function definitely did not fix it yet I can try to come up with something else, or we could try to plot some data and fit a curve through it

frontend/src/components/panels/LayerTree.vue
```diff
@@ -72,42 +82,96 @@
...
+
+				&::after {
+					content: "";
+					position: absolute;
+					width: 0;
+					height: 0;
+					top: 2px;
+					left: 3px;
+					border-style: solid;
+					border-width: 0 3px 6px 3px;
```
0HyperCube, line 116: I would have expected this to be pointed right and the expanded one to be pointed down (like in in vscode, windows file explorer etc)

frontend/src/components/panels/LayerTree.vue
```diff
@@ -25,22 +25,32 @@
 							:title="layer.visible ? 'Visible' : 'Hidden'"
 						/>
 					</div>
+					<button
+						v-if="layer.layer_type === LayerType.Folder"
+						class="node-connector"
+						:class="{ expanded: layer.layer_data.expanded }"
```
0HyperCube, line 31: This should be updated when the folder is expanded or collapsed (currently it is only updated when the layer tree modified again - by modifying the selection or by drawing or deleting a layer.

0HyperCube (commented):
Looks great; the only problems I could find where the rearranging layers to below a folder breaking the layer tree and the arrows and expanding folders not updating arrow direction. I would also say that multiple folders should be able to be expanded at once.

TrueDoctor:
> Looks great; the only problems I could find where the rearranging layers to below a folder breaking the layer tree and the arrows and expanding folders not updating arrow direction. I would also say that multiple folders should be able to be expanded at once.

Yeah we kinda gave up on implement proper rearranging in this pr and leave that for a follow up pr

editor/src/document/document_file.rs
```diff
@@ -324,6 +323,20 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+				self.layerdata_mut(&path).expanded = true;
+				responses.push_back(DocumentOperation::CreateFolder { path }.into())
+			}
+			GroupSelectedLayers => {
+				let id = generate_uuid();
+				responses.push_back(DocumentsMessage::CopySelectedLayers.into());
+				responses.push_back(DocumentMessage::DeleteSelectedLayers.into());
+				responses.push_back(DocumentOperation::CreateFolder { path: vec![id] }.into());
+				self.layerdata_mut(&[id]).expanded = true;
+				responses.push_back(DocumentsMessage::PasteLayers { path: vec![id], insert_index: -1 }.into());
```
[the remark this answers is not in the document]
TrueDoctor: Good point but this requires a lot more logic because we need special case handling if all layers are in the same folder. But I'll add a todo

editor/src/document/document_file.rs
```diff
@@ -324,6 +323,20 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					.into(),
 				)
 			}
+			CreateFolder(mut path) => {
+				let id = generate_uuid();
+				path.push(id);
+				self.layerdata_mut(&path).expanded = true;
+				responses.push_back(DocumentOperation::CreateFolder { path }.into())
+			}
+			GroupSelectedLayers => {
```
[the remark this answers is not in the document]
TrueDoctor: sortof done but still wonky because of a different part of the codebase

frontend/src/components/panels/LayerTree.vue
TrueDoctor: me to deferring to @Keavon

frontend/src/components/panels/LayerTree.vue
TrueDoctor: fixed

editor/src/input/input_mapper.rs
```diff
@@ -185,9 +187,17 @@ impl Default for Mapping {
...
 			entry! {action=DocumentMessage::Undo, key_down=KeyZ, modifiers=[KeyControl]},
 			entry! {action=DocumentMessage::DeselectAllLayers, key_down=KeyA, modifiers=[KeyControl, KeyAlt]},
 			entry! {action=DocumentMessage::SelectAllLayers, key_down=KeyA, modifiers=[KeyControl]},
+			entry! {action=DocumentMessage::CreateFolder(vec![]), key_down=KeyN, modifiers=[KeyControl, KeyShift]},
+			entry! {action=DocumentMessage::DispatchOperation (Box::new(Operation::AddRect {
+				path: vec![0],
+				insert_index: -1,
+				transform: DAffine2::default().to_cols_array(),
+				style: style::PathStyle::default(),
+			})), key_down=KeyG},
```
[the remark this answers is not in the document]
TrueDoctor: Testing - removed

TrueDoctor (approved):
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/315)* status: 0 of 83 files reviewed, 5 unresolved discussions (waiting on @0HyperCube, @Keavon, @pkupper, and @tillarnold)

PR #323, opened 2021-07-31
Change flipping to use the joint bounding box of the selection

### LINKED ISSUE #316: Improve Flip actions to work based on selection bounding box
Currently the Flip action works based on the origin point of the selected layer. It should instead work based on the union of bounding boxes of all selected layers, so that the result of the flip is the shapes flipped within that bounding box. The bounding box should be relative to the viewport rotation.

henryksloan:
Closes #316

- Replaced the `FlipLayer` message with a new `FlipSelectedLayers` message that takes a `FlipAxis` enum
- Changed the `Select` `ToolMessage`s to use this new message
- Implemented the `FlipSelectedLayers` message
    - Collects the max and min X or Y coordinates of each layer's bounding box
    - Aggregates the max's and min's to get the extrema of the bounding box in the given axis
    - Takes the average of these extrema as the flipping axis
    - Scales each layer by `-1.` along the axis of flipping, and translates it to the same distance on the other side of the axis

## Caveats
- The flipping is still in canvas space, similarly to alignment. This is because the current bounding box implementation (I believe) exclusively works in canvas space.
- The code related to joint bounding boxes is quite repetitive and verbose between the flipping and alignment procedures
    - #282 is supposed to add a merging function between two bounding boxes, but it is not yet ready to review
    - I could add the function for use in these methods, and it could be reused when work on that PR continues
    - Either way, that should probably be saved for another PR

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/323)


core/editor/src/document/document_message_handler.rs
```diff
@@ -652,13 +658,52 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
-					let scale = DVec2::new(if flip_horizontal { -1. } else { 1. }, if flip_vertical { -1. } else { 1. });
+			FlipSelectedLayers(axis) => {
+				// TODO: Handle folder nested transforms with the transforms API
+				let selected_paths = self.selected_layers_sorted();
+				if selected_paths.is_empty() {
+					return;
+				}
+
+				let selected_layers = selected_paths.iter().filter_map(|path| {
+					let layer = self.active_document().document.layer(path).unwrap();
```
Keavon, line 669: I don't know the answer, but is this a place you can use a `?` instead of `.unwrap()`?

core/editor/src/document/document_message_handler.rs
```diff
@@ -652,13 +658,52 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+
+				let (min, max) = selected_layers
+					.clone()
+					.map(|(_, extrema)| extrema)
+					.reduce(|(min_a, max_a), (min_b, max_b)| (min_a.min(min_b), max_a.max(max_b)))
+					.unwrap();
+				let middle = (min + max) / 2.;
+
+				for (path, _) in selected_layers {
+					let layer = self.active_document().document.layer(&path).unwrap();
```
Keavon, line 689: Same here.

Keavon (approved):

core/editor/src/document/document_message_handler.rs
henryksloan: Yeah, good catch. We've generally been okay with using `unwrap` for layer paths provided to messages, but we might as well use `self.active_document().document.layer(path).ok()?`

Keavon:
Before:
image

After horizontal flip:
image

Flipping it back restores it to the original design.

After vertical flip:
image

core/editor/src/document/document_message_handler.rs
henryksloan: I think not, because this scope is a function with no return value. The `?` operator basically evaluates to:
```rust
let mut f = match f {
    Ok(x) => x,
    Err(e) => return Err(e),
};
```
And we can only return `Err` when the containing function has some `Result` return type.

Keavon:
The above bugs arise from their rotated coordinate system when they are created. We'll have to fix that later.

Keavon (approved):

PR #325, opened 2021-08-01
Add support for saving and opening files

### LINKED ISSUE #324: Implement Save/Open to serialize and provide a Graphite document file
File > Save should cause a request to the editor backend for a serialized document. We can give this a `.gsvg` file extension ("Graphite SVG") which will then be replaced with `.gdd` ("Graphite Design Document") in version 0.2 when there is a proper document system and node graph. Then backend will then reply with the text of this file, which the Response Handler will receive and provide to the `download(filename: string, svgData: string)` function in Document.vue (currently used for File > Export). Maybe rename `svgData` in that function to `fileData`.

File > Open should then make the browser display a file system browse prompt and the user can pick a `.gsvg` file to open. This should then get sent to the backend and get deserialized. Then the appropriate message(s) can be sent to the frontend for creating the new tab, switching to that tab, and displaying its contents in the viewport.

This should probably be implemented before #233, so it can be considered a soft-blocker for it.

azeemba:
This is similar to the "export" functionality, except that we store all metadata needed to open the file again.

The saved file is the JSON representation of the internal Layer data structure. This creates two concerns:

- Fragility: Old files are going to break frequently as the internal data structure changes
- Security: Loading data directly into internal data structures is a frequent cause of vulnerabilities

We probably want to come up with an appropriate serialized format with proper backwards compatibility and safety guarantees. Not sure if we want to do that now, before 0.1 release or post 0.1 release.

Closes #324

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/325)


TrueDoctor:
I am not sure what you are trying to achieve with the `action`, I feel like it should be either an enum variant or just two different messages

azeemba:
I was thinking that `SerializeData` request would be used anytime the frontend needs to actually serialize the document. Since the responses comes in through a separate pipeline, the frontend wouldn't be able to keep track of what the original UI action was that triggered the backend request. So I was thinking that maybe the UI could use the `action` string to track that.

Maybe I am overengineering it and I can just implement the one event. If I remove the action argument, should I keep the request `SerlalizeData` or should I explicitly name it something like `SaveFile`?

core/document/src/document.rs
```diff
@@ -63,6 +63,12 @@ impl Document {
 		svg
 	}
 
+	// Returns string that represents the full document (not just svg)
+	pub fn serialize_document(&mut self) -> String {
+		// Manually constructing JSON. Should switch to serde_json
+		format!("{{\"version\": 0, \"svg\": \"{}\"}}", self.render_root())
```
azeemba, line 69: It would also be good to get feedback on what we think the serialized document should look like. Here I am constructing a JSON string with the `version` and `svg` keys. If we are okay with this, I can switch to constructing the JSON properly.

TrueDoctor: Shouldn"t the forma be something we can import again?

TrueDoctor:
Sorry for not properly addressing your questions earlier. The overall approach seems fine to me.
I would suggest to use serde_json to serialize the entire document (the root layer).
That would also enable us to import the document again. One caveat is that the serialization is not entirely dumb, the caches should not be serialized but instead every layer should be marked as dirty.
I am not entire happy with the current design, it would be nicer to not store the cache in the layer struct but we might get away with that for now

editor/src/document/document_message_handler.rs
```diff
@@ -164,8 +164,18 @@ impl MessageHandler<DocumentsMessage, &InputPreprocessor> for DocumentsMessageHa
 					_ => format!("Untitled Document {}", new_doc_title_num),
 				};
 
+				let new_document = match serialized_contents {
+					None => DocumentMessageHandler::with_name(name),
+					Some(content) => {
+						let res = DocumentMessageHandler::with_name_content(name, content);
+						if res.is_err() {
+							// how can we show an error message in the UI?
+							return;
```
azeemba, line 173: Is there a standard way of showing the user an error message?

graphene/src/document.rs
```diff
@@ -31,6 +31,19 @@ fn split_path(path: &[LayerId]) -> Result<(&[LayerId], LayerId), DocumentError>
...
+		let json_value = json_value.unwrap_or(serde_json::Value::default());
+		let json_value = json_value.get("root");
+		match json_value {
+			Some(root) => Ok(Self {
+				root: serde_json::from_value(root.clone()).map_err(|_| DocumentError::InvalidFile)?,
+				hasher: DefaultHasher::new(),
+			}),
+			_ => Err(DocumentError::InvalidFile),
+		}
+	}
```
azeemba, line 45: I struggled with the error handling here so probably good for folks to have a look in case I am doing something wrong.

graphene/src/document.rs
```diff
@@ -41,6 +54,15 @@ impl Document {
 		self.hasher.finish()
 	}
 
+	pub fn serialize_document(&self) -> String {
+		let doc = serde_json::json!({
+			"version": 0,
+			"root": self.root,
+			"nodes": {}
+		});
```
azeemba, line 62: This is the main "format". Open question whether this is "good enough" or we want to think of a format with better backwards compatibility and security.

editor/src/document/document_message_handler.rs
```diff
@@ -164,8 +164,18 @@ impl MessageHandler<DocumentsMessage, &InputPreprocessor> for DocumentsMessageHa
 					_ => format!("Untitled Document {}", new_doc_title_num),
 				};
 
+				let new_document = match serialized_contents {
+					None => DocumentMessageHandler::with_name(name),
+					Some(content) => {
+						let res = DocumentMessageHandler::with_name_content(name, content);
```
TrueDoctor, line 170: I feel like the name should default to the name of the file you opened and not just untitled document.
I would prefer to just add an extra message  e.g. something like `DocumentMessage::OpenFile{name: String, contents:String}`. IIRC the NewDocument handler is pretty huge and could use a bit of a modularization anyway.

graphene/src/document.rs
TrueDoctor: I think you could add a string to the enum variant if you want to pipe through the serde error

graphene/src/document.rs
```diff
@@ -31,6 +31,19 @@ fn split_path(path: &[LayerId]) -> Result<(&[LayerId], LayerId), DocumentError>
 }
 
 impl Document {
+	pub fn with_content(serialized_content: String) -> Result<Self, DocumentError> {
+		let json_value: Result<serde_json::Value, serde_json::Error> = serde_json::from_str(&serialized_content);
```
TrueDoctor, line 35: What is the reason for not just implementing/deriving Serialize for document?

TrueDoctor: We could also think about using ron instead: https://github.com/ron-rs/ron
That would probably make our exported file look nicer

TrueDoctor: We also need to manually implement serialize for Layer because we don't want to serialize the render cache

graphene/src/document.rs
TrueDoctor: I guess the format will still change a bit before 0.1. I don't think it is feasible to aim for creating a backwards compatible format, but rather providing a script to convert files to the newest version. (If we e.g. rename/add/delete fields from Layer)

editor/src/document/document_message_handler.rs
azeemba: Kept the same function but pulled out some logic and added another optional arg.

editor/src/document/document_message_handler.rs
Keavon: Added this for you here: https://github.com/GraphiteEditor/Graphite/commit/89ee35b831e2ddcd3fd196642455f2e93fa353fb

frontend/src/components/widgets/inputs/MenuBarInput.vue
```diff
@@ -155,6 +155,43 @@ const menuEntries: MenuListEntries = [
...
+				filename = file.name;
+				reader.readAsText(file);
+			}
+		},
+		{ capture: false, once: true }
+	);
+
+	// A: Trigger user selection
+	element.click();
+}
```
Keavon, line 193:
```ts
async function handleOpenClick() {
	const element = document.createElement("input");
	element.type = "file";

	element.addEventListener("change", async () => {
		if (!element.files || !element.files.length) return;

		const file = element.files[0];

		const filename = file.name;
		const content = await file.text();

		(await wasm).open_document(filename, content);
	});

	element.click();
}
```

editor/src/document/document_message_handler.rs
```diff
@@ -43,6 +44,52 @@ impl DocumentsMessageHandler {
...
+			})
+			.collect::<Vec<usize>>();
+		doc_title_numbers.sort_unstable();
+		let mut new_doc_title_num = 1;
+		while new_doc_title_num <= self.documents.len() {
+			if new_doc_title_num != doc_title_numbers[new_doc_title_num - 1] {
+				break;
+			}
+			new_doc_title_num += 1;
+		}
```
TrueDoctor, line 68: this could be probably made more readable by doing something like this:
```rust
let (new_num, _) = doc_title_numbers.enumerate().find(|(i, number)| i != number).unwrap_or(self.documents.len());
```

editor/src/document/document_message_handler.rs
```diff
@@ -43,6 +44,52 @@ impl DocumentsMessageHandler {
...
+		let name = match new_doc_title_num {
+			1 => "Untitled Document".to_string(),
```
TrueDoctor, line 70: "Untitled Document" should probably be moved to a constant that is also used in the name matching

graphene/src/document.rs
```diff
@@ -4,15 +4,17 @@ use std::{
...
 use crate::{
 	layers::{self, Folder, Layer, LayerData, LayerDataType, Shape},
 	DocumentError, DocumentResponse, LayerId, Operation, Quad,
 };
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct Document {
 	pub root: Layer,
+	#[serde(skip)]
```
TrueDoctor, line 17: we could think about including the current document hash here (hasher.finish()) but I am not sure if that would even be useful

frontend/src/components/widgets/inputs/MenuBarInput.vue
azeemba: Haha writing the original code made me miss promises. I am glad there is a better promise-based API here.

Keavon: Indeed, a JS API that returns its result through an event is pretty unsightly. I'd even settle for a callback, but this is a beautiful API— I'm very happy they simplified it!

editor/src/document/document_message_handler.rs
```diff
@@ -43,6 +44,52 @@ impl DocumentsMessageHandler {
...
 		&mut self.documents[self.active_document_index]
 	}
+	fn get_new_doc_title(&self) -> String {
+		let digits = ('0'..='9').collect::<Vec<char>>();
+		let mut doc_title_numbers = self
+			.documents
+			.iter()
+			.map(|d| {
+				if d.name.ends_with(digits.as_slice()) {
+					let (_, number) = d.name.split_at(17);
```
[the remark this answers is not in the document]
azeemba: So it looks like there are multiple concerns with this existing logic (this PR just pulls it into its own function). Is this PR the right place to upgrade it or should that be done separately?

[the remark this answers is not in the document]
Keavon: I suppose a natural consequence of it showing up in the diff is that it's fair game for code review. You, and all contributors, are always welcome to file an issue or write a TODO comment indicating that it'll should be done later if you would prefer to hurry up and get the main feature merged. That said, this looks like a moderately quick cleanup of existing code and we encourage cleaning up dirty laundry whenever you see it (as long as it doesn't involve a considerably large refactor that moves lots of code which is unrelated to the original feature, then it would be preferable to work on the two things in separate branches). But when there are strict procedures like "don't change touch anything unrelated to your current issue!" then it's unlikely that the code ever gets cleaner throughout use. So please do clean things up wherever you see something that could be improved— document any code that you stared at for a while in confusion before learning how it works, rename any undescriptive/vague/ambiguous/abbreviated variable/function/type names, move a function to a more relevant file, break out complex code into simpler functions, etc.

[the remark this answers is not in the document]
TrueDoctor: > So it looks like there are multiple concerns with this existing logic (this PR just pulls it into its own function). Is this PR the right place to upgrade it or should that be done separately?

Some changes like this one:
```rust
let (new_num, _) = doc_title_numbers.enumerate().find(|(i, number)| i != number).unwrap_or(self.documents.len());
```

are purely cosmetic and you are always welcome to fix these, especially if it is a relatively quick fix  and a solution is already provided.

Something like the magic number 17 actually requires changes because your pr would otherwise break the code.
Fixes that fix pieces of code your current pr breaks should always be done in the same pr

editor/src/document/document_message_handler.rs
azeemba: I tried implementing this but i ran into some issues so I have left it as-is for now.

frontend/src/components/panels/Document.vue
```diff
@@ -330,6 +334,10 @@ export default defineComponent({
 			const updateData = responseData as ExportDocument;
 			if (updateData) this.download("canvas.svg", updateData.document);
 		});
+		registerResponseHandler(ResponseType.SaveDocument, (responseData: Response) => {
+			const saveData = responseData as SaveDocument;
+			if (saveData) this.download("canvas.graphite", saveData.document);
```
Keavon, line 339: I think the last thing we need to change is setting the file name to the document (tab) name. So it would, for example, be `Untitled Document 3.graphite` if you tried saving the third tab you opened. I think you'll need to plumb that through as string data along with this `SaveDocument` message in the enum data.

editor/src/document/document_message_handler.rs
TrueDoctor: Ah yeah, my bad, wanted to post this
```rust
let (_, new_num) = doc_title_numbers.enumerate().find(|(i, number)| i != number).unwrap_or((0,self.documents.len()));
```
I am not sure what you mean by It would also return 1.
Oh btw I think this could actually be made more efficient by using a binary search.
```rust
// Uses binary search to find the index of the element where number is bigger than i
let new_num = doc_title_numbers.enumerate().binary_search_by(|(i,number)|number.cmp(i + 1)).map_or_else(|e|e, |v|v) + 1;
```
I think something like this should work, I knew there was some kind of optimization hiding

TrueDoctor: Oh and while it would be pretty cool to use a fancy bin search, this would be mostly to set a precedent and to write better code in the future (and inspire others reading the code to do the same). I don't think we will ever have enough elements for this to make any difference.
And I just realized this does not work because we operate on an iterator for which bin search is not implemented…
we would need:
```rust
(Make the vector consist of i32 somewhere above during the collection)
doc_title_numbers.iter_mut().enumerate.for_each(|(i, number)| number = number -i -1);
// Uses binary search to find the index of the element where number is bigger than i
let new_num = doc_title_numbers.binary_search(0).map_or_else(|e|e, |v|v) + 1;
```

frontend/src/components/panels/Document.vue
```diff
@@ -232,6 +232,7 @@ import OptionalInput from "@/components/widgets/inputs/OptionalInput.vue";
 import ToolOptions from "@/components/widgets/options/ToolOptions.vue";
 import { SectionsOfMenuListEntries } from "@/components/widgets/floating-menus/MenuList.vue";
 
+const FILE_SUFFIX = ".graphite";
```
Keavon, line 235: Could you please move this (and `.svg`) into `consts.rs` and do the concatenation in the editor backend, for both `.graphite` and `.svg`? (So the concatenated name would be passed along in the `name` variable of the frontend messages.)

frontend/src/components/panels/Document.vue
```diff
@@ -332,11 +333,22 @@ export default defineComponent({
 		});
 		registerResponseHandler(ResponseType.ExportDocument, (responseData: Response) => {
 			const updateData = responseData as ExportDocument;
-			if (updateData) this.download("canvas.svg", updateData.document);
+			if (!updateData) return;
+
+			let { name } = updateData;
+			const target_suffix = ".svg";
```
Keavon, line 339: This too.

Keavon:
I believe we're all good to merge! Gonna wait to see if @TrueDoctor gives his approval when he wakes up soon. Excellent work on this, I can't understate how important this feature is so thank you for bringing it to fruition.

editor/src/document/document_message_handler.rs
```diff
@@ -43,6 +45,52 @@ impl DocumentsMessageHandler {
...
+		let mut doc_title_numbers = self
+			.documents
+			.iter()
+			.map(|d| {
+				if d.name.starts_with(DEFAULT_DOCUMENT_NAME) && d.name.ends_with(digits.as_slice()) {
+					let (_, number) = d.name.split_at(DEFAULT_DOCUMENT_NAME.len());
+					number.trim().parse::<usize>().unwrap()
+				} else {
+					1
+				}
```
TrueDoctor, line 59: I think this code might actually break the logic, we should not assume a one when the other criteria don't match,
only if the document title matches but not the ends with slice, otherwise they should be ignored

TrueDoctor (commented):
> Reviewed 1 of 4 files at r8, 1 of 1 files at r10.
> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/325)* status: 2 of 23 files reviewed, 22 unresolved discussions (waiting on @azeemba and @Keavon)

TrueDoctor (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/325#-:-Mh287B20BxwPPtdK2hx:bnfp4nl)* status: 2 of 23 files reviewed, 22 unresolved discussions (waiting on @azeemba and @Keavon)

Keavon (approved):

PR #327, opened 2021-08-02
Change eyedropper to use correct message

### LINKED ISSUE #326: Eyedropper Tool color selection doesn't actually set the color
The Eyedropper Tool picks a color and shows it in the primary working color swatch in the bottom left. But drawing a shape with this color uses the old color, and swapping the primary and secondary working colors resets them to the original pair from before the Eyedropper Tool was used. In other words, the Eyedropper Tool is doing nothing other than showing a fake color in the swatch.

henryksloan:
Closes #326 

Diagnosed as a regression of #300 due to #305

The fix simply was due to not changing the message that the eyedropper tool dispatched. It originally used `FrontendMessage::UpdateWorkingColors`, which was the original cause of loopy behavior. When the looping tendency was fixed, this should have been replaced with `ToolMessage::SelectPrimaryColor` and `ToolMessage::SelectSecondaryColor`. This PR does that replacement, which then propagates the change with a `FrontendMessage`.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/327)

### DIFF
--- a/core/editor/src/tool/tools/eyedropper.rs
+++ b/core/editor/src/tool/tools/eyedropper.rs
@@ -1,5 +1,4 @@
 use crate::consts::SELECTION_TOLERANCE;
-use crate::frontend::FrontendMessage;
 use crate::message_prelude::*;
 use crate::tool::{ToolActionHandlerData, ToolMessage};
 use glam::DVec2;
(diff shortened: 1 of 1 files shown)


Keavon (approved):

TrueDoctor (approved):
LGTM

PR #328, opened 2021-08-02
Make scrollbars interactable

### LINKED ISSUE #307: Hook up document scrollbars and make them interactable
Scrollbars are visually implemented but they need to work based on user interaction, and need to be hooked up to the backend to display the current scroll status and make the interaction tell the backend about the scrolling. This includes dragging the bar, clicking in the track to the sides of the bar, and clicking the arrows at the ends of the bar. Some of this will be implemented in the `PersistentScrollbar.vue` component and some will be implemented in the `Document.vue` component. For an infinite canvas, the bar's width should be based on the bounding box of the artwork in the canvas (from the angle of the current rotation of the viewport, just like you'd see if you selected all the layers in the document and saw the Select Tool's selection bounding box), and it should be based on the artboard dimensions when the canvas is not an infinite canvas.

0HyperCube:
Closes #307 

- [x] Drag bar
- [x] Click in track
- [x] Use arrows
- [x] Messages to backend
- [x] Messages to frontend

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/328)


client/web/src/components/widgets/scrollbars/PersistentScrollbar.vue
```diff
@@ -2,7 +2,7 @@
 	<div class="persistent-scrollbar" :class="direction.toLowerCase()">
 		<button class="arrow decrease" @mousedown="changePosition(-50)"></button>
 		<div class="scroll-track" ref="scrollTrack" @mousedown="grabArea">
-			<div class="scroll-thumb" @mousedown="grabHandle" :class="{ dragging: Dragging }" ref="handle" :style="[thumbStart, thumbEnd, sides]"></div>
+			<div class="scroll-thumb" @mousedown="grabHandle" :class="{ dragging: dragging }" ref="handle" :style="[thumbStart, thumbEnd, sides]"></div>
```
Keavon, line 5: Replace `dragging: dragging` with `dragging`. Thanks for noticing those were PascalCase, I'm not sure why I did that.

client/web/src/components/widgets/scrollbars/PersistentScrollbar.vue
```diff
@@ -139,63 +139,67 @@ export default defineComponent({
...
 			const track = this.$refs.scrollTrack as HTMLElement;
-			return this.direction === ScrollbarDirection.Vertical ? track.clientHeight - this.HandleLength : track.clientWidth;
+			return this.direction === ScrollbarDirection.Vertical ? track.clientHeight - this.handleLength : track.clientWidth;
 		},
 		trackOffset(): number {
 			const track = this.$refs.scrollTrack as HTMLElement;
 			return this.direction === ScrollbarDirection.Vertical ? track.getBoundingClientRect().top : track.getBoundingClientRect().left;
 		},
-		updateHandlePosition(e: MouseEvent) {
+		updatehandlePosition(e: MouseEvent) {
```
Keavon, line 169: Revert (`h` should be capitalized).

client/web/src/components/widgets/scrollbars/PersistentScrollbar.vue
0HyperCube: Done

client/web/src/components/widgets/scrollbars/PersistentScrollbar.vue
0HyperCube: Done

Keavon:
Looks like this is coming together quite nicely!

I was playing with the scrollbar behavior and I think I need to alter my previous description about the algorithm a little for infinite canvases:

If the artwork is smaller (in either dimension, this description applies per-axis) than the viewport at 100% scale, we should expand the artwork bounding box to the current size of the viewport. Zooming in, then, should shrink the scrollbars because it still is using the bounds of the 100%-scaled viewport (which is larger than the artwork). Zooming out, however, needs to make the scrollbars appear bigger but never run out of space on either side: to make that happen, we'd need to change the bounds from the `max(artwork, viewport-at-100%)` to `max(artwork, viewport-at-x%)` where `x%` is a number that proportionally changes with zoomed-out-ness in order to cause the scrollbars to grow but still leave space on either side. (Then, whatever is calculated for the modified artwork bounds, it gets unioned with the current viewport location to draw the actual scrollbars, just like you're doing right now which works superbly.) If this isn't clear (it's definitely hard to describe), let's discuss it in more detail on Discord.

editor/src/document/document_file.rs
```diff
@@ -372,12 +372,29 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
 				Ok(_) => (),
 			},
-			RenderDocument => responses.push_back(
+			RenderDocument => responses.extend([
 				FrontendMessage::UpdateCanvas {
 					document: self.document.render_root(),
 				}
 				.into(),
-			),
+				FrontendMessage::UpdateScrollbars {
```
TrueDoctor, line 380: I don't like the functionality split between frontend and backend, either send the correct unmodified viewport bounds to the frontend, or just send {xpos, xscale, ypos, yscale} for the scrollbars. I think I would prefer the second approach

editor/src/document/document_file.rs
```diff
@@ -372,12 +372,29 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+						let bounds = self.active_document_mut().document.visible_layers_bounding_box();
+						let bounds = bounds.unwrap_or([glam::DVec2::ZERO, glam::DVec2::ZERO]);
+						let viewport = ipp.viewport_size.as_f64();
+						[
+							bounds[0].x.min(0.) - viewport.x * 0.5,
+							bounds[0].y.min(0.) - viewport.y * 0.5,
+							bounds[1].x.max(viewport.x) + viewport.x * 0.5,
+							bounds[1].y.max(viewport.y) + viewport.y * 0.5,
+						]
+					},
```
TrueDoctor, line 391: This is not final yet, but in general I would prefer something like this:
```rust
                        let bounds = bounds.unwrap_or([glam::DVec2::ZERO, glam::DVec2::ZERO]);
						let [pos1, pos2] = ipp.viewport_size.as_f64();
                        pos1 = pos1.min(DVec::Zero) - viewport * 0.5;
                        pos2 = pos2.max(viewport) + viewport * 0.5;
						[ pos1.x, pos1.y, pos2.x, pos2.y]

File: editor/src/document/movement_handler.rs [hunk 5]
```diff
@@ -29,6 +29,7 @@ pub enum MovementMessage {
 	WheelCanvasZoom,
 	SetCanvasRotation(f64),
 	ZoomCanvasToFitAll,
+	TranslateCanvas((f64, f64)),
```
TrueDoctor, line 32: This could probably also just use a DVec2 as an argument because we don't need to expose a FFI

editor/src/document/document_file.rs
TrueDoctor: while you're at it you could also remove the .active_document() functions to avoid future confusion. Was a bit lazy and did not fix everything the the transforms pr

0HyperCube: Done

editor/src/document/movement_handler.rs
0HyperCube: Done

editor/src/document/document_file.rs
0HyperCube: Done

editor/src/document/document_file.rs
```diff
@@ -372,12 +364,31 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+					FrontendMessage::UpdateCanvas {
+						document: self.document.render_root(),
+					}
+					.into(),
+				);
+				let viewport = ipp.viewport_size.as_f64();
+				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or_default();
+				let bounds1 = bounds1.min(glam::DVec2::ZERO) - viewport * 0.5;
+				let bounds2 = bounds2.max(viewport) + viewport * 0.5;
+				let bounds_length = (bounds2 - bounds1).abs();
```
TrueDoctor, line 378: bounds2 is always bigger than bounds1

0HyperCube: done

editor/src/document/document_file.rs
```diff
@@ -372,12 +364,31 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+				responses.push_back(
+					FrontendMessage::UpdateCanvas {
+						document: self.document.render_root(),
+					}
+					.into(),
+				);
+				let viewport = ipp.viewport_size.as_f64();
+				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or_default();
+				let bounds1 = bounds1.min(DVec2::ZERO) - viewport * 0.5;
+				let bounds2 = bounds2.max(viewport) + viewport * 0.5;
```
TrueDoctor, line 377: could you try `out 2^(-¹/₃ • x ) * SCROLLBAR_PERCENT_AT_100_ZOOM (0.5 for now)` instead of 0.5?

TrueDoctor: actually probably 1- that whole thing now that I think about it

TrueDoctor:
Added some info on discord

Keavon:
Because the scrollbars in infinite canvas mode move nonlinearly, what do you think about making the mouse drag linear even though the resulting scrollbar change would appear as per normal (nonlinearly), therefore making the user's cursor not follow the scrollbar but still behaving more expectedly?

editor/src/document/document_file.rs
0HyperCube: Done (sorry for late reply)

0HyperCube:
> 
> 
> Because the scrollbars in infinite canvas mode move nonlinearly, what do you think about making the mouse drag linear even though the resulting scrollbar change would appear as per normal (nonlinearly), therefore making the user's cursor not follow the scrollbar but still behaving more expectedly?

Done

Keavon (approved):

PR #338, opened 2021-08-10
Stop clicks on submenus from propagating to the parent menu

KeatonTech:
Fixes a bug where clicking any submenu item would trigger the default behavior on the item that opened the submenu.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/338)


KeatonTech:
This is getting pretty involved but I've been able to maintain every piece of functionality I've found so far except for `data-hover-menu-spawner`. I think that could be better served as a directive, which seems preferable to direct DOM traversal. I also think it should have explicit 'groups', since if the user is hovering over menu items they would not expect that moving their cursor over an unrelated popover button would open it. Grouping into logical sets like 'menu items', 'toolbar popover buttons' and 'color pickers' seems most logical to me.

I'm still new to Vue but my basic assumption is that data binding > $refs > $el, so I'm building most of this code around data binding.

KeatonTech:
Actually, thinking about it more, 'groups' are still basically global state. MenuBarInput already has the right behavior without the need for a new directive at all (code is in the latest commit).

The question is: should opening a floating menu and then hovering elsewhere ever be able to open an entirely different floating menu? Or should this 'spawn' functionality be limited to menus that share some parent component that does the orchestration (eg. MenuBarInput or SwatchPairInput)

Keavon:
Groups has been a planned feature! Glad you noticed, that means it'll be somewhat important :) Of course that doesn't have to be done in this PR, unless you want to. Your choice.

I noticed you changed a ref to a `v-model` binding, excellent idea.

I am not familiar enough with Vue's concept of directives, unless you are just referring to `v-model`. Do you mean custom directives or the built-in ones?

I might regret this later but I think we should be fine to assume popovers won't spawn child popovers, so only FloatingMenus will need that ability.

frontend/src/components/panels/Document.vue
```diff
@@ -1,70 +1,71 @@
 <template>
 	<LayoutCol :class="'document'">
 		<LayoutRow :class="'options-bar'">
-			<div class="left side">
-				<DropdownInput :menuEntries="documentModeEntries" v-model:selectedIndex="documentModeSelectionIndex" :drawIcon="true" />
+			<FloatingMenuToggleGroup>
+				<div class="left side">
+					<DropdownInput :menuEntries="documentModeEntries" v-model:selectedIndex="documentModeSelectionIndex" :drawIcon="true" />
```
KeatonTech, line 6: I'm frustrated at GitHub's diff here. The only change is that I wrapped everything in FloatingMenuToggleGroup which means everything else got indented.

TrueDoctor: image
looks fine in reviewable

KeatonTech:
Not sure what's up with the Cloudflare deploy. I don't think I have access to the logs.

frontend/src/components/panels/Document.vue
Keavon: GitHub's diffs are shockingly bad. I'm amazed I can't find more complaints about it on the internet; I would have thought people would have revolted by now having to live with such useless diffs. It's seriously one of the absolute worst parts of the entire GitHub experience.

Although I'm not a fan of Reviewable as a whole, I added it to our project just so we can see competent diffs. I avoid it for small code reviews because I can make do with GitHub's diffs and not complicate the code review process with the comment-based reviews. I use it for medium-sized reviews. For big reviews, I pull the branch and use VSCode's Git Graph plugin where I Ctrl+click the PR branch and master to view the diff, then I find the associated file in GitHub's code review area and write my comments on the respective line.

That said, in this particular case, you can just set GitHub's code review feature to ignore whitespace:

frontend/src/components/panels/Document.vue
```diff
@@ -1,70 +1,72 @@
...
+
+					<Separator :type="SeparatorType.Section" />
+
+					<ToolOptions :activeTool="activeTool" />
+				</div>
+				<div class="spacer"></div>
+				<div class="right side">
```
Keavon, line 13: Let's wrap the left side and right side into separate groups.

frontend/src/components/widgets/buttons/FloatingMenuToggleButton.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+	<div class="toggle-button" @mousedown.stop="toggleFloatingMenu()" @mouseenter="maybeOpenOnHover()">
```
Keavon, line 2: I'm concerned that we are using a wrapper div for our click and hover events. This is, for example, triggering outside the rounded corners of the widget. It's much nicer if we can trigger on the HTML `<button>` element itself for widgets which use that HTML element, which is what the `:action="..."` prop is for in many button/input widgets. I wonder if we can alter this approach a little in order to make the button/input widget, which is supposed to open the floating menu, pipe its `action` prop callback as the trigger. I also find that the extra double wrappers required for the two template tags inside this component's tag makes the code more complex and visually noisy. Since these are, in my opinion, a pretty big compromise and I am not sure the previous approach without this `FloatingMenuToggleButton` component was any worse, it might be worth exploring other options or chatting about whether it's worth reverting this particular component's changes back to how it was before and then finding a simpler way to clean it up. I'd love to discuss the pros and cons of each, or see if we might be able to find a cleaner way to do it, over Discord if you have any thoughts.

Keavon (changes requested):
Commenting on the first half of the files. The latter half are more complex so I'd like to chat about them in more detail after a revision or two addresses my comments so far.

frontend/src/components/widgets/buttons/FloatingMenuToggleButton.vue
KeatonTech: Yeah I take your point. I really wanted to use a Directive for this but unfortunately Vue Custom Directives are quite limited in what they can do. What this component provides is the logic to navigate between group items by hovering. That logic is needed in enough places that it seemed worthwhile to implement it once instead of having a bunch of repeated code. I'm imagining that this component would also expand in the future to help with keyboard nav and a11y, again cutting down on repeated code. And I've been thinking about logic to allow menu items to be selected with one click-and-drag gesture (eg. to change blend modes mouse down on the toggle button, drag to the one you want, and mouse up to select it).

The specific bug about rounded corners and that sort of thing could be addressed with an extra check in the event handlers (good catch on that one). Again, a bit overcomplicated, but at least that's just internal implementation details.

No argument that the template code is a bit overly verbose. Still, in my opinion it is worth it to have one component that handles all our various floating menu toggles and plays nice with everything else.

frontend/src/components/panels/Document.vue
```diff
@@ -1,70 +1,72 @@
...
-					:max="1000000"
-					:incrementBehavior="IncrementBehavior.Callback"
-					:incrementCallbackIncrease="increaseCanvasZoom"
-					:incrementCallbackDecrease="decreaseCanvasZoom"
-					:unit="`%`"
-					:displayDecimalPlaces="4"
-					ref="zoom"
-				/>
-			</div>
+			<FloatingMenuToggleGroup>
```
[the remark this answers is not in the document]
KeatonTech: I haven't heard the word 'spawner' in a UI context before but I suppose it would make sense. Although to be pedantic, it does also 'unspawn'. You could even say it 'toggles' between spawned and unspawned states :) Toggle is admittedly a bit of an overloaded term. Here I mean it as a verb, but yeah it can also be a noun referring to a specific style of button.

What about simplifying the naming all the way down to something like "MenuButton" and "MenuButtonGroup"?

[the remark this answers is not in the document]
KeatonTech: Or maybe "MenuTriggerButton"

frontend/src/components/widgets/behavior/FloatingMenuToggleGroup.vue
```diff
@@ -0,0 +1,37 @@
...
+	data() {
+		return {
+			openMenu: undefined as typeof FloatingMenu | undefined,
+		};
+	},
+	provide() {
+		// Injected by FloatingMenuToggleButton and FloatingMenu, where grouping behavior is implemented.
+		return {
+			openedMenuInGroup: toRef(this, "openMenu"),
+			notifyGroupedFloatingMenuOpened: (newOpenMenu: typeof FloatingMenu) => {
```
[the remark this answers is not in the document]
KeatonTech: Done

frontend/src/components/panels/Document.vue
KeatonTech: Done

frontend/src/components/widgets/behavior/FloatingMenuToggleGroup.vue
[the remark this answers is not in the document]
KeatonTech: Had to undo this change. In this case we do want `this` bound to the current scope

Keavon:
Sorry for the delay in continued code reviewing this! Was swamped finishing up Sprint 6 this past weekend. Reviewing this further now for you.

frontend/src/components/panels/Document.vue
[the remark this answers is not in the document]
Keavon: I'd be inclined to go with `FloatingMenuGroup` since dropdowns, popover buttons, etc. should all fit within a group.

frontend/src/components/widgets/behavior/FloatingMenuToggleGroup.vue
```diff
@@ -0,0 +1,38 @@
...
+			notifyGroupedFloatingMenuClosed: (closedMenu: typeof FloatingMenu) => {
+				if (this.openMenu === closedMenu) {
+					this.openMenu = undefined;
+				}
+			},
+			// Creates a new computed variable, similar to one created in the computed section of a component.
+			openMenuOnHover: computed(() => this.openMenu !== undefined),
+		};
+	},
+});
```
Keavon, line 37: Can you move all this logic into a new file in `frontend/src/utilities` so we are consistent in the way we use `provide/inject`? See `dialog.ts`, `documents.ts`, and `fullscreen.ts` for how that's done. Then we can import those functions into this Vue component. Being consistent about the way we use that logic would be helpful for understanding (I was confused for several minutes trying to understand this and how it related to the previously established pattern of `provide/inject` usage).

frontend/src/components/widgets/buttons/FloatingMenuToggleButton.vue
Keavon: What if we take the following approach instead of relying on a wrapper component?:

We create a TS file in utilities that acts as a floating menu manager (this is what I described above about moving the `provide/inject` logic into a utilities TS file). In every component where we have a button (or any input widget with an `:action="..."` prop) we import a function from that utility TS file and feed it to the `action` prop. For example, if the function is called `openAssociatedFloatingMenu`, a button input component could be given the prop `:action="openAssociatedFloatingMenu('some-unique-id-for-the-associated-popover')"`. Then the FloatingMenu component can be given a required prop like `:id="'some-unique-id-for-the-associated-popover'"` which is used by the FloatingMenu to register itself with the utility TS file. In addition to `:action" props, we could also add a ":hoverAction" prop to deal with the hover.

I don't know if the above description is the best, but it is one possible way to get around using a wrapper component which I still have reservations about because of multiple compromises it requires. I'd love to chat on Discord about other ways we may be able to reduce the code and minimize the complexity. I'm just concerned that the current approach in this part of the PR adds both code and complexity, even though I understand your goal is to generalize things as more components are added that need to utilize common functionality. There is probably a better way neither of us have thought about yet, we should discuss the best option. (Or perhaps break this out from the rest of the PR and get the other parts merged now as we defer on this topic specifically.)

frontend/src/components/widgets/floating-menus/FloatingMenu.vue
```diff
@@ -240,66 +262,29 @@ export default defineComponent({
...
-			});
-		},
-		enableMinWidth(minWidth: string) {
+		// Measures the width this menu should have, absent minWidth constraints.
+		async measureNaturalWidth(): Promise<number | undefined> {
 			const floatingMenuContent = this.$refs.floatingMenuContent as HTMLElement;
-			floatingMenuContent.style.minWidth = minWidth;
-		},
-		mouseMoveHandler(e: MouseEvent) {
-			const MOUSE_STRAY_DISTANCE = 100;
```
Keavon, line 275: Mouse stray behavior was lost.

KeatonTech: Oh, we chatted about that in the Discord. I thought the conclusion was that we didn't need that behavior because it doesn't match what other menu systems (eg. VS Code) do

frontend/src/components/widgets/buttons/FloatingMenuToggleButton.vue
KeatonTech: That's an interesting idea, but I think in practice more cumbersome than the current approach.  one thing it needs `:action` but also mouseenter and eventually mousedrag. For another, I don't think there's a way to directly pass injected properties into a template, so each component that includes a menu button would need to add those 2-3 functions to its local state.

This does seem like the perfect use case for a Directive, and with some more Googling I think it might be possible (although a bit hacky) with Vue custom directives. That directive could be applied to an existing button element like `<my-button v-floating-menu-trigger='"some-unique-id"'>`, and would apply all the necessary event handlers internally. Adding native event handlers is easy enough, although I'm not 100% sure there's a way to set the :action handler on the button.

Personally though I'd miss having the buttons and the menus tied together inside one component, I think it's just a little bit more idiot-proof and saves us from having to think about the case where multiple buttons toggle the same menu id.

frontend/src/components/widgets/floating-menus/FloatingMenu.vue
Keavon: Ah, I'm sorry if that was unclear. I believe we were talking about how submenus persist after un-hovering the entry which spawns the submenu (VS Code works by keeping it around for about a second, while we make it disappear instantly). The stray behavior matches how it works in Blender and it's intended behavior across all floating menus, like popovers, dropdown menus, and the color picker.

Keavon:
Sorry, accidentally closed the wrong branch.

Keavon:
Bug fixed in #516.

PR #342, opened 2021-08-11
Fix n-gon Intersection

### LINKED ISSUE #302: Fix collision detection against N-gons
N-gons (created with the Shape Tool) do not have accurate collision detection. This is a problem for selecting such shapes with the Select Tool, filling it with the Fill Tool, and sampling its color with the Eyedropper Tool, among other problems. It's especially bad for lower values of N on the N-gons, like triangles.

pkupper:
Fixes intersection of n-gons by switching to our custom kurbo fork which adresses the following kurbo issue: https://github.com/linebender/kurbo/issues/180.

Also simplifies intersection logic through better use of kurbo.

Closes #302

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/342)

### DIFF
--- a/graphene/src/intersection.rs
+++ b/graphene/src/intersection.rs
@@ -1,7 +1,7 @@
 use std::ops::Mul;
 
 use glam::{DAffine2, DVec2};
-use kurbo::{BezPath, Line, PathSeg, Point, Shape, Vec2};
+use kurbo::{BezPath, Line, PathSeg, Point, Shape};
 
 #[derive(Debug, Clone, Default, Copy)]
 pub struct Quad([DVec2; 4]);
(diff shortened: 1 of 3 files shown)


Keavon (approved):

graphene/src/intersection.rs
```diff
@@ -44,31 +54,17 @@ pub fn intersect_quad_bez_path(quad: Quad, shape: &BezPath, closed: bool) -> boo
...
-			}
-			if pos > 0 && neg > 0 {
-				return false;
-			}
-		}
+		// check if shape is entirely within the selection
+		return quad.path().contains(shape_point);
+	} else {
+		// empty shape
+		return false;
```
TrueDoctor, line 66: you can remove the returns

pkupper: Done

TrueDoctor (approved):

PR #345, opened 2021-08-14
Refactor ViewportPosition from u32 (UVec2) to f64 (DVec2)

Keavon:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/345)

### DIFF
--- a/editor/src/communication/mod.rs
+++ b/editor/src/communication/mod.rs
@@ -2,7 +2,10 @@ pub mod dispatcher;
 pub mod message;
 use crate::message_prelude::*;
 pub use dispatcher::*;
-use rand_chacha::{rand_core::{SeedableRng, RngCore}, ChaCha20Rng};
+use rand_chacha::{
+	rand_core::{RngCore, SeedableRng},
+	ChaCha20Rng,
+};
 use spin::Mutex;
 
 pub use crate::input::InputPreprocessor;
@@ -27,7 +30,7 @@ where
 }
 
 pub fn generate_uuid() -> u64 {
-    let mut lock = RNG.lock();
+	let mut lock = RNG.lock();
 	if lock.is_none() {
 		*lock = Some(ChaCha20Rng::seed_from_u64(0));
 	}
--- a/editor/src/communication/mod.rs
+++ b/editor/src/communication/mod.rs
@@ -28,7 +28,7 @@ where
 pub fn generate_hash<'a>(messages: impl IntoIterator<Item = &'a Message>, ipp: &InputPreprocessor, document_hash: u64) -> u64 {
 	let mut s = DefaultHasher::new();
 	document_hash.hash(&mut s);
-	ipp.hash(&mut s);
+	ipp.pseudo_hash();
 	for message in messages {
 		message.pseudo_hash().hash(&mut s);
 	}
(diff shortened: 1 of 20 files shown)


editor/src/communication/mod.rs
```diff
@@ -28,7 +28,7 @@ where
 pub fn generate_hash<'a>(messages: impl IntoIterator<Item = &'a Message>, ipp: &InputPreprocessor, document_hash: u64) -> u64 {
 	let mut s = DefaultHasher::new();
 	document_hash.hash(&mut s);
-	ipp.hash(&mut s);
+	ipp.pseudo_hash();
```
TrueDoctor, line 31: oh sorry, `ipp.pseudo_hash().hash(&mut s);`

editor/src/input/input_preprocessor.rs
```diff
@@ -30,13 +32,34 @@ bitflags! {
...
+	///
+	/// # Safety
+	/// This function reads from uninitialized memory but the generated value should be fine.
+	pub fn pseudo_hash(&self) -> u64 {
+		let mut s = DefaultHasher::new();
+		unsafe { self.as_slice() }.hash(&mut s);
+		s.finish()
+	}
+}
+
```
TrueDoctor, line 62: I mean we really don't need unsafe here. I am fine with it but others might disagree

frontend/wasm/src/document.rs
```diff
@@ -92,16 +92,17 @@ pub fn close_all_documents_with_confirmation() -> Result<(), JsValue> {
...
+pub fn viewport_resize(new_width: f64, new_height: f64) -> Result<(), JsValue> {
 	let ev = InputPreprocessorMessage::ViewportResize((new_width, new_height).into());
 	EDITOR_STATE.with(|editor| editor.borrow_mut().handle_message(ev)).map_err(convert_error)
 }
 
 // TODO: When a mouse button is down that started in the viewport, this should trigger even when the mouse is outside the viewport (or even the browser window if the browser supports it)
 /// Mouse movement within the screenspace bounds of the viewport
 #[wasm_bindgen]
-pub fn on_mouse_move(x: u32, y: u32, modifiers: u8) -> Result<(), JsValue> {
+pub fn on_mouse_move(x: f64, y: f64, modifiers: u8) -> Result<(), JsValue> {
```
TrueDoctor, line 103: I am not really a fan of doing the conversion here, I would prefer it to take place explicitly in the ipp.
(We might want to apply a different mapping based on the input device)

editor/src/input/input_preprocessor.rs
Keavon: Less unsafe is always better. What is needed to fix it?

frontend/wasm/src/document.rs
Keavon: The conversion from int to float? JS gives this as a float, not an int.

TrueDoctor: Yeah it just hurts to specify pixels as floats but I guess there is nothing I can do about it

Keavon: Retina/display scaling, and perhaps pen inputs that have a different resolution digitizer than the screen resolution, are cases that come to mind. Not sure how the OS or browser treats it, but the JS spec clearly has a reason for specifying floats.

PR #348, opened 2021-08-14
Improve Frontend -> Backend user input system

### LINKED ISSUE #124: Change how Vue sends input to the backend

### LINKED ISSUE #291: Fix console warning when releasing a mouse press inside the canvas started outside
Click somewhere outside the canvas (in the Vue part of the UI) and drag your mouse into the document canvas, then release the click. You'll see the following error show up each time in the JS console.

Keavon:
Includes refactor that sends coordinates of the document viewports to the backend so input is sent relative to the application window
Closes #124
Fixes #291

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/348)

### DIFF
--- a/editor/src/input/input_preprocessor.rs
+++ b/editor/src/input/input_preprocessor.rs
@@ -93,6 +93,8 @@ impl MessageHandler<InputPreprocessorMessage, ()> for InputPreprocessor {
 				responses.push_back(InputMapperMessage::KeyUp(key).into());
 			}
 			InputPreprocessorMessage::BoundsOfViewports(bounds_of_viewports) => {
+				assert_eq!(bounds_of_viewports.len(), 1, "Only one viewport is currently supported");
+
 				for bounds in bounds_of_viewports {
 					let new_size = bounds.size();
 					let existing_size = self.viewport_bounds.size();
--- a/editor/src/input/input_preprocessor.rs
+++ b/editor/src/input/input_preprocessor.rs
@@ -1,7 +1,7 @@
 use std::usize;
 
 use super::keyboard::{Key, KeyStates};
-use super::mouse::{MouseKeys, MouseState, ScrollDelta, ViewportPosition};
+use super::mouse::{EditorMouseState, MouseKeys, MouseState, ViewportBounds};
 use crate::message_prelude::*;
 use bitflags::bitflags;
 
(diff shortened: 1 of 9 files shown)


editor/src/input/input_preprocessor.rs
```diff
@@ -77,15 +92,24 @@ impl MessageHandler<InputPreprocessorMessage, ()> for InputPreprocessor {
...
-				responses.push_back(
-					graphene::Operation::TransformLayer {
-						path: vec![],
-						transform: glam::DAffine2::from_translation((size - self.viewport_size) / 2.).to_cols_array(),
-					}
-					.into(),
-				);
-				self.viewport_size = size;
+			InputPreprocessorMessage::BoundsOfViewports(bounds_of_viewports) => {
+				for bounds in bounds_of_viewports {
```
TrueDoctor, line 96: Putting this in a for loop is just confusing

TrueDoctor: I would prefer if it would at least panic if anything other than one viewport was supplied because this will otherwise lead to bugs in the future

editor/src/input/mouse.rs
```diff
@@ -3,6 +3,26 @@ use glam::DVec2;
...
+impl ViewportBounds {
+	pub fn from_slice(slice: &[f64]) -> Self {
+		Self {
+			top_left: DVec2::from_slice(&slice[0..2]),
+			bottom_right: DVec2::from_slice(&slice[2..4]),
+		}
+	}
+
+	pub fn size(&self) -> DVec2 {
+		DVec2::new(self.bottom_right.x - self.top_left.x, self.bottom_right.y - self.top_left.y)
```
TrueDoctor, line 23: `self.bottom_right - self.top_left`

editor/src/input/input_preprocessor.rs
Keavon: Why would iterating through a list in a for loop be confusing?

TrueDoctor: because you set overwrite the value in each iteration which is unexpected

frontend/src/utilities/input.ts
```diff
@@ -48,24 +52,76 @@ export async function handleKeyDown(e: KeyboardEvent) {
...
+	if (dialogIsVisible() && !inDialog) {
+		dismissDialog();
 		e.preventDefault();
 		e.stopPropagation();
 	}
+
+	if (inCanvas) viewportMouseInteractionOngoing = true;
+
+	if (viewportMouseInteractionOngoing) {
+		const modifiers = makeModifiersBitfield(e.ctrlKey, e.shiftKey, e.altKey);
```
TrueDoctor, line 89: why don't you just pass the event? I guess for cleanliness but probably performs worse

frontend/src/utilities/input.ts
```diff
@@ -48,24 +52,76 @@ export async function handleKeyDown(e: KeyboardEvent) {
...
+	const viewports = Array.from(document.querySelectorAll(".canvas"));
+	const boundsOfViewports = viewports.map((canvas) => {
+		const bounds = canvas.getBoundingClientRect();
+		return [bounds.left, bounds.top, bounds.right, bounds.bottom];
+	});
+
+	const flattened = boundsOfViewports.flat();
+	const data = Float64Array.from(flattened);
+
+	if (!boundsOfViewports.length) return;
```
TrueDoctor, line 122: Is it common practise to cast numbers to bools in js? in rust we would prefer .is_epmty() or .len() ==0

TrueDoctor (approved):

PR #351, opened 2021-08-16
Change scrollbar behaviour

0HyperCube:
Change some of the behavour from #328

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/351)

### DIFF
--- a/editor/src/document/document_file.rs
+++ b/editor/src/document/document_file.rs
@@ -406,15 +406,16 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					}
 					.into(),
 				);
-				let root = self.layerdata(&[]);
-				let viewport = ipp.viewport_bounds.size();
-				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or_default();
-				let bounds1 = bounds1.min(DVec2::ZERO) - viewport * (f64::powf(2., root.scale / 3.) * 0.5);
-				let bounds2 = bounds2.max(viewport) + viewport * (f64::powf(2., root.scale / 3.) * 0.5);
+				let scale = 0.5 + 0.5 * self.layerdata(&[]).scale;
+				let viewport_size = ipp.viewport_bounds.size();
+				let viewport_mid = ipp.viewport_bounds.mid();
+				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or([viewport_mid, viewport_mid]);
+				let bounds1 = bounds1.min(viewport_mid) - viewport_size * scale;
+				let bounds2 = bounds2.max(viewport_mid) + viewport_size * scale;
 				let bounds_length = bounds2 - bounds1;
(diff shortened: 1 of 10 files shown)


editor/src/input/mouse.rs
```diff
@@ -22,6 +22,10 @@ impl ViewportBounds {
 	pub fn size(&self) -> DVec2 {
 		self.bottom_right - self.top_left
 	}
+
+	pub fn mid(&self) -> DVec2 {
```
TrueDoctor, line 26: to stay somewhat consistent with the existing code I would prefer center

editor/src/document/document_file.rs
```diff
@@ -406,15 +406,16 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
 				);
-				let root = self.layerdata(&[]);
-				let viewport = ipp.viewport_bounds.size();
-				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or_default();
-				let bounds1 = bounds1.min(DVec2::ZERO) - viewport * (f64::powf(2., root.scale / 3.) * 0.5);
-				let bounds2 = bounds2.max(viewport) + viewport * (f64::powf(2., root.scale / 3.) * 0.5);
+				let scale = 0.5 + 0.5 * self.layerdata(&[]).scale;
+				let viewport_size = ipp.viewport_bounds.size();
+				let viewport_mid = ipp.viewport_bounds.mid();
+				let [bounds1, bounds2] = self.document.visible_layers_bounding_box().unwrap_or([viewport_mid, viewport_mid]);
```
TrueDoctor, line 412: could we do `[viewport_mid;2]` ? Afaik DAffine 2 is copy

editor/src/input/mouse.rs
0HyperCube: Done

editor/src/document/document_file.rs
0HyperCube: Done

editor/src/document/document_file.rs
```diff
@@ -406,7 +406,7 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					}
 					.into(),
 				);
-				let scale = 0.5 + 0.5 * self.layerdata(&[]).scale;
+				let scale = 0.7 + 0.3 * self.layerdata(&[]).scale;
```
Keavon, line 409: Unfortunately changing 0.5 and 0.5 to 0.7 and 0.3 doesn't seem to really fix the algorithm, which is that we need to add some extra space to the ends at all times. I suggest taking the final calculation, scaling it by 0.9, and adding 0.05. That way you take whatever your calculation was and convert it into something that is 10% smaller and you will therefore always have 5% of the scrollbar track's total width on either end.

0HyperCube: Done

Keavon (approved):
Wonderful job! The behavior feels great.

PR #355, opened 2021-08-24
Add a stroke width option to the Line Tool

### LINKED ISSUE #350: Add a stroke width option to the Line Tool
Should be pretty similar to the Shape Tool's option for the number of points of the N-gon.

henryksloan:
Closes #350

- Added a `Line` variant to the `ToolOptions` struct, with a `stroke_width` property
- Made the Line tool use the stroke width
- Allow the `NumberInput` widget to expand in width (we can tweak the default width and expansion behavior in this PR)
- Renamed the Vue `setToolOptions` method to `setShapeOptions` and added a new `setLineOptions` methods
    - This will be cleaned up when the unified layout system is implemented
- Added a new number input to the line tool, allowing line width to be specified

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/355)

### DIFF
--- a/frontend/src/components/widgets/inputs/NumberInput.vue
+++ b/frontend/src/components/widgets/inputs/NumberInput.vue
@@ -19,7 +19,7 @@
 
 <style lang="scss">
 .number-input {
-	width: 80px;
+	min-width: 80px;
 	height: 24px;
 	position: relative;
 	border-radius: 2px;
@@ -38,7 +38,7 @@
 
 	input {
 		flex: 1 1 100%;
-		width: 100%;
+		width: 30px;
 		height: 18px;
 		line-height: 18px;
 		margin: 0 8px;
--- a/frontend/src/components/widgets/inputs/NumberInput.vue
+++ b/frontend/src/components/widgets/inputs/NumberInput.vue
@@ -38,7 +38,8 @@
 
 	input {
 		flex: 1 1 100%;
-		width: 30px;
+		width: 0;
+		min-width: 30px;
 		height: 18px;
 		line-height: 18px;
 		margin: 0 8px;
--- a/frontend/src/components/widgets/options/ToolOptions.vue
+++ b/frontend/src/components/widgets/options/ToolOptions.vue
@@ -52,11 +52,11 @@ export default defineComponent({
 		},
 		async setLineOptions(newValue: number) {
 			// eslint-disable-next-line camelcase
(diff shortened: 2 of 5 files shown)


frontend/src/components/widgets/options/ToolOptions.vue
```diff
@@ -126,7 +130,8 @@ export default defineComponent({
 					props: {},
 				},
 			],
-			Shape: [{ kind: "NumberInput", callback: this.setToolOptions, props: { value: 6, min: 3, isInteger: true, label: "Sides" } }],
+			Shape: [{ kind: "NumberInput", callback: this.setShapeOptions, props: { value: 6, min: 3, isInteger: true, label: "Sides" } }],
+			Line: [{ kind: "NumberInput", callback: this.setLineOptions, props: { value: 5, min: 1, isInteger: true, label: "Stroke width" } }],
```
Keavon, line 134: I believe we're using Title Case, so "Stroke width" should be "Stroke Width".

henryksloan: Hah, I had that at first and wasn't sure. Fixing now.

frontend/src/components/widgets/options/ToolOptions.vue
```diff
@@ -126,7 +130,8 @@ export default defineComponent({
 					props: {},
 				},
 			],
-			Shape: [{ kind: "NumberInput", callback: this.setToolOptions, props: { value: 6, min: 3, isInteger: true, label: "Sides" } }],
+			Shape: [{ kind: "NumberInput", callback: this.setShapeOptions, props: { value: 6, min: 3, isInteger: true, label: "Sides" } }],
+			Line: [{ kind: "NumberInput", callback: this.setLineOptions, props: { value: 5, min: 1, isInteger: true, label: "Stroke Width" } }],
```
Keavon, line 134: `min` should probably be something like `0.001` (1 thou).

Keavon: And should have a unit of `px`.

frontend/src/components/widgets/inputs/NumberInput.vue
```diff
@@ -38,7 +38,7 @@
 
 	input {
 		flex: 1 1 100%;
-		width: 100%;
+		width: 30px;
```
Keavon, line 41: Do you think it would be better to do `width: 0; min-width: 30px;`?

Keavon (approved):

PR #356, opened 2021-08-24
(G)rab, (R)otate, and (S)cale layers with hotkeys

### LINKED ISSUE #317: [G], [R], and [S] key should grab, rotate, and scale the selected layers
This should work exactly like Blender:

Hit <kbd>G</kbd> and enter Grab mode. Mouse movements cause the selected layers to move around with the mouse. LMB and <kbd>Enter</kbd> applies the transformation while RMB and <kbd>Esc</kbd> cancel the transformation. Holding the <kbd>Shift</kbd> modifier key at any time will slow the movement down to a 1/10 rate of the mouse movement speed. (Later, when we have a grid system, <kbd>Ctrl</kbd> will cause movements to snap to the grid while held down, but this doesn't need to be added yet.) Subsequently hitting <kbd>X</kbd> or <kbd>Y</kbd> will constrain the movement to the X or Y axis, and then numbers typed should specify how many document units to move along that axis.

Hit <kbd>R</kbd> and enter Rotate mode. Mouse movements around the center point of rotation (the layer's origin, or median point of multiple layers' origins) should rotate the layer(s). Holding <kbd>Shift</kbd> makes the movement go at 1/10 the rate, and holding <kbd>Ctrl</kbd> makes it snap to 15° increments. Typing a number sets the rotation to the degrees entered by the user.

Hit <kbd>S</kbd> and enter Scale mode. Mouse movements towards and away from the center point of rotation (the layer's origin, or median point of multiple layers' origins) should scale the layer(s). Holding <kbd>Shift</kbd> makes the scaling go at 1/10 the rate, and holding <kbd>Ctrl</kbd> makes it snap to scale factors that are multiples of 0.1, such as 0.0, 0.1, 0.2, 0.3, and so on. Typing a number sets the scale factor to the number entered by the user (in real time as they type). Backspace allows the removal of digits that were typed, up to and including the first digit which sets it back to mouse-controlled mode.

0HyperCube:
Closes #317

- [x] Add new handler
- [x] Add (neat) bindings to input mapper
- [x] Handle translation mouse movement
- [x] Handle rotation mouse movement
- [x] Handle scale mouse movement
- [x] Handle constrain to axis
- [x] Handle number typing
- [x] Handle shift to slow mouse movements
- [x] Handle ctrl to snap rotation and scale
- [x] Fix 0 scale breaking things
- [x] Fix deselecting layers sometimes causing panic (trouble reproducing)

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/356)


editor/src/input/input_mapper.rs
```diff
@@ -258,6 +263,17 @@ impl Default for Mapping {
 		];
 
 		let (mut key_up, mut key_down, mut pointer_move, mut mouse_scroll) = mappings;
+		const NUMBER_KEYS: [Key; 10] = [Key::Key0, Key::Key1, Key::Key2, Key::Key3, Key::Key4, Key::Key5, Key::Key6, Key::Key7, Key::Key8, Key::Key9];
```
TrueDoctor, line 266: I think we even have a use Keys:* statement above, if not feel free to add that and shorten to [Key1, Key2, …]

Keavon:
Some UX observations/requests:
- If the user aborts (Esc/right click), it should return to the original transform even if the user successively hit <kbd>G</kbd>, <kbd>R</kbd>, and <kbd>S</kbd> to switch between modes)
- Please make it so the minus key toggles between positive and negative numbers that the user types. So they might type <kbd>-</kbd><kbd>1</kbd> to scale by negative 1, or <kbd>-</kbd><kbd>.</kbd><kbd>5</kbd><kbd>-</kbd> to scale by positive 0.5. If only a minus key is pressed, but no numbers entered, nothing should happen yet and the mouse should still have normal control over it.
- Multiple presses of the decimal point key should only count the first one, and discard subsequent ones, as far as <kbd>Backspace</kbd> is concerned. Currently if you type numbers and decimal points together, it will delete the decimal point and cause no visible change (since it's not the first decimal point) but it should instead delete the next *number* from the sequence. The minus key should also not affect the backspace key at all, it minus should just toggle signs but not be part of the queue. You might want the very last backspace (when the list of values is empty) to reset the sign back to positive, though, to avoid confusing users who accidentally hit minus and can't figure out why they are stuck with negative numbers.
- While in number entry mode, I wonder if there's a way to block all other keyboard input (so you can't, for example change tools)?

0HyperCube:
> If the user aborts (Esc/right click), it should return to the original transform even if the user successively hit G, R, and S to switch between modes)

@Keavon That is somewhat difficult with the current architecture. Looking at blender, if you are translating and then you rotate, the translate is reverted. Would this be acceptable behaviour?

0HyperCube:
> Multiple presses of the decimal point key should only count the first one, and discard subsequent ones, as far as Backspace is concerned. Currently if you type numbers and decimal points together, it will delete the decimal point and cause no visible change (since it's not the first decimal point) but it should instead delete the next number from the sequence.

Done.

0HyperCube:
> Please make it so the minus key toggles between positive and negative numbers that the user types. So they might type -1 to scale by negative 1, or -.5- to scale by positive 0.5. If only a minus key is pressed, but no numbers entered, nothing should happen yet and the mouse should still have normal control over it.

Done

0HyperCube:
> While in number entry mode, I wonder if there's a way to block all other keyboard input (so you can't, for example change tools)?

@Keavon You'd probably have to refer to @TrueDoctor but my guess is that this would not be trivial.

TrueDoctor:
> > While in number entry mode, I wonder if there's a way to block all other keyboard input (so you can't, for example change tools)?
> 
> @Keavon You'd probably have to refer to @TrueDoctor but my guess is that this would not be trivial.

Should be pretty easy, just add a InputMapper::MaskEvents message

Keavon:
> > If the user aborts (Esc/right click), it should return to the original transform even if the user successively hit G, R, and S to switch between modes)
> 
> @Keavon That is somewhat difficult with the current architecture. Looking at blender, if you are translating and then you rotate, the translate is reverted. Would this be acceptable behaviour?

Sure, that would be equally acceptable.

Keavon:
Bug: when snapping a rotation, it doesn't let me snap back to the starting rotation (it skips that 15 degree increment, you can't go to 0 degrees, only the +15 or -15 surrounding it). Same deal for scale snapping: I can't snap to 1.0, only 0.9 or 1.1.

0HyperCube:
> Bug: when snapping a rotation, it doesn't let me snap back to the starting rotation (it skips that 15 degree increment, you can't go to 0 degrees, only the +15 or -15 surrounding it). Same deal for scale snapping: I can't snap to 1.0, only 0.9 or 1.1.

Fixed

0HyperCube:
> > > If the user aborts (Esc/right click), it should return to the original transform even if the user successively hit G, R, and S to switch between modes)
> > 
> > 
> > @Keavon That is somewhat difficult with the current architecture. Looking at blender, if you are translating and then you rotate, the translate is reverted. Would this be acceptable behaviour?
> 
> Sure, that would be equally acceptable.

Done

0HyperCube:
> Should be pretty easy, just add a InputMapper::MaskEvents message

@TrueDoctor What trigger should be used and how should it know only to advertise if the layer is being transformed?

TrueDoctor:
> > Should be pretty easy, just add a InputMapper::MaskEvents message
> 
> @TrueDoctor What trigger should be used and how should it know only to advertise if the layer is being transformed?

So do you want to mask all keyboard input?
That would be something like:
(Ready, StartDragging) => {
  responses.push_back(InputMapperMessage::MaskKeiboardInputs)
  …
  }
  (Dragging, StopDragging) => {
    responses.push_back(InputMapperMessage::UnMaskKeiboardInputs)
  }

0HyperCube:
@TrueDoctor It would still need typing numbers, <kbd>x</kbd> and <kbd>y</kbd> to constrain axis etc.

TrueDoctor:
@0HyperCube @Keavon suggested adding different "input modes" that you could define. Like separate mapping definition files that you could switch between, but I don't know what the best approach for that would be.
You can think about a better solution I would need a bit of time and thought to come up with a better one

0HyperCube:
@Keavon Are you fine it this for this PR?

Keavon:
I'm happy with the features and UX. We can save input mode switching for another PR. I still want to review this for code quality though. I'll try to do that in an hour or two.

Keavon (approved):

Issue #357, opened 2021-08-26
Provide users with information in the case of a crash (panic)


Keavon:
If the Graphite Editor backend panics, the user has to open up the JS console to learn what happened. It will otherwise just look like most things have frozen. At a minimum, we should show a dialog when the backend panics, with a button to reload the page.

If can can display information about the crash, like logs that can be reported and a link to file an issue here, that would be helpful.

If it's possible, it would be even better if we can provide users a way to download their documents so their hard work isn't lost. Or automatically reload the Graphite Editor wasm backend and reopen the saved documents automatically.

PR #362, opened 2021-08-29
Show a crash dialog when the editor panics

Keavon:
Part of #357

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/362)


Keavon:
I can use try-catch in JS to filter out the redundant "unreachable" error to avoid confusing users making bug reports.

bjorn3:
I think this doesn't include the backtrace. Can you add it?

Keavon:
> I think this doesn't include the backtrace. Can you add it?

It includes a stack trace (give the demo build a try and crash it by grouping a layer, then repeatedly clicking the top group and grouping it). However it's not the full stack trace that you get when you click the expansion arrow next to the message in the JS console. I spent an hour researching how to get the *full* trace, as described in [this comment](https://stackoverflow.com/questions/2923858/how-to-print-a-stack-trace-in-node-js#comment21085596_2932410), but to no avail. (I can print it with `console.trace()` but can't figure out how to get it as a string.)

Keavon:
To expand on that— the stack trace is currently grabbed from the `console.error()` call made by the autogenerated JS code from the WASM toolchain, by intercepting the `console.error()` function in the browser. I'm doing this since I can't find a way to obtain it directly from the Rust code. Do you know how to do that?

frontend/src/components/panels/Document.vue
```diff
@@ -245,7 +246,7 @@ import OptionalInput from "@/components/widgets/inputs/OptionalInput.vue";
 import ToolOptions from "@/components/widgets/options/ToolOptions.vue";
 import { SectionsOfMenuListEntries } from "@/components/widgets/floating-menus/MenuList.vue";
 
-const wasm = import("@/../wasm/pkg");
+const wasm = import("@/../wasm/pkg").then(panicProxy);
```
henryksloan, line 249: A comment here could be useful. This is cool, though!

Also, is there a way to avoid duplicating this between files? For example, could we make a new file that just exports `import("@/../wasm/pkg").then(panicProxy)`?

frontend/src/utilities/errors.ts
```diff
@@ -32,5 +33,74 @@ registerResponseHandler(ResponseType.DisplayError, (responseData: Response) => {
...
+	};
+	const buttons = [reloadButton, copyErrorLogButton, reportOnGithubButton];
+
+	createDialog("Warning", data.title, data.description, buttons);
+});
+
+function githubUrl() {
+	const url = new URL("https://github.com/GraphiteEditor/Graphite/issues/new");
+
+	const body = `
```
henryksloan, line 65: Maybe put this text in a `.md` or `template` file and import it?

frontend/src/utilities/panic.ts
```diff
@@ -0,0 +1,49 @@
...
+			if (isClass) return interceptedFunction;
+
+			const targetValue = Reflect.get(target, propKey, receiver);
+			if (typeof targetValue === "function") {
+				// eslint-disable-next-line @typescript-eslint/no-explicit-any, func-names
+				return function (...args: any) {
+					let returned;
+					try {
+						// @ts-expect-error
+						returned = targetValue.apply(this, args);
```
henryksloan, line 19: If this is the only branch of this block that can return, why not use `return targetValue.apply(this, args);`

henryksloan (commented):
Works great as far as I've tested

frontend/src/components/panels/Document.vue
Keavon: I'll comment the `panicProxy` function in `panic.ts`.

Unfortunately I cannot find a way to re-export a dynamic module and then dynamically import it, so this was the cleanest approach I could make work.

frontend/src/utilities/errors.ts
Keavon: Do you know how I could import from an external file at build time? If I do this, I would like to put the file in `/.github/ISSUE_TEMPLATE/crash-report.yaml` that the build process would have to grab it from outside the `/frontend` directory. I don't want to add another webpack plugin as a dependency though. Maybe there's a built-in way?

frontend/src/utilities/panic.ts
Keavon: We're catching the exception, so we still need to return the function call's value.

henryksloan (approved):

PR #370, opened 2021-09-06
Auto-generate third-party license notices

### LINKED ISSUE #149: Add folders in the Layer Tree panel of the frontend

### LINKED ISSUE #294: Set up cargo-deny (Rust) and license-checker-webpack-plugin (Node)
We need to add license compatibility checkers to our CI to ensure no copyleft or Apache2-incompatible licenses are ever merged into the project, even as transient dependencies.

For Rust dependencies from crates.io, we should use Embark's [cargo-deny](https://github.com/EmbarkStudios/cargo-deny).

For Node dependencies from npm, I found [license-checker](https://www.npmjs.com/package/license-checker) looks reasonably popular, but popular alternatives would also be fine. (Edit: we ended up using Microsoft's `license-checker-webpack-plugin`)

- [x] Rust (cargo-deny)
- [x] Node.js (license-checker-webpack-plugin)

### LINKED ISSUE #371: Display open source license notices for dependencies
We need to add a way of viewing the open source licenses used by our dependencies before the public release. This should be accessible from the Help menu. It needs to support both Rust and JS packages.

Keavon:
Closes #294
Closes #371

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/370)


TrueDoctor (approved):
LGTM

PR #377, opened 2021-09-11
Do not silently install cargo-about.

### LINKED ISSUE #375: Don't install cargo-about without asking for permission
Currently we unconditionally install `cargo-about` during the build step: 
https://github.com/GraphiteEditor/Graphite/blob/master/frontend/package.json#L8

cauthmann:
On production builds, the user will be prompted to install it themselves if it isn't found.

Resolves #375

Tested on linux with and without cargo-about installed. As of now, untested on windows or github ci.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/377)


TrueDoctor:
Let's check if the ci run works

.github/workflows/ci.yml
```diff
@@ -25,6 +25,9 @@ jobs:
     - name: 🚧 Install Node dependencies
       run: cd frontend && npm install
 
+    - name: 🚧 Install cargo-about
```
Keavon, line 28: To avoid duplicate emoji, let's use: 📦

about.hbs
```diff
@@ -18,4 +19,4 @@ module.exports = [
 		],
 	},
 	{{/each}}
-];
+]
```
Keavon, line 22: I think we should still have the semicolon. Although perhaps I'm confused what's going on here (see the comment above).

frontend/vue.config.js
```diff
@@ -1,18 +1,52 @@
...
+		// Nothing to show. Error messages were printed above.
+	}
+
+	if (!rustLicenses) {
+		// This is probably caused by cargo about not being installed
+		console.error();
+		console.error("Could not run `cargo about`, which is required to generate license information.");
+		console.error("To install cargo-about on your system, you can run:");
+		console.error("  cargo install cargo-about");
+		console.error("License information is required on production builds. Aborting.");
```
Keavon, line 47: A single `console.error()` would be preferable (you can use a JS template string with backticks).

frontend/vue.config.js
```diff
@@ -1,18 +1,52 @@
...
+			}
+			throw new Error("");
+		}
+
+		// Make sure the output starts as expected, we don't want to eval an error message.
+		const CODEWORD = "GRAPHITE-LICENSES: ";
+		const stdout = result.stdout.trim();
+		if (stdout.substr(0, CODEWORD.length + 1) === `${CODEWORD}[`) {
+			// Security-wise, eval() isn't any worse than require(), but it doesn't need a temporary file.
+			rustLicenses = eval(stdout.substr(CODEWORD.length)); // eslint-disable-line no-eval
```
Keavon, line 33: Nitpick: for consistency, please put the eslint disable comment on the line above.

about.hbs
```diff
@@ -1,7 +1,8 @@
-// Be careful to prevent auto-formatting from breaking this file's indentation
-// Replace this file with JSON output once this is resolved: https://github.com/EmbarkStudios/cargo-about/issues/73
-
-module.exports = [
+{{!
+Be careful to prevent auto-formatting from breaking this file's indentation
+Replace this file with JSON output once this is resolved: https://github.com/EmbarkStudios/cargo-about/issues/73
+}}
+GRAPHITE-LICENSES: [
```
Keavon, line 5: I'm confused what this does. What is `GRAPHITE-LICENSES: `? It's not a valid variable name since it includes a hyphen, and you're using a colon, and it doesn't include a declaration like `const`, but the file is also not wrapped in `{}` so it's not a JavaScript object?

frontend/vue.config.js
```diff
@@ -1,18 +1,52 @@
...
+	}
+
+	if (!rustLicenses) {
+		// This is probably caused by cargo about not being installed
+		console.error();
+		console.error("Could not run `cargo about`, which is required to generate license information.");
+		console.error("To install cargo-about on your system, you can run:");
+		console.error("  cargo install cargo-about");
+		console.error("License information is required on production builds. Aborting.");
+		process.exit(5);
```
Keavon, line 48: What is special about exit code 5? Please comment if it has meaning.

frontend/vue.config.js
```diff
@@ -78,28 +112,31 @@ module.exports = {
...
-	// De-duplicate any licenses with the same text by merging their lists of packages
-	licenses.forEach((license, licenseIndex) => {
-		licenses.slice(0, licenseIndex).forEach((comparisonLicense) => {
-			if (license.licenseText === comparisonLicense.licenseText) {
-				license.packages.push(...comparisonLicense.packages);
-				comparisonLicense.packages = [];
-				// After emptying the packages, the redundant license with no packages will be removed in the next step's `filter()`
-			}
+	let licenses = [];
+	if (rustLicenses) {
```
Keavon, line 116: `if (!rustLicenses) return;` would help avoid a scope indentation, which is helpful for something like this which is reasonably long.

about.hbs
cauthmann: it's just a marker to make sure that cargo-about outputs something we expect, before we run the output as javascript. `cargo install cargo-about` always installs the latest version, which may not be backwards compatible with what we expect.

Probably overkill defensive programming, but it doesn't hurt. And eventually we'll replace it with JSON anyway.

frontend/vue.config.js
cauthmann: That's the value of EXIT_FAILURE on AmigaOS. Today I learned that other operating systems use different values. I've changed it to 1.

frontend/vue.config.js
cauthmann: Previously, the code would output the javascript licenses, but add a notice that rust licenses are missing. With your proposed change, it would skip all license information, including the warning.

I tried to keep the logic intact.

about.hbs
cauthmann: if we must nitpick, then `eval()` expects an expression, and an expression does not have a semicolon. But it doesn't matter either way, and it's going to be replaced with proper JSON eventually.

frontend/vue.config.js
```diff
@@ -131,7 +169,7 @@ function formatThirdPartyLicenses(jsLicenses) {
 
 	// Generate the formatted text file
 	let formattedLicenseNotice = "GRAPHITE THIRD-PARTY SOFTWARE LICENSE NOTICES\n\n";
-	if (debugMode) formattedLicenseNotice += "WARNING: Licenses for Rust packages are excluded in debug mode to improve performance — do not release without their inclusion!\n\n";
+	if (!rustLicenses) formattedLicenseNotice += "WARNING: Licenses for Rust packages are excluded in debug mode to improve performance — do not release without their inclusion!\n\n";
```
Keavon, line 172: Here we'd also check for `rustLicenses.length === 0`.

about.hbs
Keavon: I see. I thought this was some weird JS syntax that I was unfamiliar with, but now I understand we preprocess this file's string to remove it. To avoid causing the confusion I faced, I'd suggest moving this to a line above and make it into actual valid syntax: a string statement.

```js
"This JS was generated by cargo-about and the presence of this text indicates the whole file, including the following license array expression, should be valid.";

[...]
```

(This is valid syntax because `eval("'Hello World'; [123, 456]");` returns `[123, 456]`.)

You can check for that string expression (I'd suggest [startsWith()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) rather than `substr()` for simplicity) and then you don't even have to bother using substring to remove it from what `eval()` sees.

frontend/vue.config.js
```diff
@@ -1,18 +1,53 @@
...
+			// eslint-disable-next-line no-eval
+			rustLicenses = eval(stdout.substr(CODEWORD.length));
+		} else {
+			console.error("Unexpected output from cargo-about", result.stdout);
+		}
+	} catch (e) {
+		// Nothing to show. Error messages were printed above.
+	}
+
+	if (!rustLicenses) {
```
Keavon, line 42: See the comment below about changing this to `if (rustLicenses.length === 0) {`.

Keavon (commented):
Looks good! Just some comments to improve simplicity and readability/clarity.

frontend/vue.config.js
```diff
@@ -1,18 +1,53 @@
...
+			encoding: "utf8",
+			timeout: 60000, // one minute
+			shell: true,
+			windowsHide: true, // hide the DOS window on windows
+		});
+
+		if (result.status !== 0) {
+			if (result.status !== 101) {
+				// cargo returns 101 when the subcommand wasn't found
+				console.error("cargo-about failed", result.status, result.stderr);
```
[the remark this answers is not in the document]
cauthmann: If cargo-about isn't found (status 101), no message is emitted here. This message is for non-101 exit codes.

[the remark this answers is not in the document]
Keavon: Gotcha 👍

frontend/vue.config.js
```diff
@@ -1,18 +1,53 @@
...
 
-let rustLicenses = [];
-let debugMode = false;
-try {
-	// eslint-disable-next-line global-require, import/extensions, import/no-unresolved
-	rustLicenses = require("./rust-licenses");
-} catch (_) {
-	// Rust licenses are not generated by Cargo About except in release mode (`npm run build`)
-	debugMode = true;
+let rustLicenses = null;
```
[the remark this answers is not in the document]
cauthmann: There's a difference between "we don't know the rust licenses" (null) and "we know the rust licenses, but there aren't any" (empty array). While the latter state is unlikely to happen in practice, I believe that a clean distinction is more important than saving an indentation.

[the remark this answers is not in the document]
Keavon: Hmm, that's not something I had thought of. Ok, fair enough, that sounds compelling to me.

Keavon (approved):

about.hbs
Keavon: @cauthmann Sorry, I apparently marked this as resolved by accident. I just updated the branch on top of master so we're almost ready to merge, but if you could take a look at the above comment, that would be much appreciated. Sorry for the delay in the past few days, I've been out of town until now.

cauthmann:
One more time.
  * cargo-about now generates valid javascript. The prefix is a label, which is ignored by eval()
  * added a helper function to invoke cargo-about. The helper function is only called in `formatThirdPartyLicenses()` (i.e. when actually building), not everytime someone includes `vue.config.js`. Apparently the latter happens several times during a build.

PR #382, opened 2021-10-05
Update Viewport bounds on scroll

### LINKED ISSUE #381: Editor not notified on scroll
The editor is not notified about the page being scrolled. This leads to the location of editing not matching the location of the mouse. For example, if a short browser window is used and the user scrolls down half an inch to use the lower tools, edits made in that state will show up a half inch above the mouse click/drag locations.

dchiasson:
The viewport bounds were not being updated when the page was scrolled.
This led to the canvas editing location not matching the mouse location.

Closes #381

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/382)

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -1,7 +1,7 @@
 import { createApp } from "vue";
 
 import { fullscreenModeChanged } from "@/utilities/fullscreen";
-import { onKeyUp, onKeyDown, onMouseMove, onMouseDown, onMouseUp, onMouseScroll, onWindowResize } from "@/utilities/input";
+import { onKeyUp, onKeyDown, onMouseMove, onMouseDown, onMouseUp, onMouseScroll, onWindowResize, onScroll } from "@/utilities/input";
 import "@/utilities/errors";
 import App from "@/App.vue";
 import { panicProxy } from "@/utilities/panic-proxy";
@@ -32,4 +32,5 @@ const wasm = import("@/../wasm/pkg").then(panicProxy);
 	window.addEventListener("mouseup", onMouseUp);
 
 	window.addEventListener("wheel", onMouseScroll, { passive: false });
+	window.addEventListener("scroll", onScroll, true); // scroll event does not bubble for elements so we must capture it
 })();
--- a/frontend/src/utilities/input.ts
+++ b/frontend/src/utilities/input.ts
@@ -112,7 +112,7 @@ export async function onMouseScroll(e: WheelEvent) {
 	}
 }
 
-export async function onWindowResize() {
+async function updateViewportBounds() {
 	const viewports = Array.from(document.querySelectorAll(".canvas"));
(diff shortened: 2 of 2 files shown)


dchiasson:
The scrolling callback is triggered many times in succession (maybe once per pixel?). Should I add any throttling? I noticed that window resize has the same behavior.

frontend/src/main.ts
```diff
@@ -32,4 +32,5 @@ const wasm = import("@/../wasm/pkg").then(panicProxy);
 	window.addEventListener("mouseup", onMouseUp);
 
 	window.addEventListener("wheel", onMouseScroll, { passive: false });
+	window.addEventListener("scroll", onScroll, true);
```
Keavon, line 35: It would be helpful to add a comment explaining the reason that the `useCapture` argument needs to be `true`. I am also thinking we should check if the `Event.target` is the document, so this isn't triggered on various elements within the application.

Keavon:
> The scrolling callback is triggered many times in succession (maybe once per pixel?). Should I add any throttling? I noticed that window resize has the same behavior.

The spamming is happening only when actually scrolling the page (or resizing the page), right? That should be correct, since the scroll distance or window size has changed each time by at least a pixel and you therefore want to re-run the code, right? If that explanation doesn't cover what you are describing, I can take a closer look with some elaboration.

The rest of your PR looks nice and simple and clean! Thanks for solving this issue.

dchiasson:
I think the scroll event is being consumed by the `main-window` element and not bubbling up to the window or document. I hard-coded a class check in the handler, but I don't know if that is the best method. Sounds like @cauthmann's direction will be better.

> Any modern browser will throttle resize and scroll events. They're fired at most once per frame.
Ok, that is what I am seeing.

dchiasson:
Oh, I also noticed that if I draw something and then scroll the window down for the first time, the drawing will jump to the left by the amount that the horizontal scrollbar is scrolled left. I haven't had a chance to look into it yet.

Keavon:
Sorry I wasn't able to look into the specifics of this and decide the best route. However the long-term goal was to completely eliminate the full-page scrollbar because a desktop app should always responsively resize to fit its window. This has now been properly implemented in [commit ca4faabd5cdb00b160493b3d84659b0377e8b077](https://github.com/GraphiteEditor/Graphite/commit/ca4faabd5cdb00b160493b3d84659b0377e8b077) which means there should be no scrollbar to throw off the calculations. If you do notice any continued issues, please point them out! I will close this PR but I still wish to express my sincere thanks for your efforts even though it didn't ultimately get merged.

dchiasson:
No problem, glad to hear it is moving forward. Maybe I'll have better luck on my next PR 😛

Keavon:
Certainly! If you ever want some suggestions now or in the future, I'm always happy to pick out some suitable issues from the backlog based on your areas of experience, just let me know.

PR #386, opened 2021-10-25
Layer snapping

### LINKED ISSUE #200: Add an alignment/snapping system to the Select Tool
When dragging a layer around with the Select Tool, vertical lines should show when the edges or center of the layer is aligned with the edges or center of other layers in the document and it should snap when close to such points.

~~Blocked on #143~~

0HyperCube:
Closes #200 

- [ ] Display lines (probably wait for overlay system?)
- [x] Don't search for every layer each mouse move
- [x] Snap new shapes while they're being created
- [x] Hook up the snap (🧲) icon to entirely enable/disable snapping behavior (on by default)

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/386)


Keavon:
This is looking awesome! Please reduce `SNAP_TOLERANCE` from `5` to `3` which aligns better with user expectations and, especially while we have no snapping visualizations yet, avoids user confusion about why the snapping behavior is often being triggered so generously for random things.

I also added a task to the list in this main post:
> Hook up the snap (🧲) icon to entirely enable/disable snapping behavior (on by default)
If you'd prefer that I do that, let me know and I can make a commit to do that in the next few days.

0HyperCube:
> Please reduce `SNAP_TOLERANCE` from `5` to `3` 

Done

> Hook up the snap (magnet) icon to entirely enable/disable snapping behavior (on by default)

Done

> Lastly, we'll need to eventually support snapping while creating lines with the Line Tool and points with the Pen Tool. If you'd prefer to merge first and address that in a later PR, that would be fine— up to you. Perhaps just the Line Tool, for now, might be easy to do first.

Done

> Is there anything else that you know needs to be done before we begin code review and prepare to merge this? Thanks again for continuing to work on this during the past month of slow progress. Things will be picking up again in the next few weeks when I regain free time!

Thanks. I've marked it ready for review.

editor/src/tool/tools/resize.rs
```diff
@@ -10,25 +11,37 @@ pub struct Resize {
 	pub path: Option<Vec<LayerId>>,
 }
 impl Resize {
-	pub fn calculate_transform(&self, center: Key, lock_ratio: Key, ipp: &InputPreprocessor) -> Option<Message> {
-		let mut start = self.drag_start;
-		let stop = ipp.mouse.position;
+	/// Starts of a resize, assigning the snap targets and snapping and assigning the starting position.
```
Keavon, line 14: Typo: "Starts of a resize"

editor/src/tool/tools/select.rs
```diff
@@ -173,20 +173,31 @@ impl Fsm for SelectToolFsmState {
 						DrawingBox
 					};
 					buffer.into_iter().rev().for_each(|message| responses.push_front(message));
+
+					// ToDo: Clean up cloning
```
Keavon, line 177: Please change to `// TODO: ` for consistency with other usages. Specifically, because this breaks my VS Code extension which calls out TODO comments when they are formatted that way. Thanks.

editor/src/tool/tools/select.rs
```diff
@@ -173,20 +173,31 @@ impl Fsm for SelectToolFsmState {
...
+					let ignore_layers = match data.bounding_box_path.as_ref() {
+						Some(bounding_box) => Vec::from([bounding_box.to_vec()]),
+						None => Vec::new(),
```
Keavon, line 180: Can these be simplified with the `vec![]` macro? I'm actually not at all sure what the `Vec::from([bounding_box.to_vec()])` even does.

editor/src/consts.rs
```diff
@@ -22,6 +22,7 @@ pub const SLOWING_DIVISOR: f64 = 10.;
 
 // SELECT TOOL
 pub const SELECTION_TOLERANCE: f64 = 1.;
+pub const SNAP_TOLERANCE: f64 = 3.;
```
Keavon, line 25: This probably is more appropriate in the VIEWPORT section several lines up.

editor/src/document/snapping.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+impl SnapHandler {
+	/// Gets a list of snap targets for the X and Y axes in Viewport coords for the target layers (usually all layers or all non-selected layers.)
+	/// This should be cached at the start of a drag.
+	pub fn start_snap(&mut self, document: &Document, target_layers: Vec<Vec<LayerId>>, ignore_layers: &[Vec<LayerId>]) {
+		if self.snapping_enabled {
+			let targets = target_layers
+				.iter()
+				.filter(|path| !ignore_layers.contains(path))
```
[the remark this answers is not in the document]
0HyperCube: Added the comment. Iterators are lazy so with optimisations they should yield the same assembly. Also it doesn't really make sense to group the filter that discards the ignored layers with one that finds the viewport bounding box of a layer.

editor/src/document/snapping.rs
```diff
@@ -0,0 +1,108 @@
...
+		Self {
+			snap_targets: None,
+			snapping_enabled: true,
+		}
+	}
+}
+
+impl SnapHandler {
+	/// Gets a list of snap targets for the X and Y axes in Viewport coords for the target layers (usually all layers or all non-selected layers.)
+	/// This should be cached at the start of a drag.
```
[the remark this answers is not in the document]
0HyperCube: Nice spot, I meant 'called' not 'cached'.

Keavon (approved):

Keavon:
Looks great! Thanks for humoring my pedantry @0HyperCube in that round of reviews. We should be good to merge but perhaps, since the weekend is now here, we can wait to see if @TrueDoctor has the chance to take a look with his superior Rust experience. But feel free to merge anytime.

TrueDoctor:
I am not really a fan of passing a mutable document to the tools, there should be relatively easy way to avoid having to do that. If I am missing crucial benefits of using a `&mut` please correct me

0HyperCube:
> I am not really a fan of passing a mutable document to the tools, there should be relatively easy way to avoid having to do that. If I am missing crucial benefits of using a `&mut` please correct me

@TrueDoctor It is needed to cache the snap targets to avoid walking through the layer tree each frame. I am unsure where to put the cache without having pass a mutable reference to the document message handler to the tools.

TrueDoctor:
@0HyperCube Could you put the cache into the tool storage?
We will eventually try to execute different steps of the pipeline in parallel and for that we don't want different steps to depend on a mutable reference at the same time

0HyperCube:
@TrueDoctor Sure, done.

TrueDoctor:
Nice this looks promising, will take a final look and then you should be good to go

editor/src/tool/tools/resize.rs
```diff
@@ -1,34 +1,54 @@
...
-			Operation::SetLayerTransformInViewport {
-				path,
-				transform: DAffine2::from_scale_angle_translation(size, 0., start).to_cols_array(),
+			let stop = self.snap_handler.snap_position(document, ipp.mouse.position);
+
+			let mut size = stop - start;
+			if ipp.keyboard.get(lock_ratio as usize) {
+				size = size.abs().max(size.abs().yx()) * size.signum();
+			}
+			if ipp.keyboard.get(center as usize) {
```
TrueDoctor, line 33: Not at all your pr, just something I noticed:
Should we add a `ipp.keyboard.is_presseed(key: Key)` method?

TrueDoctor (approved):
Looks good to me, great work

Keavon:
> @0HyperCube Could you put the cache into the tool storage? We will eventually try to execute different steps of the pipeline in parallel and for that we don't want different steps to depend on a mutable reference at the same time

I am not sure if the tool storage is the best spot for it, since caching is a global system, not specific to one tool. The Select Tool is certainly the most obvious, but also the Path Tool will use it for snapping vertices, the <kbd>G<kbd> (grab) action will use it, and plenty others I'm not thinking of now. We could save that for a future refactor though, I'm just commenting this to mention it. Thoughts? @TrueDoctor @0HyperCube

TrueDoctor:
The main thing is that caching should not take place in the document itself and this is a fine solution for now.
having this in the tool is also kinda nice because the cache is unlikely to get invalidated during the usage of a tool which would not be the case for a global cache.

0HyperCube:
@Keavon The path tool snapping is already implemented with a separate cache in the above refactor. The cache is cleared after the tool is finished (i.e. when your line is created or when your grab is released) - this is because most of the snapping operations mutate the new snap targets. If the cache is empty then there are just 56 bytes of stack (`size_of::<Option<SnapHandler>>()`)and no bytes of heap used.

Keavon:
Sounds good. I'm happy to merge now. Do you want to have the honor of pushing the button?

PR #387, opened 2021-11-01
Upgrade Graphite to Rust 2021 edition; upgrade dependencies

Keavon:
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/387)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,4 @@
 [workspace]
-rust-version = "1.56.0"
-edition = "2021"
 members = [
 	"editor",
 	"graphene",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,6 @@
 [workspace]
+rust-version = "1.56.0"
+edition = "2021"
 members = [
 	"editor",
 	"graphene",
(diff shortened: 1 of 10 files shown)


proc-macros/src/discriminant.rs
```diff
@@ -59,16 +59,19 @@ pub fn derive_discriminant_impl(input_item: TokenStream) -> syn::Result<TokenStr
...
-						let attr: Attribute = syn::parse_quote! {
-							#[#tokens]
-						};
-						Some(attr)
-					}
-					Err(e) => {
-						attr_errs.push(syn::Error::new(a_span, e));
-						None
+				.then(|| {
+					let _ = &a;
```
Keavon, line 63: Is this necessary? @TrueDoctor

Cargo.toml
```diff
@@ -1,4 +1,6 @@
 [workspace]
+rust-version = "1.56.0"
+edition = "2021"
```
Keavon, line 3: Is this correct to be added to the `[workspace]` section?

TrueDoctor: no, it has to be added to the individual cargo toml files, working on a fix

proc-macros/src/discriminant.rs
TrueDoctor: I mean that line seem to do absolutely nothing, did you try without it?

Keavon: I did not try without it since I don't know precisely what to test for in case of breakages. I was hoping you might be able to tell if it'd theoretically be a problem. See: https://doc.rust-lang.org/edition-guide/rust-2021/disjoint-capture-in-closures.html

TrueDoctor (approved):
Good to merge if you remove that `let _ = ` line

Issue #267, opened 2021-07-15
Confirm when the browser window is closed and there is unsaved work


Keavon:
This should only happen when any open documents have unsaved work.

In `main.ts` we should add `window.addEventListener("beforeunload", onWindowClose);` (see https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onbeforeunload) and define/export `onWindowClose` right below `onWindowResize()` in `utilities/input.rs`. It should check if any documents are currently unsaved (i.e. displaying an asterisk). If so, show the prompt with the text `"Unsaved work will be lost if the web browser tab is closed. Close anyway?"` (note: some browsers like Chrome won't display that message). This browser prompt should give the user the opportunity to leave the site or stay on the page. If there are no documents that are unsaved, this `beforeunload` event should be canceled so that the page closes immediately. (See the MDN docs for how to use this event in a cross-browser compatible manner for both the prompt and canceled cases.)

Later, when the Auto-Save feature (#233) is implemented, this should only prompt if somehow the auto-save hasn't happened yet (perhaps if writing to IndexedDB is asynchronous and a large file might take some time, or if changes are batched every few seconds to be written to IndexedDB and the save hasn't happened yet). But that's a detail to be implemented in that feature, not this one.

~~Blocked on #268~~

PR #392, opened 2021-11-22
Display Asterisk on Unsaved Documents

### LINKED ISSUE #268: Display an asterisk next to document tabs and page title with unsaved work
Making changes to the document should display the asterisk. Performing undo back to the most-recently-saved ~~(at the time of writing, that means the beginning of document history since there is no saving support yet)~~ should remove the asterisk.

The page title should also update to show the name of the current document including any asterisk.

Related: #267

mfish33:
Closes #268

An asterisk will now display next to the document tab as well as the title when it is unsaved. Detection for an unsaved document is "ok" as described in the code comments. In order to have more accurate detection, there may need to be changed on the rust side as well.

In addition, I have moved wasm-pack to optional dependencies since on the m1 mac it will fail to install.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/392)


frontend/src/utilities/documents.ts
```diff
@@ -84,16 +93,49 @@ export default readonly(state);
...
 	const documentListData = responseData as UpdateOpenDocumentsList;
 	if (documentListData) {
-		state.documents = documentListData.open_documents;
-		state.title = state.documents[state.activeDocumentIndex];
+		state.documents = documentListData.open_documents.map((name) => {
+			// Can't guarantee an exact match since it is going off of names
+			// This is ok though since it is only used to mark a document as unsaved
+			const probableMatch = state.documents.find((doc) => doc.name === name);
+			if (probableMatch) {
+				return { name, unsaved: probableMatch.unsaved, layerTreeHeight: probableMatch.layerTreeHeight };
```
Keavon, line 101: Let's change it so the Rust code sends all this information instead of just the names. We can pair program this part so I can show you the Rust stuff.

frontend/src/utilities/documents.ts
```diff
@@ -84,16 +93,49 @@ export default readonly(state);
...
 	const documentData = responseData as SetActiveDocument;
 	if (documentData) {
 		state.activeDocumentIndex = documentData.document_index;
-		state.title = state.documents[state.activeDocumentIndex];
+		state.title = state.documents[state.activeDocumentIndex].name;
+	}
+});
+
+// Used to figure out if the document has been modified since the last save
+// Pitfalls: This will not detect if a new effect has been applied but just if the hierarchy has been modified
```
Keavon, line 118: This is a nice attempt working with what you have in the browser, since it has very little "context" about the state of the application. Although I believe the only way to do this "correctly" would be to edit the Rust code which has the entire application state and contextual information. Let's pair program this so I can show you the Rust codebase and we can make this work based on any change that has occurred since the document was saved.

frontend/src/utilities/documents.ts
```diff
@@ -119,6 +161,7 @@ registerResponseHandler(ResponseType.ExportDocument, (responseData: Response) =>
 
 registerResponseHandler(ResponseType.SaveDocument, (responseData: Response) => {
 	const saveData = responseData as SaveDocument;
+	state.documents[state.activeDocumentIndex].unsaved = false;
```
Keavon, line 164: Let's make this happen in the Rust code, since the frontend isn't supposed to modify state.

frontend/package.json
```diff
@@ -40,7 +40,9 @@
 		"sass-loader": "^8.0.2",
 		"typescript": "^4.4.4",
 		"vue-loader": "^16.8.2",
-		"vue-template-compiler": "^2.6.14",
+		"vue-template-compiler": "^2.6.14"
```
Keavon, line 43: Let's save this change for another PR, since it might have some unintended consequences. Such as: the change to `package-lock.json`, which might or might not be related to this change, has had 15.5k lines added to it for some reason. We should investigate this further. But for now, please discard the changes to `package-lock.json` and `package.json` right before we merge this PR.

Keavon (commented):
Nice work! Let's do some pair programming/debugging to figure out the things mentioned in the comments.

bjorn3:
If you revert all changes since the last save using ctrl+z or ctrl+shift+z does ot revert back to showing that it is already saved?

mfish33:
The revised rust implementation works properly with undo and redo since it is now hooked into the history system. In addition, Eslint needed to be upgraded because Typescript 4.0 changed the way that tuple types are represented in the AST. As a consequence of upgrading, some rules became stricter and I updated the code in a few places to comply. Finally, the reason for the large package.json is that I am on a newer version of npm. Missed this when I ran npm install for the first time:
```bash
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile 
npm WARN old lockfile This is a one-time fix-up, please be patient...
```
@Keavon Do we want to upgrade this for the project?

frontend/package.json
Keavon: I went ahead and made #395 which, with some careful experimentation, upgrades everything that we can possibly upgrade except the things that are blocked on Vue CLI 5.0 which is in the release candidate stage right now. This also cleaned up some new linter errors and things. Some of this might be a duplication of your work, but it was better suited for another PR. If you made any improvements that I didn't include also in that commit, please feel free to include them. Sorry if this means you have some merge conflicts, I am more than happy to help you sort them out.

Keavon:
> If you revert all changes since the last save using ctrl+z or ctrl+shift+z does ot revert back to showing that it is already saved?

@bjorn3 this works properly now, thanks.

Keavon:
Since it's likely to be buried in the resolved code review comment [here](https://github.com/GraphiteEditor/Graphite/pull/392#discussion_r758176890), let me copy this comment I wrote into the main thread:

I went ahead and made #395 which, with some careful experimentation, upgrades everything that we can possibly upgrade except the things that are blocked on Vue CLI 5.0 which is in the release candidate stage right now. This also cleaned up some new linter errors and things. Some of this might be a duplication of your work, but it was better suited for another PR. If you made any improvements that I didn't include also in that commit, please feel free to include them. Sorry if this means you have some merge conflicts, I am more than happy to help you sort them out.

Keavon:
@mfish33 we also assigned you to #267 which is a natural extension to this feature, and should be relatively small to tackle once this one is merged. Feel free to un-assign yourself if you'd prefer not to work on it. And of course, if you have any questions or want to pair program on it, I'm always very happy to help!

editor/src/document/document_file.rs
```diff
@@ -332,23 +335,30 @@ impl DocumentMessageHandler {
...
+		self.backup(responses);
+		self.undo(responses)
+		// TODO: Consider if we should check if the document is saved
 	}
 
-	pub fn undo(&mut self) -> Result<(), EditorError> {
-		match self.document_history.pop() {
+	pub fn undo(&mut self, responses: &mut VecDeque<Message>) -> Result<(), EditorError> {
+		// Push the GetOpenDocumentsList message to the bus in order to update the save status of the open documents
+		responses.push_back(DocumentsMessage::GetOpenDocumentsList.into());
```
TrueDoctor, line 359: It feels a bit weird for these function to unconditionally push an update to the message queue, I'm wondering if there are situations where you would want to use the history operations without a message queue. But this probably helps to avoid errors

TrueDoctor: (but you could return the message as the Ok() part, then clippy is going to complain if you don't use the `result but you could still use the function without a message queue)`

editor/src/document/document_message_handler.rs
```diff
@@ -96,6 +98,11 @@ impl DocumentsMessageHandler {
 			responses.push_back(DocumentMessage::LayerChanged(layer.clone()).into());
 		}
 	}
+
+	// Returns an iterator over the open documents in order
+	pub fn ordered_document_iterator(&self) -> impl Iterator<Item = &DocumentMessageHandler> {
+		self.document_ids.iter().map(|id| self.documents.get(id).expect("document id is not found in the document hashmap"))
```
TrueDoctor, line 104: @Keavon this sounds a bit off, shouldn't it be `was not found` or `is not in`?

editor/src/document/document_file.rs
```diff
@@ -359,7 +369,10 @@ impl DocumentMessageHandler {
 		}
 	}
 
-	pub fn redo(&mut self) -> Result<(), EditorError> {
+	pub fn redo(&mut self, responses: &mut VecDeque<Message>) -> Result<(), EditorError> {
+		// Push the GetOpenDocumentsList message to the bus in order to update the save status of the open documents
```
TrueDoctor, line 373: The GetOpenDocumentsList Message sholud probably be renamed to UpdateOpenDocumentsList, righ?

editor/src/document/document_file.rs
```diff
@@ -368,13 +381,23 @@ impl DocumentMessageHandler {
...
 				Ok(())
 			}
 			None => Err(EditorError::NoTransactionInProgress),
 		}
 	}
 
+	pub fn current_history_hash(&self) -> u64 {
+		// We can use the last state of the document to serve as the hash to compare against
+		// This is useful since when the document is empty the hash will be 0
+		self.document_undo_history.last().map(|(graphene_document, _)| graphene_document.hash()).unwrap_or(0)
```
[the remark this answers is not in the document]
mfish33: Is this expensive? calling:
```rs
graphene_document.hash()
```
results in just calling:
```rs
self.hasher.finish()
```
does hasher.finish have a time complexity greater than O(1)? This could be replaced with a counter that starts at 0 when saved and increments on a redo or a new change and then decrements on an undo. However, this moves the responsibility of maintaining the counter to several places inside document_file instead of a single source of truth. However, if you think that it will help performance I can change it.

editor/src/document/document_file.rs
mfish33: That's a good point. However, do we want to allow this function to be run without the message queue? Currently, we want to send the GetOpenDocumentsList message at all times, so it would lead to code duplication that if missed at a later time could create a hard-to-track down bug.

editor/src/document/document_message_handler.rs
mfish33: Good catch

editor/src/document/document_file.rs
mfish33: changed

editor/src/document/document_file.rs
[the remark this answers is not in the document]
TrueDoctor: Oh, yeah was wondering what sort of hash function you were using… lets say there is a reason for the hashing function containing unsafe code, the "hash" is not "really" as hash, it is more of a hash with some randomnes sprinkled in.
I thought I had removed that impl when transitioning to the uuid based system.
The implementation you are currently using is not a performance concern, but well it kinda relies on a bit of undefined behaviour and `std::mem::transmute_copy`'s.
If there is enough value in the current system, we should keep it, otherwise we should probably replace it with something sensible.
The main issue with the current system is that hash(x) might not be equal to hash(x), meaning x == y does not imply hash(x) == hash(y)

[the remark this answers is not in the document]
TrueDoctor: You can also take a look at the hash implementation yourself and judge whether you can reconcile this with your conscience. Float are just a pain to deal with

editor/src/document/document_file.rs
TrueDoctor: Yeah its fine as it is we can change it should it become a problem

editor/src/document/document_file.rs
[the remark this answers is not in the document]
mfish33: I went to implement it as a counter, however, it becomes a problem since once a new change is made to the document after undoing needs to invalidate the counter permanently. There are ways that this can be done but it starts to feel like the incorrect solution. The hash I think is safe to use since I am never using it to compare two different documents. The documents saved in document_undo_history are used immutable so I am using the hash to just identify a single state of a document.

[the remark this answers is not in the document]
mfish33: It might make sense to change the name from hash to identifier. This would alleviate some of the concerns around the fact that x == y does not imply hash(x) == hash(y)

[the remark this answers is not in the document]
TrueDoctor: Yeah that's a good idea, I thought I had already done that and then crossed it off my mental todo list. If you want you can append the change to this pr, otherwise I'll create an issue for it.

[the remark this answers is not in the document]
mfish33: OK. I added those changes plus a note about the identifier to try and prevent misconceptions in the future.

TrueDoctor (approved):
Looks good from the rust side, the package.lock file is a bit concerning I thought that was fixed in master?

Keavon:
> Looks good from the rust side, the package.lock file is a bit concerning I thought that was fixed in master?

Yeah, it still needs to be merged with those changes in master. Happy to help with that if you have any troubles @mfish33.

TrueDoctor:
> Yeah, it still needs to be merged with those changes in master. Happy to help with that if you have any troubles @mfish33.

Is it an option to just remove the file and regenerate it?

mfish33:
Just finished, just did it in two steps, because I did not want to manually make sure that package-lock.json was merged correctly since my fork had a different package.json

frontend/src/utilities/documents.ts
```diff
@@ -83,17 +84,14 @@ export default readonly(state);
...
-		state.documents = documentListData.open_documents;
-		state.title = state.documents[state.activeDocumentIndex];
-	}
+	state.documents = documentListData.open_documents.map(({ name, isSaved }) => `${name}${isSaved ? "" : "*"}`);
 });
 
 registerResponseHandler(ResponseType.SetActiveDocument, (responseData: Response) => {
 	const documentData = responseData as SetActiveDocument;
 	if (documentData) {
+		console.log(`Setting active index to ${documentData.document_index}`);
```
Keavon, line 93: This is left over from debugging

Keavon (approved):

PR #394, opened 2021-11-23
Change responses to use classes instead of interfaces

### LINKED ISSUE #399: Change Frontend Responses to use Classes
Currently, the frontend uses a reactive system to alert different parts of the frontend about editor changes. In order to do this, all responses are described by interfaces and have a corresponding new function.
```ts
export interface SaveDocument {
	document: string;
	name: string;
}
function newSaveDocument(input: any): SaveDocument {
	return {
		document: input.document,
		name: input.name,
	};
}
```
Switching to a class-based system would have various benefits
- Classes stay in after compilation. This would allow us to offer better type checking on the main subscription API
```ts
function registerResponseHandler(responseType: ResponseType, callback: ResponseCallback)
```
to
```ts
function subscribeJsMessage<T extends JsMessage>(responseType: Constructs<T>, callback: JsMessageCallback<T>)
```
The advantage of this change is that the callback given to subscribeJsMessage automatically has the correct type for the input parameter and does not need to be cast. Removing this cuts down on boilerplate code where mistakes can be made.

- The other advantage of a class-based system is we can remove the new function to create a given type. Instead, we can make use of a library [class-transformer](https://github.com/typestack/class-transformer) in order to automate this process, further cutting down on code duplication.

- Finally while implementing the class solution, naming should be refactored to better align with the backend message system

mfish33:
For some reason, Github closed this issue while rebasing to get rid of the commit from my other PR. This is not related to an active issue, but I saw an opportunity to try and improve the codebase. This PR is not done and needs to be cleaned up a little. Is this change something that is worth pursuing further? The advantage of this pattern is that it makes the subscribing to responses more explicit as well as cuts down on repeated code.

Closes #399
Workaround for #396

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/394)


Keavon:
For future reference, you can reopen a closed PR or issue instead of creating a new one. But no worries, we'll use this one from now on.

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,260 @@
...
+			const popped = currentFolderStack.pop();
+			if (!popped) throw Error("Too many negative indents in the folder structure");
+			if (popped) currentFolder = popped;
+		}
+	}
+
+	return currentFolder;
+}
+
+// NOTE: Why was it newPath(input.data.path) instead of input.path. Seems to work at input.path?
```
mfish33, line 177: Do you know why it is this way?

Keavon: It looks like `editor\src\frontend\frontend_message_handler.rs:18` defines this as `UpdateLayer { path: Path, data: LayerPanelEntry },` and that `LayerPanelEntry` also has a `path` field. I am not sure precisely what the difference is without further digging, but I'd suggest leaving it alone to prevent a regression, unless you believe that was already a mistake. We can as TrueDoctor for his thoughts, if so.

TrueDoctor: Currently they are the same thing, but I am actually not sure if that will always be the case.
It is basically a key value pair in which the value also contains the key. But that does not currently make a difference.

Keavon: Is one of the two a safer option to use? @TrueDoctor

TrueDoctor: Too late rn will try not to forget to answer tororow

TrueDoctor: I don't think there is a good reason to keep both of them in there, we should just get rid of one of them

mfish33: Ok, I removed path from the UpdateLayer Message

frontend/src/components/panels/Document.vue
```diff
@@ -318,46 +319,40 @@ export default defineComponent({
 		},
 	},
 	mounted() {
-		registerResponseHandler(ResponseType.UpdateCanvas, (responseData: Response) => {
-			const updateData = responseData as UpdateCanvas;
+		registerJsMessageHandler(UpdateCanvas, (updateData) => {
 			if (updateData) this.viewportSvg = updateData.document;
```
mfish33, line 323: What is the point of this if statement? As long as updateData is an object this will return true. The old system as well as this one should not emit a message unless it is nonnull.

Keavon: To satisfy typescript, if I recall correctly. If removing it works, to ahead.

Keavon:
Looks good. Still needs a rebase though. After rebasing, do you want me to review it?

mfish33:
Ready for review

frontend/src/components/panels/Document.vue
```diff
@@ -332,50 +333,34 @@ export default defineComponent({
 		},
 	},
 	mounted() {
-		registerResponseHandler(ResponseType.UpdateCanvas, (responseData: Response) => {
-			const updateData = responseData as UpdateCanvas;
-			if (updateData) this.viewportSvg = updateData.document;
+		registerJsMessageHandler(UpdateCanvas, (updateData) => {
```
Keavon, line 336: Would you mind renaming the arrow function's parameter as the camelCased version of its name name, e.g. `updateCanvas` for this one, throughout?

frontend/src/components/panels/LayerTree.vue
```diff
@@ -425,18 +423,17 @@ export default defineComponent({
...
-				this.setOpacityForSelectedLayers();
+		registerJsMessageHandler(UpdateLayer, (updateData) => {
+			const responsePath = updateData.data.path;
+			const responseLayer = updateData.data;
+
+			const layer = this.layerCache.get(responsePath.toString());
+			if (layer) Object.assign(this.layerCache.get(responsePath.toString()), responseLayer);
+			else {
+				this.layerCache.set(responsePath.toString(), responseLayer);
 			}
```
Keavon, line 434: Did the autoformatter do this? I try to keep consistency with the if and else cases, so they either both have the curlies or both don't. If you have control over it, and the autoformatter doesn't change it back, could you switch it so both cases either do or don't have them? (They are arguably kinda long lines, so I don't have a strong preference either way if we use or omit curlies.)

frontend/src/utilities/js-message-dispatcher.ts
```diff
@@ -0,0 +1,94 @@
...
+
+	if (callback && message) {
+		callback(message);
+	} else if (message) {
+		// eslint-disable-next-line no-console
+		console.error(`Received a Response of type "${responseType}" but no handler was registered for it from the client.`);
+	}
+}
+
+export function registerJsMessageHandler<T extends JsMessage>(responseType: Constructs<T>, callback: JsMessageCallback<T>) {
```
Keavon, line 92: I'm thinking it might be a clearer name to just call this `registerJsMessage`, and would also reduce some code length in many places. Thoughts?

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,256 @@
...
+
+	@To255Scale
+	green!: number;
+
+	@To255Scale
+	blue!: number;
+
+	alpha!: number;
+
+	toRgb() {
```
Keavon, line 29: `toRgba` would be a more accurate name.

frontend/src/components/widgets/inputs/SwatchPairInput.vue
```diff
@@ -135,20 +136,16 @@ export default defineComponent({
...
-			this.primaryColor = { r: primary.red, g: primary.green, b: primary.blue, a: primary.alpha };
-			let color = this.primaryColor;
+			this.primaryColor = primary.toRgb();
 			let button = this.getRef<HTMLButtonElement>("primaryButton");
-			button.style.setProperty("--swatch-color", `rgba(${color.r}, ${color.g}, ${color.b}, ${color.a})`);
+			button.style.setProperty("--swatch-color", primary.toString());
 
-			this.secondaryColor = { r: secondary.red, g: secondary.green, b: secondary.blue, a: secondary.alpha };
-			color = this.secondaryColor;
+			this.secondaryColor = secondary.toRgb();
```
Keavon, line 146: Might as well group this line together with the `this.primaryColor = primary.toRgb();` line above.

frontend/src/utilities/js-message-dispatcher.ts
Keavon: Or would `subscribe` be a better word than `register` perhaps?

frontend/src/main.ts
```diff
@@ -1,5 +1,5 @@
+import "reflect-metadata";
```
Keavon, line 1: Could you please write a comment describing why this has to be imported in `main.ts` if it's not used in this file?

frontend/src/utilities/js-message-dispatcher.ts
mfish33: I like subscribe

frontend/src/utilities/js-message-handler-binding.ts
```diff
@@ -1,4 +1,4 @@
 // This file is instantiated by wasm-bindgen in `/frontend/wasm/src/lib.rs` and re-exports the `handleResponse` function to
 // provide access to the global copy of `response-handler.ts` with its shared state, not an isolated duplicate with empty state
 
-export { handleResponse } from "@/utilities/response-handler";
+export { handleResponse } from "@/utilities/js-message-dispatcher";
```
Keavon, line 4: This file should be called `js-message-dispatcher-binding.ts` (or maybe `js-message-dispatcher-reexport.ts` if you think that's more clear) since it's meant to shadow the name of the "real" file (the reexported one, `js-message-dispatcher.ts`).

frontend/src/utilities/js-message-handler-binding.ts
```diff
@@ -1,4 +1,4 @@
 // This file is instantiated by wasm-bindgen in `/frontend/wasm/src/lib.rs` and re-exports the `handleResponse` function to
 // provide access to the global copy of `response-handler.ts` with its shared state, not an isolated duplicate with empty state
```
Keavon, line 2: Update the name in this comment too, please.

frontend/src/components/panels/LayerTree.vue
mfish33: Im an always curly bracer :)

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,256 @@
...
+		return { r: this.red, g: this.green, b: this.blue, a: this.alpha };
+	}
+
+	toString() {
```
Keavon, line 33: `toRgbaCSS()` might be slightly preferable to imply what kind of string it is: a CSS RGBA string.

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,256 @@
...
+			const layerId = layerIdsSection.getBigUint64(layersEncountered * 8, true);
+			layersEncountered += 1;
+
+			const childLayer = new DisplayFolderTreeStructure(layerId, []);
+			currentFolder.children.push(childLayer);
+		}
+
+		// Check the sign of the MSB, where a 1 is a negative (outward) indent
+		const subsequentDirectionOfDepthChange = (msbSigned & msbMask) === BigInt(0);
+		// debugger;
```
Keavon, line 160: Please delete this line.

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,256 @@
...
+	| "hard-light"
+	| "difference"
+	| "exclusion"
+	| "hue"
+	| "saturation"
+	| "color"
+	| "luminosity";
+
+function newOpacity(input: number): number {
+	return input * 100;
```
Keavon, line 215: Do you think this would, for consistency, be a good place for a transformer?

frontend/src/utilities/js-messages.ts
```diff
@@ -0,0 +1,256 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable camelcase */
+/* eslint-disable max-classes-per-file */
+
+import { Transform, Type } from "class-transformer";
+
+export class JsMessage {
+	static responseMarker = true;
```
Keavon, line 8: What does `responseMarker` do? Please comment its purpose, thanks.

frontend/src/utilities/js-messages.ts
mfish33: I was going back and forth on that one. I can put it inside the transformer

frontend/src/utilities/js-message-dispatcher.ts
```diff
@@ -0,0 +1,94 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+import { reactive } from "vue";
+import { plainToInstance } from "class-transformer";
```
Keavon, line 4: Nitpick: please put a separator newline between the first two imports (of libraries) and the other imports (of source code files).

frontend/src/utilities/js-messages.ts
mfish33: added the comment, "The marker provides a way to check if an object is a sub-class constructor for a jsMessage."

frontend/src/utilities/js-message-dispatcher.ts
```diff
@@ -0,0 +1,94 @@
...
+} as const;
+
+export type JsMessageType = keyof typeof responseMap;
+
+function isResponseConstructor(fn: ConstructsJsMessage | ((data: any) => JsMessage)): fn is ConstructsJsMessage {
+	return (fn as ConstructsJsMessage).responseMarker !== undefined;
+}
+
+export function handleResponse(responseType: JsMessageType, responseData: any) {
+	const MessageMaker = responseMap[responseType];
```
Keavon, line 68: Do you think it's correct to use PascalCase here? The variable *contains* a class, but the variable *itself* is not a class. I don't have much of an opinion on the matter, but I thought I'd bring it up to ask yours.

mfish33: I could go either way as well. The more I look at it the more I think it should be camelCase

Keavon (approved):

Keavon:
This is a rockstar PR. You know TS way better than me so I really, really appreciate your simplification of my janky code. Ready to merge after CI is done!

Keavon:
Actually, before you merge, could you really quickly file an issue describing this change and assign it to the current sprint, then edit this PR's description to say it closes that issue number? I want it to count as one of the big features of this sprint and that's how we keep track of the sprint accomplishments (versus small commits for little fixes that aren't tracked in the task board).

mfish33:
> This is a rockstar PR. You know TS way better than me so I really, really appreciate your simplification of my janky code. Ready to merge after CI is done!

Thank you for the kind words

mfish33:
oops

Keavon:
> oops

Ah, did you hit Close instead of Merge? I'm surprised reopening it requires another CI run. Kind of annoying, oh well.

Keavon:
Haha, that's kind of funky: two sets of each check passed, since it had to rerun them on the same commit.

PR #397, opened 2021-12-04
Confirm when the browser window is closed and there is unsaved work

### LINKED ISSUE #267: Confirm when the browser window is closed and there is unsaved work

mfish33:
- browser confirmation on page exit
- prompt to save a document when closed will only trigger when unsaved

Closes #267

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/397)

### DIFF
--- a/frontend/src/utilities/documents.ts
+++ b/frontend/src/utilities/documents.ts
@@ -16,12 +16,17 @@ import { panicProxy } from "@/utilities/panic-proxy";
 
 const wasm = import("@/../wasm/pkg").then(panicProxy);
 
+class DocumentState {
+	readonly displayName: string;
+
+	constructor(readonly name: string, readonly isSaved: boolean) {
+		this.displayName = `${name}${isSaved ? "" : "*"}`;
+	}
+}
+
 const state = reactive({
-	documents: [] as string[],
+	documents: [] as DocumentState[],
 	activeDocumentIndex: 0,
-	get activeDocument() {
-		return this.documents[this.activeDocumentIndex];
-	},
 });
 
 export async function selectDocument(tabIndex: number) {
@@ -29,9 +34,13 @@ export async function selectDocument(tabIndex: number) {
 }
 
 export async function closeDocumentWithConfirmation(tabIndex: number) {
-	selectDocument(tabIndex);
+	const targetDocument = state.documents[tabIndex];
+	if (targetDocument.isSaved) {
+		(await wasm).close_document(tabIndex);
+		return;
+	}
 
-	const tabLabel = state.documents[tabIndex];
+	const tabLabel = targetDocument.name;
 
 	createDialog("File", "Save changes before closing?", tabLabel, [
 		{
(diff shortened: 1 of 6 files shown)


frontend/src/utilities/documents.ts
```diff
@@ -16,22 +16,34 @@ import { panicProxy } from "@/utilities/panic-proxy";
...
+	if (targetDocument.isSaved) {
+		(await wasm).close_document(tabIndex);
+		return;
+	}
+
+	// Show the document is being prompted to close
+	await selectDocument(tabIndex);
 
-	const tabLabel = state.documents[tabIndex];
+	const tabLabel = targetDocument.name;
```
Keavon, line 46: Let's use `.displayName` here so we see the asterisk.

frontend/src/utilities/documents.ts
```diff
@@ -16,22 +16,34 @@ import { panicProxy } from "@/utilities/panic-proxy";
...
 	(await wasm).select_document(tabIndex);
 }
 
 export async function closeDocumentWithConfirmation(tabIndex: number) {
-	selectDocument(tabIndex);
+	const targetDocument = state.documents[tabIndex];
+	if (targetDocument.isSaved) {
+		(await wasm).close_document(tabIndex);
+		return;
+	}
```
Keavon, line 41: When will this occur? As I recall the situation here, this response/message `closeDocumentWithConfirmation` should only be sent off from the backend in the case where a tab is being closed which was confirmed by the backend to be unsaved. This doesn't hurt, but it seems unnecessary. However my memory of exactly when this occurs could be spotty... does this code actually change behavior at all?

Keavon:
Looks great!

frontend/src/utilities/documents.ts
mfish33: closeDocumentWithConfirmation is called by the Vue frontend. The if statement stops the popup from being displayed if the document is already saved

Keavon: Oh, gotcha. Yes, that makes sense. I had gotten that confused with the other messages (`DisplayConfirmationToCloseDocument` and `DisplayConfirmationToCloseAllDocuments`) which is sent from the backend. This is a good chance 👍. Ready to merge!

Keavon (approved):

PR #401, opened 2021-12-11
Implement outline view mode

caleb-ad:
Early first draft.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/401)


Keavon:
Looking good! Here's a few improvements that would be great to see before merging:

- The scale of the wireframe should always be 1px regardless of canvas zoom. Look at how this is done with the overlays.
- For now, we should draw all shape outlines with a black stroke regardless of their fill or stroke color. In the future, we can give the user the option of all black, preference to the fill (if it exists, otherwise stroke if it exists, otherwise black), or preference to stroke, and an option to scale light colors to some minimum darkness so white outlines aren't invisible on the white page background. But for now, all black is a safe default that would match the capabilities of other software, and those handy other options aren't urgent.
- Please run `cargo fmt` so your code passes CI. Ideally, VS Code should do this automatically for you when you hit save if you're using that editor.

Great work so far! I'll do a code review once these things are done and hopefully it can be a relatively quick merge process. Welcome!

caleb-ad:
Thanks for the welcome!
Everything seems to be working properly, and I remembered to cargo fmt :-).

One thing I have not messed with is the layer thumbnails. It seems like it might be better for the thumbnail to always reflect the normal view? but maybe not... Currently the thumbnails don't update immediately on a view mode change, but they will update when something else asks for an update.

Keavon:
Ah, yes, the thumbnails should never be shown in wireframe view.

Keavon:
A quick bug I'll mention here before I forget: <kbd>Ctrk</kbd><kbd>0</kbd> doesn't handle the resize.

editor/src/document/document_file.rs
```diff
@@ -562,6 +564,17 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+			SetViewMode(mode) => {
+				self.graphene_document.update_view_mode(mode);
+				let mut mode_update_func = |s: &mut GrapheneShape| s.style.view_mode(mode);
+				GrapheneDocument::visit_all_shapes(&mut self.graphene_document.root, &mut mode_update_func);
+				responses.push_back(
+					FrontendMessage::UpdateCanvas {
+						document: self.graphene_document.render_root(),
+					}
+					.into(),
+				);
```
TrueDoctor, line 576: This should be changed to a `RenderDocument` Message to allow for lazy evaluation

editor/src/document/document_file.rs
```diff
@@ -694,7 +706,12 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					.into(),
 				);
 			}
-
+			ReRenderDocument => {
+				//When the document is zoomed the wireframes must be re-rendered in order to maintain wireframe stroke width
+				if self.graphene_document.view_mode() == ViewMode::WireFrame {
+					GrapheneDocument::visit_all_shapes(&mut self.graphene_document.root, &mut (|_s: &mut GrapheneShape| {}));
+				}
+			}
```
TrueDoctor, line 714: This seems weird to me, I would expect something like this:
1. Receive SetViewMode Message
2. Send operation to recursively apply view mode
3. Apply view mode in the document
4. Send RenderDocument Message

graphene/src/document.rs
```diff
@@ -36,6 +39,7 @@ impl Document {
 
 	/// Wrapper around render, that returns the whole document as a Response.
 	pub fn render_root(&mut self) -> String {
+		log::debug!("{:?}", self.root.transform);
```
TrueDoctor, line 42: log statement

graphene/src/document.rs
```diff
@@ -50,6 +54,14 @@ impl Document {
...
 	}
 
+	pub fn view_mode(&self) -> ViewMode {
+		self.view_mode
+	}
+
+	pub fn update_view_mode(&mut self, mode: ViewMode) {
+		self.view_mode = mode;
+	}
+
```
TrueDoctor, line 64: Mutations of the document should usually not be done through function calls but rather Operations but I think it is fine in this case because the contents of the document don't really change

graphene/src/document.rs
```diff
@@ -154,6 +166,27 @@ impl Document {
 		Ok(())
 	}
 
+	/// Visit each layer recursively, applies modify_shape to each non-overlay Shape
+	/// Currently used to swap between viewmodes
+	pub fn visit_all_shapes<F: FnMut(&mut Shape) -> ()>(layer: &mut Layer, modify_shape: &mut F) -> bool {
```
TrueDoctor, line 171: It would be nice if there was a function that also cleans the upstream caches in case you want to work on anything but the root

graphene/src/layers/style/mod.rs
```diff
@@ -83,16 +107,27 @@ impl PathStyle {
...
+			match (self.view_mode, self.fill) {
+				(ViewMode::WireFrame, _) => Fill::none().render(),
+				(_, Some(fill)) => fill.render(),
+				(_, None) => String::new(),
 			},
-			match self.stroke {
-				Some(stroke) => stroke.render(),
-				None => String::new(),
+			match (self.view_mode, self.stroke) {
+				(ViewMode::WireFrame, _) => Stroke::new(Color::BLACK, WIRE_FRAME_STROKE_WIDTH).render(),
```
TrueDoctor, line 128: These should probably be exported as constants somewhere

TrueDoctor:
Just to have that written down somewhere, it would be conceptually cleaner to separate the document storage from the rendering, the viewmode information does not really belong there (e.g. you wouldn't want to save a document that way)
It would essentially nice to have a document.render and a document.render_wireframe function.
But the implementation overhead of such a dual structure would be to big, ideally we'd want to implement some kind of bridge pattern to deal with that, but this should be fine for now. Will hopefully have the time to review this tomorro

editor/src/document/movement_handler.rs
```diff
@@ -199,6 +203,7 @@ impl MessageHandler<MovementMessage, (&mut LayerData, &Document, &InputPreproces
 					layerdata.scale *= new_scale;
 					responses.push_back(FrontendMessage::SetCanvasZoom { new_zoom: layerdata.scale }.into());
 					responses.push_back(ToolMessage::SelectedLayersChanged.into());
+					responses.push_back(DocumentMessage::DispatchOperation(Box::from(DocumentOperation::SetViewMode { mode: data.1.view_mode() })).into());
```
TrueDoctor, line 206: Couldn't you just do `DocumentOperation::SetViewmode {…}.into()` ?

graphene/src/layers/style/mod.rs
```diff
@@ -83,16 +108,26 @@ impl PathStyle {
 	pub fn clear_stroke(&mut self) {
 		self.stroke = None;
 	}
+	pub fn view_mode(&mut self, new_mode: ViewMode) {
+		self.view_mode = new_mode;
+	}
+	pub fn get_view_mode(&self) -> ViewMode {
+		self.view_mode
+	}
+
```
TrueDoctor, line 117: The rust naming convention discourages explicit `get` functions in most cases.
These should probably be `view_mode` and `update_view_mode` to stay consistent with the naming on Document

TrueDoctor:
Could you modify the render function to take a `RenderMode` parameter?
I think it makes more sense to have the thing behave differently during render that tho change the property of the layer.  That would also simplify dealing with thumbnails

graphene/src/layers/style/mod.rs
```diff
@@ -1,6 +1,9 @@
 use crate::color::Color;
 use serde::{Deserialize, Serialize};
 const OPACITY_PRECISION: usize = 3;
+pub const WIRE_FRAME_STROKE_WIDTH: f32 = 1.0;
+pub const EMPTY_FILL: Fill = Fill::none();
+pub const THIN_BLACK_STROKE: Stroke = Stroke::new(Color::BLACK, WIRE_FRAME_STROKE_WIDTH);
```
0HyperCube, line 6: Perhaps change to WIREFRAME_FILL and WIREFRAME_STROKE to be more descriptive about what it is used for (and to make sense if the wireframe colour is changed)

graphene/src/layers/style/mod.rs
```diff
@@ -83,16 +108,26 @@ impl PathStyle {
...
+				(_, Some(fill)) => fill.render(),
+				(_, None) => String::new(),
 			},
-			match self.stroke {
-				Some(stroke) => stroke.render(),
-				None => String::new(),
+			match (self.view_mode, self.stroke) {
+				(ViewMode::WireFrame, _) => THIN_BLACK_STROKE.render(),
+				(_, Some(stroke)) => stroke.render(),
+				(_, None) => String::new(),
```
0HyperCube, line 130: This is quite hard to read, perhaps nesting would add clarity:
```rs
match self.view_mode{
	ViewMode::WireFrame => format!("{}{}",EMPTY_FILL.render(),THIN_BLACK_STROKE.render()),
	_ => format!(
		"{}{}",
		match self.fill {
			Some(fill) => fill.render(),
			None => String::new(),
		},
		match self.stroke {
			Some(stroke) => stroke.render(),
			None => String::new(),
		}),
}
```

caleb-ad:
@TrueDoctor I modified the render functions to take the view mode as an argument. This made things far simpler, and I probably should have done it this way from the start. I also took this opportunity to effect @TrueDoctor 's suggestion, and move the view mode out of the graphene::Document and into the DocumentMessageHandler, which seems like a much more natural place for it.

And as a bonus I think the layer thumbnails are behaving properly now!

TrueDoctor:
That sounds awesome! Glad that worked out :smile:

graphene/src/layers/style/mod.rs
caleb-ad: I changed THIN_BLACK_STROKE to WIREFRAME_STROKE (tbh that was a kinda weird sounding name), but I kept EMPTY_FILL because it is a fairly descriptive name which may be used for other things besides wireframe.

editor/src/document/document_file.rs
```diff
@@ -127,6 +129,7 @@ pub enum DocumentMessage {
 	DocumentHistoryBackward,
 	DocumentHistoryForward,
 	ClearOverlays,
+	SetViewMode((ViewMode, bool)),
```
Keavon, line 132: This is already a tuple so it doesn't need a tuple inside a tuple.

graphene/src/layers/style/mod.rs
```diff
@@ -83,16 +98,23 @@ impl PathStyle {
 	pub fn clear_stroke(&mut self) {
 		self.stroke = None;
 	}
-	pub fn render(&self) -> String {
+
+	pub fn render(&self, mode: (ViewMode, bool)) -> String {
+		// change stroke rendering so solid paths don't dissapear
```
Keavon, line 103: Is this a `// TODO;` comment, or is this describing what this function does, or what a part of this function does?

graphene/src/document.rs
```diff
@@ -363,6 +383,8 @@ impl Document {
 		operation.pseudo_hash().hash(&mut self.state_identifier);
 		use DocumentResponse::*;
 
+		// intercept operations here for wireframe view
```
Keavon, line 386: Should this comment be removed?

Keavon:
Looks good, thanks for the big simplification by switching to the approach of informing the render engine, not each layer.

To speed things along and get this merged quickly, I went ahead and made a few polishing changes in a commit I just pushed. Hopefully this doesn't step on your efforts, but I figured it's faster to make the tweaks directly instead of typing up changes for you to apply in the code review. I did, however, leave two code review questions above asking about the status of two comments that might be out of date.

Here's a summary of my changes:
- Renamed "wireframe" to "overlay" since I was inconsistent about my usage of the terminology, however "overlay" should be the user-facing name to avoid confusion with UX wireframing.
- Broke out `SetViewMode((ViewMode, bool))` into `SetViewMode(ViewMode)` plus `DirtyRenderDocument` and `DirtyRenderDocumentInOutlineView`. This helps avoid confusion about what the bool was meant to do, since there's no way of naming a tuple, and it clarifies the purpose of that message where calling `SetViewMode` with a `false` value for "update" didn't actually *set the view mode*, it just told it to rerender. So now the calls where "update" was `false` go to `DirtyRenderDocumentInOutlineView`, which in turn calls `DirtyRenderDocument` only if the editor is in the outline view mode.
- Created a `consts.rs` file in Graphene, since the Editor backend already has an equivalent file, and moved `LAYER_OUTLINE_STROKE_COLOR` and `LAYER_OUTLINE_STROKE_WIDTH` in there.
- Changed the render function's `mode: (ViewMode, bool)` argument so it doesn't take a confusing tuple (which has no names to describe its arguments) and only takes `ViewMode`. The bool (`mode.1`) was used to perform the highly case-specific logic of `let view_mode = if mode.1 { ViewMode::Normal } else { mode.0 };` which the render function doesn't directly need to know, so I moved that out in order for its callers to perform that logic: `if self.overlay { ViewMode::Normal } else { view_mode }`.

Please let me know if you disagree with anything, you're most welcome to change stuff if you still want. But with these changes, I give you my approval in the code review process and we are ready to merge once we figure out what's going on with those two potentially stray comments I commented on above.

Thank you and congratulations on your first PR!

Keavon (approved):

graphene/src/layers/style/mod.rs
caleb-ad: I think that was a ```// TODO: ``` from before my time, I'm not quite sure what it means actually.

graphene/src/document.rs
caleb-ad: yeah, that comment is old

caleb-ad:
The changes look good

PR #406, opened 2021-12-18
Remove all use of document indices

### LINKED ISSUE #417: Avoid referencing documents in tabs by their index
This finishes the incomplete work done in #225

mfish33:
Closes #417

This removes all uses of document_index. This will hopefully help prevent the backend and the frontend from getting out of sync. 

I wanted to use the generate_uuid to make the document id's. However, wasm-bindgen does not support u64 types inside structs. I can not find anything directly saying this but internally wasm-bindgen uses JSON.parse which will return a number rather than a BigInt.

To remedy this I created another UUID generator that uses i32's which can be represented by js numbers correctly. Alternatively, We could replace JSON.parse with https://www.npmjs.com/package/json-bigint and force Bigint if the number is small but going to be used with rust API's that require u64's.

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/406)


TrueDoctor:
We ran into this issue already, the reason for the error is, that the json spec does not support u64 integers.
one way to get it working without too much effort is to just split the u64 into two u32, send them and reassemble them in the frontend. Thats how we currently transfer paths.

mfish33:
Ok, I see that. From looking around I found this project https://github.com/cloudflare/serde-wasm-bindgen. I think it would make sense to use this instead of the default wasm-Bingen serializer. I have opened a PR to add Bigint support since it seems like there is already interest in it. In the meantime, we can use the fork on my Github. This will simplify some existing code as well as make it more seamless when we transition to a native frontend.

Keavon:
I believe this PR fixes a bug I just discovered that exists in master where the editor crashes if you close the rightmost tab while it is saved (doesn't display an asterisk and thus doesn't prompt to save or discard the document) and isn't the currently-active tab (and thus has to switch to it before displaying the close confirmation prompt). I'll skip filing a crash report since this will soon be fixed, nice job!

Keavon:
I'll review this tomorrow after it's rebased.

frontend/src/dispatcher/js-messages.ts
```diff
@@ -127,23 +139,25 @@ export class DisplayFolderTreeStructure extends JsMessage {
 }
 
 interface DataBuffer {
-	pointer: number;
-	length: number;
+	pointer: bigint;
+	length: bigint;
```
Keavon, line 143: Why lowercase rather than `BigInt`?

editor/src/frontend/frontend_message_handler.rs
```diff
@@ -4,16 +4,23 @@ use crate::tool::tool_options::ToolOptions;
 use crate::Color;
 use serde::{Deserialize, Serialize};
 
+#[derive(PartialEq, Clone, Deserialize, Serialize, Debug)]
+pub struct FrontendDocumentState {
```
Keavon, line 8: What would you think about renaming this from -`State` to -`Status` or -`Details`, or perhaps you can think of another better name? I'm a little worried "state" implies statefulness as it pertains to the application architecture, but really this is more of just a "status" or its current "details" not related to statefulness.

frontend/src/state/documents.ts
```diff
@@ -8,63 +8,57 @@ import {
...
-		const targetDocument = state.documents[tabIndex];
-		if (targetDocument.isSaved) {
-			editor.instance.close_document(tabIndex);
+	const closeDocumentWithConfirmation = async (documentId: BigInt) => {
+		// Assume we receive a correct document_id
+		const targetDocument = state.documents.find((doc) => doc.id === documentId) as FrontendDocumentState;
+		if (targetDocument.is_saved) {
+			editor.instance.close_document(targetDocument.id);
 			return;
 		}
```
Keavon, line 35: I'm realizing this check should probably be moved to the backend.

frontend/src/state/documents.ts
```diff
@@ -8,63 +8,60 @@ import {
...
+		const targetDocument = state.documents.find((doc) => doc.id === documentId) as FrontendDocumentState;
+
 		// Close automatically if it's already saved, no confirmation is needed
-		const targetDocument = state.documents[tabIndex];
-		if (targetDocument.isSaved) {
-			editor.instance.close_document(tabIndex);
+		if (targetDocument.is_saved) {
+			editor.instance.close_document(targetDocument.id);
 			return;
 		}
```
Keavon, line 37: I'm realizing this check should probably be moved to the backend.

Keavon (approved):
Looks good!

mfish33:
I want to wait until https://github.com/cloudflare/serde-wasm-bindgen/pull/24 gets merged. Think it should only be a day or two

Keavon:
> I want to wait until [cloudflare/serde-wasm-bindgen#24](https://github.com/cloudflare/serde-wasm-bindgen/pull/24) gets merged. Think it should only be a day or two

Sounds good, but I'm also happy to merge sooner then swap out the crate source after. Of course, the few changes I suggested should be fixed also, please.

frontend/src/dispatcher/js-messages.ts
```diff
@@ -41,9 +41,7 @@ export class UpdateInputHints extends JsMessage {
 	readonly hint_data!: HintData;
 }
 
-export class HintGroup extends Array<HintInfo> {}
```
mfish33, line 32: @Keavon how do you feel about this being a type def vs class extension. I think it makes the code a little bit easier to read being just a type

Keavon: Chrome devtools gives it a name when it's a class. But I don't have a strong preference.

mfish33: I think for continuity for the rest of the other messages I will make it a typedef. If there is a strong need in the future we can switch the rest over as well.

Keavon: Sure, that's fine.

PR #410, opened 2021-12-20
Move layer selection logic from vue to editor

### LINKED ISSUE #297: Move layer range selection logic to the Rust editor codebase
#137 added layer range selection in JS (using <kbd>Shift</kbd> and <kbd>Ctrl</kbd> modifier keys to expand a selection in the Layer Tree panel). See [LayerTree.vue](https://github.com/GraphiteEditor/Graphite/blob/master/frontend/src/components/panels/LayerTree.vue) for the `handleControlClick()`, `handleShiftClick()`, `handleClick()`, `deselectAllLayers()`, `fillSelectionRange()`, `clearSelection()`, and `sendSelectedLayers()` methods which pertain to this. Currently range selection mostly works (although there are bugs that can cause index out of bounds exceptions in JS), but it is very limited and buggy. Changing the list of layers (adding, duplicating, moving, etc.) causes problems with the selection context. Range selection will also need to work across folders. Another problem is that range selection does not preserve a shared context between layer selection in the Layer Tree panel and within the viewport. This shared context needs to understand which layer was most recently selected to know which range to expand from.

Another small detail is that <kbd>Shift</kbd> is the only modifier key used for range selection of layers in the viewport, since but <kbd>Ctrl</kbd> is also used in the Layer Tree panel. That's because the viewport only adds/removes layers from the selection while the Layer Tree does that and also does range selection from the recent selected layer context.

otdavies:
Closes #297

Future improvements when flattened data structure for layers is persisted:
* Remove reliance on layer_is_between(..)

---
This change is [Reviewable](https://reviewable.io/reviews/graphiteeditor/graphite/410)


frontend/wasm/src/api.rs
```diff
@@ -306,6 +306,11 @@ impl Editor {
 		self.dispatch(message);
 	}
 
+	pub fn select_layer(&self, paths: Vec<LayerId>, ctrl: bool, shift: bool) {
```
TrueDoctor, line 309: this does not look ffi safe, I would expect there to be an issue with bigints not being sendable across the js/rs boundary

otdavies: This was how this was being done previous as far as I am aware. I just checked prior to my commit and similar examples seem to exist. Recommendations on how to proceed?

TrueDoctor: Actually, nevermind, I think this could work: https://rustwasm.github.io/wasm-bindgen/reference/types/number-slices.html
(If not, we would have to turn paths into &[LayerId])
This just doesn't work for nested arrays

TrueDoctor: The wasm-bindgen docs don't explicitly mention Vec as a supported type (or at least I didn't see it) but that might just be a lack of documentation

TrueDoctor: You can leave it as it its I just got confused by the documentation

TrueDoctor (commented):
I'll have to review this more thoroughly unfortunately didn't really have enough time today

TrueDoctor:
This is definitely a change of behavior, when shift selecting layers, the point from where the selection starts is always kept the same, even if it is no longer selected, this seems a bit weird to me

otdavies:
> 

I've attempted to copy window's behavior when it comes to selection. If it deviates from that, then that is unintended. In windows, if you continue to hold shift and click elsewhere it will _not_ choose a new start point for selection even if you deselect that folder with ctrl or deselect all together. 

If it doesn't behave correctly in that context, I want to immediately remedy that!

graphene/src/document.rs
```diff
@@ -119,13 +120,60 @@ impl Document {
...
+
+			// If these are equal it means we are comparing folder elements
+			if a == b {
+				continue;
+			}
+
+			// If this is smaller, it is closer to the root
+			if a < b {
+				return true;
+			}
```
TrueDoctor, line 142: I am not really sure how this makes sense, if the layer id is a uuid, the ordering of ids doesn't really make sense here.
the proper way would probably be to find the ordering in the corresponding folder

TrueDoctor:
> > 
> 
> I've attempted to copy window's file explorer's behavior (by Keavon's request) when it comes to selection. If it deviates from that, then that is unintended. In window's file explorer, if you continue to hold shift and click elsewhere it will _not_ choose a new start point for selection even if you deselect that file/folder with ctrl or deselect all together.
> 
> If it doesn't behave correctly in that context, I want to immediately remedy that!

That seems to be the intended behaviour, we can probably still tweak the ux afterwards with something like a highlight for the last selected layer

Keavon:
Please fix the behavior to match Windows Explorer. Namely, a <kbd>Ctrl</kbd> click should change the range selection reference layer.

graphene/src/document.rs
otdavies: The id here is the indices, not a UUID. Check the above lines.

otdavies:
> Please fix the behavior to match Windows Explorer. Namely, a Ctrl click should change the range selection reference layer.

Done

graphene/src/document.rs
otdavies: I clarified the code / comments on this feedback, it would be confusing to anyone reading it.

TrueDoctor: Thanks a lot, admittedly was pretty tired while reading the code but more comments are always welcome

graphene/src/document.rs
```diff
@@ -119,13 +120,68 @@ impl Document {
...
+				// path_a is closer
+				return true;
+			}
+
+			// If neither of the above are true, path_b is closer
+			break;
+		}
+
+		// path_b is closer
+		return false;
```
TrueDoctor, line 154: I was a bit confused by all the loop manipulation.
coldnt lines 144-154 just be replaced by
```rust
return index_a < index_b
```
? But i guess it makes sense in terms of putting comments inbetween

editor/src/document/document_file.rs
```diff
@@ -568,6 +574,45 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+				let mut paths = vec![];
+				let last_selection_exists = !self.layer_range_selection_reference.is_empty();
+
+				// If we don't have ctrl selected, clear last selection
+				if !ctrl {
+					// Todo: This clears the selection, figure out how to improve this
+					self.layer_data.iter_mut().filter(|(_, layer_data)| layer_data.selected).for_each(|(path, layer_data)| {
+						layer_data.selected = false;
+						responses.push_back(LayerChanged(path.clone()).into())
+					});
```
TrueDoctor, line 587: responses.push_back(DeselectAllLayers.into());

editor/src/document/document_file.rs
```diff
@@ -568,6 +574,45 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
...
+						// This allows toggling selection when holding ctrl
+						let layer = self.layerdata_mut(&selected);
+						layer.selected = !layer.selected;
+						responses.push_back(LayerChanged(selected.clone()).into());
+					} else {
+						paths.push(selected.clone());
+					}
+
+					// Set our last selection reference
+					self.layer_range_selection_reference = selected;
```
TrueDoctor, line 610: I think this could probably be made a little bit simpler by using the existing messages for selection manipulition

TrueDoctor (approved):
@Keavon this is a preliminary approval. I've left some suggestions and would like Oliver to read through them before merging but it would be fine if it stayed as is

editor/src/document/document_file.rs
otdavies: It does use the existing functions, just at the end when it has accumulated all the selection to be made..

graphene/src/document.rs
otdavies: I think this simplification is a good suggestion - it is like this due to breaking down the problem into simplified pieces in an earlier stage. Reconsolidating it now makes sense!

editor/src/document/document_file.rs
otdavies: I would love this to work, but this is what I tried initially and it simply doesn't propagate the message in time it seems. It happens too late and deselects everything right after you select.

TrueDoctor: have you tried enabling the debug output to trace the messages?

TrueDoctor: Ah that is due to the push_front you did down there

otdavies:
> @Keavon this is a preliminary approval. I've left some suggestions and would like Oliver to read through them before merging but it would be fine if it stayed as is

I've implemented the suggestions I could. I left some comments regarding why I couldn't implement them all.

I really appreciate all of your thoughtful review and guidance, thank you!

editor/src/document/document_file.rs
otdavies: Oh, this one? _responses.push_front(AddSelectedLayers(paths).into());_. Let me try switching that around.

otdavies: I was following the structure of 
```
SetSelectedLayers(paths) => {
self.layer_data.iter_mut().filter(|(_, layer_data)| layer_data.selected).for_each(|(path, layer_data)| {
	layer_data.selected = false;
	responses.push_back(LayerChanged(path.clone()).into())
});

responses.push_front(AddSelectedLayers(paths).into());
}
```

otdavies: Hmm, I tried reordering the responses there. Making both push back, and other combinations with push front. The selection is somewhat functional when they both push back, but the shift select doesn't work properly. I'll investigate this more.

TrueDoctor: you can always press iirc `3` to open the message debugging (in the dev console) that should help to figure this out

otdavies: Resolved this issue by leaning on the function, _SetSelectedLayers_, that already has the clear logic, ha! 

```
// Add or set our selected layers
if ctrl {
	responses.push_front(AddSelectedLayers(paths).into());
} else {
	responses.push_front(SetSelectedLayers(paths).into());
}
```

As opposed to clearing before hand.

TrueDoctor:
looks good to me

editor/src/document/document_file.rs
```diff
@@ -601,11 +601,14 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 					self.layer_range_selection_reference = selected;
 				}
 
-				// Add or set our selected layers
-				if ctrl {
-					responses.push_front(AddSelectedLayers(paths).into());
-				} else {
-					responses.push_front(SetSelectedLayers(paths).into());
+				// Don't create messages for empty operations
```
otdavies, line 604: I realized, based on your astute suggestion @TrueDoctor, to monitor the messages with '3' + log,  that we were sending empty messages on occasion. Resolved that.

PR #414, opened 2021-12-20
Fix undefined behavior in the editor wasm bridge

mfish33:
This should remove the possible undefined behavior, described by @bjorn3, that was introduced in #379.


Keavon:
@bjorn3 would you mind reviewing this change? Thanks 😊

bjorn3:
I think this indeed fixes the issue.

Keavon:
Awesome, thanks for taking a look. I'll give my approval. Waiting on @TrueDoctor but we could merge early if needed.

Keavon (approved):

frontend/wasm/src/api.rs
```diff
@@ -31,8 +31,11 @@ pub struct Editor {
...
 	pub fn new(handle_response: js_sys::Function) -> Editor {
+		let editor_id = generate_uuid();
+		let editor = editor::Editor::new();
+		EDITOR_INSTANCES.with(|instances| instances.borrow_mut().insert(editor_id, editor));
 		Editor {
-			editor: UnsafeCell::new(editor::Editor::new()),
+			editor_id,
 			instance_received_crashed: Cell::new(false),
 			handle_response,
 		}
```
TrueDoctor, line 41: Could we make this naming a little less confusing? Having two `Editor`s is a bit confusing, had to read the function twice to understand what was going on.

TrueDoctor (commented):
this seems a little bit weird to me, if we already have a hashmap of editors if would expect the dispatch function to just take an editor id. Then the wasm::Editor would be stored in the hashmap directly

mfish33:
> this seems a little bit weird to me, if we already have a hashmap of editors if would expect the dispatch function to just take an editor id. Then the wasm::Editor would be stored in the hashmap directly

Can you explain further on this? The point of the hashmap behind the thread-local is to be able to modify an instance of an editor with a shared reference rather than a mutable one

TrueDoctor:
> Can you explain further on this? The point of the hashmap behind the thread-local is to be able to modify an instance of an editor with a shared reference rather than a mutable one. I also updated the naming of Editor to JsEditorHandle to hopefully make it less confusing.

we are using a hashmap to support multiple editors right?
But we still keep one global instance of the JsEditorHandle.
I feel like it would make sense to go all the way and store these editor handles in the hashmap itself, otherwise I don't really see the point of adding a hashmap

TrueDoctor (approved):
Missed the last pr getting merged, this looks all good

mfish33:
After Discussion @TrueDoctor and myself agree that this PR is ready to go

PR #419, opened 2021-12-21
Make Vue modifier keys work on the Mac keyboard layout

Keavon:


Keavon:
Closed in favor of #736

PR #421, opened 2021-12-21
Fix new document position

0HyperCube:
A new document had its (0,0) in the top left of the screen instead of in the centre until the document was moved. This fixes that.

### DIFF
--- a/editor/src/document/document_file.rs
+++ b/editor/src/document/document_file.rs
@@ -158,6 +158,7 @@ impl DocumentMessageHandler {
 			saved_document_identifier: 0,
 			name,
 			layer_data: vec![(vec![], LayerData::new(true))].into_iter().collect(),
+			layer_range_selection_reference: Vec::new(),
 			movement_handler: MovementMessageHandler::default(),
 			transform_layer_handler: TransformLayerMessageHandler::default(),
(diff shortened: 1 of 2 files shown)


Keavon:
Thanks for the quick PR!

After glancing at the code real quick, I'm wondering why the original default document correctly is centered but other new documents require a special new function specifically for a centered transform? Shouldn't the initial document and later new documents be the same, and this problem thus fixed the same way the initial document is centered?

0HyperCube:
@Keavon The js is sending a resize event (`bounds_of_viewports`) which adjusts the position to be in the centre.

Keavon:
Ah, I see. Could you please explain the reasoning for adding a new constructor function, versus just fixing the behavior in `with_name()` and `with_name_and_content()`? I can't imagine we'd ever want to call either of those constructor functions with the incorrect centering behavior.

Please also move those constructor functions to the top of the `impl` block, I notice right now they are weirdly hidden in the middle of the other functions.

0HyperCube:
@Keavon My reasoning was that decoding the string didn't need the document to be centred (as its position would be overridden anyway) but it doesn't really matter either way.

Keavon (approved):

PR #422, opened 2021-12-25
Implement IndexedDB document auto-save

### LINKED ISSUE #233: "Auto-Save" persistent document storage
We should store the open documents and the artwork within each document in ~~`localStorage` or~~ `IndexedDB`. Other state, like the current tool and working colors, would be a nice bonus. This helps restore previous sessions if the user accidentally closes the browser tab (for example, while intending to use a keyboard shortcut to close a document tab).

The UX of this should work based on the "Auto-Save" feature. When enabled, every change is saved locally and automatically restored on the next visit. When disabled, any saves are discarded and not saved, so the next visit will not reopen any saved documents.

`IndexedDB` should be used, not `localStorage`, because we want to store proper binary data and not be limited to a string. This is how VS Code's online editor (at vscode.dev) does it.

~~Soft-blocked on #324.~~

mfish33:
Closes #233 

WIP. Pretty close but needs an hour or two more to be ready for review.


editor/Cargo.toml
```diff
@@ -29,3 +31,6 @@ package = "graphite-graphene"
 
 [dev-dependencies]
 env_logger = "0.8.4"
+
+[features]
+js = ["getrandom/js"]
```
bjorn3, line 36: Does this work
```suggestion
[target.'cfg(target_arch = "wasm")'.dependencies]
getrandom = { version = "0.2.3", features = ["js"] }
```
? I think you will need to add `resolver = 2` to the `[workspace]` section of the root `Cargo.toml` for this to work.

mfish33: This seems to work ok. It forwards the js feature to getrandom. Documented here: https://doc.rust-lang.org/cargo/reference/features.html. Let me know if I am incorrect since I don't have a lot of experience with cargo features.

mfish33:
@Keavon I think it is ready for the first round of review. I have not implemented a toggle in the file dropdown. Is this worth implementing now or when we make a more comprehensive settings system that restores between sessions?

Keavon:
The ability to control the checkbox toggle in the menu bar menu is not currently built in a way that will really let you control that. So I would suggest that fixing the checkbox is something we give to a person who's knowledgeable in Vue and interested in helping with that as a separate PR.

editor/src/document/document_file.rs
```diff
@@ -413,6 +413,14 @@ impl DocumentMessageHandler {
 		self.current_identifier() == self.saved_document_identifier
 	}
 
+	pub fn saved(&mut self, state: bool) {
+		if state {
```
Keavon, line 417: It's unclear to me what a boolean called `state` means.

editor/src/document/document_file.rs
```diff
@@ -493,7 +501,8 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 				)
 			}
 			SaveDocument => {
-				self.saved_document_identifier = self.current_identifier();
+				self.saved(true);
```
Keavon, line 504: `saved` also seems like a weird name for a void function. Could you use a few more words to describe what this function does?

editor/src/document/document_message_handler.rs
```diff
@@ -76,11 +84,19 @@ impl DocumentsMessageHandler {
...
-		let new_id = generate_uuid();
-		self.active_document_id = new_id;
-		self.document_ids.push(new_id);
-		self.documents.insert(new_id, new_document);
+	fn load_document(&mut self, new_document: DocumentMessageHandler, document_id: u64, replace_first_empty: bool, responses: &mut VecDeque<Message>) {
+		// Special case when loading a document on an empty page
+		if replace_first_empty && self.document_ids.len() == 1 {
+			let first_id = self.document_ids[0];
+			let first_document = self.documents.get(&first_id).unwrap();
+			if first_document.serialize_root().len() == 2 {
```
Keavon, line 92: This magic number `2` seems fragile. If `serialize_root()` ever changes, this will break and it'll be hard to notice until eventually somebody finds the regression. Could we create a temporary empty document in here, call `.serialize_root().len()` on it, and compare it to `first_document.serialize_root().len()`?

editor/src/document/document_message_handler.rs
```diff
@@ -219,16 +235,32 @@ impl MessageHandler<DocumentsMessage, &InputPreprocessor> for DocumentsMessageHa
...
 				let new_document = DocumentMessageHandler::with_name(name, ipp);
-				self.load_document(new_document, responses);
+				self.load_document(new_document, generate_uuid(), false, responses);
 			}
 			OpenDocument => {
 				responses.push_back(FrontendMessage::OpenDocumentBrowse.into());
 			}
-			OpenDocumentFile(name, serialized_contents) => {
-				let document = DocumentMessageHandler::with_name_and_content(name, serialized_contents, ipp);
+			OpenDocumentFile(document_name, document) => responses.push_back(
```
Keavon, line 243: For readability and consistency, I think it's better to use a `{ }` block instead of directly `=> responses.push_back(...),`

editor/src/document/document_message_handler.rs
```diff
@@ -219,16 +235,32 @@ impl MessageHandler<DocumentsMessage, &InputPreprocessor> for DocumentsMessageHa
...
+				document_id,
+				document,
+				document_is_saved,
+			} => {
+				let document = DocumentMessageHandler::with_name_and_content(document_name, document, ipp);
 				match document {
-					Ok(document) => {
-						self.load_document(document, responses);
+					Ok(mut document) => {
+						log::debug!("Setting Saved: {}", document_is_saved);
```
Keavon, line 261: Remember to remove this debug statement before merging.

editor/src/document/document_file.rs
mfish33: I changed it to `set_save_state` it is meant to be an abstraction over the way that we actually track whether a document is saved. @TrueDoctor has mentioned that it is generally bad in rust to use get/set. Is there a better name that you can think of?

Keavon: That's true for getters, not setters. https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

mfish33: Ok, thanks for the clarification. Is the name change better?

Keavon: Yes, thanks

editor/src/document/document_file.rs
otdavies: Might be worth changing `state` to `is_saved`

frontend/src/lifetime/auto-save.ts
```diff
@@ -0,0 +1,77 @@
...
+	dbOpenRequest.onerror = () => {
+		console.error("Graphite IndexedDb error:", dbOpenRequest.error);
+	};
+
+	dbOpenRequest.onsuccess = () => {
+		resolve(dbOpenRequest.result);
+	};
+});
+
+export function createAutoSave(editor: EditorState, documents: DocumentsState) {
```
otdavies, line 30: I would recommend changing the name of `createAutoSave` to `createAutoSaveManager` to maintain consistently with the neighboring files. This will also help reduce confusion.

graphene/src/consts.rs
```diff
@@ -1,5 +1,8 @@
 use crate::color::Color;
 
+// Document
+pub const GRAPHITE_DOCUMENT_VERSION: &'static str = "0.0.1";
```
otdavies, line 4: Preferably `GRAPHENE_DOCUMENT_VERSION`.

editor/Cargo.toml
otdavies: We ended up removing the dep entirely.

Keavon (approved):

PR #423, opened 2021-12-25
Bug Fix: duplication with Ctrl+D didn't properly duplicate

caleb-ad:
### Fixed Bug:
From the discord code-todo-list: "`Ctrl+D to duplicate a shape doesn't update the layer panel to show the new shape, then dragging it crashes the editor`"


TrueDoctor (approved):
I'm not really a fan of the rust fmt on this one, but the code looks good!
You could think about adding one or two more variables to make the code a bit shorter and more readable

PR #429, opened 2021-12-27
Fix Ctrl+scroll wheel zoom point relative to the mouse position

### LINKED ISSUE #428: Ctrl+mouse wheel zoom does not zoom relative to the exact mouse position
Fix the point relative to which Ctrl+scroll wheel zoom zooms into so that it is relative to the mouse position (or maybe just the center of the viewport).

0HyperCube:
Closes #428

### DIFF
--- a/editor/src/document/movement_handler.rs
+++ b/editor/src/document/movement_handler.rs
@@ -161,7 +161,7 @@ impl MessageHandler<MovementMessage, (&mut LayerData, &Document, &InputPreproces
 				let new_viewport_bounds = viewport_bounds * (1. / zoom_factor);
 				let delta_size = viewport_bounds - new_viewport_bounds;
 				let mouse_percent = mouse / viewport_bounds;
-				let delta = (delta_size * -2.) * (mouse_percent - DVec2::splat(0.5));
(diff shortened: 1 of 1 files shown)


editor/src/document/movement_handler.rs
```diff
@@ -161,7 +161,7 @@ impl MessageHandler<MovementMessage, (&mut LayerData, &Document, &InputPreproces
 				let new_viewport_bounds = viewport_bounds * (1. / zoom_factor);
 				let delta_size = viewport_bounds - new_viewport_bounds;
 				let mouse_percent = mouse / viewport_bounds;
-				let delta = (delta_size * -2.) * (mouse_percent - DVec2::splat(0.5));
+				let delta = (-delta_size) * (mouse_percent - DVec2::splat(0.5));
```
Keavon, line 164: Can be simplified to:
```rs
let delta = (DVec2::splat(0.5) - mouse_percent) * delta_size;
```

Keavon (approved):

Keavon:
Thanks! This is so much nicer now that it's fixed. I'm surprised it was such an easy solution :)

PR #432, opened 2021-12-27
Fix console error caused by the Pen Tool

0HyperCube:
The pen tool was causing a `DocumentError: LayerNotFound` because it was trying to delete the preview layer that did not yet exist.


TrueDoctor:
I think I would prefer if we were to split the make operation into two parts and only do the delete explicitly where it is needed, that should help with readability

Keavon (approved):

PR #433, opened 2021-12-27
Snap while dragging with select tool

### LINKED ISSUE #402: Hold Shift to constrain dragging to 45° increments with Select Tool
Adding the <kbd>Shift</kbd> key while dragging selected layer(s) should constrain the movement to the X, Y, and 45° diagonal axes.

We probably also want the <kbd>Shift</kbd> modifier to cause a box selection if the drag's mousedown begins while over an existing shape. So the axis aligned drag should only occur when it is added after a drag has already begun.

This should be viewport-relative, so when the canvas is rotated, it should still be the horizontal, vertical, and diagonal directions of the screen.

0HyperCube:
Closes #402

Hold shift whilst dragging with the select tool to snap to 45 degree increments.


TrueDoctor:
This does not really feel natural yet. @Keavon I think constraining it to 90ˇ increments might be better. especially when hovering close to the origen the shape changes it's snap direction quite frequently.
Another approach would be to constrain movement to the first axis it is moved in when starting to drag with having to release shift and press it again to snap a different angle

0HyperCube:
I think the 90 degree increments are do indeed work better.

Keavon:
It'll help to have a line visualizing the snapping (perhaps a lower transparency line showing the other snap directions also). But that should wait until the overlay system is done. This is ready to merge right away, just give me a minute to code review it.

Keavon (approved):

PR #434, opened 2021-12-27
Drag to rearrange layers in the layer panel

### LINKED ISSUE #141: Drag to rearrange layers in the stack
In the layer panel, we need to drag the selection to a point in-between other layers (by inserting where a temporary line appears showing the destination). Need to consider folders if that's implemented yet. Depends on #137.

Complexity: 2
Involves: Frontend (JS), Editor (Rust), Document (Rust)

0HyperCube:
Closes #141 

TODO:
- [x] Better styling on line
- [x] Stop crash when a folder is dragged inside itself (What is the desired behaviour here?)


otdavies (commented):
Amazing work, already quite usable. 

My one comment so far: Folders that are already expanded should remain expanded after being dragged.

otdavies:
It appears that the yellow insertion line doesn't appear when the number of layers exceeds the available view space (we have to scroll).

Example: 
image

Keavon:
The missing insert line is fixed by giving the insert marker the CSS rule `flex: 0 0 auto;`

Keavon:
I pushed some CSS tweaks to use the blue hovered accent color and avoid shifting the content height. Soon I'll need to tweak some more of the layer panel's CSS to simplify it in general, like the expand/collapse circular buttons and stuff. They are the remnants of an old design mockup.

Keavon:
Much better! Last problem I notice: <kbd>Shift</kbd>-clicking layers in the viewport no longer adds to the selection, it instead replaces the selection.

Keavon:
This works awesomely (besides the crash, although this isn't the only crash originating from doing funky stuff in the layer panel). I'm very impressed how cleanly the code is an how nicely it works!

The only thing is that we ideally want to keep all our business logic in the Rust editor backend. That's why @otdavies also recently ported over the layer range selection code from JS to the Rust backend in #410. However, I'd be happy to file a separate issue for porting that over from JS to Rust after we merge this, since it's nice to quickly prototype out a feature then deal with polishing it later on.

We're also planning to do the big Rust codebase refactor throughout the next 24 hours so it would be good to avoid a lot of merge conflicts. If you think you can port it over and we can get it merged by tomorrow before the refactor starts in about ~14 hours, that'll avoid the problem, or we can merge the JS approach now and deal with the porting to Rust after the refactor. Got any preferences?

0HyperCube:
@Keavon Most of the logic is in Rust, the JS just reports which layer was dragged and which layer it was dragged next to and if it was inserted below or above that layer.

Keavon:
Let me look closer in the morning and I can see if there are specific concerns or if you are right about it being good as is with the separation of concerns between JS and Rust. I admittedly hadn't looked very closely and probably came to some assumptions that might not be fully true.

Keavon (approved):

PR #437, opened 2021-12-28
Populate layer entry cache

### LINKED ISSUE #430: An opened file doesn't populate its layers, causing a crash when they're clicked in the viewport
This is a regression since the *File > Open* system used to work. Now this is more urgent with the auto-save system enabled, causing this bug to be encountered by default (and potentially causing a broken state where the only fix is to clear the site's IndexedDB storage).

0HyperCube:
Closes #430 

The layer entry cache in the frontend was not populated when the document was opened because `UpdateLayer` was never called. This MR just creates the layerdata for all of the layers and then calls `UpdateLayer` on them.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -101,6 +101,7 @@ dependencies = [
  "log",
  "rand_chacha",
  "serde",
+ "serde_json",
  "spin",
  "thiserror",
 ]
@@ -113,7 +114,6 @@ dependencies = [
  "kurbo",
  "log",
  "serde",
- "serde_json",
 ]
 
 [[package]]
@@ -157,9 +157,9 @@ checksum = "9a3a5bfb195931eeb336b2a7b4d761daec841b97f947d34394601737a7bba5e4"
 
 [[package]]
 name = "itoa"
-version = "0.4.8"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b71991ff56294aa922b450139ee08b3bfc70982c6b2c7562771375cf73542dd4"
+checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "js-sys"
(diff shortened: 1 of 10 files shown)


TrueDoctor:
I think it would make more sense to serialize the layerdata along with the document and restore it from there, otherwise you loose contextual information such as if a folder was expanded

0HyperCube:
@TrueDoctor Could we serialise the whole `DocumentMessageHandler`?

0HyperCube:
I don't know if it would make sense to serialise the undo history though?

Keavon:
This is useful for regenerating missing data, although @TrueDoctor is right that we should write the layerdata to the file also in order to preserve that expanded/collapsed information in the file. Although come to think of it, we probably don't need to store the selected layers in the file. We also should probably phase out our usage of the translation/rotation/scale in the LayerData since (after True Doctor talked with me about it yesterday) it sounds like this is a temporary thing until we have a Transform node in the node graph, and I don't mind small floating point errors from the back-and-forth conversion into/from a matrix within the Graphene document. So I propose we get rid of LayerData and store two vectors: a selection list and an expanded/collapsed list. The expanded/collapsed list should be serialized to the saved document.

TrueDoctor:
> This is useful for regenerating missing data, although @TrueDoctor is right that we should write the layerdata to the file also in order to preserve that expanded/collapsed information in the file. Although come to think of it, we probably don't need to store the selected layers in the file. We also should probably phase out our usage of the translation/rotation/scale in the LayerData since (after True Doctor talked with me about it yesterday) it sounds like this is a temporary thing until we have a Transform node in the node graph, and I don't mind small floating point errors from the back-and-forth conversion into/from a matrix within the Graphene document. So I propose we get rid of LayerData and store two vectors: a selection list and an expanded/collapsed list. The expanded/collapsed list should be serialized to the saved document.

We can just add the #[serde_skip] attribute on the selected field of the layerdata struct to achieve the same result

Keavon:
> We can just add the #[serde_skip] attribute on the selected field of the layerdata struct to achieve the same result

I was sort of also hinting at a general simplification refactor. Thoughts on that proposal?

TrueDoctor:
> I was sort of also hinting at a general simplification refactor. Thoughts on that proposal?

If you can explain to me how that improves anything?

Keavon:
We found that the `LayerData` struct was confusing for multiple reasons:
- We assumed it had to match every layer exactly
- We assumed it was a tree
- We didn't understand the precise nature and reasoning behind the transform, rotation, and scale properties
- The fact that selection state and expanded state are stored as booleans per-layer feels weird compared to just storing a list of selected and a list of expanded layers
- The name `LayerData` is perhaps misleading since it sounds much more important than it is, although I can't think of a better name for it

These could certainly be commented to provide more explanation, and we'll do that if we don't refactor it as such. But cleaning up and simplifying the code (just adding a field for selected layers and another for expanded layers on the main `DocumentMessageHandler`) sounds much cleaner to me.

TrueDoctor:
I suppose layer panel entry or layer metadata would describe it better.
If you split up the layerdata into two seperate vectors you just made the problem twice as bad, e.g. when deleting a layer you have to remove it from two vectors instead of just removing it from the one

TrueDoctor (approved):

Keavon (approved):

PR #439, opened 2021-12-29
Remove transformations from LayerData

0HyperCube:
Removes some transformation attributes from the layerdata (as they were only used on the root) and moves them to the MovementHandler.


TrueDoctor (approved):
This all looks good

PR #442, opened 2021-12-30
Fix one item folders

0HyperCube:
Properly displays folders with one item.


Keavon (approved):
I am a huge fan of one-character fixes.

PR #443, opened 2021-12-30
Resize panels

### LINKED ISSUE #194: Add support for dragging the boundaries between panels
This should bind an event listener on the document for drag events after a mousedown event on the gutter between tabs (where it's currently set to make the CSS apply a drag arrow icon to the user's cursor). The updated values should be pixel amounts for the flex basis, accounting for the gaps. This task can start as frontend-only, but eventually a follow-up task should save this to the editor backend. A related task should make it possible to drag around and dock panels and tabs.

0HyperCube:
Closes #194


Keavon:
Nice! It looks like it leaves some black area when growing the viewport area (after the window was resized by dragging your browser window bounds), but that can easily be solved by just calling the `onWindowResize(...)` function in `input.ts`. There should also probably be some reasonable minimum proportion for each panel direction.

Keavon:
* Shrink the viewport (drag its border to the left).
* Resize the browser window (like unmaximizing then re-maximizing it) a bit to make it update the sizing.
* Expand the viewport (drag its border to the right).
* Notice the black area in the right of the viewport and incomplete ruler markings.

Keavon (approved):

PR #451, opened 2022-01-02
Path Tool: Implement anchor point dragging

### LINKED ISSUE #82: Implement the Path Tool for Bezier curve editing
* Create a system for dealing with points.
* The visualization part eventually involves the Viewport Overlays system (#99), but we can temporarily use the already-implemented system for "working layers" (shapes drawn to the canvas in a preview state before they get committed to the document).
* De-conflicting clicks when multiple points are very close together (using Blender's method of picking the one based on distance from the mouse to the edge connected to that point, so you move your mouse closer to one side or the other of the path edge to prioritize one point or the other)
* Points are extracted from all layers but not occluded by the fill of other layers (that's a complexity of behavior we can address later.)

Goes hand-in-hand with #185, the Pen Tool, which creates the Bezier paths similarly to how the Path Tool edits them.

Complexity: 3
Involves: Editor (Rust), Document (Rust)

leonardpauli:
Closes #82


Keavon (approved):

editor/src/tool/tools/path.rs
```diff
@@ -211,6 +223,170 @@ impl Fsm for PathToolFsmState {
...
+								ref layer_path,
+								shape_offset,
+							} => {
+								// TODO make this work for the handles, right now just selects the anchors
+								data.selected_shapes = shapes_to_draw;
+								let shape = &data.selected_shapes[shape_offset];
+								let path = shape.path.clone();
+								let bez: Vec<PathEl> = (&path).into_iter().collect();
+								let transformed = shape.transform.inverse().transform_point2(input.mouse.position);
+								data.selection.bez_segment_id = closest_anchor(&bez, Vec2::new(transformed.x, transformed.y));
```
leonardpauli, line 327: Use new fn `closest_handle` instead? // Seems like it checks bezier curve endpoints (anchor points) but not its other control points (direction handles)?  Note how, on an ellipse, pressing down on a handle marks it correctly, but then when dragging, it drags the closest anchor instead.

otdavies: I took a very slap dash approach to getting this working, the cloest_anchors function doesn't even consider Handles yet. Overall it does have incorrect behavior with anything regarding handles or the close path invisible point. Would love a better way of finding the indexes without brute forcing etc. 

I was just pushing to get this closer to the finish line for game dev meetup this morning. I would be happy to sit down and work on it today!

Issue #444, opened 2021-12-30
Rearranging multiple selected layers loses selection


Keavon:
This is a regression that was broken in 7b28a1754ca922f8bc159884e3a68d531fc169a5 back on August 20. The test called `move_selection` was broken at the time so it didn't catch it, but it was just fixed in #418 which revealed this regression from months ago. The test has been disabled, so fixing this issue should involve un-commenting that test.

To reproduce: make a bunch of layers, select a group of three contiguous ones, then use <kbd>Ctrl</kbd><kbd>[</kbd> or <kbd>Ctrl</kbd><kbd>]</kbd> to rearrange them. This results in the operation succeeding, but only one layer ends up selected after that.

PR #455, opened 2022-01-03
Fix group creation parenting

### LINKED ISSUE #411: Grouping selection that contains layers from multiple folder depths can cause crash
In certain situations when selecting multiple layers from various folder depths, like so: 
image
grouping will cause the editor to crash.

Also grouping two top level groups when they are side by side / minimized also seems to cause a crash.

### LINKED ISSUE #412: Nest newly created group properly when selected layers are within an existing group
When creating a new group make sure it goes under the same parent that the layers selected are already under.
image
As opposed to what it does now: 
image
(It just places it at the bottom)
It should also go between the layers from which the selected layers were grouped from, not just at the bottom of the group or hierarchy.

### LINKED ISSUE #453: Copy & pasting folders should paste as sibling
Currently pastes under the original folder.

### LINKED ISSUE #454: Copying & pasting will cause crash with complex selection across folders

otdavies:
In the process of solving #412 several other issues were resolved.

Closes #412
Closes #411 
Closes #454

Doesn't resolve #453. This needs more discussion.


otdavies:
Regarding #453, the current solution conflicted with the intended functionality of pasting _into_ your selection. This occurs when Copy/Pasting, the copy operation does _not_ deselect the layer after the copy. Thus if you Copy + Paste a folder pasts into itself. My walked back my original "fix" as it doesn't make sense here in that context.

otdavies:
I am seeing a test fail intermittently. It fails roughly 50% of the time. Similar to the previous test we had trouble with (https://github.com/GraphiteEditor/Graphite/pull/444) 
Similar to the test referenced in #444, it is using let mut editor = create_editor_with_three_layers();

otdavies:
Resolution on this PR:

For now we are going to merge the initial less efficient version as the tests work and functions correctly demonstrably in the editor. The more optimized version, with all the ordering fixes included, still doesn't behave correctly in the editor or with the tests. This will be revisited at a later time when optimizations are more important, or whenever anyone wants to take on this TODO.

I accidentally removed you from the PR @TrueDoctor, I just meant to add @Keavon. I added you back. 

Thanks!

Keavon (approved):

Issue #313, opened 2021-07-28
Polish interactivity of the NumberInput widget


henryksloan:
The NumberInput widget should support more common ways of modifying numerical widgets, namely:

1) Holding down the mouse button on the arrow to change it multiple times. This should probably work like key repeat, with an initial delay and a constant rate of repetition.
2) Clicking and dragging on the clickable region of the widget to change the value. The change should either be proportional to `new_x - initial_x` or `(new_x - initial_x) + (new_y - initial_y)`, based on what other editors (GIMP) do, as well as what feels better in Graphite.

piersdeseilligny:
May I suggest that in addition to `Hover mouse over it and Ctrl+scroll wheel to increment/decrement the value` there should also be `Hover mouse over it and scroll horizontally (with trackpad or horizontal scroll wheel) to increment/decrement the value`?

Keavon:
We can certainly make it work with the horizontal wheel while holding <kbd>Ctrl</kbd>, however it is too dangerous to accidentally disrupt values just while scrolling around the interface. Perhaps clicking the input, then horizontally scrolling, would be another acceptable alternative that requires some explicit interaction to avoid accidental disruption.

makoConstruct:
yeah, having Draggy Number Fields would be really important to me.

Some thoughts regarding that; it's best implemented with pointer warping (or, pointer pinning), so that the mouse pointer doesn't move while the button is held down and the value is being dragged, this allows the user to drag it as far in any direction as they want without being worried about hitting the edge of the screen and stalling out, and it's also mildly beneficial for the mouse to remain where it is just in itself, as the direction the user wants to drag the value in doesn't really meaningfully correlate with where they're going to want the mouse pointer to be once they're done dragging, and often, due to grouping related UIs in the same place, they're going to want it to be pretty much where it started.

I just went looking, and it seems like that is actually supported in web browsers now, which is neat. https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API
It causes an ugly popup, but I'm sure users will be appreciative of it.

Nikita-str:
@Keavon 

* `Unit conversions in the input boxes`:
Maybe make unit input as another dropdown list element?  
`|  ← value  → | unit |`
And in case when it can have only one unit value then it would be unchangeable div? And if there no units, just value, then hide the unit element?
[...]

PR #457, opened 2022-01-03
Better decimal rounding in the NumberInput widget

mfish33:
Part of #313

This makes the number input use a sig fig-like rounding system in order to provide a more visually appealing experiance


frontend/src/components/widgets/inputs/NumberInput.vue
```diff
@@ -240,13 +242,21 @@ export default defineComponent({
...
 
-			const roundingPower = 10 ** this.displayDecimalPlaces;
-			const displayValue = Math.round(sanitized * roundingPower) / roundingPower;
+			this.setText(sanitized);
+		},
+		setText(value: number) {
+			// Find the amount of digits on the left side of the number
+			// 10.25 == 2
+			// 1.23 == 3
+			// 0.23 == 0 - Reason for the slightly more complicated code
```
Keavon, line 255: These examples really don't make any sense to me. `10.25 -> 2` leads me to believe "left side digits" means left of the decimal point. `1.23` leads me to believe it means all the digits. `0.23` leads me to believe it means left of the decimal point. What does "left side digits" and "digits on the left side of the number" mean?

mfish33: I messed it up. It's supposed to say `1.23 == 1`

Keavon: Comment should also say left side of the **decimal point**.

mfish33: I think once CI passes this should be good to go

Keavon (approved):

frontend/src/components/widgets/inputs/NumberInput.vue
Keavon: Looks good. I was specifically referring to the first line of the comment, where it confusingly says "left side of the number" rather than "left side of the decimal" but your additions to the subsequent lines can probably clear up any confusion that causes. Feel free to fix or merge now, I gave my approval.

PR #461, opened 2022-01-04
Refine navigate tool zoom

### LINKED ISSUE #448: Refine the zoom behavior of the Navigate Tool
Extends upon the work on #155.

It would be great to slightly refine this behavior and make it so zooming is relative to the position where you clicked down, and also a click without dragging should zoom in to the next increment (`IncreaseCanvasZoom`) while holding <kbd>Shift</kbd> and clicking does `DecreaseCanvasZoom`. Except unlike `IncreaseCanvasZoom`/`DecreaseCanvasZoom` it zooms relative to the place you click.

0HyperCube:
Closes #448


Keavon:
Regression: while snapping rotation (in the navigate tool and the global behavior), it now shows the fine-grained numbers (not the snapped numbers) in the zoom percent text field at top.

For the navigate tool's snap, we should use <kbd>Ctrl</kbd> not <kbd>Shift</kbd> since snapping should generally be Ctrl. This also made me realize that we can actually make snapping for the global rotation be more consistent: once you begin the rotation with <kbd>Ctrl</kbd><kbd>MMB</kbd> drag, let's switch to letting go of the <kbd>Ctrl</kbd> key and then pressing and holding it again (while <kbd>MMB</kbd> is still down) to snap. This way we never use <kbd>Shift</kbd> to snap for much better consistency.

0HyperCube:
> Regression: while snapping rotation (in the navigate tool and the global behavior), it now shows the fine-grained numbers (not the snapped numbers) in the zoom percent text field at top.

Good spot - I'd been polishing up some of the todos to use the `SetCanvasRotation` message and didn't notice that it passed the raw angle instead of `self.snapped_angle` to the frontend.

> For the navigate tool's snap, we should use <kbd>Ctrl</kbd> not <kbd>Shift</kbd>

@Keavon I thought this was odd, but you had shift in the issue. Changed now.

> This also made me realize that we can actually make snapping for the global rotation be more consistent: once you begin the rotation with <kbd>Ctrl</kbd><kbd>MMB</kbd> drag, let's switch to letting go of the <kbd>Ctrl</kbd> key and then pressing and holding it again (while <kbd>MMB</kbd> is still down) to snap. This way we never use <kbd>Shift</kbd> to snap for much better consistency.

Done

Keavon:
> I thought this was odd, but you had shift in the issue.

Indeed, sometimes it really just needs to be tried before deciding on s better UX option. Thank you for handling the changes :)

I'll review this when I wake up in 4 hours but I imagine it's about ready to merge. 🤠

Keavon:
For zoom snapping with the Navigate Tool:
- Please update the hint to use the <kbd>Ctrl</kbd> not <kbd>Shift</kbd> key and change the text to "Snap to Increment"
- The panning part of the zoom-relative-to-mouse-position should also be snapped, it shouldn't jarringly drift while the zoom portion is snapped

Please also implement <kbd>LMB</kbd> click to zoom, which upon mouseup (assuming no drag occurred) zooms in to the next increment. The <kbd>Shift</kbd> modifier zooms out instead of in. We'll have a magnifying glass cursor in the future to help make this more obvious (see #463).

Keavon:
Would you mind also adding another feature request? A global pan: <kbd>Spacebar</kbd><kbd>LMB drag</kbd> equivalent to <kbd>MMB Drag</kbd>.

editor/src/document/movement_handler.rs
```diff
@@ -17,16 +17,27 @@ use std::collections::VecDeque;
 #[impl_message(Message, DocumentMessage, Movement)]
 #[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub enum MovementMessage {
-	MouseMove { snap_angle: Key },
+	MouseMove {
+		snap_angle: Key,
+		wait_for_snap_angle_release: bool,
+		snap_zoom: Key,
+		zoom_from_viewport: Option<DVec2>,
```
Keavon, line 24: When reviewing the code, I've been rather stumped by this variable name. Can you think of any clearer name for it? Or if you want to describe for me what it does, I can perhaps suggest a name.

0HyperCube: It is an optional viewport position around which to zoom (defaults to the centre).

0HyperCube:
> Would you mind also adding another feature request? A global pan: SpacebarLMB-drag (equivalent to the existing MMB-drag panning.)

@Keavon This has already been implemented into master

Keavon:
> @Keavon This has already been implemented into master

Oh, indeed it is. I never knew about that, I must have missed when that was included.

editor/src/document/movement_handler.rs
otdavies: zoom_viewport_pivot?

Keavon (approved):

PR #462, opened 2022-01-04
Add alignment overlays

### LINKED ISSUE #450: Alignment visualization lines when snapping
This completes the main features needed for #200. 1px blue horizontal and/or vertical lines stretching across the full viewport should be drawn along any snapping point to visualize what's going on with snapping. Now that we have the new overlays system, this is all ready to be implemented.

0HyperCube:
Closes #450


otdavies:
Great start on this. I made a few small changes to where the start_snapping was called from. I also removed the ignored_layers. Your comment regarding the overlays eliminating the need for that was accurate I believe.

I am wondering, should we consider moving some of the start snapping / stop snapping operations into messages? This way they can be queued like most other operations in the editor. This would potentially resolve some of the issues with when start snapping is called / cleared.

Keavon:
With those bugs fixed, this is looking very good. The last thing before merging is that we need to show only the relevant lines so we aren't spamming the user with dozens or hundreds of lines. In the simple case: show only the horizontal and/or vertical line being snapped to. In the fancy case which can come now or later depending on your interests: we also show the nearby lines per-axis so if you're within, say, 100px of other snapping lines it will show those also. If you are 100px away they fade to 0% opacity and if you are 0px away (but it's not the active snapping target) it becomes 50% opacity, and the line that becomes the active snapping target is then drawn with 100% opacity. So nearby potential snapping lines linearly fade with distance to the left/center/right and top/middle/bottom.

0HyperCube:
@Keavon I think I have added the desired behaviour.

Keavon:
Looks great! I haven't had a chance to review the code yet but perhaps @TrueDoctor can help with that. The only problem I'm noticing with the UX isn't actually part of this feature, but part of the bounding box, so I'm happy feature-wise to merge this. The problem is that we don't seem to snap the bounding box to pixel increments on all sides which causes some antialiasing. And additionally, sometimes we might be snapping but the snapped edge doesn't line up with the alignment lines from this PR and it's instead on the neighboring pixel which causes two side-by-side blue lines (one from the snapping alignment and one from the bounding box). We'd like for those to be aligned where they fill the same pixel. Unless you want to try your hand at a quick fix, it's probably a better use of your time if we save it for the future and I just make a little note of it to be touched up someday.

Keavon:
Shiny! They line up much nicer now with the snapping alignment lines. It also looks like the bounding box AA is fixed on the tops and bottoms of shapes but there is still some AA on the left and right sides of the bounding box. Here's a screenshot of a shape, if you zoom in you will see that the top and bottom edges have crisp lines but the left and right have fuzzier lines where the AA is giving some color to the next column of over of pixels.



Not a big deal to spend time fighting this AA now if it isn't a quick fix. It's pretty close to perfect already.

The last thing that should be fixed before merging though (besides code review, which I'll try to do later today):

> currently we seem to be snapping to hidden layers

Keavon (approved):
All good to merge, looks and works great.

PR #464, opened 2022-01-06
Tidy up path handling in document_file

TrueDoctor:
* Improve solution for #455
* Reduce the number of allocations in document_file.rs

### DIFF
--- a/editor/src/communication/dispatcher.rs
+++ b/editor/src/communication/dispatcher.rs
@@ -339,17 +339,17 @@ mod test {
 		init_logger();
 		let mut editor = create_editor_with_three_layers();
 
-		fn to_vec<'a>(paths: Vec<&'a [LayerId]>) -> Vec<Vec<LayerId>> {
+		fn map_to_vec<'a>(paths: Vec<&'a [LayerId]>) -> Vec<Vec<LayerId>> {
 			paths.iter().map(|layer| layer.to_vec()).collect::<Vec<_>>()
 		}
-		let sorted_layers = to_vec(editor.dispatcher.documents_message_handler.active_document().all_layers_sorted());
+		let sorted_layers = map_to_vec(editor.dispatcher.documents_message_handler.active_document().all_layers_sorted());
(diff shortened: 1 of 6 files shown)


editor/src/communication/dispatcher.rs
```diff
@@ -339,10 +339,19 @@ mod test {
 		init_logger();
 		let mut editor = create_editor_with_three_layers();
 
-		let sorted_layers = editor.dispatcher.documents_message_handler.active_document().all_layers_sorted();
+		fn to_vec<'a>(paths: Vec<&'a [LayerId]>) -> Vec<Vec<LayerId>> {
```
0HyperCube, line 342: A bit confusing with the std `.to_vec` (copying outer vec). Perhaps `map_to_vec`?

editor/src/communication/dispatcher.rs
```diff
@@ -339,10 +339,19 @@ mod test {
...
+			paths.iter().map(|layer| layer.to_vec()).collect::<Vec<_>>()
+		}
+		let sorted_layers = to_vec(editor.dispatcher.documents_message_handler.active_document().all_layers_sorted());
 		println!("Sorted layers: {:?}", sorted_layers);
```
0HyperCube, line 346: Probably should be debug / trace.

TrueDoctor: Thought the same thing, but afaik this is only used for debugging test cases which use std out

TrueDoctor: We could put it into a #[cfg(test)] though

editor/src/communication/dispatcher.rs
TrueDoctor: yeah wasn't happy about the name either

otdavies (approved):
Heading read over this PR with @TrueDoctor, tested it locally and gone over it multiple times, it seems like an overall positive improvement. It has my go-ahead!

PR #465, opened 2022-01-06
Ungrouping layers

### LINKED ISSUE #446: Ungroup with Ctrl+Shift+G
If a group is selected, <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>G</kbd> should eliminate the group and put its former children in the location where the group sat in the tree. If multiple groups are selected, it should do the same for each group. However if any of the multiple selected groups are descendants of a parent one, only ungroup the eldest parent. (If this behavior is too had to implement, alternate behaviors may be acceptable so don't put too much time into making this edge case work precisely as described.)

otdavies:
Closes #446

### DIFF
--- a/editor/src/communication/dispatcher.rs
+++ b/editor/src/communication/dispatcher.rs
@@ -339,7 +339,7 @@ mod test {
 		init_logger();
 		let mut editor = create_editor_with_three_layers();
 
-		fn map_to_vec<'a>(paths: Vec<&'a [LayerId]>) -> Vec<Vec<LayerId>> {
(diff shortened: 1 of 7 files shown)


otdavies:
@TrueDoctor could you take a look at the undo issue here? For some reason when Undoing the ungroup it selects all the folders incorrectly. You'll see what I mean. I tried push_front the StartTransaction and all that _big brain ordering_, but doesn't seem to change the behavior. This makes me a little concerned about this way of manipulating the messaging system, seems to cause instability with existing message ordering.

TrueDoctor (approved):

PR #466, opened 2022-01-06
Feature: Artboards

### LINKED ISSUE #289: Artboards
We only need to implement one artboard at first, so the user can pick an infinite canvas (like it is now) or a bounded canvas (1 artboard with the specified dimensions). The Crop Tool adjusts the corners of the artboard. The white background only exists within the artboard, and a dark (`#111111`) background is used behind everything else.

mfish33:
- graphite document artboard implementation
- autosave document load hitch fix (Inconsistency on root transform when the document loaded)
- Autosave will delete saved files when graphite document version changes

The scrollbars have not been changed when an artboard has been added since I need a little guidance on what exactly @Keavon you are looking for.

Closes #289


Keavon:
Before I forget about it: we'll also want the rulers to start at (0, 0) in the top left corner instead of the center.

Keavon:
Here some elaboration on the goal for scrollbars (which should hopefully, in theory, provide a reasonably good UX).

Currently the scrollbars are based on an AABB of the visible layers. (Then that AABB goes into the algorithm that determines the hybrid linear/logarithmic scaling of the scrollbars.) When a document has an artboard instead of being an infinite canvas, I'm suggesting you alter the source of the AABB to be the AABB of the artboard(s).

For clarity, this means that when not tilted, the AABB would be the same as the artboard bounds, but when tilted the AABB would be larger to encompass the tilt. And when multiple artboards are involved it would be the AABB of all of them.

Feed this AABB of the tilted artboard(s) into the scrollbar system instead of the AABB of the visible layers, which it uses currently.

This means that any visible layers outside the bounds of the artboards will be intentionally excluded from the scrollbar calculation, unlike in infinite canvas mode.

Keavon:
<kbd>Ctrl</kbd><kbd>0</kbd> works based on the AABB of the visible layers in the document for an infinite canvas. But with artboards, will want the <kbd>Ctrl</kbd><kbd>0</kbd> and center-and-scale-down-if-unable-to-fit-within-the-viewport messages to use the same AABB described in the scrollbar AABB source. That is:
* From the visible layers in infinite canvas mode
* From the (tilted) artboards in artboard mode

Keavon:
It seems I described my request about the (0, 0) origin too briefly and caused you to misunderstand my intent (rereading what I wrote, your interpretation was the obvious one). I'm very sorry to waste your time implementing a change we unfortunately need to revert, since the affected behavior was as intended. Here's a more descriptive change, and please let me know if you need any clarification since I want to make sure I avoid wasting your time doing the wrong task again.

When a bounded document is created (by clicking *File* > *New 1920x1080*), that document should be created such that its origin (0, 0) point is its top left corner. That means the rulers count from 0 to the right, and from 0 to the bottom. In other words, within the infinite expanse of the document's X and Y axes, we just need to shift the default placement of the artboard by half its width and height. That artboard placement could be any arbitrary position in space (since the user can use the Crop Tool to reposition and create new artboards in random places within the coordinate system), but the useful default for creating one artboard is the position where the top left corner is the origin of the infinite document coordinates. This should not affect infinite canvas mode at all, which should still begin with the origin point loaded in the center of the viewport. Once the artboard is initialized with its 1920x1080 bounds positioned at (0, 0), that means the default behavior of loading a document with its origin centered within the viewport will result in the new document centering on the top left corner of the artboard, with blank space in the upper and left quadrants and the document hanging off the viewport in the bottom right quadrant. That is why it's desirable to fire the message to recenter and scale the document, equivalent to hitting <kbd>Ctrl</kbd>0</kbd> to center and scale the document to fit the viewport, followed by <kbd>Ctrl</kbd><kbd>1</kbd> to zoom to 100% scale ONLY IF this would cause a zoom **out** to 100% to show the entire document in the viewport but not **in** to 100% to hide the edges of the document beyond the bounds of the viewport.

Keavon:
Turns out half of this was a very, very easy fix. From `editor.instance.create_artboard(-1920 / 2, -1080 / 2, 1920, 1080);` to `editor.instance.create_artboard(0, 0, 1920, 1080);`. This puts the top left corner in the center of the viewport as expected, so the next step is adding the recenter and scale message we need. I'll file a separate issue for that since I'm merging this into master now.

PR #469, opened 2022-01-08
Migrate to using MoveSelectedLayersTo

### LINKED ISSUE #444: Rearranging multiple selected layers loses selection

### LINKED ISSUE #447: In the layer panel, drag layer(s) into a folder
This extends upon #141 where we can now drag to rearrange layers in the layer panel and insert between other layers. But if a folder is empty, it isn't possible to insert between any children of that empty folder, so it's impossible to put it in that layer (without resorting to cut/paste). This should be altered to allow inserting the dragged layers into a folder itself, in addition to inserting between layers. That means the current behavior of putting the blue insertion marker line between the layers closest to the mouse's vertical location will need to be changed to check if the mouse is over a folder, and only show the blue line when close (say, 10px) from the top/bottom but when the mouse is in the center area of the folder's vertical area, outline the folder in blue.

0HyperCube:
Migrate to using `MoveSelectedLayersTo`. 

@TrueDoctor This currently does not work as intended. With a flat tree of 3 layers, if I drag the top layer between the bottom two layers (index 2 at on the left of the image) then the layer is deleted (to be re-added in another location) causing index 2 to become the bottom of the tree.

This is only a problem when dragging the layers down the tree but I'm not really sure how best to solve it.
image

Part of #447
Closes #444


TrueDoctor:
Interesting, I can look into that

TrueDoctor:
So the issue is with temporal consistency assumptions in subsequent operations, we can try to think about a time consistent description for where to insert the layer, or we could bandaid the existing solution by doing the calculation of where the layers will end up during the creation of the operation

TrueDoctor:
Also one thing @Keavon and I noticed when looking at this noted, was that the selection should nod be effected by the layer movement, this complicates things a little bit

frontend/wasm/src/api.rs
```diff
@@ -383,8 +383,9 @@ impl JsEditorHandle {
 	}
 
 	/// Move a layer to be next to the specified neighbor
-	pub fn move_layer_in_tree(&self, layer: Vec<LayerId>, insert_above: bool, neighbor: Vec<LayerId>) {
-		let message = DocumentMessage::MoveLayerInTree { layer, insert_above, neighbor };
+	pub fn move_layer_in_tree(&self, path: Vec<LayerId>, insert_index: isize) {
+		log::info!("Move layers in tree, Path {:?}, insert index {}", &path, insert_index);
```
TrueDoctor, line 387: Log can be removed

editor/src/document/document_file.rs
```diff
@@ -842,10 +847,12 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 							}
 							DocumentResponse::LayerChanged { path } => responses.push_back(LayerChanged(path.clone()).into()),
 							DocumentResponse::CreatedLayer { path } => {
-								self.layer_metadata.insert(path.clone(), LayerMetadata::new(false));
+								if !self.layer_metadata.contains_key(path) {
```
TrueDoctor, line 850: I think this should be an error right?
could we maybe add a log::warn! in the else case? I don't think this should ever not be true

TrueDoctor (approved):
This looks pretty good to me, you can resolve/dismiss the comments and then merge if you want

editor/src/document/document_file.rs
0HyperCube: This should never be true, I just misread the PasteIntoFolder as populating the layermetadata before the InsertData (when actually it is push_front)

PR #470, opened 2022-01-08
Boolean shape operations

caleb-ad:
# Boolean Shape Operations


frontend/src/components/widgets/options/ToolOptions.vue
```diff
@@ -134,11 +134,11 @@ export default defineComponent({
 
 				{ kind: "Separator", props: { type: "Section" } },
 
-				{ kind: "IconButton", tooltip: "Boolean Union", callback: (): void => this.dialog.comingSoon(197), props: { icon: "BooleanUnion", size: 24 } },
-				{ kind: "IconButton", tooltip: "Boolean Subtract Front", callback: (): void => this.dialog.comingSoon(197), props: { icon: "BooleanSubtractFront", size: 24 } },
-				{ kind: "IconButton", tooltip: "Boolean Subtract Back", callback: (): void => this.dialog.comingSoon(197), props: { icon: "BooleanSubtractBack", size: 24 } },
-				{ kind: "IconButton", tooltip: "Boolean Intersect", callback: (): void => this.dialog.comingSoon(197), props: { icon: "BooleanIntersect", size: 24 } },
-				{ kind: "IconButton", tooltip: "Boolean Difference", callback: (): void => this.dialog.comingSoon(197), props: { icon: "BooleanDifference", size: 24 } },
+				{ kind: "IconButton", tooltip: "Boolean Union", callback: () => this.editor.instance.boolean_operation("Union"), props: { icon: "BooleanUnion", size: 24 } },
```
0HyperCube, line 137: Add return type i.e. `callback: (): void => ...`

graphene/src/intersection.rs
```diff
@@ -69,3 +71,284 @@ pub fn get_arbitrary_point_on_path(path: &BezPath) -> Option<Point> {
...
+/// \/                               \/
+/// Bezier Curve Intersection algorithm
+/// \/                               \/
+
+/// each intersection has two curves, which are distinguished between using this enum
+/// TODO: refactor so actual curve data and Origin aren't seperate
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Origin {
+	Alpha,
+	Beta,
```
0HyperCube, line 84: Often called Subject and Intersect polygon.

graphene/src/intersection.rs
```diff
@@ -69,3 +71,284 @@ pub fn get_arbitrary_point_on_path(path: &BezPath) -> Option<Point> {
...
 }
+
+/// \/                               \/
+/// Bezier Curve Intersection algorithm
+/// \/                               \/
+
+/// each intersection has two curves, which are distinguished between using this enum
+/// TODO: refactor so actual curve data and Origin aren't seperate
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum Origin {
```
0HyperCube, line 82: Perhaps PolygonType would be more fitting.

graphene/src/intersection.rs
```diff
@@ -69,3 +71,284 @@ pub fn get_arbitrary_point_on_path(path: &BezPath) -> Option<Point> {
...
+		}
+	}
+}
+
+pub struct Intersect {
+	pub point: Point,
+	pub t_a: f64,
+	pub t_b: f64,
+	pub a_seg_idx: usize,
+	pub b_seg_idx: usize,
```
0HyperCube, line 102: Expand abbreviations. `time_alpha` and `alpha_segement_id`.

graphene/src/intersection.rs
```diff
@@ -69,3 +71,284 @@ pub fn get_arbitrary_point_on_path(path: &BezPath) -> Option<Point> {
...
+		}
+		let (a1, a2) = a.split(0.5);
+		let (b1, b2) = b.split(0.5);
+		intersections.append(&mut path_intersections(&a1, &b1, recursion));
+		intersections.append(&mut path_intersections(&a1, &b2, recursion));
+		intersections.append(&mut path_intersections(&a2, &b1, recursion));
+		intersections.append(&mut path_intersections(&a2, &b2, recursion));
+	}
+	intersections
+}
```
0HyperCube, line 265: https://pomax.github.io/bezierinfo/#curveintersection - you should not be getting duplicates. Code from me trying to understand bezier curves:
```rust
let mut curves = vec![utils::Pair::new(self, other)];

const ITERATIONS: usize = 20;
for _ in 0..ITERATIONS {
	curves = curves
		.into_iter()
		.flat_map(|p| p.permutations(p.subject.split(0.5), p.intersect.split(0.5)))
		.filter(
			|utils::Pair {
					subject_time: _,
					subject,
					intersect,
				}| {
				utils::bounding_box_intersect(
					subject.bounding_box(),
					intersect.bounding_box(),
				)
			},
		)
		.collect();
}

utils::filter(curves)
```
and in utils
```rust
pub fn bounding_box_intersect([a1, a2]: [DVec2; 2], [b1, b2]: [DVec2; 2]) -> bool {
	let [a_min, a_max] = [a1.min(a2), a1.max(a2)];
	let [b_min, b_max] = [b1.min(b2), b1.max(b2)];
	a_min.x <= b_max.x && a_max.x >= b_min.x && a_min.y <= b_max.y && a_max.y >= b_min.y
}
pub fn filter<Sub: Curve, Int: Curve>(mut curves: Vec<Pair<Sub, Int>>) -> Vec<DVec2> {
	let mut subject_times = HashSet::new();
	let mut filtered_curves = Vec::new();

	while let Some(c) = curves.pop() {
		if !(subject_times.contains(&c.subject_time)
			|| subject_times.contains(&(c.subject_time + 1))
			|| subject_times.contains(&(c.subject_time - 1)))
		{
			subject_times.insert(c.subject_time);
			filtered_curves.push(c);
		}
	}
	filtered_curves
		.into_iter()
		.map(|x| x.subject.start())
		.collect()
}

#[derive(Debug)]
pub struct Pair<Sub: Curve, Int: Curve> {
	pub subject_time: u64,
	pub subject: Sub,
	pub intersect: Int,
}
impl<Sub: Curve, Int: Curve> Pair<Sub, Int> {
	pub fn new(subject: Sub, intersect: Int) -> Self {
		Self {
			subject_time: 0,
			subject,
			intersect,
		}
	}
	pub fn permutations(&self, [s1, s2]: [Sub; 2], [i1, i2]: [Int; 2]) -> [Self; 4] {
		[
			Pair {
				subject_time: self.subject_time * 2,
				subject: s1,
				intersect: i1,
			},
			Pair {
				subject_time: self.subject_time * 2,
				subject: s1,
				intersect: i2,
			},
			Pair {
				subject_time: self.subject_time * 2 + 1,
				subject: s2,
				intersect: i1,
			},
			Pair {
				subject_time: self.subject_time * 2 + 1,
				subject: s2,
				intersect: i2,
			},
		]
	}
}

```

graphene/src/intersection.rs
```diff
@@ -69,3 +71,284 @@ pub fn get_arbitrary_point_on_path(path: &BezPath) -> Option<Point> {
...
+pub fn overlap(a: &Rect, b: &Rect) -> bool {
+	fn in_range(n: f64, range: &[f64]) -> bool {
+		n >= range[0] && n <= range[1]
+	}
+	fn in_range_e(n: f64, range: &[f64]) -> bool {
+		n > range[0] && n < range[1]
+	}
+	(in_range(b.x0, &[a.x0, a.x1]) || in_range(b.x1, &[a.x0, a.x1]) || in_range_e(a.x0, &[b.x0, b.x1]) || in_range_e(a.x1, &[b.x0, b.x1]))
+		&& (in_range(b.y0, &[a.y0, a.y1]) || in_range(b.y1, &[a.y0, a.y1]) || in_range_e(a.y0, &[b.y0, b.y1]) || in_range_e(a.y1, &[b.y0, b.y1]))
+}
```
0HyperCube, line 348: Simplify (also from me trying to understand curves):
```rust
pub fn bounding_box_intersect([a1, a2]: [DVec2; 2], [b1, b2]: [DVec2; 2]) -> bool {
	let [a_min, a_max] = [a1.min(a2), a1.max(a2)];
	let [b_min, b_max] = [b1.min(b2), b1.max(b2)];
	a_min.x <= b_max.x && a_max.x >= b_min.x && a_min.y <= b_max.y && a_max.y >= b_min.y
}
```

graphene/src/document.rs
```diff
@@ -92,6 +95,52 @@ impl Document {
...
+		// skip the root layer, whose transform doesn't effect the coordinates
+		if let LayerDataType::Folder(root) = &self.root.data {
+			match root.layer(path[0]) {
+				Some(layer) => helper(layer, DAffine2::IDENTITY, 1, path),
+				None => Err(DocumentError::InvalidPath),
+			}
+		} else {
+			Err(DocumentError::NotAFolder)
+		}
+	}
```
0HyperCube, line 129: Duplicated behaviour from graphene_document.multiply_transforms

graphene/src/intersection.rs
```diff
@@ -1,7 +1,9 @@
 use std::ops::Mul;
 
-use glam::{DAffine2, DVec2};
-use kurbo::{BezPath, Line, PathSeg, Point, Shape};
+use glam::{DAffine2, DMat2, DVec2};
+use kurbo::{BezPath, Line, ParamCurve, ParamCurveExtrema, PathSeg, Point, Rect, Shape};
+
+pub const F64PRECISION: f64 = f64::EPSILON * 1000.0; // for f64 comparisons
```
0HyperCube, line 6: Perhaps should be defined in the consts file (I know this is not possible as consts is in the editor - but perhaps a file for document consts should be added), Also probably describe more what the cont does e.g. `CURVE_INTERSECTION_PRECISION`

0HyperCube, line 6: Perhaps should be defined in the consts file (I know this is not possible as consts is in the editor - but perhaps a file for document consts should be added), Also probably describe more what the cont does e.g. `CURVE_INTERSECTION_PRECISION`

graphene/src/document.rs
```diff
@@ -92,6 +95,52 @@ impl Document {
...
+
+	/// Return vector of immutable references to each shape specified in paths
```
0HyperCube, line 131: Not immutable - cloned.

graphene/src/document.rs
```diff
@@ -92,6 +95,52 @@ impl Document {
...
+	/// If any path is not a shape, or does not exist, DocumentError::InvalidPath is returned
+	fn shapes_as_seen(&self, paths: &Vec<Vec<LayerId>>) -> Result<Vec<Shape>, DocumentError> {
```
0HyperCube, line 133: Perhaps `projected_shapes`

0HyperCube (commented):
@caleb-ad Looks great so far 👍🏽 Handling edge cases e.g. with a curve just touching a line and not crossing needs thought and this could use extensive tests.

graphene/src/intersection.rs
Keavon: (or `index` not `id` in this case, from `idx`)

For context, we encourage full names instead of abbreviations because code readability is very important, and the mental burden of expanding abbreviations is more taxing than reading a few extra characters of code. Even if you end up with pretty long variable or function names, descriptivity makes that worthwhile. Soon I'll find the time to write a style/best practices guide so I don't have to mention this to contributors individually! :)

graphene/src/intersection.rs
Keavon: We recently added a `consts.rs` file to Graphene also.

Keavon:
This is very fun and exciting to play around with!

graphene/src/document.rs
caleb-ad: Good catch! at one point, long long ago (in a galaxy far far away), it was a vector of immutable reference :)

graphene/src/document.rs
caleb-ad: `shapes_as_seen` definitely isn't a great name. I'm thinking maybe `transformed_shapes`, because typically a projection is onto something

graphene/src/intersection.rs
caleb-ad: What do you mean I shouldn't be getting duplicates?

graphene/src/intersection.rs
caleb-ad: Sounds good! I think I'm gonna hold off on making this change for a little while however, I'm considering doing some refactoring concerning the Intersect structure, so when I do that i'll make the changes.

frontend/wasm/src/type_translators.rs
```diff
@@ -127,6 +128,17 @@ pub fn translate_key(name: &str) -> Key {
 	}
 }
 
+pub fn translate_boolean_operation(operation: &str) -> Option<BooleanOperation> {
+	match operation {
+		"Union" => Some(BooleanOperation::Union),
+		"Difference" => Some(BooleanOperation::Difference),
+		"Intersection" => Some(BooleanOperation::Intersection),
+		"SubFront" => Some(BooleanOperation::SubFront),
+		"SubBack" => Some(BooleanOperation::SubBack),
```
Keavon, line 137: Please use the full word "Subtract": `SubtractFront` and `SubtractBack`.

graphene/src/intersection.rs
0HyperCube: > Once the sub-curves we form are so small that they effectively occupy sub-pixel areas, we consider an intersection found, noting that we might have a cluster of multiple intersections at the sub-pixel level, out of which we pick one to act as "found" t value (we can either throw all but one away, we can average the cluster's t values, or you can do something even more creative).

graphene/src/document.rs
0HyperCube: It is projecting the shapes into viewport space, however `transformed_shapes` is also a good name.

graphene/src/intersection.rs
caleb-ad: Right now the algorithm occasionally finds duplicates I think when an intersection is near a point where the curve is split, because the recursive "paths" the algorithm takes can't communicate with each other. I think to solve this there either needs to be reconciliation at the end (where we do what you commented and average, or discard, intersections "close" together), or maybe restructure the algorithm to be iterative. 
While it's possible that these intersections may not actually be duplicates, I think the best policy is to treat them as if they are and average them together. Another idea, instead of something like averaging, is an intersection quality metric, which we could use to pick the (hopefully) best intersection from a set of possibilities.

Keavon:
Do you need help resolving the merge conflicts? I'd like to wait until it's up to date with master before beginning a review.

graphene/src/document.rs
```diff
@@ -519,6 +540,34 @@ impl Document {
...
+				// TODO: proper style selection (done?)
+				// TODO: should generate symmetrical code
+				// TODO: Operations on any number of shapes
+				// TODO: boolean ops on any number of shapes
+				// TODO: handle overlapping identical curve case
+				// TODO: precision reached without intersection bug (maybe caused by separating a closed path, or dragging handles)
+				// TODO: click on shape should drag the shape
+				// TODO: add ability to undo
+				let mut responses = Vec::new();
+				if selected.len() > 1 && selected.len() < 3 {
```
TrueDoctor, line 554: so `selected.len() == 2`?

graphene/src/boolean_ops.rs
```diff
@@ -0,0 +1,744 @@
...
+}
+
+/// A simple and idiomatic way to write short "if let Some(_)" statements which do nothing in the None case
+macro_rules! do_if {
+	($option:expr, $name:ident{$todo:expr}) => {
+		if let Some($name) = $option {
+			$todo
+		}
+	};
+}
```
TrueDoctor, line 36: Why not use `bar.map(|x| foo(x))`?

PR #472, opened 2022-01-08
Fix rotation input

0HyperCube:
The rotation input field was clamped between 0 and 1 after `Better decimal rounding in the NumberInput widget` #457


mfish33 (commented):
Thank you for catching this

mfish33 (approved):

PR #473, opened 2022-01-09
Fix bounds with artboards for zoom-to-fit and scrollbar scaling

mfish33:
- document load keeps position
- zoom to fit
- scrollbars use artboard dimensions


editor/src/document/document_file.rs
```diff
@@ -82,7 +82,6 @@ pub struct DocumentMessageHandler {
 	#[serde(with = "vectorize_layer_metadata")]
 	pub layer_metadata: HashMap<Vec<LayerId>, LayerMetadata>,
 	layer_range_selection_reference: Vec<LayerId>,
-	#[serde(skip)]
```
Keavon, line 85: Was this intentional?

editor/src/document/document_file.rs
```diff
@@ -1063,6 +1068,18 @@ impl MessageHandler<DocumentMessage, &InputPreprocessor> for DocumentMessageHand
 			SetSnapping(new_status) => {
 				self.snapping_enabled = new_status;
 			}
+			ZoomCanvasToFitAll => {
+				if let Some(bounds) = self.document_bounds() {
+					responses.push_back(
+						MovementMessage::FitViewportToBounds {
+							bounds,
+							padding_scale_factor: Some(1.05),
```
Keavon, line 1076: Let's move this into `consts.rs` as `VIEWPORT_ZOOM_TO_FIT_PADDING_SCALE_FACTOR`.

editor/src/input/input_mapper.rs
```diff
@@ -134,6 +134,7 @@ impl Default for Mapping {
 		// it as an available action in the respective message handler file (such as the bottom of `document_message_handler.rs`)
 		let mappings = mapping![
 			// Higher priority than entries in sections below
+			entry! {action=DocumentMessage::ZoomCanvasToFitAll, key_down=Key0, modifiers=[KeyControl]},
```
Keavon, line 137: This doesn't need to be in its top priority position.

editor/src/document/movement_handler.rs
```diff
@@ -321,7 +334,6 @@ impl MessageHandler<MovementMessage, (&Document, &InputPreprocessor)> for Moveme
 			IncreaseCanvasZoom,
 			DecreaseCanvasZoom,
 			WheelCanvasTranslate,
-			ZoomCanvasToFitAll,
```
Keavon, line 324: Should this be removed, or just renamed to its new name of `FitViewportToBounds`?

Keavon (approved):
Approved, pending fixes to code review comments. Merge when ready.

Keavon:
One more thing to be changed: when exporting an SVG, we need to set the viewBox based upon the same bounding box logic.

editor/src/document/document_file.rs
mfish33: Yes, this allows the document position to be restored from saved files as well as autosave

editor/src/document/movement_handler.rs
mfish33: Yes since the document now gets the keyboard shortcut and sends a message to the movement handler

PR #480, opened 2022-01-15
Set the mouse cursor in the canvas based on the current tool and its state

### LINKED ISSUE #463: Set mouse cursor in the viewport based on tool state
Add a FrontendMessage to set the cursor to an enum value with variants like `Default`, `ZoomIn`, `ZoomOut`, `Grabbing`, and `Crosshair`. Then apply the [CSS `cursor` rule](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) and have the document viewport `<div class="canvas" ref="canvas">` apply the cursor style property like `<div class="canvas" ref="canvas" :style="{ cursor: canvasCursor }">` where the default cursor is the JS `undefined` value (so Vue applies no style value).

The Navigate tool should display `ZoomIn` by default since <kbd>LMB</kbd> clicking causes a zoom in operation, and <kbd>LMB</kbd> dragging zooms in and out. While holding the <kbd>Shift</kbd> modifier, switch to the `ZoomOut` cursor since <kbd>LMB</kbd> clicking with that held down zooms out. When <kbd>MMB</kbd> dragging, switch to the `Grabbing` cursor (both in the Navigate tool and the global behavior). For rotating the document, that can stay the default cursor until we design a custom SVG cursor for that (and probably all icons but the default cursor icon).

The shape drawing tools (Line, Rectangle, Ellipse, and Shape) should all get the `Crosshair` cursor.

The Select, Path, Pen, Eyedropper, Fill, and other future tools will eventually get their own SVG icons also. @Keavon needs to design those in a format that's optimized for a mouse cursor.

Ideally, the cursor should stay the same when the mouse leaves the viewport `.canvas` element while a drag is occurring. [See this](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) for how.

asyncth:
Closes #463

### DIFF
--- a/editor/src/frontend/utility_types.rs
+++ b/editor/src/frontend/utility_types.rs
@@ -8,7 +8,7 @@ pub struct FrontendDocumentDetails {
 }
 
 #[derive(Clone, Copy, Debug, Eq, Deserialize, PartialEq, Serialize)]
-pub enum FrontendMouseCursor {
+pub enum MouseCursorIcon {
 	Default,
 	ZoomIn,
 	ZoomOut,
--- a/editor/src/frontend/utility_types.rs
+++ b/editor/src/frontend/utility_types.rs
@@ -6,3 +6,12 @@ pub struct FrontendDocumentDetails {
 	pub name: String,
 	pub id: u64,
 }
+
+#[derive(Clone, Copy, Debug, Eq, Deserialize, PartialEq, Serialize)]
+pub enum FrontendMouseCursor {
+	Default,
+	ZoomIn,
+	ZoomOut,
+	Grabbing,
+	Crosshair,
+}
--- a/frontend/src/components/panels/Document.vue
+++ b/frontend/src/components/panels/Document.vue
@@ -123,7 +123,7 @@
 						<CanvasRuler :origin="rulerOrigin.y" :majorMarkSpacing="rulerSpacing" :numberInterval="rulerInterval" :direction="'Vertical'" />
 					</LayoutCol>
 					<LayoutCol :class="'canvas-area'">
-						<div class="canvas" ref="canvas">
+						<div class="canvas" ref="canvas" :style="{ cursor: canvasCursor }">
 							<svg class="artboards" v-html="artboardSvg" :style="{ width: canvasSvgWidth, height: canvasSvgHeight }"></svg>
 							<svg class="artwork" v-html="artworkSvg" :style="{ width: canvasSvgWidth, height: canvasSvgHeight }"></svg>
 							<svg class="overlays" v-html="overlaysSvg" :style="{ width: canvasSvgWidth, height: canvasSvgHeight }"></svg>
@@ -261,6 +261,7 @@ import {
 	UpdateCanvasRotation,
 	ToolName,
 	UpdateDocumentArtboards,
+	UpdateMouseCursor,
 } from "@/dispatcher/js-messages";
 
 import LayoutCol from "@/components/layout/LayoutCol.vue";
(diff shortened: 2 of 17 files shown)


asyncth:
Forgot to add the `setCapture` thing, will do tomorrow.

editor/src/frontend/utility_types.rs
```diff
@@ -6,3 +6,12 @@ pub struct FrontendDocumentDetails {
 	pub name: String,
 	pub id: u64,
 }
+
+#[derive(Clone, Copy, Debug, Eq, Deserialize, PartialEq, Serialize)]
+pub enum FrontendMouseCursor {
```
Keavon, line 11: "Frontend" is perhaps redundant here. Maybe `MouseCursorIcon` would be more descriptive.

frontend/src/components/panels/Document.vue
```diff
@@ -123,7 +123,7 @@
 						<CanvasRuler :origin="rulerOrigin.y" :majorMarkSpacing="rulerSpacing" :numberInterval="rulerInterval" :direction="'Vertical'" />
 					</LayoutCol>
 					<LayoutCol :class="'canvas-area'">
-						<div class="canvas" ref="canvas">
+						<div class="canvas" ref="canvas" :style="{ cursor: canvasCursor }" v-on:pointerdown="canvasPointerDown">
```
Keavon, line 126: `@pointerdown="(e: PointerEvent) => canvasPointerDown(e)"`

frontend/src/components/panels/Document.vue
```diff
@@ -338,6 +339,10 @@ export default defineComponent({
 		resetWorkingColors() {
 			this.editor.instance.reset_colors();
 		},
+		canvasPointerDown(event: PointerEvent) {
```
Keavon, line 342: `e`, not `event`, is convention.

frontend/src/components/panels/Document.vue
asyncth: What's wrong with `@pointerdown="canvasPointerDown"`?

frontend/src/dispatcher/js-messages.ts
```diff
@@ -39,6 +39,8 @@ export class FrontendDocumentDetails extends DocumentDetails {
 	readonly id!: BigInt;
 }
 
+export type MouseCursorIcon = string;
```
Keavon, line 42: Change the type from `string` to `"default" | "zoom-in" | "zoom-out" | "grabbing" | "crosshair"`

frontend/src/dispatcher/js-messages.ts
```diff
@@ -201,6 +203,33 @@ export class UpdateDocumentRulers extends JsMessage {
...
+
+		case "Grabbing":
+			return "grabbing";
+
+		case "Crosshair":
+			return "crosshair";
+
+		default:
+			return "default";
+	}
```
Keavon, line 225: A dictionary lookup would be cleaner here:
```ts
const cssNames = {
    ZoomIn: "zoom-in",
    ZoomOut: "zoom-out",
    Grabbing: "grabbing",
    Crosshair: "crosshair",
};
return cssNames[value] || "default";
```

frontend/src/dispatcher/js-messages.ts
Keavon: And please move this down next to where it's used. Right above its first usage (in the `Transform(...)` would be best.

frontend/src/components/panels/Document.vue
Keavon: We're being explicit with our type usage by convention.

Keavon (approved):

PR #481, opened 2022-01-15
Do not transform layers when transforming a parent folder

### LINKED ISSUE #482: Children of folders, when both are selected, move at double the transformation

0HyperCube:
Closes #482

transforms-applied

### DIFF
--- a/editor/src/communication/dispatcher.rs
+++ b/editor/src/communication/dispatcher.rs
@@ -161,10 +161,10 @@ mod test {
 		let mut editor = create_editor_with_three_layers();
 
 		let document_before_copy = editor.dispatcher.message_handlers.portfolio_message_handler.active_document().graphene_document.clone();
-		editor.handle_message(PortfolioMessage::Copy(Clipboard::User));
+		editor.handle_message(PortfolioMessage::Copy { clipboard: Clipboard::User });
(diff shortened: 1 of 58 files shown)


Keavon:
@TrueDoctor @otdavies thoughts about this being `O(n^2)`? Any less expensive way to do this? It's called every mousemove which is rather a frequent.

otdavies:
> @TrueDoctor @otdavies thoughts about this being `O(n^2)`? Any less expensive way to do this? It's called every mousemove which is rather a frequent.

Yes, there are functions in the graphene document that will likely make this faster / easier. Take a look at shallowest common folder and it's neighboring utility functions. I will provide an explicit example when I am able to pull the PR and take a look at it.

otdavies:
> @TrueDoctor @otdavies thoughts about this being `O(n^2)`? Any less expensive way to do this? It's called every mousemove which is rather a frequent.

I believe in this scenario the function you want is called: shallowest_unique_layers, in document.rs. It only returns unique layers in the selection, aka it excludes children of folders but doesn't exclude layers that are neighbors of the folder. This is the exact situation that @0HyperCube  is solving here I think.

Keavon (approved):

PR #484, opened 2022-01-16
Make the toggle overlays button functional

### LINKED ISSUE #458: Hook up the toggle overlays button
When you click this button, overlays should be entirely hidden or shown.
image

Now that overlays are part of a separate canvas, we can hide or show it by rendering the overlays SVG or an empty SVG. In the future, the popover button next to it will provide options for selectively hiding or showing specific overlays (so this button is the master switch to hide/show the enabled set of overlays).

Keavon:
Closes #458

### DIFF
--- a/editor/src/document/document_message.rs
+++ b/editor/src/document/document_message.rs
@@ -91,7 +91,7 @@ pub enum DocumentMessage {
 	SetOpacityForSelectedLayers {
 		opacity: f64,
 	},
-	SetOverlaysVisible {
+	SetOverlaysVisibility {
 		visible: bool,
 	},
 	SetSelectedLayers {
--- a/editor/src/document/document_message.rs
+++ b/editor/src/document/document_message.rs
@@ -91,6 +91,9 @@ pub enum DocumentMessage {
 	SetOpacityForSelectedLayers {
 		opacity: f64,
 	},
+	SetOverlaysVisible {
+		visible: bool,
+	},
 	SetSelectedLayers {
 		replacement_selected_layers: Vec<Vec<LayerId>>,
 	},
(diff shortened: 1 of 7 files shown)


editor/src/document/overlays_message.rs
```diff
@@ -10,6 +10,7 @@ use serde::{Deserialize, Serialize};
 pub enum OverlaysMessage {
 	ClearAllOverlays,
 	DispatchOperation(Box<DocumentOperation>),
+	Redraw,
```
0HyperCube, line 13: We are not redrawing as that would involve the overlay document being cleared and all of the tools resubmitting their draw calls. Perhaps calling it `Rerender` or similar would be more informative?

0HyperCube (approved):
LGTM

editor/src/document/document_message.rs
```diff
@@ -91,6 +91,9 @@ pub enum DocumentMessage {
 	SetOpacityForSelectedLayers {
 		opacity: f64,
 	},
+	SetOverlaysVisible {
```
TrueDoctor, line 94: SetOverlayVisibility?

editor/src/document/document_message_handler.rs
```diff
@@ -41,6 +41,7 @@ pub struct DocumentMessageHandler {
 	artboard_message_handler: ArtboardMessageHandler,
 	#[serde(skip)]
 	transform_layer_handler: TransformLayerMessageHandler,
+	pub overlays_visible: bool,
```
TrueDoctor, line 44: Why not store this in the overlay manager?

Keavon: I started with that approach but changed it to this because we'd ideally like to not serialize it (i.e. `#[serde(skip)]`). It's a transient thing, not a part of the document, so it doesn't belong in the file written to disk. However the options bar document state should be saved per-document, and we should eventually move towards those all being grouped into a struct.

PR #492, opened 2022-01-22
Implement the Text Tool and text layer MVP

### LINKED ISSUE #153: New Tool: Text Tool
Clicking somewhere creates a Text Layer at that location and begins letting the user type. This translates to a [`<text>` SVG tag](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text) on the canvas. While editing the text, the user can type stuff. <kbd>Return</kbd> is a new line but <kbd>Enter</kbd> (on the numpad) applies the text, as does <kbd>Ctrl</kbd><kbd>Return</kbd>. The Text Tool also lets you click anywhere over an existing text layer and begin editing that text in the location clicked.

0HyperCube:
Closes #153


graphene/src/layers/text/to_kurbo.rs
```diff
@@ -0,0 +1,168 @@
...
+Test for really long word: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"#;
+	let svg = to_kurbo(text, buzz_face, 20., Some(400.)).to_svg();
+
+	let mut file = File::create("src/layers/text/SourceSansPro/font_text.svg").unwrap();
+	write!(
+		&mut file,
+		r#"<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><path d="{}" /></svg>"#,
+		svg
+	)
+	.unwrap();
```
TrueDoctor, line 167: Could we create a temp file here instead? using a relative path might also potentially break I think?

editor/src/input/input_mapper.rs
```diff
@@ -59,6 +61,10 @@ impl Default for Mapping {
 			// Eyedropper
 			entry! {action=EyedropperMessage::LeftMouseDown, key_down=Lmb},
 			entry! {action=EyedropperMessage::RightMouseDown, key_down=Rmb},
+			// Text
+			entry! {action=TextMessage::Interact, key_up=Lmb},
+			entry! {action=TextMessage::Abort, key_down=KeyEscape},
+			entry! {action=TextMessage::CommitText, key_down=KeyControl, modifiers=[KeyControl]},
```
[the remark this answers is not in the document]
0HyperCube: Good spot, this is meant to be Ctrl+enter

graphene/src/layers/text/to_kurbo.rs
TrueDoctor: Actually, I'm a bit confused, is this a test that is supposed to be checked manually? If so it should probably not run as part of the regular test suite

0HyperCube: This was just me messing around with harfbuzz. I will remove it.

TrueDoctor (approved):

PR #497, opened 2022-01-22
Add folder outline insert marker and clean up layer insertion Vue code

### LINKED ISSUE #447: In the layer panel, drag layer(s) into a folder

0HyperCube:
Closes #447


frontend/src/components/panels/LayerTree.vue
```diff
@@ -361,6 +366,9 @@ export default defineComponent({
 			// Insert index
 			let insertIndex = -1;
 
+			// Weather you are inserting into a folder and should show the folder outline
```
Keavon, line 369: Spelling: Whether*

frontend/src/components/panels/LayerTree.vue
```diff
@@ -414,35 +426,59 @@ export default defineComponent({
...
 
-				if (insertAboveNode.parentNode) {
+				if (this.draggingData.highlightedFolder) {
+					this.draggingData.highlightedFolder.classList.remove("insert-folder");
+				}
+
+				// Highlight folder
+				if (highlightFolder && insertAboveNode instanceof HTMLDivElement) {
+					this.draggingData.highlightedFolder = insertAboveNode;
+					insertAboveNode.classList.add("insert-folder");
```
Keavon, line 466: Please use Vue rather than directly modifying DOM elements, which conflicts with Vue.

Keavon:
One issue is the location where the insert line gets put when dragging to the bottom of the list, it puts the line under the folder (as if the folder was collapsed) rather than below all its children at the bottom of the list. It also feels like `RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT` isn't working anymore. Other than that bug, this looks good to go!

0HyperCube:
@Keavon This is the behaviour from putting the line below the layer row that corresponds to the insert index. I'm not really sure how to resolve this.

Also `RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT` does appear to be working?

Keavon:
> @Keavon This is the behaviour from putting the line below the layer row that corresponds to the insert index. I'm not really sure how to resolve this.

Yeah, not really an easy solution. Is there somewhere we can write a special case that conditionally offsets the index or something to mimic the old behavior? I'm not familiar enough with the logic that the line placement, I basically just worked on the CSS to draw the line. Let me know if you'd like to bounce ideas off or have me investigate something in more detail for you.

> Also `RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT` does appear to be working?


Isn't that supposed to make it so the first 20 pixels below the expanded bottom folder cause it to appear within the folder instead of at the bottom of root? Actually that doesn't even make sense to me at this moment so I think I forgot what that does.

0HyperCube:
> Isn't that supposed to make it so the first 20 pixels below the expanded bottom folder cause it to appear within the folder instead of at the bottom of root? Actually that doesn't even make sense to me at this moment so I think I forgot what that does.

@Keavon No - it means that when you drag below the last layer (if the last layer is a the folder) then the layer is inserted at the bottom of the root.

Keavon:
Code looks good. I'll test it out when the build is live.

Keavon:
Works well. I'll have a chance to glance over the code in a few hours but I suspect it'll be all good. I notice there's a merge conflict, I'm happy to resolve that in a few hours unless you want to get to it first.

frontend/src/components/panels/LayerTree.vue
```diff
@@ -293,9 +295,12 @@ const blendModeEntries: SectionsOfMenuListEntries<BlendMode> = [
 	],
 ];
 
-const RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT = 40;
-const LAYER_LEFT_MARGIN_OFFSET = 28;
+const RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT = 20;
 const LAYER_LEFT_INDENT_OFFSET = 16;
+const LAYER_LEFT_MARGIN_OFFSET = 4 + 32 + LAYER_LEFT_INDENT_OFFSET;
+const INSERT_MARK_OFFSET = 2;
```
Keavon, line 301: I couldn't figure out based on this name what this does. Could you please pick a more descriptive name? If it ends up requiring too many words, a name that's a summary would be good paired with a comment here.

Also a comment above `RANGE_TO_INSERT_WITHIN_BOTTOM_FOLDER_NOT_ROOT`, and possibly a rename if you can think of a more appropriate name, would be helpful since I'm still confused what that one does. (If I recall correctly, I renamed it some time ago but I may have had a mistaken understanding of what it did.) Thanks :)

frontend/src/components/panels/LayerTree.vue
```diff
@@ -346,22 +357,26 @@ export default defineComponent({
 				layer.entry.layer_metadata.selected = false;
 			});
 		},
-		closest(tree: HTMLElement, clientY: number): { insertFolder: BigUint64Array; insertIndex: number; insertAboveNode: Node } {
+		closest(tree: HTMLElement, clientY: number): DraggingData {
```
Keavon, line 360: A more descriptive name for `closest()` would also be pretty helpful, I'm a bit confused what this does overall based on its name.

Keavon:
All good to merge after clearing up my minor remaining confusion about some names, see the two review comments above. Merge when ready and great work on this sir!

Keavon (approved):

PR #498, opened 2022-01-25
Overhaul Path Tool

### LINKED ISSUE #471: Path Tool: allow dragging handles, not just anchors
Create + use new fn `closest_handle` instead? (or better, restructure code with path control-point refs such that multiple disconnected iterations/brute force checks won't be necessary).

Seems like it checks bezier curve endpoints (anchor points) but not its other control points (direction handles)?  Note how, on an ellipse, pressing down on a handle marks it correctly, but then when dragging, it drags the closest anchor instead.

_Originally posted by @leonardpauli in https://github.com/GraphiteEditor/Graphite/pull/451#r780704320_

### LINKED ISSUE #476: Path Tool: Select and modify multiple anchor points at the same time
We want to be able to: 
- Drag box select multi anchor points
- Move the anchor points by grabbing any of the selected points and dragging

otdavies:
Closes #471 
Closes #476 

Still in progress: 
- [x]  Associate overlays with VectorShapes / VectorAnchors / VectorControlPoints.
- [x]  Resolve inconsistencies with ALT toggle mirroring.
- [x]  Use overlays to show which point is selected.
- [x]  Handle corner cases introduced by the addition of text paths.
- [x]  Remove overlay pooling.
- [x]  More comments and cleanup.
- [x]  Resolve "snapback" some shapes not submitting their changes with multiple shapes selected
- [x]  Resolve start point handle issues with open curves

This is an MVP. We should open new issues for additional functionality.


TrueDoctor:
I am not actually sure overlay pooling improves performance that much, we could try to get rid of it and benchark

otdavies:
> I am not actually sure overlay pooling improves performance that much, we could try to get rid of it and benchmark

I agree with this, I am planning on associating anchors / handles overlays with the VectorManipulatorShape / VectorManipulatorShape instead. I was going to leave the pooling in since it at least avoids a bunch of message spam / layer creation / deletion on selection change.

otdavies:
I've heavily refactored all of the systems we use to represent the state of kurbo in our *own* data structures. It has a long way to go, but I've built the scaffold so that adding new features should be a magnitude faster. This also means when we abandon kurbo, we'll have something to start with. I've also pulled as much out of path.rs as possible since it has significant overlay with the pen tool. This PR represents *much* more than just getting handles draggable, it's now in a place to be built on top of, not just tech debt on top of debt tech.

It is by no means perfect and I would love some scrutiny, but I feel it is a massive step in the right direction.

otdavies:
I merged in master and the overlays aren't working properly for text. Resolving now.

otdavies:
I am resolving some issues in the cases where we have open curves. The first point's handle isn't working properly.

editor/src/viewport_tools/shape_manipulation.rs
```diff
@@ -0,0 +1,1197 @@
...
+					/		         \
+			   VectorAnchor ...  VectorAnchor <- VectorShape contains many VectorAnchors
+
+
+					VectorAnchor <- Container for the anchor metadata and optional VectorControlPoints
+						/
+			[Option<VectorControlPoint>; 3] <- [0] is the anchor's draggable point (but not metadata), [1] is the handle1's draggable point, [2] is the handle2's draggable point
+			/				|				\
+		"Anchor"		"Handle1" 		 "Handle2" <- These are VectorControlPoints and the only editable / draggable "primitive"
+*/
```
TrueDoctor, line 40: A lot of these comments can be turned into doc comments (///) to be visible in the auto generated docs

TrueDoctor:
Could you update the PR description/name to better reflect the changes?

editor/src/viewport_tools/vector_editor/shape_editor.rs
```diff
@@ -0,0 +1,202 @@
...
+					/                 \
+			   VectorAnchor ...  VectorAnchor <- VectorShape contains many VectorAnchors
+
+
+					VectorAnchor <- Container for the anchor metadata and optional VectorControlPoints
+						  /
+			[Option<VectorControlPoint>; 3] <- [0] is the anchor's draggable point (but not metadata), [1] is the handle1's draggable point, [2] is the handle2's draggable point
+			 /              |                      \
+		"Anchor"        "Handle1"          "Handle2" <- These are VectorControlPoints and the only editable / draggable "primitive"
+*/
```
TrueDoctor, line 16: Just a reminder to turn these into doc comments

TrueDoctor:
I have to admit I am a bit confused why it treats the diagram as a doc test, I thought that only worked with code blocks (```).
But we can still figure that out after this is merged, otherwise, this all looks probably as good as it can get given the circumstances :stuck_out_tongue_winking_eye:

TrueDoctor (approved):

Issue #147, opened 2021-05-23
Standardized layout definition system


Keavon:
We need to create a reusable system for defining input widgets for the frontend to render and communicate changes back to Rust. This will be used in the Document Panel's options bar for tools (#114) in one row but grouped into sections, as well as the Properties Panel for a multi-row and column-based layout likely broken into sections and possibly tabs (#146), plus other future uses. This will involve defining a flexible API spec (likely JSON) for interface elements and their layout, and creating a modular, reusable system.

Keavon:
Some examples of needed layout capabilities:

![Properties panel](https://files.keavon.com/-/ReadyWellinformedStilt/capture.png)

![Options bar](https://files.keavon.com/-/SomberUnrealisticGreatdane/capture.png)

Keavon:
Notes on this:

Layout system depends on a lot of things. We need to define some layout API approach that can properly encode all the designs and layouts we need throughout Graphite. This requires some additional design work to fully understand the needs. This also touches upon the extension system, since extensions should be able to add layouts. It also gets into the question of how far we want to go with this because it leans into the territory of the original XML-based layout system. That system was meant for defining composable components (much like Vue is used now for) and could be resurrected with the web and Vue as UI renderers, until we port that system over to a custom WGPU-based render backend. This could be a path towards gently transitioning to the native editor client.

TrueDoctor:
some pseudocode:
```rust
trait Component {
	fn render(&self) -> HTML;
}

struct Checkbox {
	bool: checked,
}

impl Component for Checkbox {
	fn render(&self) -> HTML {
		html!("<box>{}</box>", self.checked)
	}
}
struct BarSeparator;
impl Component for BarSeparator {
	fn render(&self) -> HTML {
		html!("|", self.checked)
	}
}

type Icon = String;
const LEFT_ALIGN: Icon = "left_align";

struct AlignButton {
    icon: Icon,
    tooltip: Option<String>,
	callback: String,
}

fn callback(m: Message) -> HTML {
	html!("<div onclick=dispatch_message({})>", serde_json::to_string(m))
}

let shape_options = …;
shape_options.render()

const LEFT_ALIGN_BUTTON: AlignButton = AlignButton{icon: LEFT_ALIGN, tooltip:"left_align"};

#[derive(Component)]
#[Component(direction:horizontal)]
struct ShapeOptions { // 3c | 3b | 3c
    stroke_bitmask: Row<Checkbox>,
[...]

PR #499, opened 2022-01-25
Layout system implementation and applied to tool options bar

### LINKED ISSUE #284: Add a unified messaging system for tool options
#283 completed all of #184 except for:
> A unified approach to routing options changes to the backend, and the state of those options back to the frontend

Each tool should probably have an associated options struct on the frontend. Each tool option widget should probably have an `onChange` or analogous callback that updates the struct and then sends it to the backend. Updates to a tool's options within the backend should probably send a `Response` to the frontend, allowing it to update the tool's options struct. Persisting the option widgets should probably occur at this response handling step (see zoom and rotate).

The main question is how to associate each widget with a field in the struct. The easiest thing would probably be to somehow identify the field within the option object (e.g. `NumberOption { ... option: "shape_type.Polygon.new_value" }`), or possibly a choice between this approach and a custom callback.

Buttons should most likely be implemented by associating a string with the button, that will be translated to a variant of a new struct like `OptionButtonMessage`. It would be nice to have a button messages enum for each tool.

### LINKED ISSUE #485: Send the viewport state info to the widgets when document loaded/switched
The widgets on the right of the viewport's options bar needs to be updated to display the state of those widgets when switching between document tabs or loading the app (which loads an autosaved document).

Options bar widgets

This should hopefully be a general system, not just a one-off fix per widget.

mfish33:
Closes #284
Fixes #485
Part of #147

This is a big PR so there are some small things I still might change but I think the majority of it is in a pretty good state.


editor/src/viewport_tools/tool.rs
```diff
@@ -35,14 +37,16 @@ pub trait Fsm {
...
-type SubToolMessageHandler = dyn for<'a> MessageHandler<ToolMessage, ToolActionHandlerData<'a>>;
+pub trait ToolCommon: for<'a> MessageHandler<ToolMessage, ToolActionHandlerData<'a>> + PropertyHolder {}
+impl<T> ToolCommon for T where T: for<'a> MessageHandler<ToolMessage, ToolActionHandlerData<'a>> + PropertyHolder {}
+
+type Tool = dyn ToolCommon;
 
 pub struct ToolData {
 	pub active_tool_type: ToolType,
-	pub tools: HashMap<ToolType, Box<SubToolMessageHandler>>,
+	pub tools: HashMap<ToolType, Box<Tool>>,
```
otdavies, line 49: I feel like this rename doesn't really improve readability, it is still a message handler correct? I could be wrong, so feel free to correct me.

otdavies (approved):
Overall, I think this looks very good. I really like the trait requiring per-tool options be impl'd. This puts the options near the tool logic. My one consideration would be duplicate code if some tools have similar options bars, but that feels like something to solve then. Also we will need to update the messaging diagram to add the concept of a layout message.

I think a similar approach would be beneficial for the bottom control hints bar and the properties panel. I appreciate that it is relatively extensible and will likely be able to extend into other uses. I like the implementation of the widget concept and think it could be leveraged in the property panel as well. 

Great change, I look forward to leveraging it in future work on the tools.

I would request you don't merge this until @TrueDoctor and @Keavon give their ok aswell.

Keavon (approved):
Amazing work, seriously.

editor/src/viewport_tools/tool.rs
Keavon: This has a different meaning, we talked about this and I think `Tool` is the best option.

PR #500, opened 2022-01-27
Simplify point storage for pen tool

TrueDoctor:
Store points as vectors instead of representing them as transforms to
avoid confusion and improve efficiency


0HyperCube (approved):
Looks good.

PR #501, opened 2022-01-29
Implement layer renaming

### LINKED ISSUE #445: Renaming layers
Double click a layer name to rename it. Clicking outside the text entry or hitting Enter or Return should apply the change and Escape should cancel the change.

We should also probably make it so empty names default to the name of the layer (node) type, such as Ellipse, Rectangle, Shape, Line, or Path, written in italics and darker text. Then we get rid of the "Unnamed Shape" default text and use empty default text, which becomes the Ellipse/Rectangle/Shape/Line/Path, etc.

The best way to edit the text might be [`contenteditable`](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Editable_content). Or maybe we switch the HTML to use an `<input>` element and style it when not in editing mode to appear just like regular text and not be editable until it's double clicked and Vue switches it to editable.

Keavon:
Closes #445

### DIFF
--- a/editor/src/document/document_message.rs
+++ b/editor/src/document/document_message.rs
@@ -99,6 +99,10 @@ pub enum DocumentMessage {
 		layer_path: Vec<LayerId>,
 		set_expanded: bool,
 	},
+	SetLayerName {
+		layer_path: Vec<LayerId>,
+		name: String,
+	},
 	SetOpacityForSelectedLayers {
 		opacity: f64,
 	},
--- a/editor/src/document/layer_panel.rs
+++ b/editor/src/document/layer_panel.rs
@@ -21,8 +21,7 @@ impl LayerMetadata {
 }
 
 pub fn layer_panel_entry(layer_metadata: &LayerMetadata, transform: DAffine2, layer: &Layer, path: Vec<LayerId>) -> LayerPanelEntry {
-	let layer_type: LayerDataTypeDiscriminant = (&layer.data).into();
-	let name = layer.name.clone().unwrap_or_else(|| format!("Unnamed {}", layer_type));
+	let name = layer.name.clone().unwrap_or_else(|| String::from(""));
 	let arr = layer.data.bounding_box(transform).unwrap_or([DVec2::ZERO, DVec2::ZERO]);
 	let arr = arr.iter().map(|x| (*x).into()).collect::<Vec<(f64, f64)>>();
 
--- a/frontend/src/components/panels/LayerTree.vue
+++ b/frontend/src/components/panels/LayerTree.vue
@@ -379,7 +379,7 @@ export default defineComponent({
 			listing.editingName = true;
 			const tree = (this.$refs.layerTreeList as typeof LayoutCol).$el as HTMLElement;
 			this.$nextTick(() => {
-				(tree.querySelector("input:not([disabled]") as HTMLInputElement).select();
+				(tree.querySelector("input:not([disabled])") as HTMLInputElement).select();
 			});
 		},
 		async onEditLayerNameChange(listing: LayerListingInfo, inputElement: EventTarget | null) {
--- a/frontend/src/components/panels/LayerTree.vue
+++ b/frontend/src/components/panels/LayerTree.vue
@@ -29,21 +29,21 @@
 			</PopoverButton>
 		</LayoutRow>
 		<LayoutRow class="layer-tree" :scrollableY="true">
-			<LayoutCol class="list" ref="layerTreeList" @click="() => deselectAllLayers()" @dragover="updateInsertLine($event)" @dragend="drop()">
+			<LayoutCol class="list" ref="layerTreeList" @click="() => deselectAllLayers()" @dragover="(e) => draggable && updateInsertLine(e)" @dragend="draggable && drop()">
 				<LayoutRow
 					class="layer-row"
 					v-for="(listing, index) in layers"
 					:key="String(listing.entry.path.slice(-1))"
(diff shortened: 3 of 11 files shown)


editor/src/document/layer_panel.rs
```diff
@@ -21,8 +21,7 @@ impl LayerMetadata {
 }
 
 pub fn layer_panel_entry(layer_metadata: &LayerMetadata, transform: DAffine2, layer: &Layer, path: Vec<LayerId>) -> LayerPanelEntry {
-	let layer_type: LayerDataTypeDiscriminant = (&layer.data).into();
-	let name = layer.name.clone().unwrap_or_else(|| format!("Unnamed {}", layer_type));
+	let name = layer.name.clone().unwrap_or_else(|| String::from(""));
```
0HyperCube, line 24: Can be simplified to `.unwrap_or(String::new()))`

Keavon: Thanks, I suspected there ought to be a better way but didn't know what.

frontend/src/components/panels/LayerTree.vue
```diff
@@ -254,7 +287,7 @@ import NumberInput from "@/components/widgets/inputs/NumberInput.vue";
 import IconLabel from "@/components/widgets/labels/IconLabel.vue";
 import Separator from "@/components/widgets/separators/Separator.vue";
 
-type LayerListingInfo = { entry: LayerPanelEntry; bottomLayer: boolean; folderIndex: number };
+type LayerListingInfo = { folderIndex: number; bottomLayer: boolean; editingName: boolean; entry: LayerPanelEntry };
```
0HyperCube, line 290: I believe `bottomLayer` can be removed as it is not used.

editor/src/document/layer_panel.rs
Keavon: Wait, no, Clippy asks me to switch it back to avoid the greedy call.

frontend/src/components/workspace/Workspace.vue
```diff
@@ -23,12 +23,12 @@
 					ref="documentsPanel"
 				/>
 			</LayoutCol>
-			<LayoutCol class="workspace-grid-resize-gutter" @pointerdown="resizePanel($event)"></LayoutCol>
+			<LayoutCol class="workspace-grid-resize-gutter" @pointerdown="(e) => resizePanel(e)"></LayoutCol>
```
0HyperCube, line 26: What are the benefits of using `(e) => fn(e)` over `fn($event)`?

graphene/src/document.rs
```diff
@@ -641,6 +641,13 @@ impl Document {
 				layer.visible = *visible;
 				Some([vec![DocumentChanged], update_thumbnails_upstream(path)].concat())
 			}
+			Operation::SetLayerName { path, name } => {
+				self.mark_as_dirty(path)?;
```
0HyperCube, line 645: Why are we marking it as dirty? It shouldn't need to be rerendered.

0HyperCube (approved):
Looks good :)

editor/src/document/layer_panel.rs
0HyperCube: even better: `unwrap_or_default()`

Keavon: Ah, yes, of course!

frontend/src/components/workspace/Workspace.vue
Keavon: Purely standardization with the rest of the code. It's sort of an arbitrary choice, we could also switch them all to `$event`.

frontend/wasm/src/api.rs
```diff
@@ -397,6 +398,12 @@ impl JsEditorHandle {
 		self.dispatch(message);
 	}
 
+	/// Set the name for the layer
+	pub fn set_layer_name(&self, layer_path: Vec<LayerId>, name: String) {
+		let message = DocumentMessage::SetLayerName { layer_path, name };
+		self.dispatch(message);
+	}
+
```
TrueDoctor, line 406: Why didn't you just call the existing rename_layer() function?

editor/src/document/document_message.rs
```diff
@@ -99,6 +99,10 @@ pub enum DocumentMessage {
 		layer_path: Vec<LayerId>,
 		set_expanded: bool,
 	},
+	SetLayerName {
```
HansAuger, line 102: I noticed a `RenameLayer` operation already exists. See if that can be used/reused/removed to avoid duplication.

frontend/src/components/panels/LayerTree.vue
```diff
@@ -336,6 +371,36 @@ export default defineComponent({
...
 		},
+		onEditLayerName(listing: LayerListingInfo) {
+			if (listing.editingName) return;
+
+			this.draggable = false;
+
+			listing.editingName = true;
+			const tree = (this.$refs.layerTreeList as typeof LayoutCol).$el as HTMLElement;
+			this.$nextTick(() => {
+				(tree.querySelector("input:not([disabled]") as HTMLInputElement).select();
```
HansAuger, line 382:
```suggestion
				(tree.querySelector("input:not([disabled])") as HTMLInputElement).select();
```

editor/src/document/document_message.rs
Keavon: Thanks for noticing this! It never came up in my search. All the more reason for nice documentation to be a goal where we can generate a Rustdoc documentation website to preview all these things nicely :) Thanks for the code review.

PR #502, opened 2022-01-29
Transformation cage

### LINKED ISSUE #491: Turn the Select Tool's bounding box into a transformation cage
Currently the Select Tool displays a rectangle around the bounds of the selected layers in the viewport. We should add squares (or ideally, triangles that point locally up) to the centers of the rectangle's edges as well as the corners which can be used for dragging to transform the selection. The cursor should become an up/down or left/right arrow when dragging anywhere along the edges of the rectangle. It should become a diagonal cursor when dragging near the corners of the rectangle.

And when the cursor is a bit further away from the edges and corners, but still within a reasonable proximity, it should display a rotational cursor (CSS doesn't provide one but we'll make one, you can start out by using the horizontal/vertical/diagonal ones parallel to the direction of rotational "torque"). Dragging when such a cursor is shown should cause rotation around the origin. (We'll add the ability to drag the origin point in a future issue.)

One complication: when the canvas is rotated in the viewport, we currently draw the bounding rectangle aligned upright in viewport space, which means it rotates in document space. This transformation is arguably a document space action which means it might be nice to display it in document space. One approach is to make the bounding box work in document space, the other option is to have it work in viewport space like it is now but transformation would then be applied from document to viewport space. Ideally, both are implemented and we figure out which is nicer and either make it a user preference or a toggle in the Select Tool's options bar.

0HyperCube:
Closes #491


Keavon:
It looks like this branch's [build](https://30d5fcc8.graphite-master.pages.dev/) also has the right side of the options bar missing. @mfish33 do you have any ideas why multiple Cloudflare builds now would have caused this randomly, unrelated to the code itself, and not in editor.graphite.design or localhost (yet?)?

0HyperCube:
@Keavon My best guess is that it is something to do with compiling the wasm / js with production optimisations. Can you try localhost with optimised wasm compiling?

0HyperCube:
Also @Keavon are there any more features that you would like in this before merging?

Keavon:
> Also @Keavon are there any more features that you would like in this before merging?

I think the current set of features is a good checkpoint for the MVP implementation. We'll have more stuff like a draggable origin point and an options bar checkbox for local or global transformation (this is currently global where the bounding box is always viewport-relative not layer-relative). But those can all probably wait till after the alpha release. Current priority is getting the items on the sprint board done, like fonts and gradients. Let me code review this real quick then I think we can go ahead and merge.

editor/src/viewport_tools/tools/select.rs
```diff
@@ -296,6 +303,63 @@ impl SelectToolData {
...
+	top: bool,
+	bottom: bool,
+	left: bool,
+	right: bool,
+}
+
+impl SelectedEdges {
+	fn new(top: bool, bottom: bool, left: bool, right: bool, bounds: [DVec2; 2], pivot: &mut DVec2) -> Self {
+		// Calculate the pivot for the operation (the opposite point to the edge dragged)
+		*pivot = {
```
Keavon, line 319: Is there any other clean way we could do this that doesn't involve "returning data" to the arguments of the caller? This feels too much like a C hack which Rust's superior feature set should help us avoid. If you think this is the cleanest way, no worries.

editor/src/viewport_tools/tools/select.rs
```diff
@@ -309,10 +373,150 @@ fn add_bounding_box(responses: &mut Vec<Message>) -> Vec<LayerId> {
...
 }
 
+/// Contains info on the overlays for the bounding box and transform handles
+#[derive(Clone, Debug, Default)]
+struct BoundingBoxOverlays {
+	pub bounding_box: Vec<LayerId>,
+	pub transform_handles: [Vec<LayerId>; 8],
+	pub bounds: [DVec2; 2],
+	pub selected_edges: Option<SelectedEdges>,
+	pub origional_transforms: OriginalTransforms,
```
Keavon, line 420: Typo: `origional` -> `original`

editor/src/viewport_tools/tools/select.rs
```diff
@@ -309,10 +373,150 @@ fn add_bounding_box(responses: &mut Vec<Message>) -> Vec<LayerId> {
...
+		outside_bounds & inside_extended_bounds
+	}
+
+	pub fn get_cursor(&self, input: &InputPreprocessorMessageHandler) -> MouseCursorIcon {
+		if let Some(directions) = self.check_selected_edges(input.mouse.position) {
+			match directions {
+				(true, false, false, false) | (false, true, false, false) => MouseCursorIcon::NSResize,
+				(false, false, true, false) | (false, false, false, true) => MouseCursorIcon::EWResize,
+				(true, false, true, false) | (false, true, false, true) => MouseCursorIcon::NWSEResize,
+				(true, false, false, true) | (false, true, true, false) => MouseCursorIcon::NESWResize,
```
Keavon, line 499: This is a pretty cool use of Rust's match statement.

Keavon:
> Also @Keavon are there any more features that you would like in this before merging?

Oh yeah, that one other feature request about removing the corner handles if the width or height of the layer (in viewport pixels) is too small to fit the three handles. Although if you want to do that in a followup PR, that's fine too.

editor/src/viewport_tools/tools/select.rs
```diff
@@ -440,6 +683,49 @@ impl Fsm for SelectToolFsmState {
...
+							selected.update_transforms(delta);
+						}
+					}
+					ResizingBounds
+				}
+				(RotatingBounds, MouseMove { snap_angle, .. }) => {
+					if let Some(bounds) = &mut data.bounding_box_overlays {
+						let angle = {
+							let start_vec = data.drag_start - bounds.pivot;
+							let end_vec = input.mouse.position - bounds.pivot;
```
Keavon, line 707: Nitpick: using the full word `vector` would help avoid confusion with Rust's `Vec` data type.

Keavon (approved):
Good to merge after resolving comments! The editor is becoming more and more like a real graphics suite every day largely thanks to your awesome work.

editor/src/viewport_tools/tools/select.rs
0HyperCube: Also similar to C# out parameters :). This probably should have been refactored when I moved the pivot out of the SelectedEdges struct.

editor/src/viewport_tools/tools/select.rs
0HyperCube: I think end_offset is probably a better name for this. I have renamed all the other occurrences of this pattern too.

Keavon: Perfect, even better.

PR #503, opened 2022-01-29
Implement Freehand Tool

### LINKED ISSUE #495: Implement the Freehand Tool
Duplicate the current Pen Tool implementation and make it the Freehand Tool. Make it so it draws a polyline segment every mousemove. Someday we can bother with smoothing and spline/bezier interpolation, but that day is not today.

pkupper:
Naive implementation of Freehand tool based on Pen Tool. Adds a new line segment for every PointerMove event.

Closes #495


editor/src/input/input_mapper.rs
```diff
@@ -94,6 +94,10 @@ impl Default for Mapping {
 			entry! {action=PenMessage::Confirm, key_down=Rmb},
 			entry! {action=PenMessage::Confirm, key_down=KeyEscape},
 			entry! {action=PenMessage::Confirm, key_down=KeyEnter},
+			// Pen
```
0HyperCube, line 97: Comment should be freehand

editor/src/viewport_tools/tools/freehand.rs
```diff
@@ -0,0 +1,192 @@
...
+	fn default() -> Self {
+		FreehandToolFsmState::Ready
+	}
+}
+#[derive(Clone, Debug, Default)]
+struct FreehandToolData {
+	points: Vec<DVec2>,
+	weight: u32,
+	path: Option<Vec<LayerId>>,
+	layer_exists: bool,
```
0HyperCube, line 83: Shouldn't this always be the case when path is some? You appear not to be using this in your code anyway.

graphene/src/document.rs
```diff
@@ -485,6 +485,17 @@ impl Document {
...
+				path,
+				insert_index,
+				points,
+				transform,
+				style,
+			} => {
+				let points: Vec<glam::DVec2> = points.iter().map(|&it| it.into()).collect();
+				self.set_layer(path, Layer::new(LayerDataType::Shape(Shape::poly_line(points, *style)), *transform), *insert_index)?;
+				Some([vec![DocumentChanged, CreatedLayer { path: path.clone() }], update_thumbnails_upstream(path)].concat())
+			}
```
0HyperCube, line 498: This is duplicated behaviour. Consider (for now) merging this with the `Operation::AddPen` and calling it `Operation::AddPolyline` or similar to reflect the actual shape type that is created.

0HyperCube (commented):
Works brilliantly 👍🏽

editor/src/viewport_tools/tools/freehand.rs
pkupper: Removed this from pen tool as well. Was also unused there.

graphene/src/document.rs
pkupper: Consolidated this to Operation::AddPolyline as suggested.

Keavon:
Please add the keyboard shortcut <kbd>N</kbd> in `input_mapper.rs`.

editor/src/viewport_tools/tools/freehand.rs
```diff
@@ -0,0 +1,190 @@
...
+		} else {
+			self
+		}
+	}
+
+	fn update_hints(&self, responses: &mut VecDeque<Message>) {
+		let hint_data = match self {
+			FreehandToolFsmState::Ready => HintData(vec![HintGroup(vec![HintInfo {
+				key_groups: vec![],
+				mouse: Some(MouseMotion::Lmb),
```
Keavon, line 160: Let's use `MouseMotion::LmbDrag`, I think that'd be more accurate.

editor/src/viewport_tools/tool_message.rs
```diff
@@ -25,6 +25,9 @@ pub enum ToolMessage {
 	Fill(FillMessage),
 	#[remain::unsorted]
 	#[child]
+	Freehand(FreehandMessage),
+	#[remain::unsorted]
+	#[child]
```
Keavon, line 30: Please put below `Pen`. Actually some others of these are in the wrong order. Would you mind rearranging them in the order that they appear in the tool shelf of the app?

editor/src/viewport_tools/tools/freehand.rs
```diff
@@ -0,0 +1,190 @@
...
+				label: String::from("Draw Freehand"),
```
Keavon, line 161: For the moment, while we have no smoothing, let's call this `"Draw Polyline"`.

editor/src/viewport_tools/tool.rs
```diff
@@ -267,6 +267,7 @@ pub fn standard_tool_message(tool: ToolType, message_type: StandardToolMessageTy
 			ToolType::Path => Some(PathMessage::Abort.into()),
 			ToolType::Navigate => Some(NavigateMessage::Abort.into()),
 			ToolType::Pen => Some(PenMessage::Abort.into()),
+			ToolType::Freehand => Some(FreehandMessage::Abort.into()),
```
Keavon, line 270: Same request here as mentioned in the other comment: if you don't mind filling in (commented out) additional lines for all of the tools, then order them according to their order of appearance in the tool shelf (the order used in the block right above this block), that would be much appreciated!

Keavon (approved):
Looks and works great! Just a few comments where you can clean up the code a little (most of this isn't even your fault, you just happen to be the victim to touch those lines of code so I thought I'd ask you to do a little cleanup while you're there 😉). Thank you for this! Smash that "merge" button once you're made the requested changes.

PR #505, opened 2022-01-30
Fix overlays on document switching

0HyperCube:
When switching documents, the tool would try and delete the overlay layer from the previous document after switching to the new document (resulting in an error as well as the overlay remaining permanently on the old document)


Keavon:
Was this PR meant to fix the issue shown in this video? https://files.keavon.com/-/SeashellButteryOwlbutterfly/capture_7_.mp4

0HyperCube:
@Keavon Yes it is, and it worked for switching between documents but not for new documents, however that is now resolved with the latest commit (I had to make the new document message a proxy so as to process the tool abort before changing the active_document_id).

Keavon:
> (I had to make the new document message a proxy so as to process the tool abort before changing the active_document_id).

Instead of proxying the message, couldn't you just do it like before but change the `self.active_document_id = document_id;` line to a call to the `SelectDocument` message?

0HyperCube:
@Keavon I think that this would break when there is no active document (i.e. the document list is empty)

Keavon:
Can SelectDocument be modified to allow that case? I just don't love that we have to use a wrapper that serves no external purpose besides use by one other message. It seems like we should be able to use the existing machinery and generally aim to reduce our usage of function calls (which happen immediately) mixed in with message dispatches (which happen in the queue after the function body returns). Maybe we can work towards purifying our message system to achieve that separation in the future.

0HyperCube:
@Keavon I have cleaned up the messages as well as applying the fix to where the active document is deleted.

Keavon (approved):
Much nicer! This is a very clean approach. Sorry this dropped from my radar for the past two days. Ready to merge!

PR #506, opened 2022-01-31
Clean up text input

0HyperCube:
The text input on firefox was sometimes not displaying the text.


Keavon:
It would appear on Chrome that the cursor doesn't blink now with empty text. Strange, I'm surprised this CSS affects that.

TrueDoctor (approved):
The 10000px magic number seems a bit strange but that seems to fix the issue

0HyperCube:
> It would appear on Chrome that the cursor doesn't blink now with empty text. Strange, I'm surprised this CSS affects that.

@Keavon Apparently `overflow: visible` fixes that.

0HyperCube:
@TrueDoctor It is somewhat of a hack as it appears that `overflow: visible` along with a small size on the parent foreignobject causes the text to be clipped sometimes in firefox.

Keavon:
Large-numbered widths are a well-established, but unfortunate, hack in CSS. Best avoided when possible but this seems fine here.

graphene/src/layers/text/to_kurbo.rs
```diff
@@ -74,7 +74,7 @@ pub fn to_kurbo(str: &str, buzz_face: rustybuzz::Face, font_size: f64, line_widt
 		path: BezPath::new(),
 		pos: Point::ZERO,
 		offset: Vec2::ZERO,
-		ascender: buzz_face.ascender() as f64,
+		ascender: (buzz_face.ascender() as f64 / buzz_face.height() as f64) * font_size / scale,
```
Keavon, line 77: Descenders below the bounding box is to be expected. This change seems to add a large basically a full line below the text for the bounding box, which isn't ideal.

Keavon:
Somehow, this CI build is showing no widgets on the right side of the viewport options bar. That isn't the case when I build it myself, and obviously the code didn't cause the breakage. I wonder what happened. Perhaps something to keep an eye on if this ever ends up disappearing again.

graphene/src/layers/text/to_kurbo.rs
0HyperCube: > This change seems to add a large basically a full line below the text for the bounding box, which isn't ideal.

No this does not. I am not sure what you are talking about. This does not influence the bounding box at all.

0HyperCube:
@Keavon It is showing them for me in firefox, but not in Chromium for some reason.

graphene/src/layers/text/to_kurbo.rs
0HyperCube: I've reverted the baseline change, however the new line issue probably still remains.

Keavon:
I figured out the problem. Multiple lines cause there to be a large gap underneath the text, but this is also the case in master. I thought that had to do with this baseline change which is where I first noticed the issue. But it actually had nothing to do with it.

Keavon:
> @Keavon It is showing them for me in firefox, but not in Chromium for some reason.

If you're referring to the missing right side of the options bar, both Firefox and Chrome do this for me. Quite weird. Not related to the PR though.

graphene/src/layers/text/to_kurbo.rs
Keavon: I figured out the problem. Multiple lines cause there to be a large gap underneath the text, but this is also the case in master. I thought that had to do with this baseline change which is where I first noticed the issue. But it actually had nothing to do with it.



Sorry for the confusion.

0HyperCube: @Keavon No problem, it appears that chrome returns a random trailing `\n` char in some situations with `textContent`. Converting to `innerText` appears to have solved this issue?

Keavon: Awesome, looks good. The only text issues I'm seeing now is that we seem to have *nothing*(?) highlighted when you first click to create a new text field and this shows up as a block of blue (you can zoom into it with Ctrl+scroll). Then typing something and deleting it again makes it so that goes away. Also it seems that zooming into the document in general causes all the text to get selected. But both of these can be eventually resolved when we have more time to polish, they aren't important enough for current priorities.

0HyperCube:
@mfish33 I can't seem to find the code that dispatches `UpdateLayout` on startup?

frontend/src/components/panels/Document.vue
```diff
@@ -379,19 +379,29 @@ export default defineComponent({
 			this.canvasCursor = updateMouseCursor.cursor;
 		});
 		this.editor.dispatcher.subscribeJsMessage(TriggerTextCommit, () => {
-			if (this.textInput) this.editor.instance.on_change_text(this.textInput.textContent || "");
+			if (this.textInput) {
```
Keavon, line 382: Please add a comment explaining the motivation and details for this since it'll probably be confusing to people in the future.

frontend/src/components/panels/Document.vue
```diff
@@ -379,19 +379,29 @@ export default defineComponent({
...
+			this.textInput.innerText = `${displayEditableTextbox.text}\n`;
 			this.textInput.contentEditable = "true";
 			this.textInput.style.width = displayEditableTextbox.line_width ? `${displayEditableTextbox.line_width}px` : "max-content";
 			this.textInput.style.height = "auto";
 			this.textInput.style.fontSize = `${displayEditableTextbox.font_size}px`;
 			this.textInput.oninput = (): void => {
-				if (this.textInput) this.editor.instance.update_bounds(this.textInput.textContent || "");
+				if (this.textInput) {
+					let text = this.textInput.innerText;
+					if (text[text.length - 1] === "\n") text = text.slice(0, -1);
```
Keavon, line 401: This seems to be duplicate code as above. Can you break this out into a function?

Keavon (approved):
I left some comments and replied to the thread we had going about the bounding box. Feel free to fix what you'd like and leave what else you'd prefer to save for eventual polishing, then merge. You have my approval and True Doctor's now. Thanks for keeping a close eye on Firefox and Chromium, since I too often just test in Chrome. You're teaching me to test elsewhere (and if I had a Mac, I'd also test on Safari...).

graphene/src/layers/text/to_kurbo.rs
0HyperCube: > Awesome, looks good. The only text issues I'm seeing now is that we seem to have nothing(?) highlighted when you first click to create a new text field and this shows up as a block of blue

Resolved.

frontend/src/components/panels/Document.vue
0HyperCube: The shared code is 2 lines used three times (one time has a different editor function) I'm not really sure it is worth sharing them?

Keavon: I'd normally agree with your thoughts that they are too minimal to share, but these two lines are relatively high-entropy and error-prone when doing maintenance on them. Like they could get easily out of sync with one another by accident when maintaining this code someday. So for that reason I feel like a function would be better in this case, even though most average triple-duplicate pairs of lines aren't worth it.

frontend/src/utilities/cleanup_text_input.ts
```diff
@@ -0,0 +1,5 @@
+// Necessary because innerText puts an extra newline character at the end when the text is more than one line.
```
Keavon, line 1: Pease use `kebab-case` for TypeScript files per the existing convention. Although you can probably avoid using a single-purpose file in `/utilities` by just exporting it from `input.ts` since it's such a short function and is already used in `input.ts` (in addition to `Document.vue`).

Keavon: I committed the change so we could get this merged while you're asleep.

PR #507, opened 2022-02-04
Double click on a shape to transition to path tool

0HyperCube:
From #code-todo-list


pkupper (approved):
You might want to consider adding a hint for the double-click action (Although there doesn't seem to be an appropriate MouseMotion Variant yet). Otherwise LGTM!

PR #510, opened 2022-02-05
Fix shape not closed after using Fill Tool

### LINKED ISSUE #509: Allow a filled polyline to be a draggable shape
Currently if you give a fill to a polyline (from the Pen Tool or Freehand Tool), you can only drag the line but can't drag the filled portion of the shape.

pkupper:
Closes #509
Well that was simpler than expected. @Keavon Let me know if the shape outline should be closed as well when using the fill tool.

### DIFF
--- a/editor/src/document/document_message_handler.rs
+++ b/editor/src/document/document_message_handler.rs
@@ -151,7 +151,7 @@ impl DocumentMessageHandler {
 			};
 
 			let (path, closed) = match &layer.ok()?.data {
-				LayerDataType::Shape(shape) => Some((shape.path.clone(), shape.closed)),
+				LayerDataType::Shape(shape) => Some((shape.path.clone(), shape.path.elements().last() == Some(&kurbo::PathEl::ClosePath))),
 				LayerDataType::Text(text) => Some((text.to_bez_path_nonmut(), true)),
 				_ => None,
 			}?;
--- a/editor/src/document/document_message_handler.rs
+++ b/editor/src/document/document_message_handler.rs
@@ -151,6 +151,7 @@ impl DocumentMessageHandler {
 			};
 
 			let (path, closed) = match &layer.ok()?.data {
+				// TODO: This ClosePath check does not handle all cases, fix this soon
 				LayerDataType::Shape(shape) => Some((shape.path.clone(), shape.path.elements().last() == Some(&kurbo::PathEl::ClosePath))),
 				LayerDataType::Text(text) => Some((text.to_bez_path_nonmut(), true)),
 				_ => None,
--- a/graphene/src/document.rs
+++ b/graphene/src/document.rs
@@ -699,7 +699,10 @@ impl Document {
 			Operation::SetLayerFill { path, color } => {
 				let layer = self.layer_mut(path)?;
 				match &mut layer.data {
-					LayerDataType::Shape(s) => s.style.set_fill(layers::style::Fill::new(*color)),
+					LayerDataType::Shape(s) => {
+						s.style.set_fill(layers::style::Fill::new(*color));
+						s.closed = true;
+					}
 					_ => return Err(DocumentError::NotAShape),
 				}
 				self.mark_as_dirty(path)?;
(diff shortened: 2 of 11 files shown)


Keavon:
Somehow the click target is now actually the filled shape, and I can't see any obvious shape it's forming. In this video, every time I move my mouse I click afterwards showing that I'm often not clicking the shape's click target at all.

https://files.keavon.com/-/TriangularRedNutria/capture_3_.mp4

graphene/src/document.rs
```diff
@@ -699,7 +699,10 @@ impl Document {
 			Operation::SetLayerFill { path, color } => {
 				let layer = self.layer_mut(path)?;
 				match &mut layer.data {
-					LayerDataType::Shape(s) => s.style.set_fill(layers::style::Fill::new(*color)),
+					LayerDataType::Shape(s) => {
+						s.style.set_fill(layers::style::Fill::new(*color));
+						s.closed = true;
```
Keavon, line 704: I'm surprised this isn't causing the stroke to be drawn from the starting point to the ending point, since I thought that was what the `closed` property meant. However we don't actually want to change the stroke when filling the shape, since the shape is still open and it just happens to have a fill. Perhaps this `closed` property does what we want but it might be deserving of a rename. Pining @otdavies for thoughts.

pkupper: Whether the path is closed is determined by whether or not the `kurbo::BezPath` ends with a `ClosePath` element. This closed field is only used for selection and should probably be called something like `solid` instead.

Keavon: `filled` would probably be clearest, although couldn't we just remove the field entirely and check for whether it has a `Some(...)` or `None` fill color?

pkupper:
> Somehow the click target is now actually the filled shape, and I can't see any obvious shape it's forming. In this video, every time I move my mouse I click afterwards showing that I'm often not clicking the shape's click target at all.
> 
> https://files.keavon.com/-/TriangularRedNutria/capture_3_.mp4

When I try to replicate the shape from the video I have the same issue. Strangely enough, for all other shapes I have tried it works just fine. Will investigate.

Keavon:
> When I try to replicate the shape from the video I have the same issue. Strangely enough, for all other shapes I have tried it works just fine. Will investigate.

Indeed, it's pretty strange. Glad you're able to reproduce the problem and thank you for looking into it.

editor/src/document/document_message_handler.rs
```diff
@@ -151,7 +151,7 @@ impl DocumentMessageHandler {
 			};
 
 			let (path, closed) = match &layer.ok()?.data {
-				LayerDataType::Shape(shape) => Some((shape.path.clone(), shape.closed)),
+				LayerDataType::Shape(shape) => Some((shape.path.clone(), shape.path.elements().last() == Some(&kurbo::PathEl::ClosePath))),
```
Keavon, line 154: I thought the goal of removing `.closed` was to treat it as having a filled color, not something related to the Kurbo geometry?

pkupper: The `closed` field was used incorrectly here. Probably due to the confusing naming. Here `closed` actually means whether the outline is closed. So the name fits.

Keavon: Gotcha. Checking with @otdavies: does this cause any conflicts with your branch? And is this robust enough when multiple sub-shapes (separated by `moveto`s) are involved?

otdavies: I will remove my usage of .closed and thus won't be an issue. I didn't use it before and I only recently incorporated it.

Keavon: @pkupper: @otdavies and I discussed this and it sounds like the best course of action here is to put a comment above this line: `// TODO: This ClosePath check does not handle all cases, fix this soon` and then we can go ahead and merge your PR right away.

Keavon (approved):

otdavies (approved):
I don't see anything concerning here - I agree with these changes since it appears that closed was used incorrectly prior to these changes. Good work.

The TODO comment gives enough context to resolve in future situations.

PR #512, opened 2022-02-05
Implement the Spline Tool

### LINKED ISSUE #496: Implement the Spline Tool
If Kurbo provides an easy way to implement splines, or there's a simple algorithm to be found with a little research for converting a set of points that would form a spline into a cubic bezier curve approximation (perhaps [this](https://math.stackexchange.com/questions/3770662/convert-cubic-spline-to-b%C3%A9zier-curve-and-get-control-points) is a good starting point), then it would be relatively easy to take the Pen Tool's code and adapt it into code for the Spline Tool. The tool should work by drawing points and the resulting curve should smoothly go through each point. Bonus points if the user can choose on tool's options bar if they want interpolation (smoothly goes through each point) or approximation (smoothly fits the set of points but doesn't usually go through them).

pkupper:
Basic Spline Tool implementation. Works just like the Pen Tool but creates a smooth bezier curve through the given points.
Algorithm used to determine control points: https://www.particleincell.com/2012/bezier-splines/

Closes #496


editor/src/viewport_tools/tools/spline.rs
```diff
@@ -0,0 +1,270 @@
...
+					responses.push_back(make_operation(data, tool_data, true));
+
+					Drawing
+				}
+				(Drawing, DragStop) => {
+					let snapped_position = data.snap_handler.snap_position(responses, input.viewport_bounds.size(), document, input.mouse.position);
+					let pos = transform.inverse().transform_point2(snapped_position);
+
+					// TODO: introduce comparison threshold when operating with canvas coordinates (https://github.com/GraphiteEditor/Graphite/issues/100)
+					if data.points.last() != Some(&pos) {
```
TrueDoctor, line 177: issues #100 is already closed this should probably be adressed here

editor/src/viewport_tools/tools/spline.rs
```diff
@@ -0,0 +1,270 @@
...
+	fn update_cursor(&self, responses: &mut VecDeque<Message>) {
+		responses.push_back(FrontendMessage::UpdateMouseCursor { cursor: MouseCursorIcon::Default }.into());
+	}
+}
+
+fn remove_preview(data: &SplineToolData) -> Message {
+	Operation::DeleteLayer { path: data.path.clone().unwrap() }.into()
+}
+
+fn make_operation(data: &SplineToolData, tool_data: &DocumentToolData, show_preview: bool) -> Message {
```
TrueDoctor, line 256: more descriptive name than `make_operation`? something like `add_spline` or something like that?

graphene/src/layers/simple_shape.rs
```diff
@@ -140,4 +140,63 @@ impl Shape {
...
+			p1[n - 1] = r[n - 1] / b[n - 1];
+			for i in (0..n - 1).rev() {
+				p1[i] = (r[i] - c[i] * p1[i + 1]) / b[i];
+			}
+
+			for i in 0..n - 1 {
+				p2[i] = 2.0 * points[i + 1] - p1[i + 1];
+			}
+			p2[n - 1] = 0.5 * (points[n] + p1[n - 1]);
+
```
TrueDoctor, line 183: some comments would be very much appreciated

TrueDoctor: This could probably be reimplemented in a more functional style but I thinkt that is probably a pretty low priority rn because this pice of code will not be touched a lot

editor/src/viewport_tools/tools/spline.rs
pkupper: Addressed this TODO for all tools.

editor/src/viewport_tools/tools/spline.rs
pkupper: Done!

graphene/src/layers/simple_shape.rs
pkupper: Added some comments

Keavon (approved):
I had a chance to look over all the code in moderate detail. Everything looks really clean! I'm merging this now.

PR #518, opened 2022-02-06
Add the TextInput widget

Keavon:
Part of #508


mfish33:
Looks good to me. I really like the more explicit emit events as well as changing the on changes to arrow functions. We are one step closer to the properties panel!

mfish33 (approved):

PR #519, opened 2022-02-06
Implement the Crop Tool for artboard resizing

### LINKED ISSUE #493: Implement the Crop Tool for resizing the artboard
Right now it should be fine to just drag and draw a rectangle with the Crop Tool which becomes the new artboard bounds. <kbd>Shift</kbd> to constrain to a square. Bonus points if this can be used to create multiple artboards. When the viewport's view of the canvas is rotated, the artboard remains axis-aligned in document space, meaning it stays tilted in viewport space (unlike, say, a rectangle if you decided to draw one with the Rectangle Tool). Additional bonus points if we can have little square blue drag handles overlaid at the corners and edge-centers of the artboard so dragging those would perform resizing along those edges (no rotation though). The resulting document bounds (width, height, X position, and Y position) should each be rounded to integer values in document space.

0HyperCube:
Closes #493

TODO:

- [x] Size to square with shift and centre with alt
- [x] Snap artboards & layers to other artboards
- [x] Fix resize behaviour with a rotated canvas - we want to calculate the bounding box with the shape transformed with just that layer's transform, and then transform the resulting box.
- [x] Allow artboards to be dragged and moved.

### DIFF
--- a/editor/src/document/artboard_message_handler.rs
+++ b/editor/src/document/artboard_message_handler.rs
@@ -75,13 +75,10 @@ impl MessageHandler<ArtboardMessage, ()> for ArtboardMessageHandler {
 			}
 			ResizeArtboard { artboard, position, size } => {
 				responses.push_back(
-					ArtboardMessage::DispatchOperation(
-						DocumentOperation::SetLayerTransform {
-							path: artboard,
-							transform: DAffine2::from_scale_angle_translation(size.into(), 0., position.into()).to_cols_array(),
-						}
-						.into(),
-					)
+					ArtboardMessage::DispatchOperation(Box::new(DocumentOperation::SetLayerTransform {
+						path: artboard,
+						transform: DAffine2::from_scale_angle_translation(size.into(), 0., position.into()).to_cols_array(),
+					}))
 					.into(),
 				);
 
--- a/editor/src/document/artboard_message_handler.rs
+++ b/editor/src/document/artboard_message_handler.rs
@@ -5,7 +5,7 @@ use graphene::document::Document as GrapheneDocument;
 use graphene::layers::style::{self, Fill, ViewMode};
 use graphene::Operation as DocumentOperation;
 
-use glam::{DAffine2, DVec2};
+use glam::DAffine2;
 use serde::{Deserialize, Serialize};
 use std::collections::VecDeque;
 
--- a/editor/src/document/artboard_message_handler.rs
+++ b/editor/src/document/artboard_message_handler.rs
@@ -36,7 +36,7 @@ impl MessageHandler<ArtboardMessage, ()> for ArtboardMessageHandler {
 			},
 
 			// Messages
-			AddArtboard { top, left, height, width } => {
+			AddArtboard { position, size } => {
 				let artboard_id = generate_uuid();
 				self.artboard_ids.push(artboard_id);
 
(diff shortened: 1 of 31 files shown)


editor/src/document/artboard_message_handler.rs
```diff
@@ -73,6 +73,20 @@ impl MessageHandler<ArtboardMessage, ()> for ArtboardMessageHandler {
...
 				}
 			}
+			ResizeArtboard { artboard, position, size } => {
+				responses.push_back(
+					ArtboardMessage::DispatchOperation(
+						DocumentOperation::SetLayerTransform {
+							path: artboard,
+							transform: DAffine2::from_scale_angle_translation(size.into(), 0., position.into()).to_cols_array(),
+						}
+						.into(),
```
TrueDoctor, line 83: is this into even required here?

editor/src/document/document_message_handler.rs
```diff
@@ -226,6 +230,22 @@ impl DocumentMessageHandler {
...
+				.filter_map(|path| self.graphene_document.viewport_bounding_box(path).ok()?)
+				.collect(),
+			self.artboard_message_handler
+				.artboard_ids
+				.iter()
+				.filter(|id| Some(**id) != ignore_artboard)
+				.filter_map(|path| self.artboard_message_handler.artboards_graphene_document.viewport_bounding_box(&[*path]).ok()?)
+				.collect::<Vec<[DVec2; 2]>>(),
+		]
+		.concat()
```
TrueDoctor, line 246: would it be possible to return a chained iterator instead?

editor/src/viewport_tools/tools/crop.rs
```diff
@@ -1,30 +1,334 @@
...
+	DocumentIsDirty,
 
 	// Tool-specific messages
-	MouseMove,
+	MouseDown,
+	MouseMove {
+		axis_align: Key,
+		centre: Key,
+	},
+	MouseUp,
```
TrueDoctor, line 40: We should probably use the more general Pointer move here, the usage kinda differs throughout he codebase buf iirc we agreed that it should be PointerMove etc

editor/src/document/artboard_message_handler.rs
0HyperCube: @TrueDoctor It boxes the operation.

0HyperCube: I have converted it to a manual `Box::new` to add clarity

editor/src/document/document_message_handler.rs
0HyperCube: Sure - forgot about that :)

editor/src/document/artboard_message_handler.rs
TrueDoctor: Ah, right, always forget that

editor/src/viewport_tools/tools/crop.rs
0HyperCube: I have cleaned up in the rest of the codebase.

TrueDoctor: Awesome!

Keavon:
Giving my approval because I dig it! Feature works splendidly and is very production-ready. However I suggest we have someone review the code before merging if there is time. I won't have time to review code before alpha.

Keavon (approved):

TrueDoctor (approved):
As far as I can tell this all looks good!

PR #521, opened 2022-02-07
Set text fill color to the primary color

pkupper:
Fix issue where non-editable text has no fill after #510.

### DIFF
--- a/graphene/src/layers/text/mod.rs
+++ b/graphene/src/layers/text/mod.rs
@@ -52,7 +52,7 @@ impl LayerData for Text {
 					.collect::<String>(),
 				match self.style.fill() {
 					Some(fill) => format!("#{}", fill.color().rgba_hex()),
-					None => "#00000000".to_string(),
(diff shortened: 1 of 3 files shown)


Keavon (approved):

graphene/src/layers/text/mod.rs
```diff
@@ -43,13 +43,17 @@ impl LayerData for Text {
...
+				r#"<foreignObject transform="matrix({})" style="color: {}"></foreignObject>"#,
 				transform
 					.to_cols_array()
 					.iter()
 					.enumerate()
 					.map(|(i, entry)| { entry.to_string() + if i == 5 { "" } else { "," } })
 					.collect::<String>(),
+				match self.style.fill() {
+					Some(fill) => format!("#{}", fill.color().rgba_hex()),
+					None => "#00000000".to_string(),
```
Keavon, line 55: Why 8 `0`s instead of 6? Won't this be transparent?

PR #522, opened 2022-02-07
Experimental Overlay renderer

TrueDoctor:
Closes #

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -477,6 +477,8 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "25f1af7423d8588a3d840681122e72e6a24ddbcb3f0ec385cac0d12d24256c06"
 dependencies = [
  "cfg-if",
+ "serde",
+ "serde_json",
  "wasm-bindgen-macro",
 ]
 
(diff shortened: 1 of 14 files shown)


Keavon:
In this branch, didn't you do something with the event system to speed up overlay rendering? Could you please backport that to master?

TrueDoctor:
> In this branch, didn't you do something with the event system to speed up overlay rendering? Could you please backport that to master?

I did two changes, the first one was to use the message deduplicaton provided by the messaging system, the other one was to switch to using request animation frame.
The first one should be relatively easy to backport: https://github.com/GraphiteEditor/Graphite/pull/522/commits/2ba44922e10a0f196b29b81631a1c7bd5d39e06a

Keavon:
If you could make a PR for that when you get the chance, that would be splendid.

TrueDoctor:
> If you could make a PR for that when you get the chance, that would be splendid.

Just checkd, the first option was already fixed in https://github.com/GraphiteEditor/Graphite/pull/522#issuecomment-1105763627
the other performance improvements are non trivial. We could try to use lyon to triangulate the overlays because I had a few issues with the sdf approach in regards to non rounded corners and blending

Keavon:
Closing this for now but keeping the branch around.

Issue #490, opened 2022-01-16
Draw new shapes into the selected folder


Keavon:
If a folder is selected, new shapes created with the drawing tools should be placed into that folder. If a non-folder is selected, it should be placed into the folder which contains it (so the shape becomes a sibling of the selected layer). If multiple layers are selected (either folder or shape), it should pick the closest ancestor folder to place within (so, for example, if two sibling layers are selected, it is placed as another sibling of those; or if two sibling folders are selected, it is placed into the folder containing both of those which means it's a sibling of the two folders). We have a function for finding the deepest common parent of a selection so be sure to reuse that. Pasting copied layers already picks the target based on this logic so it can be referenced.

Keavon:
Closed by #526.

PR #526, opened 2022-02-09
place new layers based on layer selection

Wuelle:
Implements #490.
I wasn't sure where to place the new `get_new_layer_location` fn since im not too familiar with the codebase yet,
so its sitting in `editor/src/viewport_tools/tools.rs` for now. Maybe some kind of utility module would be better.

### DIFF
--- a/editor/src/viewport_tools/tool.rs
+++ b/editor/src/viewport_tools/tool.rs
@@ -272,3 +272,13 @@ pub fn update_working_colors(document_data: &DocumentToolData, responses: &mut V
 		.into(),
 	);
 }
+
+/// Calculate the path that new layers should be inserted to.
+/// Depends on the selected layers as well as their types (Folder/Non-Folder)
+pub fn get_new_layer_location(document: &DocumentMessageHandler) -> Vec<u64> {
+	// If the selected layers dont actually exist, a new uuid for the
+	// root folder will be returned
+	let mut path = document.graphene_document.shallowest_common_folder(document.selected_layers()).map_or(vec![], |v| v.to_vec());
+	path.push(generate_uuid());
+	path
+}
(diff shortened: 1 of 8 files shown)


editor/src/viewport_tools/tool.rs
```diff
@@ -272,3 +272,13 @@ pub fn update_working_colors(document_data: &DocumentToolData, responses: &mut V
...
+
+/// Calculate the path that new layers should be inserted to.
+/// Depends on the selected layers as well as their types (Folder/Non-Folder)
+pub fn get_new_layer_location(document: &DocumentMessageHandler) -> Vec<u64> {
+	// If the selected layers dont actually exist, a new uuid for the
+	// root folder will be returned
+	let mut path = document.graphene_document.shallowest_common_folder(document.selected_layers()).map_or(vec![], |v| v.to_vec());
+	path.push(generate_uuid());
+	path
+}
```
TrueDoctor, line 284: I mean you pass it a reference to the DocumentMessageHandler, the most natural place to but it would be there.
I'm not entirely sure what I think about the function name yet, It might be nice if the name somehow contained "path" instead of location, but I'll have to think about that a bit
Other than that this all looks very good!

Wuelle: Hmm makes sense. I renamed the function to `get_path_for_new_layer` and moved it into `document_message_handler.rs`.

Wuelle:
https://github.com/GraphiteEditor/Graphite/pull/526/commits/7976f3ab49207ef9bf282b91b187c0ee90652496 implements the changes suggested by @TrueDoctor and also applies the new insert functionality to the text tool (#490 didnt ask for this but i think it makes sense)

TrueDoctor (approved):
Looks good to merge

Issue #146, opened 2021-05-23
Implement the Properties panel with editable shape parameters


Keavon:
The Properties panel will let us adjust the location and properties of shapes like width, height, radius, number of n-gon points, etc. We'll need to use #147 to send from Rust the JSON layout of input widgets, then respond with changes to input widgets and update the Document accordingly. Depends on #137 and #147.

Complexity: 4
Involves: Frontend (JS), Editor (Rust), Document (Rust)

PR #527, opened 2022-02-09
Layer properties

mfish33:
Part of #146
This PR is incomplete but has the transform shape properties working. The next steps are to add styling and the rest of the modifiable properties.


editor/src/document/properties_panel_message.rs
```diff
@@ -0,0 +1,23 @@
+use crate::message_prelude::*;
+
+use serde::{Deserialize, Serialize};
+
+#[remain::sorted]
+#[impl_message(Message, DocumentMessage, PropertiesPanel)]
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub enum PropertiesPanelMessage {
+	ClearSelection,
+	MaybeDelete { path: Vec<LayerId> },
```
0HyperCube, line 10: Perhaps `Check` rather than `Maybe`

0HyperCube, line 10: Also `Delete` and then a path might be misinterpreted as deleting a layer, perhaps `clear` would be better?

editor/src/document/properties_panel_message_handler.rs
```diff
@@ -0,0 +1,342 @@
...
+						}
+						.into(),
+					)
+				}
+			}
+		}
+	}
+
+	fn actions(&self) -> ActionList {
+		actions!(ArtboardMessageDiscriminant;)
```
0HyperCube, line 149: `ArtboardMessageDiscriminant`?

frontend/src/components/widgets/inputs/NumberInput.vue
```diff
@@ -237,7 +237,7 @@ export default defineComponent({
 			// 10.25 == 2
 			// 1.23 == 1
 			// 0.23 == 0 (Reason for the slightly more complicated code)
-			const leftSideDigits = Math.max(Math.floor(value).toString().length, 0) * Math.sign(value);
+			const leftSideDigits = Math.abs(Math.max(Math.floor(value).toString().length, 0) * Math.sign(value));
```
0HyperCube, line 240: Why multiply by sign and then abs?

editor/src/layout/widgets.rs
```diff
@@ -281,3 +287,17 @@ pub struct RadioEntryData {
 	#[derivative(Debug = "ignore", PartialEq = "ignore")]
 	pub on_update: WidgetCallback<()>,
 }
+
+#[derive(Clone, Serialize, Deserialize, Derivative, Debug, PartialEq)]
+pub struct IconLabel {
+	pub icon: String,
+	#[serde(rename = "gapAfter")]
```
0HyperCube, line 294: Can this not be done in js? It is fine if not, the js constructor things confuse me ;)

frontend/src/components/widgets/WidgetRow.vue
```diff
@@ -17,13 +18,16 @@
 			<OptionalInput v-if="component.kind === 'OptionalInput'" v-bind="component.props" @update:checked="(value: boolean) => updateLayout(component.widget_id, value)" />
 			<RadioInput v-if="component.kind === 'RadioInput'" v-bind="component.props" @update:selectedIndex="(value: number) => updateLayout(component.widget_id, value)" />
 			<Separator v-if="component.kind === 'Separator'" v-bind="component.props" />
+			<TextLabel v-if="component.kind === 'TextLabel'" v-bind="component.props">{{ component.props.value }}</TextLabel>
+			<IconLabel v-if="component.kind === 'IconLabel'" v-bind="component.props" />
+			<NumberInput v-if="component.kind === 'TextInput'" v-bind="component.props" @update:value="(value: number) => updateLayout(component.widget_id, value)" />
```
0HyperCube, line 23: Should be `value: string` I believe

0HyperCube (approved):
Looks good :)

Great work on this, makes it feel like an actual graphics editor.

frontend/src/components/widgets/WidgetRow.vue
0HyperCube, line 23: You are using NumberInput for text input? Also updateLayout should be `value: number`

Keavon: This was a placeholder until I merged my TextInput a few minutes ago.

0HyperCube: Ah ok, somewhat confused by that :)

mfish33: It's hooked up now!!

editor/src/layout/widgets.rs
mfish33: Which part of it? The rename. It could be but I like having it centralized in one location.

editor/src/document/properties_panel_message_handler.rs
mfish33: oops

editor/src/layout/widgets.rs
0HyperCube: Ok

editor/src/document/properties_panel_message_handler.rs
```diff
@@ -0,0 +1,379 @@
...
+	fn update_width(self, new_width: f64) -> Self;
+	fn height(&self) -> f64;
+	fn update_height(self, new_height: f64) -> Self;
+	fn x(&self) -> f64;
+	fn update_x(self, new_x: f64) -> Self;
+	fn y(&self) -> f64;
+	fn update_y(self, new_y: f64) -> Self;
+	fn rotation(&self) -> f64;
+	fn update_rotation(self, new_rotation: f64) -> Self;
+}
```
TrueDoctor, line 28: Could you potentially split this into Mat2Utils and DAffine2 Utils?
We don't have to do this pre launch though

editor/src/document/properties_panel_message_handler.rs
```diff
@@ -0,0 +1,379 @@
...
+
+impl DAffine2Utils for DAffine2 {
+	fn width(&self) -> f64 {
+		self.transform_vector2((1., 0.).into()).length()
+	}
+
+	fn update_width(self, new_width: f64) -> Self {
```
TrueDoctor, line 35: I would expect a function called update to actually update something by taking a &mut ref

frontend/src/components/panels/Properties.vue
```diff
@@ -1,15 +1,64 @@
...
 </template>
 
-<style lang="scss"></style>
+<style lang="scss">
+.properties {
+	height: 100%;
+
+	.widget-layout {
+		flex: 1 1 100%;
+		margin: 0 4px;
```
TrueDoctor, line 18: image
When zoomed in, the margin seems to be off, is this intended?

mfish33: @Keavon ?

Keavon: Noticed that. Need to find a more robust CSS solution for the labels on the left and widgets on the right. After alpha.

editor/src/document/properties_panel_message_handler.rs
mfish33: I agree however for most of the functions it is easier to just create Daffine instead of modifying since it is copy. Is there a better name that you can suggest? It also helps in the message handler below to have a consistent call signature on the update functions.

editor/src/document/properties_panel_message_handler.rs
mfish33: I think most of this will end up being removed anyways by @0HyperCube when he updates this code to properly use the bounding boxes

editor/src/document/properties_panel_message_handler.rs
TrueDoctor: Maybe something like `update_with_transform`?
Another way of going about this would be to ditch the update mentality and just recreate the complete transform every time.
But that is technically doing more work than necessary

PR #533, opened 2022-02-10
Allow deleting artboards

0HyperCube:
You can delete selected artboards with <kbd>Delete</kbd> or <kbd>Backspace</kbd>


Keavon (approved):
Works well and code LGTM 👍

Issue #524, opened 2022-02-08
Revamp the editor initialization flow


Keavon:
The current way that things are initialized when the editor web page loads is very janky. Different things hook into different points and cause problems. We should have a single "initialize" message get sent to the editor once the page is loaded and everything that needs it can handle that one message. This should fix issues like the right side of the options bar not showing up when a new default document is created, the ruler being offset wrong at times when a new document is created, the fact that the document is never initially resized which causes the bounds on the canvas to never be sent and thus it falls back to a non-integer 100% value on the SVG element which produces antialiasing, and other issues.

PR #535, opened 2022-02-11
Properly show document bar on initialization

### LINKED ISSUE #530: Right side of options bar in a document viewport is empty on page load
The right side of the options bar with the snapping, rotation and zoom, etc. is blank when the Graphite editor is loaded. This is only true for new documents (not loaded auto-saved documents). It also shows up when you try to zoom or rotate or switch document tabs.

mfish33:
Closes #530


mfish33:
This is a temporary fix until #524 lands

Keavon (approved):

PR #548, opened 2022-02-13
Reduce cage rotate tolerance

0HyperCube:
Reduce cage rotate tolerance


Keavon:
Thank you, this large tolerance had been getting in my way.

Keavon (approved):

Issue #529, opened 2022-02-10
Add additional layer-specific sections to Properties panel


Keavon:
- [x] Fill
- [x] Stroke
- [x] Text (stretch goal): font size, text (needs textarea widget in #508), ideally also boldness, italics, line spacing, and font

PR #551, opened 2022-02-14
Can set stroke and fill on text and shapes

mfish33:
Part Of #529

This implementation is prior to the upcoming color picking widget and could act as a stop gap


editor/src/document/properties_panel_message_handler.rs
```diff
@@ -429,25 +454,77 @@ fn node_section_transform(layer: &Layer) -> LayoutRow {
...
+			}],
+		},
+		Fill::LinearGradient(gradient) => LayoutRow::Section {
+			name: "Fill".into(),
+			layout: vec![
+				LayoutRow::Row {
+					name: "".into(),
+					widgets: vec![
+						WidgetHolder::new(Widget::TextLabel(TextLabel {
+							value: "0%".into(),
```
Keavon, line 485: I've been thinking about the best way to label this and I came up with `Gradient: 0%` and `Gradient: 100%`.

Keavon (approved):
Awesome! Now it's possible to basically do whatever you need to do in Graphite. The possibilities are limitless! Now we just need actual artists to make some beautiful stuff :)

TrueDoctor:
Okay, finally got to the bottom of the issue, the callback not being an FnMut and having a lifetime of `'static` certainly makes this a bit tricky. I guess other than self referential structs there isn't really a way to satisfiy these constraints

PR #552, opened 2022-02-14
Add documentation to many parts of the Rust codebase

Wuelle:
I added lots of documentation, mostly for `graphite-graphene::layers`. There is definitely still a lot of room for improvement but its better than nothing. I also added the `From<ShapeLayer>`, `From<TextLayer>` and `From<FolderLayer>` impls for `Layer` that i talked about in the discord.
 (Some functions are documented even though their name is self-explanatory - because it makes the documentation generated by `rustdoc` easier to read)


graphene/src/intersection.rs
```diff
@@ -10,14 +10,17 @@ use glam::{DAffine2, DMat2, DVec2};
 use kurbo::{BezPath, CubicBez, Line, ParamCurve, ParamCurveExtrema, PathSeg, Point, QuadBez, Rect, Shape, Vec2};
 
 #[derive(Debug, Clone, Default, Copy)]
+/// A quadriliteral defined by four vertices.
```
Keavon, line 13: "rectangle" would be clearer

graphene/src/error.rs
```diff
@@ -1,6 +1,7 @@
 use super::LayerId;
 use crate::boolean_ops::BooleanOperationError;
 
+/// A representation of different errors that can occur when using graphene.
```
Keavon, line 4: Nitpick: Graphene should be capitalized

graphene/src/intersection.rs
Wuelle: hmmm is it a rectangle? i dont think a quad necessarily has 90° corners

Keavon: Oh, you're right. My bad. Then I'd suggest calling it "quad" then since the full word, quadrilateral, is not very common. (It's also misspelled at the moment, but easier to just shorten it to its more common form.)

graphene/src/error.rs
Wuelle: done.

TrueDoctor (commented):
Really great progress so far!

graphene/src/intersection.rs
```diff
@@ -74,6 +83,7 @@ pub fn intersect_quad_bez_path(quad: Quad, shape: &BezPath, filled: bool) -> boo
 	get_arbitrary_point_on_path(&shape).map(|shape_point| quad.path().contains(shape_point)).unwrap_or_default()
 }
 
+/// Returns any point on `path`.
```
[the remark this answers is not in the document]
Wuelle: I didnt want to get any more specific because there is no **guarantee** that in the future, it will always be the first point on the path, even though it is right now. Would
```
/// Returns a point on `path`.
/// This function will usually return the first point from the path's first segment,
/// but callers should not rely on this behaviour.
``` 
be better?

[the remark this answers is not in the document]
TrueDoctor: Sounds good :+1:

Wuelle:
Hm forgot to comment again so i'll do it now: I added all your suggestions, re-review please?

TrueDoctor:
From a technical point of view this all looks good, will wait for @Keavon to give his approval after spell checking + linting the text

Keavon:
> From a technical point of view this all looks good, will wait for @Keavon to give his approval after spell checking + linting the text

Sure, I can review it for that sort of thing pretty quickly if I don't have to actually think about the logic of it. Let me get started shortly.

TrueDoctor (approved):
We should probably reslove the few leftover todos at some point but I think it would make a lot of sense to merge this pr, thoughts @Keavon?

Keavon:
My bad! I totally forgot about this. I was supposed to read through it, like, a month ago. I'm terribly sorry for dropping the ball on this! I'll get this done today.

Wuelle:
All good, im currently pretty busy irl anyways. No need to hurry

Keavon:
FINALLY done. So sorry for the wait. GDC then school starting got in the way of my last promised time to review it. But now I'm merging it. I guess I've found one of the types of things I am prone to procrastinate: code reviewing lots of non-code. Thank you for the contribution and patience!

PR #557, opened 2022-02-16
Fix Crash When Shape Does Not Have Fill

mfish33:
Part of #529

### DIFF
--- a/editor/src/document/properties_panel_message_handler.rs
+++ b/editor/src/document/properties_panel_message_handler.rs
@@ -258,16 +258,16 @@ fn register_layer_properties(layer: &Layer, responses: &mut VecDeque<Message>) {
 			vec![]
 		}
 		LayerDataType::Shape(shape) => {
-			vec![
-				node_section_transform(layer),
-				node_section_fill(&shape.style.fill()),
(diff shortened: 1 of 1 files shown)


editor/src/document/properties_panel_message_handler.rs
```diff
@@ -258,16 +258,16 @@ fn register_layer_properties(layer: &Layer, responses: &mut VecDeque<Message>) {
...
+				vec![node_section_transform(layer), fill_layout, node_section_stroke(&shape.style.stroke().unwrap_or_default())]
+			} else {
+				vec![node_section_transform(layer), node_section_stroke(&shape.style.stroke().unwrap_or_default())]
+			}
 		}
 		LayerDataType::Text(text) => {
 			vec![
 				node_section_transform(layer),
-				node_section_fill(&text.style.fill()),
+				node_section_fill(&text.style.fill()).expect("Text should have fill"),
```
Keavon, line 270: Why should text always have fill? Or does that mean something else (which could perhaps be clarified)?

Keavon:
If it's not too much extra work, can you also add a checkbox next to the fill and stroke colors to enable/disable their color? In the future, the color widget will let you pick "no color" as a color choice.

Keavon (approved):

PR #564, opened 2022-02-21
Add embedable images

### LINKED ISSUE #188: Placing bitmap images in the document
We need to add support for the SVG [<image>](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image) tag to let users drag in images. Example use case: create a meme by dragging in an image and using the Text Tool to add text over the top.

This should work by dragging an image into the editor or pasting from the clipboard.

0HyperCube:
Closes #188 

This is really slow for large images as we are cloning the base64 string every render then sending it to js.

TODO:

- [x] Convert to blob url
- [x] <kbd>Ctrl</kbd> + <kbd>V</kbd> to paste (I couldn't get this to work on the svg Edit: Turns out we have prevent default on the key up event that was blocking it)
- [x] Add properties
- [x] Rename to image
- [x] Properly handle bounding box
- [x] Remove logging

image


Keavon:
Sweet! To fix the base64 slowness, you can try sending the URL to the frontend just once and turn it into a [blob URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL)` then reference the blob URL as the image path.

Keavon:
One neat thing I just discovered: this even works with SVGs! So maybe a better name than "Bitmap" would be "Image". I can also provide you with an icon for this layer type.

Keavon:
@0HyperCube I added an image node type icon for you in 97eca6d99f338e959c30de8286b5d09977083bd3 :)

Keavon:
Copying and pasting image layers causes the new image layers to be zero in size, or somehow otherwise invisible with a zero-sized bounding box.

0HyperCube:
@Keavon Forgot to send the image data to the frontend on paste (which sends back both the blob url and the image dimensions). That should now be resolved.

Keavon:
@0HyperCube awesome. Are we good to rebase and then begin code review?

0HyperCube:
@Keavon Sure, I will just try and fix the failing test.

frontend/src/components/panels/Document.vue
```diff
@@ -312,6 +328,23 @@ export default defineComponent({
 			if (rulerHorizontal) rulerHorizontal.handleResize();
 			if (rulerVertical) rulerVertical.handleResize();
 		},
+		async pasteFile(e: DragEvent) {
```
Keavon, line 331: Does this need to be `async`? I don't see any `await` prefixes in the function implementation. Excuse me if I'm missing something obvious which is probably the case.

frontend/src/components/panels/Document.vue
```diff
@@ -312,6 +328,23 @@ export default defineComponent({
...
+			const { dataTransfer } = e;
+			if (!dataTransfer) return;
+			e.preventDefault();
+
+			for (let index = 0; index < dataTransfer.items.length; index += 1) {
+				const item = dataTransfer.items[index];
```
Keavon, line 337:
```js
dataTransfer.items.forEach((item) => {
```
would be preferred

frontend/src/components/panels/Document.vue
```diff
@@ -457,6 +499,19 @@ export default defineComponent({
...
 		this.editor.dispatcher.subscribeJsMessage(TriggerViewportResize, this.viewportResize);
 
+		this.editor.dispatcher.subscribeJsMessage(UpdateImageData, (updateImageData) => {
+			updateImageData.image_data.forEach((element) => {
+				// Using updateImageData.image_data.buffer returns undefined for some reason?
+				const blob = new Blob([new Uint8Array(element.image_data.values()).buffer], { type: element.mime });
+
+				const url = URL.createObjectURL(blob);
+
+				createImageBitmap(blob).then((i) => {
```
Keavon, line 509: Nitpick: `img` or `image` would avoid any possibility of confusion with the `i`ndex variable usually named `i`.

frontend/src/components/panels/LayerTree.vue
```diff
@@ -75,8 +75,8 @@
 					>
 						<LayoutRow class="layer-type-icon">
 							<IconLabel v-if="listing.entry.layer_type === 'Folder'" :icon="'NodeFolder'" title="Folder" />
-							<IconLabel v-else-if="listing.entry.layer_type === 'Image'" :icon="'NodeImage'" title="Path" />
-							<IconLabel v-else-if="listing.entry.layer_type === 'Shape'" :icon="'NodeShape'" title="Path" />
+							<IconLabel v-else-if="listing.entry.layer_type === 'Image'" :icon="'NodeImage'" title="Image" />
+							<IconLabel v-else-if="listing.entry.layer_type === 'Shape'" :icon="'NodeShape'" title="Shape" />
```
Keavon, line 79: Good catch

frontend/src/lifetime/input.ts
```diff
@@ -208,6 +212,33 @@ export function createInputManager(editor: EditorState, container: HTMLElement,
 		}
 	};
 
+	const onPaste = (e: ClipboardEvent): void => {
+		const dataTransfer = e.clipboardData;
+		if (!dataTransfer) return;
+		e.preventDefault();
+
+		for (let index = 0; index < dataTransfer.items.length; index += 1) {
+			const item = dataTransfer.items[index];
```
Keavon, line 221: `.forEach()` preferred

frontend/wasm/src/api.rs
```diff
@@ -17,6 +17,7 @@ use editor::Color;
 use editor::Editor;
 use editor::LayerId;
 
+use graphene::Operation;
```
Keavon, line 20: Please move one line up, into the previous section (which is for first-party crates).

frontend/wasm/src/api.rs
```diff
@@ -384,19 +385,19 @@ impl JsEditorHandle {
...
 	pub fn copy(&self) {
-		let message = PortfolioMessage::Copy { clipboard: Clipboard::User };
+		let message = PortfolioMessage::Copy { clipboard: Clipboard::Device };
 		self.dispatch(message);
 	}
 
-	/// Paste selected layers
-	pub fn paste(&self) {
-		let message = PortfolioMessage::Paste { clipboard: Clipboard::User };
+	/// Paste layers
```
Keavon, line 398: Is this comment still accurate? Perhaps if you can describe what "serialized data" means for the informed reader, that would be helpful.

graphene/src/layers/layer_info.rs
```diff
@@ -24,6 +26,7 @@ impl LayerDataType {
 			LayerDataType::Shape(s) => s,
 			LayerDataType::Folder(f) => f,
 			LayerDataType::Text(t) => t,
+			LayerDataType::Image(b) => b,
```
Keavon, line 29: Nitpick: `b` rather than `i` is inconsistent.

graphene/src/layers/layer_info.rs
```diff
@@ -32,6 +35,7 @@ impl LayerDataType {
...
```
Keavon, line 38: And here.

graphene/src/document.rs
```diff
@@ -284,7 +285,7 @@ impl Document {
 					}
 				}
 			}
-			LayerDataType::Text(_) => layer.cache_dirty = true,
+			_ => layer.cache_dirty = true,
```
Keavon, line 288: Does this just make the whole match statement's various `layer.cache_dirty = true` statements redundant, rather than just setting it after the match statement? I'm not following the logic super closely but perhaps you can take a look if what I'm saying is accurate or not.

graphene/src/document.rs
```diff
@@ -690,6 +704,14 @@ impl Document {
 				self.mark_as_dirty(&path)?;
 				Some([vec![DocumentChanged], update_thumbnails_upstream(&path)].concat())
 			}
+			Operation::SetImageBlobUrl { path, blob_url, dimentions } => {
```
Keavon, line 707: Typo: `dimensions`

frontend/wasm/src/api.rs
```diff
@@ -490,6 +491,20 @@ impl JsEditorHandle {
 		self.dispatch(message);
 	}
 
+	/// Sends the blob url generated by js
+	pub fn set_image_blob_url(&self, path: Vec<LayerId>, blob_url: String, width: f64, height: f64) {
+		let dimentions = (width, height);
```
Keavon, line 496: Typo: `dimensions`

Keavon (approved):
Looking great! Just a few review comments for you. Feel free to merge when those are addressed. This was a great feature to have when making the Yosemite Valley artwork!

editor/src/communication/dispatcher.rs
```diff
@@ -64,6 +64,12 @@ impl Dispatcher {
 				#[remain::unsorted]
 				NoOp => {}
 				Frontend(message) => {
+					// Image data should be immediatly handled
+					if let FrontendMessage::UpdateImageData { .. } = message {
+						self.responses.push(message);
```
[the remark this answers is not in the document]
0HyperCube: No - it returns from the function in this case.The check needs to be before pushing the message to prevent a clone.

frontend/src/components/panels/Document.vue
0HyperCube: A DataTransferItemList does not have a `forEach` method for some reason: https://developer.mozilla.org/en-US/docs/Web/API/DataTransferItemList

Keavon: Can you wrap it in `Array.from(...)`?

PR #569, opened 2022-03-15
Fix type error in Brave browser

TrueDoctor:
When using brave with the security shield turned on, the keyboard lock api is blocked which was causing a null type error.
The fix was tested to confirm it resolves the issue.


Keavon (approved):
So Brave apparently *declares* it was a valid property but makes it null or undefined? I hate this. Why.

Thanks for figuring this out and fixing it!

PR #573, opened 2022-04-02
Fix panic dialogue when handling a message

### LINKED ISSUE #574: The editor sometimes panics without displaying a crash dialog

0HyperCube:
The crash dialogue is not triggered when a panic happens when handling a message because the `instances` refcell is borrowed and so cannot be borrowed again in order to get the `JsEditorHandle` and display the dialogue. This caused the panic handler to panic.

This MR splits `JsEditorHandle`s and rust `Editor`s into separate refcells to fix this (testable currently by pressing delete while drawing with pen tool).

Fixes #574

### DIFF
--- a/frontend/src/lifetime/errors.ts
+++ b/frontend/src/lifetime/errors.ts
@@ -71,11 +71,12 @@ function githubUrl(panicDetails: string): string {
 		${browserVersion()}, ${operatingSystem()}
 
 		**Stack Trace**
-		Copied from the crash dialog in the Graphite Editor:`;
+		Copied from the crash dialog in the Graphite Editor:
+	`;
 
 	body += "\n\n```\n";
-	body += panicDetails;
-	body += "```";
+	body += panicDetails.trimEnd();
+	body += "\n```";
 
 	const fields = {
 		title: "[Crash Report] ",
--- a/frontend/src/lifetime/errors.ts
+++ b/frontend/src/lifetime/errors.ts
@@ -52,7 +52,7 @@ export function initErrorHandling(editor: EditorState, dialogState: DialogState)
 function githubUrl(panicDetails: string): string {
 	const url = new URL("https://github.com/GraphiteEditor/Graphite/issues/new");
 
-	const body = stripIndents`
+	let body = stripIndents`
 		**Describe the Crash**
 		Explain clearly what you were doing when the crash occurred.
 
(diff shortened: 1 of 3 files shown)


Keavon (approved):
Elegant fix, thank you for investigating this!

frontend/src/lifetime/errors.ts
```diff
@@ -71,12 +71,11 @@ function githubUrl(panicDetails: string): string {
...
-		Copied from the crash dialog in the Graphite Editor:
+		Copied from the crash dialog in the Graphite Editor:`;
 
-		\`\`\`
-		${panicDetails}
-		\`\`\`
-		`;
+	body += "\n\n```\n";
+	body += panicDetails;
+	body += "```";
```
Keavon, line 78: Missing a `\n` between `panicDetails` and the three backticks.

frontend/src/lifetime/errors.ts
```diff
@@ -71,12 +71,11 @@ function githubUrl(panicDetails: string): string {
 		${browserVersion()}, ${operatingSystem()}
 
 		**Stack Trace**
-		Copied from the crash dialog in the Graphite Editor:
+		Copied from the crash dialog in the Graphite Editor:`;
```
Keavon, line 74: It's slightly stylistically preferable to put the closing backtick on the next line. I think that also means you can avoid the double `\n` on the next concatenation.

frontend/src/lifetime/errors.ts
0HyperCube: panicDetails has a trailing `\n`

frontend/src/lifetime/errors.ts
0HyperCube: You do trim.

frontend/src/lifetime/errors.ts
Keavon: Maybe we should `.trim()` `panicDetails` then and manually add the `\n` if there's ever inconsistency. Maybe it's even browser-dependent. This is on Chrome.

frontend/src/lifetime/errors.ts
Keavon: Gotcha, no worries then.

frontend/src/lifetime/errors.ts
0HyperCube: It does appear to be browser dependant.

Keavon:
Looks perfect now! Ready to merge when CI finishes.

Issue #579, opened 2022-04-02
Add stroke properties for line corner styling


Keavon:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linecap
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit

Ask me to make some icons for these and I can whip them up to use with the radio button input, or just use text for now.

These should be part of the stroke section of the Properties panel.

Similar task: #578

PR #582, opened 2022-04-06
Add additional stroke properties

### LINKED ISSUE #578: Add stroke properties for dashed lines
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dasharray
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-dashoffset

These should be part of the stroke section of the Properties panel.

Similar task: #579

0HyperCube:
Closes #578 & #579

### DIFF
--- a/editor/src/document/properties_panel_message_handler.rs
+++ b/editor/src/document/properties_panel_message_handler.rs
@@ -558,7 +558,7 @@ fn node_section_fill(fill: &Fill) -> Option<LayoutRow> {
 }
 
 fn node_section_stroke(stroke: &Stroke) -> LayoutRow {
-	// TODO: Cleanup
+	// We have to make multiple variables because they get moved into different closures.
 	let internal_stroke1 = stroke.clone();
 	let internal_stroke2 = stroke.clone();
 	let internal_stroke3 = stroke.clone();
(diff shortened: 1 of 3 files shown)


editor/src/document/properties_panel_message_handler.rs
```diff
@@ -564,8 +558,19 @@ fn node_section_fill(fill: &Fill) -> Option<LayoutRow> {
...
+	let internal_stroke2 = stroke.clone();
+	let internal_stroke3 = stroke.clone();
+	let internal_stroke4 = stroke.clone();
+	let internal_stroke5 = stroke.clone();
+	let internal_stroke6 = stroke.clone();
+	let internal_stroke7 = stroke.clone();
+	let internal_stroke8 = stroke.clone();
+	let internal_stroke9 = stroke.clone();
+	let internal_stroke10 = stroke.clone();
+	let internal_stroke11 = stroke.clone();
```
0HyperCube, line 572: Somewhat non ideal but I'm not sure of a better way to do this.

TrueDoctor: Could you maybe add a short comment explainig why this has to be done?
"We have to create multiple variables because they get moved in the closure" or something like that

Keavon (approved):

PR #583, opened 2022-04-07
All shapes now have a Fill in the properties panel; color inputs are now optional

0HyperCube:
All colour inputs are now optional.

Colours are not saved after being disabled.

I also (possibly) fixed the menu system when there is scrolling in the properties, and I don't think I broke anything else.

### DIFF
--- a/frontend/src/components/widgets/floating-menus/FloatingMenu.vue
+++ b/frontend/src/components/widgets/floating-menus/FloatingMenu.vue
@@ -177,8 +177,6 @@
 <script lang="ts">
 import { defineComponent, PropType, StyleValue } from "vue";
 
-import { clamp } from "@/utilities/math";
-
 import LayoutCol from "@/components/layout/LayoutCol.vue";
 
 export type MenuDirection = "Top" | "Bottom" | "Left" | "Right" | "TopLeft" | "TopRight" | "BottomLeft" | "BottomRight" | "Center";
@@ -209,7 +207,7 @@ export default defineComponent({
 		};
 	},
 	// Gets the client bounds of the elements and apply relevant styles to them
-	// TODO: Use the vue :style attribute more whilst not causing recursive updates
+	// TODO: Use the Vue :style attribute more whilst not causing recursive updates
 	updated() {
 		const workspace = document.querySelector("[data-workspace]");
 		const floatingMenuContainer = this.$refs.floatingMenuContainer as HTMLElement;
@@ -217,7 +215,7 @@ export default defineComponent({
 		const floatingMenuContent = floatingMenuContentComponent && (floatingMenuContentComponent.$el as HTMLElement);
 		const floatingMenu = this.$refs.floatingMenu as HTMLElement;
 
-		if (!floatingMenuContainer || !floatingMenuContentComponent || !floatingMenuContent || !workspace || !floatingMenu) return;
+		if (!workspace || !floatingMenuContainer || !floatingMenuContentComponent || !floatingMenuContent || !floatingMenu) return;
 
 		this.workspaceBounds = workspace.getBoundingClientRect();
 		this.floatingMenuBounds = floatingMenu.getBoundingClientRect();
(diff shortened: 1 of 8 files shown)


frontend/src/components/widgets/floating-menus/FloatingMenu.vue
```diff
@@ -214,6 +214,16 @@ export default defineComponent({
...
+		if (floatingMenuBounds.bottom + this.windowEdgeMargin <= workspaceBounds.bottom) {
+			const { top } = (this.$refs.floatingMenu as HTMLElement).getBoundingClientRect();
+			if (this.type === "Popover") {
+				floatingMenuContent.style.top = `${top + 10}px`;
+				if (this.$refs.tail) (this.$refs.tail as HTMLElement).style.top = `${top}px`;
+			} else {
+				floatingMenuContent.style.top = `${top}px`;
+			}
+		}
+
```
Keavon, line 226: Based on reading this real quick, I don't see what this change is doing. What does it solve that wasn't already implemented before? It's also better to let Vue handle the adding/removing of CSS styles with `:style="..."` instead of using a `ref` to forcibly manipulate the DOM.

frontend/src/components/widgets/inputs/OptionalInput.vue
```diff
@@ -1,7 +1,7 @@
 <template>
-	<LayoutRow class="optional-input">
+	<div class="optional-input">
 		<CheckboxInput :checked="checked" @input="(e) => $emit('update:checked', (e.target as HTMLInputElement).checked)" :icon="icon" />
-	</LayoutRow>
+	</div>
```
Keavon, line 4: Is this change necessary? It's done like this for uniformity with similar widgets and to generally avoid using `<div>`s for layout purposes (the general goal is to use divs for small styled elements rather than containers or things that are used for layout).

graphene/src/layers/style/mod.rs
```diff
@@ -115,8 +116,8 @@ impl Fill {
 		match self {
 			Self::None => Color::BLACK,
 			Self::Solid(color) => *color,
-			// TODO: Should correctly sample the gradient
```
Keavon, line 118: I think this is a merge conflict regression.

Keavon (approved):
See (and possibly address or discuss) comments first, but this is overall fine to merge when you'd like. We'll definitely do some further tweaks to the color input in the future but having this functionality exposed earlier is better and polishing can come when I have time to do some more design iterations with UX refinement in mind.

frontend/src/components/widgets/floating-menus/FloatingMenu.vue
0HyperCube: Previously, if you scroll in the properties and then open the colour picker widget, it opened in the wrong place (as below when I clicked on stroke input on the main branch build) This is because we are using `position: fixed` for the menu content ( https://github.com/GraphiteEditor/Graphite/blob/master/frontend/src/components/widgets/floating-menus/FloatingMenu.vue#L48 ). This was a quick fix based upon the dom manipulation below.
Wrong position

frontend/src/components/widgets/inputs/OptionalInput.vue
0HyperCube: Necessary to avoid this behaviour:
image

Keavon: It's better to just override the CSS property that is causing that problem.

frontend/src/components/widgets/floating-menus/FloatingMenu.vue
Keavon: Oh, gotcha. That should be fine, but please switch it to not use a `ref` and comment that code, perhaps with a TODO to have a more robust solution in the future.

Keavon:
Hmm, it seems that this branch causes a callstack overflow crash and/or crash of the entire web page when you try to use the color picker to modify the fill or stroke color of a shape by clicking the ColorInput widget's color picker square and then trying to pick a color from the HSV floating menu.

0HyperCube:
> Hmm, it seems that this branch causes a callstack overflow crash and/or crash of the entire web page when you try to use the color picker to modify the fill or stroke color of a shape by clicking the ColorInput widget's color picker square and then trying to pick a color from the HSV floating menu.

Had to switch back to using a ref for the tail style because the change event is called any time a style is recomputed and the change event updates the bounding box which triggers the style update.

Keavon:
Ah, interesting little paradox. This whole floating menu system is unfortunately quite a fragile mess and really needs a rewrite sooner than later, but that should be fine for now going back to the `ref` approach. It's certainly an improvement! Feel free to merge now.

PR #585, opened 2022-04-16
Font selection for text layers

### LINKED ISSUE #504: Changing fonts with the Text Tool
The easiest good solution is to probably just load all the fonts from Google Fonts using their [API](https://developers.google.com/fonts/docs/developer_api), however that requires an API key so maybe that should be proxied through a Cloudflare Worker (and cached to avoid too many hits on the API). During development of this feature, it should be sufficient to just go to that link above and use the live demo functionality to copy the entire contents of the API response into a file in the static assets directory of the project and use that instead of the proxied API.

The actual UI should probably work by just feeding the entire ~thousand fonts into a dropdown input widget (hopefully the performance is acceptable) sorted alphabetically, then when one is selected it can be fetched from Google Fonts and loaded into the Text Tool/text layer for shaping and rendering. During that delay while the font is downloaded, I'm not sure what the ideal thing to do is for user visibility into this. Perhaps just wait and swap in the new font once it's available, showing the old font until then.

This feature should hopefully also remove the hardcoded default font that gets embedded into the WASM memory with `include_bytes!()`, since that bloats the WASM module size. Perhaps some fallback would be useful though, however I suggest we self-host that one but sticking it into our static assets folder.

### LINKED ISSUE #508: Add text input and text area input widgets

### LINKED ISSUE #529: Add additional layer-specific sections to Properties panel

0HyperCube:
Closes #504
Closes #529
Closes #508 

- [x] Fetch fonts and put them into dropdown in the properties panel
- [x] TextArea to add to text properties
- [x] Download selected font and render.


Keavon:
Glad to see this coming along, nice work! Just some quick comments while I notice them (you're probably already aware of these, but here are some nice checkboxes to serve as reminders):
- [ ] We'll want the tool's options bar to have the font selection also
- [ ] And that means renaming the font size from "Font" to "Size"
- [ ] And in the Properties panel, we also want a Size field
Excited to see this come along!

Keavon:
This works great!

- How feasible would it be to remove the built-in font now? This would reduce our wasm build size by a couple megabytes. But I realize that means it has nothing to default to when you first start typing. Although I suppose that's no different than first picking a different font before you start typing. So that's probably the best tradeoff?
- Can we have a "Style" dropdown to pick the different weights/italics/other variants? (Both in the options bar and properties panel.)

0HyperCube:
> - How feasible would it be to remove the built-in font now? This would reduce our wasm build size by a couple megabytes. But I realize that means it has nothing to default to when you first start typing. Although I suppose that's no different than first picking a different font before you start typing. So that's probably the best tradeoff?

I have removed the font (although it was only one style and was ~200kB) and instead load it when the editor is initialised (so we don't hang if the user goes offline and then types text for the first time).

> - Can we have a "Style" dropdown to pick the different weights/italics/other variants? (Both in the options bar and properties panel.)

Done

Keavon:
- It would be great if we can apply the (specific variant of the) font to the editing mode (foreign element in the SVG).
- The term "Style" not "Variant" is more standard.
- Options bar is missing a horizontal separator between font face and its style.

We should talk about how to make the <kbd>Ctrl</kbd><kbd>B</kbd> and <kbd>Ctrl</kbd><kbd>I</kbd> shortcuts work to switch into/out of bold and italic styles. We'll need some way to detect a weight (like figuring out where `"regular"` fits into the numbered weights which is usually `400` but doesn't have to be) and italics/oblique-or-not value and then intelligently pick sane default weights for bolded and non-bolded. Mixing font weights/styles in the same line of text will get even more interesting. Not in scope for this PR though.

This branch is reminding me that I definitely also need to prioritize some work on improving the Properties panel's horizontal scrolling and layout flexbox behavior to prevent horizontal scrollbars and the mess they cause with floating menus. That'll also improve with the fix you made in #583 once merged. For now, the user can make it usable by just widening the Properties/Layer panels column.

I'll also build a Cloudflare Workers worker to proxy the API, with some caching, so you don't have to include your API key in the final build.

graphene/src/layers/text_layer/mod.rs
```diff
@@ -82,14 +90,24 @@ impl LayerData for TextLayer {
...
+
+		if intersect_quad_bez_path(quad, &self.bounding_box(&self.text, buzz_face).to_path(0.), true) {
 			intersections.push(path.clone());
 		}
 	}
 }
 
 impl TextLayer {
+	pub fn load_face<'a>(&self, font_cache: FontCache<'a>) -> Face<'a> {
+		const DEFAULT_FONT: &str = "http://fonts.gstatic.com/s/merriweather/v28/u-440qyriQwlOrhSvowK_l5OeyxNV-bnrw.ttf";
```
Keavon, line 104: I'm concerned about having to include this both here and in `Document.vue`, as well as the fact that I don't trust this URL to be valid forever in the future. I'd much prefer if we can preload the font list and then the default font (Merriweather). Does that seem feasible?

Keavon:
I'm also noticing that it takes a bit of time (maybe ~1/3 of a second) to display a populated FontInput widget, during which time the UI hangs. This happens when switching to the Text tool (so the options bar displays it) or selecting a text layer (so the Properties panel displays it). I assume this is time spent serializing and deserializing in the process of sending the data from the backend to the frontend. I wonder if there's any way we can change the architecture so that data lives readily available on the frontend, or it gets streamed to the widget on scroll, or it gets loaded asynchronously in order to not hang the main UI thread.

graphene/src/layers/text_layer/mod.rs
0HyperCube: You now just specify the font name and variant once in `fonts.ts` and it sets a rust property for the default font name.

0HyperCube:
> This branch is reminding me that I definitely also need to prioritize some work on improving the Properties panel's horizontal scrolling and layout flexbox behavior to prevent horizontal scrollbars and the mess they cause with floating menus. That'll also improve with the fix you made in https://github.com/GraphiteEditor/Graphite/pull/583 once merged. For now, the user can make it usable by just widening the Properties/Layer panels column.

I also can't seem to get the height for the textarea to be more than one row without breaking the css @Keavon.

0HyperCube:
> I'm also noticing that it takes a bit of time (maybe ~1/3 of a second) to display a populated FontInput widget, during which time the UI hangs. This happens when switching to the Text tool (so the options bar displays it) or selecting a text layer (so the Properties panel displays it). I assume this is time spent serializing and deserializing in the process of sending the data from the backend to the frontend. I wonder if there's any way we can change the architecture so that data lives readily available on the frontend, or it gets streamed to the widget on scroll, or it gets loaded asynchronously in order to not hang the main UI thread.

@Keavon The fonts and variants list is kept in js in a local variable inside `fonts.ts`. The rust only sends and receives strings for the variant, the font name and the file path. I would assume the slowness comes from adding ~8400 dom nodes but I'm not really sure what to do about it.

0HyperCube:
> We should talk about how to make the <kbd>Ctrl</kbd><kbd>B</kbd> and <kbd>Ctrl</kbd><kbd>I</kbd> shortcuts work to switch into/out of bold and italic styles. We'll need some way to detect a weight (like figuring out where `"regular"` fits into the numbered weights which is usually `400` but doesn't have to be) and italics/oblique-or-not value and then intelligently pick sane default weights for bolded and non-bolded. Mixing font weights/styles in the same line of text will get even more interesting. Not in scope for this PR though.

In the font list, the regular variant is always called `regular` and italic always ends in `italic` with regular italic just being `italic`. [CSS always maps bold to 700 and normal to 400](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values) so I don't think we should complicate it further as long as we don't stop the user from choosing any variant. We can fake italic by applying a skew as well as faking bold if the variant does not exist. Mixed styles is just a case of loading a new font and laying out the text span in that font but starting from the last position, although we also need to handle if there are fonts with different line heights on the same line. The mixed styles can be stored as `Vec<{span_content: String, font: String}>`. Making this work with content editable whilst you are typing would be challenging though.

graphene/src/layers/text_layer/mod.rs
Keavon: Splendid!

Keavon:
> I also can't seem to get the height for the textarea to be more than one row without breaking the css

I was working on improving that by making TextAreaInput reuse FieldInput (which I extended to use both <input> and <textarea>), with CSS to handle proper visual display. However I broke the event handler functionality and didn't have time to debug it. I can push that if you'd be willing to help find the cause of the event handling no longer working.

> The fonts and variants list is kept in js in a local variable inside `fonts.ts`. The rust only sends and receives strings for the variant, the font name and the file path. I would assume the slowness comes from adding ~8400 dom nodes but I'm not really sure how to profile it or what to do about it.

Indeed, I profiled this just now and it looks like it is mostly a bunch of Vue functions running. That'll be something to improve in the future. I also would love the FontInput widget to display previews of the fonts someday, although that would require a lot of loading. To reduce the loading, we'd make it load only the ones visible at the moment, and we can use that same idea to only populate the DOM elements for the ones visible.

> In the font list, the regular variant is always called `regular` and italic always ends in `italic` with regular italic just being `italic`. [CSS always maps bold to 700 and normal to 400](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#values) so I don't think we should complicate it further as long as we don't stop the user from choosing any variant. We can fake italic by applying a skew as well as faking bold if the variant does not exist. Mixed styles is just a case of loading a new font and laying out the text span in that font but starting from the last position, although we also need to handle if there are fonts with different line heights on the same line. The mixed styles can be stored as `Vec<{span_content: String, font: String}>`. Making this work with content editable whilst you are typing would be challenging though.

It looks like you are right about this, it's more standardized than I realized. (Very thin fonts on Google Fonts, [like this one](https://fonts.google.com/specimen/Wire+One), which only have one style should maybe be weight 100 in reality, but it still lists it as 400/regular.) So the code should be pretty simple. Replace `"regular"` with `"400"` and `"italic"` with `"400italic"`, then find the ones ending in `"italic", and use 400 for regular and 700 for bold. <kbd>Ctrl</kbd><kbd>B</kbd> switches the current weight to 700 if currently lower than 700 or to 400 if current at or above 700. <kbd>Ctrl</kbd><kbd>I</kbd> switches from the non-italics version to the italics version of the closest weight to the current weight.

graphene/src/layers/layer_info.rs
```diff
@@ -86,11 +89,11 @@ pub trait LayerData {
 	/// let quad = Quad::from_box([DVec2::ZERO, DVec2::ONE]);
 	/// let mut intersections = vec![];
 	///
-	/// shape.intersects_quad(quad, &mut vec![shape_id], &mut intersections);
+	/// shape.intersects_quad(quad, &mut vec![shape_id], &mut intersections, &Default::default());
 	///
 	/// assert_eq!(intersections, vec![vec![shape_id]]);
 	/// ```
-	fn intersects_quad(&self, quad: Quad, path: &mut Vec<LayerId>, intersections: &mut Vec<Vec<LayerId>>);
+	fn intersects_quad(&self, quad: Quad, path: &mut Vec<LayerId>, intersections: &mut Vec<Vec<LayerId>>, font_cache: &FontCache);
```
Keavon, line 96: For this and all these other functions related to transformation, it seems pretty inappropriate to have to pass the font cache here. Can the cache instead live as a reference on `struct Layer`? Presumably that means lifetimes may get tricky, or perhaps that's even illogical to make Graphene handle that. Maybe give a copy of the data to Graphene? I'm not really sure what the best solution is here. @TrueDoctor perhaps you have any thoughts?

0HyperCube: @Keavon Don't think so (unless we use something from std::sync), as we need to get mutable access to the font cache (i.e. dropping all other references) when we load a new font.

frontend/src/components/widgets/inputs/TextAreaInput.vue
```diff
@@ -0,0 +1,90 @@
...
+			v-model="inputValue"
+			:spellcheck="true"
+			:disabled="disabled"
+			@focus="() => onTextFocused()"
+			@blur="() => onTextChanged()"
+			@change="() => onTextChanged()"
+			@keydown.esc="() => onCancelTextChange()"
+		/>
+		<label v-if="label" :for="`field-input-${id}`">{{ label }}</label>
+	</LayoutRow> -->
```
Keavon, line 17: We can delete this now (I left it for reference while fixing the bug).

Keavon (approved):
I'm so pumped for this feature, it's really, really awesome.

PR #589, opened 2022-04-20
Tranform Bound Fixes

0HyperCube:
Closes #

- [x] Allow input system to handle mousedown while dragging
- [ ] Abort should actually abort
- [ ] Preserve aspect ratio should use the aspect ratio at the start of the drag

### DIFF
--- a/editor/src/input/input_mapper.rs
+++ b/editor/src/input/input_mapper.rs
@@ -45,6 +45,7 @@ impl Default for Mapping {
 			entry! {action=SelectToolMessage::PointerMove { axis_align: KeyShift, snap_angle: KeyControl, center: KeyAlt }, message=InputMapperMessage::PointerMove},
 			entry! {action=SelectToolMessage::DragStart { add_to_selection: KeyShift }, key_down=Lmb},
 			entry! {action=SelectToolMessage::DragStop, key_up=Lmb},
+			entry! {action=SelectToolMessage::DragStop, key_down=KeyEnter},
 			entry! {action=SelectToolMessage::EditLayer, message=InputMapperMessage::DoubleClick},
 			entry! {action=SelectToolMessage::Abort, key_down=Rmb},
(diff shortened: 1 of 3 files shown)


Keavon:
Works great! A couple polishing requests, if you don't mind:
- [ ] Can you also make it do Esc/RMB for dragging the shape, not just resizing it?
- [ ] When resizing from an edge rather than a corner, <kbd>Shift</kbd> should constrain aspect ratio relative to the opposite side's center

editor/src/viewport_tools/tools/shared/transformation_cage.rs
```diff
@@ -18,11 +18,22 @@ pub struct SelectedEdges {
 	bottom: bool,
 	left: bool,
 	right: bool,
+	// Aspect ratio in the form: x : 1 = width : height
```
Keavon, line 21: I think of aspect ratio as `width / height` or `height / width` (usually the former). Could you translate this form to also include a `width / height` form? It's easier for me, and possibly some other future maintainers, to understand.

Keavon (approved):

editor/src/viewport_tools/tools/shared/transformation_cage.rs
0HyperCube: If we want the height side of the ratio to be one then you obviously need to do `width/height` for the width side.

Keavon: Ah, that makes sense. Too much thinking for my 1 AM brain :)

0HyperCube: Should I state this explicitly in the comment?

Keavon: Maybe just say something like:
> Aspect ratio in the form of `width/height`, or in other words `x:1 = width:height`
That way future people (or future me) won't need to think through the logic when it's also a future 1 AM, haha.

PR #601, opened 2022-04-21
Change stroke weight from ints to floats

### LINKED ISSUE #587: Allow line weights to be floats
We currently are using `u32` line weights because `f32` and `f64` don't implement `std::hash::Hash` which is needed for the options bar UI widgets.

Hint: a starting point for going at this issue is searching the project for `"Weight"` (in quotes), changing `is_integer: false` to `true`, and then trying to go replace all the `u32` instances with `f32` and `f64`. We probably want to stick with `u64` for the UI widgets and then maybe use `f32` for line weights or maybe switch that all over to `f64` also. The hard part is finding a solution to the `Hash` trait, since oddly floats in Rust don't implement it.

Keavon:
Closes #587


graphene/src/layers/style/mod.rs
```diff
@@ -188,19 +188,19 @@ pub struct Stroke {
 	/// Stroke color
 	color: Option<Color>,
 	/// Line thickness
-	width: f32,
+	weight: f64,
```
0HyperCube, line 191: Breaking change to file format.

0HyperCube (approved):
Looks good.

graphene/src/layers/style/mod.rs
Keavon: Thanks, I'll bump it.

PR #607, opened 2022-04-22
De-duplicate Overlay render messages

TrueDoctor:
Backported from #522


Keavon:
Thanks for this. Although trying out the preview branch, it appears there are no overlays appearing at all.

TrueDoctor:
Actually I think I already applied the first fix, gona close this pr

PR #609, opened 2022-04-22
Draw the outlines of shapes on hover and selection

### LINKED ISSUE #599: Select tool should display the vector outline on hover
To help visualize the click target, when hovering over a shape (when clicking it would result in its selection) it should display the vector outline of the shape (similar to what you see with the Path tool active, but without the anchor points).

And to make it easier to click unfilled shapes (outlines), it would be helpful to also increase the click target thickness.

0HyperCube:
Closes #599


editor/src/consts.rs
```diff
@@ -27,8 +27,9 @@ pub const SCALE_SNAP_INTERVAL: f64 = 0.1;
 pub const SLOWING_DIVISOR: f64 = 10.;
 
 // Select tool
-pub const SELECTION_TOLERANCE: f64 = 1.;
+pub const SELECTION_TOLERANCE: f64 = 5.;
 pub const SELECTION_DRAG_ANGLE: f64 = 90.;
+pub const SELECTION_HOVER_WEIGHT: f64 = 3.;
```
Keavon, line 32: I think `2` would be good, `3` is a bit chonky.

Keavon:
Can you also please make it so that selected layers (in addition to their transform cage bounding box) always display the outline (regardless of whether they're hovered)?

editor/src/consts.rs
Keavon: `2` is a good weight now 👍. But it's a different weight from the Path Tool's outline, which seems to be a little thinner. Can you please rename this to `PATH_OUTLINE_WEIGHT` and make the Path Tool's outline also use it?

editor/src/viewport_tools/tools/shared/path_outline.rs
```diff
@@ -0,0 +1,115 @@
...
+use graphene::layers::layer_info::LayerDataType;
+use graphene::layers::style::{self, Fill, Stroke};
+use graphene::{LayerId, Operation};
+
+use glam::DAffine2;
+use kurbo::{BezPath, Shape};
+use std::collections::VecDeque;
+
+use crate::consts::{COLOR_ACCENT, SELECTION_OUTLINE_WEIGHT};
+use crate::message_prelude::*;
```
Keavon, line 12: Our (arbitrary, but consistent) convention is to put the local (`crate::` and `super::`) imports first, followed then by a block for our first-party imports from other crates in the project, then followed by `std::` and third-party dependencies.

Keavon (approved):

PR #611, opened 2022-04-22
Improve rendering performance by isolating canvas SVG elements in a Shadow DOM

Keavon:
This aims to reduce the time spent in the "Recalculate Style" part of the "Rendering" slice of the page framerate performance in the Chome DevTools profiler. However it does seem to come at the cost of additional time spent on "Parse HTML". All in all, it seems to be roughly faster by 10-20 FPS when the frames are taking around 250-300ms to render with a stress test showing tons of overlays.

### DIFF
--- a/frontend/src/components/panels/Document.vue
+++ b/frontend/src/components/panels/Document.vue
@@ -79,21 +79,11 @@
 							class="canvas"
 							data-canvas
 							ref="canvas"
-							:style="{ cursor: canvasCursor }"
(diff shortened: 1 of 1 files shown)


Keavon:
We should also probably isolate the thumbnails in a shadow DOM, and perhaps rate limit their updating when performance is tanking.

Keavon:
I may or may not revisit this someday and attempt to extend this to also handle the thumbnails that are potentially also adding overhead of the similar sort, and then benchmark it again to see if the results aren't a wash like they are in the current state.

PR #620, opened 2022-04-26
Tidy up the full frontend codebase and use optional chaining where possible

Keavon:
We can finally use the optional chaining operator (`?.`) in JS with WebPack 5. I read through every TS and Vue file's JS code and made stylistic changes to employ optional chaining and other general tidiness improvements.

Code reviewers: please read through the diffs and carefully double-check my logic to ensure things are functionally equivalent. (De Morgan's Law is helpful here.)


frontend/src/dispatcher/js-dispatcher.ts
```diff
@@ -30,26 +31,26 @@ export function createJsDispatcher() {
 			return;
 		}
 
+		// TODO: Provide an explanatory comment here
```
0HyperCube, line 34: // Checks if the provided `messageConstructor` is a class extending `JsMessage`. All classes inheriting from `JsMessage` will have a static readonly `jsMessageMarker` which is true.

0HyperCube (approved):
Looks good, other than the comments to add comments in the js-dispatcher. I have (hopefully) added some useful comments there. I can't spot any broken functionality either.

frontend/src/components/widgets/inputs/NumberInput.vue
```diff
@@ -162,11 +162,13 @@ export default defineComponent({
...
 				Callback: (): void => {
-					if (direction === "Increase" && this.incrementCallbackIncrease) this.incrementCallbackIncrease();
-					if (direction === "Decrease" && this.incrementCallbackDecrease) this.incrementCallbackDecrease();
+					if (direction === "Increase") this.incrementCallbackIncrease?.();
+					if (direction === "Decrease") this.incrementCallbackDecrease?.();
 				},
 				None: (): void => undefined,
-			}[this.incrementBehavior]());
+			};
+			const action = actions[this.incrementBehavior];
```
mfish33, line 170: Big win of readability over the old version. I remember going huh when first reading it

frontend/src/dispatcher/js-dispatcher.ts
```diff
@@ -30,26 +31,26 @@ export function createJsDispatcher() {
...
 
-		const isJsMessageConstructor = (fn: typeof messageConstructor): fn is typeof JsMessage => "jsMessageMarker" in fn;
-		let message: JsMessage;
-		if (isJsMessageConstructor(messageConstructor)) {
-			message = plainToInstance(messageConstructor, unwrappedMessageData);
-		} else {
-			message = messageConstructor(unwrappedMessageData, wasm, instance);
-		}
+		// TODO: Provide an explanatory comment here
+		const message = messageIsConstructor ? plainToInstance(messageConstructor, unwrappedMessageData) : messageConstructor(unwrappedMessageData, wasm, instance);
```
mfish33, line 44: Is it possible to put this on multiple lines?

frontend/src/dispatcher/js-dispatcher.ts
```diff
@@ -20,6 +20,7 @@ export function createJsDispatcher() {
 	};
 
 	const handleJsMessage = (messageType: JsMessageType, messageData: Record<string, unknown>, wasm: WasmInstance, instance: RustEditorInstance): void => {
+		// TODO: Provide an explanatory comment here
 		const messageConstructor = messageConstructors[messageType];
```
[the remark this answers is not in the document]
mfish33: I think it might be worth mentioning that a message constructor does not have to be a class constructor but can also be a function in the case of `DisplayDocumentLayerTreeStructure`. On second thought it might make sense to change `messageConstructor` to `messageMaker` since the returned type is of the map is `MessageMaker`

frontend/src/dispatcher/js-messages.ts
```diff
@@ -496,16 +498,14 @@ function createWidgetLayout(widgetLayout: any[]): LayoutRow[] {
 				return { widget_id, kind, props };
 			});
 
-			return {
-				name: rowOrSection.Row.name,
-				widgets,
-			};
+			return { name, widgets };
```
mfish33, line 501: Widget rows no longer have names so this should be removed

frontend/src/lifetime/input.ts
```diff
@@ -43,8 +43,9 @@ export function createInputManager(editor: EditorState, container: HTMLElement,
 
 		// Don't redirect user input from text entry into HTML elements
 		const { target } = e;
-		if (key !== "escape" && !(key === "enter" && e.ctrlKey) && target instanceof HTMLElement && (target.nodeName === "INPUT" || target.nodeName === "TEXTAREA" || target.isContentEditable))
+		if (key !== "escape" && !(key === "enter" && e.ctrlKey) && target instanceof HTMLElement && (target.nodeName === "INPUT" || target.nodeName === "TEXTAREA" || target.isContentEditable)) {
```
mfish33, line 46: Is it possible to put the condition on multiple lines?

frontend/src/lifetime/input.ts
```diff
@@ -314,13 +316,13 @@ function keyCodeToKey(code: string): string | null {
 		NumpadAdd: "+",
 		NumpadDecimal: ".",
 	};
-	if (code in mapping) return mapping[code];
+	if (code in MAPPING) return MAPPING[code];
```
mfish33, line 319: Do not use in since it checks the prototype chain. Instead
```ts
if (MAPPING[code]) return MAPPING[code];
```
In addition the type of mapping is probably better described as `Record<string, string | undifined>`

frontend/src/lifetime/input.ts
```diff
@@ -626,5 +628,5 @@ function isKeyPrintable(key: string): boolean {
 		"ZoomToggle",
 	];
 
-	return !allPrintableKeys.includes(key);
+	return !ALL_PRINTABLE_KEYS.includes(key);
```
mfish33, line 631: Nitpick. `ALL_PRINTABLE_KEYS` should be defined above the function since it is doing an array allocation per function call when it should just be a constant value. In addition, you could use a `Set<string>` to have O(1) lookup

frontend/src/components/widgets/floating-menus/ColorPicker.vue
```diff
@@ -161,13 +161,13 @@ export default defineComponent({
 		},
 		onPointerDown(e: PointerEvent) {
 			const saturationPicker = this.$refs.saturationPicker as typeof LayoutCol;
-			const saturationPickerElement = saturationPicker && (saturationPicker.$el as HTMLElement);
+			const saturationPickerElement = saturationPicker?.$el as HTMLElement | undefined;
```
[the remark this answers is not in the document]
Keavon: Thanks for the reminder, we want to avoid using `as` wherever possible (since `as` is basically unsafe/unchecked casting in TS).

frontend/src/dispatcher/js-dispatcher.ts
Keavon: Nope. Autoformatter will put it back.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -112,34 +112,45 @@ export default defineComponent({
...
 		},
 		selectFont(newName: string) {
-			if (this.isStyle) this.$emit("update:fontStyle", newName);
-			else this.$emit("update:fontFamily", newName);
-
-			{
-				const fontFamily = this.isStyle ? this.fontFamily : newName;
-				const fontStyle = this.isStyle ? newName : getFontStyles(newName)[0];
-				const fontFile = getFontFile(fontFamily, fontStyle);
-				this.$emit("changeFont", { fontFamily, fontStyle, fontFile });
```
otdavies, line 122: You decided to remove the ternary operator here? I felt it was much more readable before. Is this just a stylistic choice or am I missing something?

frontend/src/lifetime/input.ts
otdavies: I second this, this one is a little too long.

otdavies (commented):
Overall seems fine, I agree with @0HyperCube and @mfish33's suggestions. I am also only moderately experienced with JS/TS, thus the nuances are better left with them.

frontend/src/components/widgets/inputs/FontInput.vue
Keavon: It checked the case three times, inconsistently mixed between an `if/else` pair and two ternaries checking the same case. It felt very strange to segregate the three checks into different styles and basically interlace the checks based on the same condition. Extending the existing `if/else` block so it more clearly segregates the two possible situations and applies all the control flow for each branch in order means that the intent is less obfuscated. It also prevents bugs arising from accidentally toying with the two ternaries or the one-liner `if/else` blocks individually causing inconsistent behavior. We want to do the check exactly once and branch into the outcomes from the situation, rather than branching three times for the three outcomes.

otdavies: Fair enough. I still stand by it's readability having gone down, but I can agree with your assessment and reasoning.

PR #621, opened 2022-05-01
Snapping system improvements and refactor

### LINKED ISSUE #550: Improve the snapping system
The current approach is very MVP.

0HyperCube:
Closes #550 

ToDo:
- [x] Add snapping to point
- [x] Cleanup snapping to a shape
- [x] Snap to appropriate points and anchors of external paths.


Keavon:
Looking really, really good! Some UX feedback after playing with it:

- [ ] The fade-in of vertical/horizontal lines has always had too much tolerance, can we cut that distance in half?
- [ ] The fade-in of potential snapping anchor/handle points is very cool, but it fades into the fully opaque overlay color when at close proximity; it should fade to the partial opacity like the horizontal/vertical lines do
- [ ] When dragging an anchor point, can we possibly (without too much effort or blockage before merging #617) skip the own anchor + both handles from being candidate snapping points? And when dragging a handle, skip just the handles. Unless they're split, then just the own handle but its opposing handle which isn't moving should be a candidate. Or in simpler terms: if the anchor/handle point is actively moving, don't use its former position as a candidate snapping point.
- [ ] Ideally, when you drag a handle and snap it to its anchor, this should remove the anchor point entirely (turning a cubic into a quadratic bezier, or a quadratic bezier into a line segment). This probably will want to wait for #617 though).
- [ ] While it's cool that a point being dragged expands in radius when it gets snapped, it should be sufficient instead to just make it turn solid and fully opaque (no need to grow in size). For anchors, this should still be a (selected/filled) square not a circle.
- [ ] It'd probably too difficult for this PR, but I'd really love to be able to snap to edges, not just points.
- [ ] It's hard to tell, but I think the handle points, when being dragged and snapped to things, are snapping a few pixels vertically offset from true. Try dragging a handle point and snapping it to the bottom of a neighboring shape and notice how it's below the bottom. This might also be the case for anchor points, but instead of vertically it's some other sort of offset. Try out the snapping of anchor and control points to other shapes then zoom in to see if they are precisely located in the same exact location (down to many decimal places before floating point error comes into play).

These are just comments for things we'd eventually like to polish up. As always, I'm more than happy to merge soon and polish later if you ever prefer taking that route. It's certainly ready to shine at this stage, but polishing will just make it even sweeter.

0HyperCube:
> The proximity used for the fade-in of vertical/horizontal lines has always had too much tolerance, can we cut that distance in half?

Sure (you can tweek it further in `consts.rs` if you want)

> When dragging an anchor point, can we possibly (without too much effort or blockage before merging https://github.com/GraphiteEditor/Graphite/pull/617) skip the own anchor + both handles from being candidate snapping points?

This and the next point would involve quite a bit of messing around with the current path representation and so would cause more merge conflicts for @otdavies 

> While it's cool that a point being dragged expands in radius when it gets snapped, it should be sufficient instead to just make it turn solid and fully opaque (no need to grow in size).

Ok - this was mainly for easier debugging.

> For anchors, this should still be a (selected/filled) square not a circle.

Do you mean the overlay should always be a square? Or should there only be an overlay for handles when we are actually snapping to them?

> I'd really love to be able to snap to edges, not just points.

Do you mean finding the closest point to a curve https://pomax.github.io/bezierinfo/#projections ? Binary searching for this is probably reasonably easy but we can save this for a later PR if you want to get this one merged (as I'm quite busy this weekend).

> It's hard to tell, but I think the handle points, when being dragged and snapped to things, are snapping a few pixels vertically offset from true. Try dragging a handle point and snapping it to the bottom of a neighboring shape and notice how it's below the bottom. This might also be the case for anchor points, but instead of vertically it's some other sort of offset.

Yes, if you drag a point with the path tool, it will snap your mouse position and then add the offset dictated by how far your mouse was originally away from the centre of the point. Also would cause more merge conflicts for the path refactor.

> Ideally, the horizontal and vertical lines wouldn't go all the way across the screen (unless they're snapping to something that's off-screen), instead they would terminate at the dot which indicates their snapping target. And even more ideally, if multiple snapping candidates are aligned in an axis, we'd draw the dots for each of them at the same time and draw the line to the furthest one

Ok

Keavon:
> > For anchors, this should still be a (selected/filled) square not a circle.
> 
> Do you mean the overlay should always be a square? Or should there only be an overlay for handles when we are actually snapping to them?

Currently, when you drag around an anchor point, it is square. But if you drag it somewhere that it snaps, the anchor point becomes (or is covered up by?) the big blue circle. Once you remove the big circle, this shouldn't be a small circle, it should be just the square shape of the anchor point (which should be a solid filled blue color since it's selected). In other words, that big circle should be removed entirely when snapping anchor points and nothing special should show up for the dragged anchor point. That's also true for dragged handle points, since they are also supposed to be a filled solid blue circle and, once the big circle is removed, they should look like they already look when snapping or not snapping to another point (the only difference is that when it snaps, it teleports away from the mouse position to live on top of the snap candidate location).

0HyperCube:
@Keavon So the snap to point markers should be removed?

Keavon:
> @Keavon So the snap to point markers should be removed?

I think that is correct, yes, assuming you mean the markers that show up on the location of the point you're currently dragging once that point has teleported from under your mouse to the snap location. Because they are blocking the point itself which you're dragging. The markers are useful for showing what is the active candidate along an axis (connected by a line). The mere fact that the point you're dragging has teleported (and likely isn't exactly under your mouse) should be sufficient to convey to the user that the point is being snapped to a specific point in space.

0HyperCube:
The snap to point markers should remain when they are not the active snap target then?

Keavon:
> The snap to point markers should remain when they are not the active snap target then?

Correct (as they fade in), since they are not blocking the point at that time.

0HyperCube:
Done

Keavon (approved):

PR #629, opened 2022-05-07
Add the File > Export dialog and PNG/JPG downloading

### LINKED ISSUE #626: Export as PNG/JPG
https://stackoverflow.com/questions/3975499/convert-svg-to-image-jpeg-png-etc-in-the-browser

We should probably have a dialog that shows up and lets you pick a resolution or resolution multiplier and lets you pick the output format (SVG, PNG, JPG, possibly others...).

0HyperCube:
Closes #626

### DIFF
--- a/frontend/src/state/documents.ts
+++ b/frontend/src/state/documents.ts
@@ -44,7 +44,7 @@ export function createDocumentsState(editor: EditorState) {
 		if (!ctx) return;
 
 		// Fill the canvas with white if jpeg (does not support transparency and defaults to black)
-		if (triggerRasterDownload.mime.endsWith("jpeg")) {
+		if (triggerRasterDownload.mime.endsWith("jpg")) {
 			ctx.fillStyle = "white";
 			ctx.fillRect(0, 0, triggerRasterDownload.size.x, triggerRasterDownload.size.y);
 		}
--- a/frontend/src/state/documents.ts
+++ b/frontend/src/state/documents.ts
@@ -45,7 +45,7 @@ export function createDocumentsState(editor: EditorState) {
 
 		// Fill the canvas with white if jpeg (does not support transparency and defaults to black)
 		if (triggerRasterDownload.mime.endsWith("jpeg")) {
-			ctx.fillStyle = "rgb(255,255,255)";
+			ctx.fillStyle = "white";
 			ctx.fillRect(0, 0, triggerRasterDownload.size.x, triggerRasterDownload.size.y);
 		}
 
@@ -64,7 +64,6 @@ export function createDocumentsState(editor: EditorState) {
 
 			// Cleanup resources
 			URL.revokeObjectURL(url);
-			document.removeChild(canvas);
 		};
 		img.src = url;
 	});
--- a/frontend/src/state/documents.ts
+++ b/frontend/src/state/documents.ts
@@ -1,9 +1,9 @@
 /* eslint-disable max-classes-per-file */
 import { reactive, readonly } from "vue";
 
-import { TriggerFileDownload, FrontendDocumentDetails, TriggerFileUpload, UpdateActiveDocument, UpdateOpenDocumentsList } from "@/dispatcher/js-messages";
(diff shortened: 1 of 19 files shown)


Keavon:
It looks like, unfortunately, you'll have to do away with the unstable features before this compiles in CI and on everyone's machines. Feel free to leave a comment about restoring the code back to the simpler way once it's stabilized.

frontend/src/state/documents.ts
```diff
@@ -34,6 +34,41 @@ export function createDocumentsState(editor: EditorState) {
...
+			ctx?.drawImage(img, 0, 0, triggerRasterDownload.size.x, triggerRasterDownload.size.y);
+
+			// Convert the canvas to an image of the correct mime
+			const imgURI = canvas.toDataURL(triggerRasterDownload.mime);
+			// Download our canvas
+			downloadBlob(imgURI, triggerRasterDownload.name);
+
+			// Cleanup resources
+			URL.revokeObjectURL(url);
+			document.removeChild(canvas);
```
Keavon, line 67: Since `canvas` was never attached to the DOM, there's no need to remove it from the DOM.

frontend/src/state/documents.ts
```diff
@@ -34,6 +34,41 @@ export function createDocumentsState(editor: EditorState) {
...
+		// https://stackoverflow.com/questions/3975499/convert-svg-to-image-jpeg-png-etc-in-the-browser
+		const canvas = document.createElement("canvas");
+		canvas.width = triggerRasterDownload.size.x;
+		canvas.height = triggerRasterDownload.size.y;
+		const ctx = canvas.getContext("2d");
+		if (!ctx) return;
+
+		// Fill the canvas with white if jpeg (does not support transparency and defaults to black)
+		if (triggerRasterDownload.mime.endsWith("jpeg")) {
+			ctx.fillStyle = "rgb(255,255,255)";
```
Keavon, line 48: Nitpick: may as well use the shorter/clearer shorthand of `"white"`. Obviously this unimportant but I thought I'd mention it.

frontend/src/utilities/files.ts
```diff
@@ -11,6 +8,14 @@ export function download(filename: string, fileData: string): void {
 	element.click();
 }
 
+export function download(filename: string, fileData: string): void {
+	const type = filename.endsWith(".svg") ? "image/svg+xml;charset=utf-8" : "text/plain;charset=utf-8";
+	const blob = new Blob([fileData], { type });
+	const url = URL.createObjectURL(blob);
```
Keavon, line 14: I think we're currently forgetting to clean up with `URL.revokeObjectURL(url);`.

frontend/src/state/documents.ts
0HyperCube: It doesn't actually fail silently, an error is logged in the console.

Keavon: Oh interesting! All the more reason to remove it then 😉

Keavon:
Very cool that you added the Export Area selection! It would probably work better as a DropdownInput though. And then the first option can be called "Artwork Bounds" while the others would be called "Artboard: [ARTBOARD NAME or Untitled 1/2/3/...]".

For naming conventions, please use Title Case for labels (in general). So "Export type" -> "Export Type". Although a better name would be "File Type". Also a better name for "Resolution" would be "Scale Factor". And "Export area" could be just called "Bounds".

Lastly, could you please make it export the JPEG files as `.jpg` instead of `.jpeg`? (The three-letter extension is more common in files even though the file format itself is typically referred to by its longer name.)

Everything's looking super good!

Keavon (approved):

0HyperCube:
> It looks like, unfortunately, you'll have to do away with the unstable features before this compiles in CI and on everyone's machines. Feel free to leave a comment about restoring the code back to the simpler way once it's stabilized.

Sorry, I thought it was stabilized because it was no longer feature gated in nightly.

PR #638, opened 2022-05-10
UniqueElements structure

caleb-ad:
I think that we can make the UniqueElements structure much simpler. The main advantage of this approach is that implementing `Deref` allows us to call all the methods of `HashMap` without any indirection. For example if `elements` is a `UniqueElements` then I can use  `elements.keys()` to call `HashMap::keys()`. The drawback of the way I did it here is that we can no longer access elements by index. However I think I'm not sure why we would want to access elements both by index and by ID.

The main value of indexing is that it keeps the elements ordered. To accomplish this we can use `Vec<ElementID>`, Then access by ID can be done directly through the hashmap and access by index is done by indexing into the vector to get the ID, and then into the hashmap. Even with the extra vec I think this approach is still simpler than the current implementation.


otdavies:
I am all for simplifying this struct. It's original purpose was to be used to provide unique identification for folders / layers and was being generalized for usage on VectorShape for VectorAnchors. All of the prior exposed functions are needed, so as long as we keep them, along with how we generate the IDs originally, other changes are fine. Moving toward a single Vec<ElementID> + hashmap is fine in my book.

otdavies:
For now, after some discussion, we aren't going to merge this. These changes can be backported as needed.

PR #639, opened 2022-05-10
Add basic support for radial gradients

### LINKED ISSUE #555: Add support for radial gradients
This should involve adding a radio toggle input to the Gradient tool options bar, the Properties Panel for shapes with gradients, and updating the SVG string generator to support radial gradients.

hannahli2010:
Add support for radial gradients 
- Add a radio toggle input to the Gradient tool options bar and the Properties Panel for shapes with gradients
- Update the SVG string generator to support radial gradients.
- Click and drag defines the centre of the radial gradient and the radius respectively

Closes #555


editor/src/document/properties_panel_message_handler.rs
```diff
@@ -785,6 +785,93 @@ fn node_section_font(layer: &TextLayer) -> LayoutRow {
...
+			.into()
+		},
+		Fill::LinearGradient(_) => move |new_gradient: Gradient| {
+			PropertiesPanelMessage::ModifyFill {
+				fill: Fill::LinearGradient(new_gradient),
+			}
+			.into()
+		},
+		// This case shouldn't ever happen, will set to None
+		_ => move |_| PropertiesPanelMessage::ModifyFill { fill: Fill::None }.into(),
```
0HyperCube, line 840: I think `_ => unimplemented!()` is cleaner

graphene/src/layers/style/mod.rs
```diff
@@ -61,7 +61,7 @@ impl Gradient {
 	}
 
 	/// Adds the gradient def with the uuid specified
-	fn render_defs(&self, svg_defs: &mut String, multiplied_transform: DAffine2, bounds: [DVec2; 2], transformed_bounds: [DVec2; 2]) {
+	fn render_defs(&self, svg_defs: &mut String, multiplied_transform: DAffine2, bounds: [DVec2; 2], transformed_bounds: [DVec2; 2], fill: &Fill) {
```
0HyperCube, line 64: Perhaps it would be better to store the gradient type (currently in `gradient_tool`) as a property of the `Gradient` in the style?

0HyperCube (approved):
Looks good.

graphene/src/layers/style/mod.rs
hannahli2010: That sounds like a good idea! I'll make that change and also move the `GradientType` definition to this file instead since I think that would make more sense.

RobNadal: @Hypercube0 we tried adding GradientType as a property to the Gradient struct and a default for GradientType (set to Linear). Despite this, we saw the below error pop up on the Graphite UI while testing our changes, which seems to imply the default isn't being applied to Gradient structs loaded up from existing projects. We're not sure if this is an issue from the Default, Serialize, or Deserialize traits. 
image

0HyperCube: @RobNadal This is due to the autosaved document failing to load (due to the json format being changed). Feel free to ignore this.

Keavon: @RobNadal for future reference, that happens commonly when the document serialization format changes from what is saved in your browser between page loads by the auto-save system. You just need to close the open document tab and it'll add a new blank tab that follows the new document format. Nothing to worry about!

graphene/src/layers/style/mod.rs
```diff
@@ -86,23 +101,35 @@ impl Gradient {
...
 ///
-/// Can be None, a solid [Color], a linear [Gradient], or potentially some sort of image or pattern in the future
+/// Can be None, a solid [Color], a linear [Gradient], a radial [Gradient] or potentially some sort of image or pattern in the future
 #[repr(C)]
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum Fill {
 	None,
 	Solid(Color),
-	LinearGradient(Gradient),
+	Gradient(Gradient),
```
hannahli2010, line 132: Since we are storing the linear vs radial type in the `Gradient`, it was unnecessary to differentiate between the two here. We're open to better naming conventions!

hannahli2010: Resolving for now and rewriting this comment after the git rebase

hannahli2010, line 132: Since we are storing the linear vs radial type in the `Gradient`, it was unnecessary to differentiate between the two here. We're open to better naming conventions!

0HyperCube (approved):
Looks good

otdavies:
You should be good to merge into master once the checks pass. Changes look good and @0HyperCube approved, nicely done.

Keavon:
A little UI/UX feedback before merging.

 In the Properties panel:
image

All widgets should be on the right side of a label. "Type" or "Spread" would be reasonable name choices. (@otdavies do you have a preference between the two names, or other suggestions?)

otdavies:
> All widgets should be on the right side of a label. "Type" or "Spread" would be reasonable name choices. (@otdavies do you have a preference between the two names, or other suggestions?)

I think that **Gradient Type** or just **Type** would be a descriptive label.

Keavon:
Let's call it Gradient Type if there's room for it to fit. If not, then just Type.

otdavies:
@hannahli2010 Noticed a small issue with the gradient tool swapping between linear / radial unexpectedly.

Reproducible Steps:
1. Draw shape
2. Add linear gradient to shape
3. Set gradient to radial
4. Drag one of the gradient handles
5. Issue will present, it will swap the gradient type unexpectedly.

I assume this is unintentional, just another of the many of cases to keep an eye out for. 😄

Keavon:
Specifically, it reverts to what the tool is set to. So it does it from linear to radial also.

otdavies (approved):
UI changes look good and the gradient switch bug is resolved. I think this is good to go. You agree @Keavon?

PR #641, opened 2022-05-12
Restore the zoom to fit behaviour

0HyperCube:
#623 broke the zoom to fit all behavior with an empty document (creating a new finite document no longer centered the document in the user's view)


Keavon:
Good eye.

Keavon (approved):

PR #644, opened 2022-05-16
Fix polygon overflow

### LINKED ISSUE #612: Overflow on polygon with 256 sides
Setting the sides to 256 in the shape tool and then drawing causes an overflow panic (assuming you are running with debug asserts). This can be fixed by either casing to a larger integer (i.e. u32) or by subtracting 1 rather than adding one (giving the same result)

Androxium:
Closes #612 

Previously, setting the number of sides in the polygon to 256 (and the drawing) causes an overflow panic. To resolve this issue, we change the polygon side type to 'u32'. Also, we increased the max polygon side value.

Work done
- change the polygon side type to `u32`
- set the max polygon side value to 1000


graphene/src/operation.rs
```diff
@@ -215,8 +215,8 @@ impl Operation {
 	/// # Safety
 	/// This function reads from uninitialized memory!!!
 	/// Only use if you know what you are doing
-	unsafe fn as_slice(&self) -> &[u8] {
-		core::slice::from_raw_parts(self as *const Operation as *const u8, std::mem::size_of::<Operation>())
+	unsafe fn as_slice(&self) -> &[u32] {
+		core::slice::from_raw_parts(self as *const Operation as *const u32, std::mem::size_of::<Operation>())
```
0HyperCube, line 219: Why are we changing this?

0HyperCube (approved):
Looks good

graphene/src/operation.rs
Androxium: Just tested it again after reverting this change and feature still works as intened, committing the revert

0HyperCube (approved):
Looks good.

PR #647, opened 2022-05-17
Move layouts definitions to backend and fix Firefox overlay scrollbars

### LINKED ISSUE #398: Fix overlaid scrollbars on Firefox
Since Firefox doesn't support the non-standard WebKit overlaid scrollbars CSS properties, and its "standard" API that it does support is terrible, the scrollbars end up displacing the content which breaks the layout, especially on horizontally scrolling content. I tried integrating the Overlay Scrollbars and Simplebar JS libraries but they do some crazy things to the DOM which massively breaks our styling. We might need a custom solution, or to use JS, or find some other workaround.

### LINKED ISSUE #646: Move Document Mode, Layer Tree panel's options bar, and tool shelf buttons from Vue to Rust
Currently these layouts are defined in Vue. Move them to the new backend system for layout definitions.

Keavon:
Closes #646
Closes #398


frontend/wasm/src/api.rs
```diff
@@ -552,16 +488,13 @@ impl JsEditorHandle {
...
+		self.dispatch(message);
+		let message = ToolMessage::ActivateTool { tool_type: ToolType::Select };
 		self.dispatch(message);
-	}
-
-	// TODO(mfish33): Replace with initialization system Issue:#524
-	pub fn init_document_bar(&self) {
-		let message = PortfolioMessage::UpdateDocumentBar;
-		self.dispatch(message)
 	}
```
Keavon, line 498: @mfish33 this didn't seem to be doing anything so far as I could tell. But perhaps I missed something and caused a regression by deleting it?

editor/src/document/utility_types.rs
```diff
@@ -32,3 +32,30 @@ pub enum TargetDocument {
...
 }
+
+#[derive(PartialEq, Clone, Copy, Debug, Serialize, Deserialize)]
+pub enum DocumentMode {
+	DesignMode,
+	SelectMode,
+	GuideMode,
+}
+
+impl ToString for DocumentMode {
```
0HyperCube, line 43: Rust docs state:
> ToString shouldn't be implemented directly: Display should be implemented instead, and you get the ToString implementation for free.

editor/src/layout/widgets.rs
```diff
@@ -170,7 +188,7 @@ pub enum Widget {
 #[derive(Clone, Serialize, Deserialize, Derivative)]
 #[derivative(Debug, PartialEq, Default)]
 pub struct NumberInput {
-	pub value: f64,
+	pub value: Option<f64>,
```
0HyperCube, line 191: Why should this be an option? We appear to just unwrap it in most cases anyway.

editor/src/viewport_tools/tool.rs
```diff
@@ -65,6 +66,53 @@ impl ToolData {
...
+
+	fn register_properties(&self, responses: &mut VecDeque<Message>, layout_target: LayoutTarget) {
+		responses.push_back(
+			LayoutMessage::SendLayout {
+				layout: self.properties(),
+				layout_target,
+			}
+			.into(),
+		)
+	}
```
0HyperCube, line 113: This doesn't need to be implemented manually (as it is the same as the default implementation)

frontend/wasm/src/api.rs
```diff
@@ -552,16 +488,13 @@ impl JsEditorHandle {
...
 
-	/// Creates an empty folder at the document root
-	pub fn create_empty_folder(&self) {
-		let message = DocumentMessage::CreateEmptyFolder { container_path: vec![] };
+	// TODO: Replace with initialization system, issue #524
+	pub fn init_app(&self) {
+		// Switch away from, and back to, the Select Tool to make it display the correct hints in the status bar
+		let message = ToolMessage::ActivateTool { tool_type: ToolType::Path };
+		self.dispatch(message);
+		let message = ToolMessage::ActivateTool { tool_type: ToolType::Select };
```
0HyperCube, line 499: Can we not use `ToolMessage::UpdateHints` directly?

frontend/wasm/src/api.rs
```diff
@@ -591,17 +525,24 @@ pub fn intentional_panic() {
...
 pub fn file_save_suffix() -> String {
 	FILE_SAVE_SUFFIX.into()
 }
 
-/// Get the constant FILE_SAVE_SUFFIX
+/// Get the constant `FILE_SAVE_SUFFIX`
 #[wasm_bindgen]
 pub fn graphite_version() -> String {
 	GRAPHITE_DOCUMENT_VERSION.to_string()
 }
```
0HyperCube, line 547: Comment is incorrect.

editor/src/layout/widgets.rs
Keavon: `None` equates to a dash in the NumberInput GUI.

frontend/wasm/src/api.rs
Keavon: Comment was out of date, the real issue is tool initialization. I've updated the comment like so:

> // Switch away from, and back to, the Select tool to make it initialize the active tool correctly
> // TODO: Fix the root cause of this issue to avoid this hacky workaround

If you have any ideas for a quick fix, that'd be great. Or we can leave it as is for now and deal with the TODO eventually.

editor/src/document/utility_types.rs
Keavon: Thanks for letting me know!

Keavon:
And @0HyperCube are we good to merge now, or any other changes you noticed or you'd like to poke into closer?

frontend/src/dispatcher/js-messages.ts
```diff
@@ -320,19 +293,11 @@ export class UpdateImageData extends JsMessage {
...
-	readonly factor!: number;
-}
-
-export class UpdateCanvasRotation extends JsMessage {
-	readonly angle_radians!: number;
-}
-
 export type BlendMode =
 	| "Normal"
 	| "Multiply"
```
Keavon, line 303: I don't think we actually need this definition on the frontend. Or in other words, `LayerPanelEntry` doesn't need a lot of its metadata like `blend_mode` and `opacity`. Should we just `#[serde(skip)]` these and remove them from `js-messages.ts`? @0HyperCube

0HyperCube: Sure

0HyperCube (approved):
Looks good 👍🏽

PR #650, opened 2022-05-19
Add font menu previews and virtual scrolling

### LINKED ISSUE #596: Preview fonts
When choosing a font there is currently no way to see if it looks nice other than selecting it which is slow. There are three ways I can think of to preview fonts. The first one is to show the font name in the font itself. The second one is to show the a sample next to the font name and the third one is to temporarily chance the font of the text when hovering over a font and reverting back if you click away from the font selection dropdown.

0HyperCube:
Closes #596
fontpreview


Keavon:
I'm heavily refactoring these files right now so hopefully this won't be too hard to port over the changes once I'm ready to merge mine.

0HyperCube:
Oh right sorry.

Keavon:
Nothing to apologize for! Just giving you an update on the status. This looks like an exciting feature!

Keavon:
For the record, but not relevant in this PR:

My plan for the design of the FontInput widget will involve using a single component that encapsulates two DropdownInputs. (And from a UI/UX standpoint, we'll tailor it more closely to provide useful font-related things, like a search box to filter the list and a button to upload fonts).

By making the one component use the two DropdownInput components as children, we can avoid duplicating code used by the DropdownInput component and avoid needing to check `this.isStyle` to change behavior depending on the intention of this FontInput component instance.

Just mentioning this for future reference.

Keavon:
It would also be a cool feature to display the font weight as a preview.

Keavon:
If you get a chance to fix the merge conflicts, I think this should be pretty quick to get merged since it has a great UX/behavior already. (I'll just skim over the code but there's probably nothing major to change.)

editor/src/dialog/dialogs/export_dialog.rs
```diff
@@ -86,6 +86,7 @@ impl PropertyHolder for Export {
 			WidgetHolder::new(Widget::DropdownInput(DropdownInput {
 				selected_index: Some(index as u32),
 				entries,
+				interactive: false,
```
Keavon, line 89: What's this for?

editor/src/document/document_message_handler.rs
```diff
@@ -767,6 +768,7 @@ impl DocumentMessageHandler {
 					entries: blend_mode_menu_entries,
 					selected_index: blend_mode.map(|blend_mode| blend_mode as u32),
 					disabled: blend_mode.is_none() && !blend_mode_is_mixed,
+					interactive: true,
```
Keavon, line 771: True is the default.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -1,17 +1,26 @@
...
+					:class="{ active: entry === highlighted }"
+					@click="selectFont(entry.name)"
+				>
+					<link v-if="!isStyle" rel="stylesheet" :href="entry.url?.toString()" />
+
+					<span class="entry-label" :style="{ fontFamily: `${isStyle ? 'inherit' : entry.name}` }">{{ entry.name }}</span>
+				</LayoutRow>
+				<LayoutRow class="spacer" :style="{ height: `${totalHeight - endIndex * 20}px` }"></LayoutRow>
+			</LayoutCol>
+		</FloatingMenu>
```
Keavon, line 23: I'm concerned about getting rid of the usage of MenuList which aims to avoid code duplication by encapsulating the concepts of list rendering. Can you add the new features needed here directly to MenuList rather than replicating MenuList's behavior and modifying it here?

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -69,26 +65,42 @@
...
-
-			svg {
-				fill: var(--color-8-uppergray);
-			}
 		}
 	}
 
 	.menu-list .floating-menu-container .floating-menu-content {
 		max-height: 400px;
+
```
Keavon, line 73: Remove blank line please.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -125,50 +162,90 @@ export default defineComponent({
 				this.$emit("update:fontFamily", newName);
 
 				fontFamily = newName;
-				fontStyle = (await this.fonts.getFontStyles(newName))[0];
+				fontStyle = "Normal (400)";
```
Keavon, line 165: It's preferable if we can define this only in that one location of the constants file in the backend, rather than spread across the frontend and backend. What necessitated the change here? (I admittedly haven't looked very closely at what the old line was and why that old line doesn't work anymore and had to be changed to this.)

Keavon (changes requested):
Welcome back from your vacation! :)

frontend/src/components/widgets/inputs/FontInput.vue
0HyperCube: Previously, when you selected a new font, then it would select the first style (often a light one). This changes it so it loads the normal weight when you select a new font.

editor/src/dialog/dialogs/export_dialog.rs
0HyperCube: Do we want it to be true?

Keavon: This is just the artwork bounds dropdown menu, right? My mental model was that the behavior for dropdown menus always are interactive, unless there are specific reasons (like performance) where they really shouldn't be. That way users can learn from that behavior and rely on it everywhere. Unless I'm forgetting something, I don't see why the artwork bounds in this dropdown menu can't be interactive. (If I am forgetting something, please let me know.)

frontend/src/components/widgets/inputs/FontInput.vue
Keavon: Ah, I see, that's helpful.

frontend/src/components/widgets/inputs/FontInput.vue
0HyperCube: Converting the only showing visible parts of the list requires that all MenuLists have a known size.

Keavon: Can that be only a requirement when enabled via a prop?

0HyperCube: What if we wanted a search box in the floating menu for just the font input - wouldn't it make sense to separate it out?

Keavon: I still think that would be better to include in the same MenuList code. It's fine if that's a very feature-rich component with lots of props to enable the many optional features. It's better than duplicating code. Otherwise we wouldn't have MenuList at all, we'd just use FloatingMenu for the menu bar, dropdowns, etc.

0HyperCube: This makes the MenuList extremely bloated, extremely convoluted and extremely difficulty to work with.

0HyperCube: I just find it difficult to work with single vue components that can have a lot of different behaviours.

0HyperCube: I think I managed to do this but I've probably broken some things.

frontend/src/state-providers/fonts.ts
```diff
@@ -58,14 +64,13 @@ export function createFontsState(editor: Editor) {
...
 				});
-				state.fontNames = result.map((value) => value.family);
 
 				resolve(result);
 			});
 	});
 
 	return {
-		state: readonly(state) as typeof state,
+		fontNames,
```
Keavon, line 73: If we aren't returning a state, I think this file should be moved out of `/state-providers`, maybe restructured to fit within either `/io-managers` or `/utility-functions`?

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -6,44 +6,51 @@
 		:type="'Dropdown'"
 		:windowEdgeMargin="0"
 		:escapeCloses="false"
-		v-bind="{ direction, scrollableY, minWidth }"
+		v-bind="{ direction, minWidth, scrollableY: !fixedHeight && scrollableY }"
```
Keavon, line 9: Nitpick, feel free to ignore since it's so minor: it might be slightly more readable and consistent to put the `scrollableY` before the `&&` (i.e. swap the two sides of the `&&`).

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -6,44 +6,51 @@
...
-					:open="entry.ref?.open || false"
-					:direction="'TopRight'"
-					:entries="entry.children"
-					v-bind="{ defaultAction, minWidth, drawIcon, scrollableY }"
-					:ref="(ref: typeof FloatingMenu) => ref && (entry.ref = ref)"
-				/>
-			</LayoutRow>
-		</template>
+		<LayoutCol ref="scroller" :scrollableY="scrollableY && fixedHeight" @scroll="onScroll" :style="{ minWidth: fixedHeight ? `${minWidth}px` : `inherit` }">
+			<LayoutRow v-if="fixedHeight" class="spacer" :style="{ height: `${startIndex * fixedHeight}px` }"></LayoutRow>
```
Keavon, line 14: Can you please rename the `fixedHeight` prop to `virtualScrolling` and the class `spacer` to `virtual-scrolling-spacer`?

Keavon: Wait, sorry, I didn't realize `fixedHeight` is a number not a bool. Maybe `virtualScrollingTotalHeight`?

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -182,13 +191,15 @@ const MenuList = defineComponent({
...
 		scrollableY: { type: Boolean as PropType<boolean>, default: false },
+		fixedHeight: { type: Number as PropType<number>, default: 0 },
 		defaultAction: { type: Function as PropType<() => void>, required: false },
 	},
 	data() {
 		return {
 			isOpen: this.open,
 			keyboardLockInfoMessage: this.fullscreen.keyboardLockApiSupported ? KEYBOARD_LOCK_USE_FULLSCREEN : KEYBOARD_LOCK_SWITCH_BROWSER,
 			highlighted: this.activeEntry as MenuListEntry | undefined,
+			entriesStart: 0,
```
Keavon, line 202: Here too, `entriesStartVirtualScrolling` or `virtualScrollingEntriesStart` would help the reader know it's part of the virtual scrolling implementation.

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -6,44 +6,51 @@
...
-					@naturalWidth="(newNaturalWidth: number) => $emit('naturalWidth', newNaturalWidth)"
-					:open="entry.ref?.open || false"
-					:direction="'TopRight'"
-					:entries="entry.children"
-					v-bind="{ defaultAction, minWidth, drawIcon, scrollableY }"
-					:ref="(ref: typeof FloatingMenu) => ref && (entry.ref = ref)"
-				/>
-			</LayoutRow>
-		</template>
+		<LayoutCol ref="scroller" :scrollableY="scrollableY && fixedHeight" @scroll="onScroll" :style="{ minWidth: fixedHeight ? `${minWidth}px` : `inherit` }">
```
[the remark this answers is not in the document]
Keavon: Sounds good! Feel free to even copy and paste what you just wrote into a comment above that code if you think that would be valuable knowledge to included in the project.

frontend/src/components/widgets/inputs/FontInput.vue
Keavon: `MenuList.vue` seems pretty clean to me! The other excellent advantage is that you just implemented virtual scrolling, not only for fonts, but for all types of menu lists which is something we would have needed to do someday anyways.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -69,23 +58,35 @@
...
+			height: 20px;
+			align-items: center;
+			white-space: nowrap;
+			position: relative;
+			flex: 0 0 auto;
+		}
+
+		.spacer {
+			flex: 0 0 auto;
+		}
```
Keavon, line 78: If I understand correctly, these belong in `MenuList.vue`.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -125,50 +151,95 @@ export default defineComponent({
...
+	},
+	computed: {
+		totalHeight() {
+			return this.entries.length * 20;
+		},
+		startIndex() {
+			return Math.floor(this.entriesStart / 20);
+		},
+		endIndex() {
+			return Math.min(this.entries.length, this.startIndex + 1 + 400 / 20);
```
Keavon, line 236: I think a lot of the code in here is now unused since it was moved to `MenuList.vue`?

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -101,17 +102,42 @@ export default defineComponent({
...
+		},
+		scroller() {
+			return ((this.$refs.menulist as typeof MenuList).$refs.scroller as typeof LayoutCol)?.$el as HTMLElement;
+		},
+		setOpen() {
+			this.open = true;
+			// Reset the highlighted entry to the active one
+			this.setHighlighted(this.activeEntry);
+			// Scroll to the active entry (the scroller div does not yet exist so we must wait for vue to render)
+			nextTick((): void => {
```
Keavon, line 129: `await nextTick();` is a slightly nicer syntax. (You'll need to prefix `setOpen() {` with `async`.)

frontend/src/io-managers/fonts.ts
```diff
@@ -58,19 +64,18 @@ export function createFontsState(editor: Editor) {
...
 				resolve(result);
 			});
 	});
 
 	return {
-		state: readonly(state) as typeof state,
+		fontNames,
 		getFontStyles,
 		getFontFileUrl,
 	};
```
Keavon, line 76: I may have misdirected you about this being an I/O manager. The format for all I/O managers is that they must return from their constructor either nothing or a single destructor function. Not an object with several functions. I don't think `utility-functions` is the correct folder for this either, though, because those are all stateless single-purpose functions. Feel free to read `frontend/src/README.md` for a bit more detail on each of these types. Do you think this could maybe be restructured to fit the conventions for any of the others? I assume you can't make this stateless to work with `utility-functions`. Maybe the getter functions could be changed into entries for the `state` object and this could become a state provider again? Happy to discuss this further, I don't mean to misdirect you again.

Keavon:
Done with this round of code review comments. I'll look more closely at `FontInput.vue` once you move some of the functions in there to `MenuList.vue` or remove them if they're redundant. I'm very happy we get to implement virtual scrolling in this PR! We were needing that for performance reasons.

frontend/src/components/widgets/inputs/FontInput.vue
```diff
@@ -69,23 +58,27 @@
...
 		}
 	}
 
 	.menu-list .floating-menu-container .floating-menu-content {
 		max-height: 400px;
+		padding: 4px 0;
+
+		.spacer {
+			flex: 0 0 auto;
+		}
```
Keavon, line 70: This too. (Not sure about the `padding: 4px 0;`, but the `.spacer` isn't used here but you may need to move it to the other file.)

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -6,44 +6,53 @@
...
-					:open="entry.ref?.open || false"
-					:direction="'TopRight'"
-					:entries="entry.children"
-					v-bind="{ defaultAction, minWidth, drawIcon, scrollableY }"
-					:ref="(ref: typeof FloatingMenu) => ref && (entry.ref = ref)"
-				/>
-			</LayoutRow>
-		</template>
+		<!-- If we put the scrollableY on the layoutcol for non-font dropdowns then for some reason it always creates a tiny scrollbar.
+		However when we are using the fixedHeight then we need the layoutcol to be scrolling so we can bind the events without using $refs. -->
```
Keavon, line 14: fixedHeight was renamed

frontend/src/io-managers/fonts.ts
0HyperCube: Why should they become entries in a state object? This would be additional complexity for no gain.

Keavon: I'm just trying to figure out how we can make it match the code style of the other files to avoid confusing inconsistency. My idea about `state` was that, perhaps instead of exposing these getter functions, we can instead just bind the current value to the `state` object so users can replace their getter function calls with reading the state values? I don't know if that is the right approach here or not, but just an idea. If you'd like to talk more about this on Discord, we can probably settle it out real quick, I just want to make sure we're using our frontend architecture in a not-inconsistent way.

Keavon (approved):

Keavon:
Hmm...

https://user-images.githubusercontent.com/4388688/173144074-8cecbd44-5c23-42c0-9bb1-47f9b3e7dd5b.mp4

Keavon:
Awesome! There's only one very minor polishing behavior I'd like to eventually fix, but it's probably worth leaving for a later date. When you scroll, once you reach the bottom of a loaded chunk, it bottoms out and doesn't let you keep scrolling with the mouse wheel until you move your cursor by a pixel. Kind of strange. But I'm happy to merge this now unless you prefer to take a poke at that real quick.

0HyperCube:
> When you scroll, once you reach the bottom of a loaded chunk, it bottoms out and doesn't let you keep scrolling with the mouse wheel until you move your cursor by a pixel. Kind of strange. But I'm happy to merge this now unless you prefer to take a poke at that real quick.

That is quite odd - this does not happen on firefox but I can reproduce it on chromium. The loaded chunk is just the 21 visible entries (you can see 21 x 20px entries in the 400px height if the scroll is not offset by a multiple of 20) so this doesn't happen when you reach the end of that.

PR #659, opened 2022-05-22
Refactor font loading from per-document to the portfolio

### LINKED ISSUE #655: Move font loading from per-document to global on the editor
When we load the default font, we'd like to be able to do that in `init_app()` of `api.rs` rather than in `load_default_font()` in `document_message_handler.rs`. Loaded fonts should be stored on the editor (either in the dispatcher struct or maybe the portfolio struct), not in each document. When we serialize a document, then we can include the font or maybe just store its URL (pro: smaller save files, con: requires internet and error handling).

Search for this string as a starting point:
```
// TODO: Loading the default font should happen on a per-application basis, not a per-document basis
```

0HyperCube:
Closes #655 

- `font_cache` per document -> font_cache in portfolio (involved passing the `font_cache` to all tools)
- `TriggerFontLoad` called from backend with default font
- `tool_data` (primary/secondary colour) -> global_tool_data
- `data` (for specific tool) -> tool_data
- Reduce number of arguments to `transition` by using `ToolActionHandlerData`
- Fix clippy lints (deriving `Eq` for many `PartialEq` structs / enums)
- Merge conflict also resolved issue with clicking on menu lists.


frontend/src/state-providers/fonts.ts
```diff
@@ -46,11 +46,11 @@ export function createFontsState(editor: Editor) {
 
 		const response = await fetch(fontFileUrl);
 		const responseBuffer = await response.arrayBuffer();
```
Keavon, line 48: For consistency with the one below, can you combine these into a single line for me? Thanks.

frontend/wasm/src/api.rs
```diff
@@ -320,8 +320,16 @@ impl JsEditorHandle {
...
-		let message = DocumentMessage::FontLoaded { font_file_url, data, is_default };
+	pub fn on_font_load(&self, font_file_url: String, data: Vec<u8>) -> Result<(), JsValue> {
+		let message = DocumentMessage::FontLoaded { font_file_url, data };
+		self.dispatch(message);
+
+		Ok(())
+	}
+
+	/// The default font has been downloaded
+	pub fn on_default_font_load(&self, font_file_url: String, data: Vec<u8>) -> Result<(), JsValue> {
```
Keavon, line 331: This basically does the same thing as `on_font_load()`. Is there a reason we need to distinguish between the default and any ordinary font load request? Why can't we just send the editor a font with `DocumentMessage::FontLoaded` and let it cache that?

frontend/src/state-providers/fonts.ts
```diff
@@ -46,11 +46,11 @@ export function createFontsState(editor: Editor) {
...
-		editor.instance.on_font_load(fontFileUrl, new Uint8Array(responseBuffer), true);
+		editor.instance.on_default_font_load(fontFileUrl, new Uint8Array(responseBuffer));
 	});
 	editor.subscriptions.subscribeJsMessage(TriggerFontLoad, async (triggerFontLoad) => {
```
Keavon, line 51: Can we rename the `TriggerFontLoad` message to `TriggerFontLoadUrl`, then rename `TriggerFontLoadDefault` to `TriggerFontLoad` and make it pass along a `font_family` and `font_style`? Then the backend can request the default font by family/style instead of URL. Meaning we'd move these into the backend:
```ts
const DEFAULT_FONT = "Merriweather";
const DEFAULT_FONT_STYLE = "Normal (400)";
```

frontend/wasm/src/api.rs
0HyperCube: This one sends a portfolio message that populates the state in the portfolio

Keavon: Shouldn't all loaded fonts populate the cache the portfolio? It seems weird to me that we'd only have the default font do that.

0HyperCube: So we no longer store the font cache in the Document?

0HyperCube: This would be quite a large refactor.

Keavon: I'm inclined to answer that it doesn't make sense to have a per-document font cache. It seems like an editor-centric concept, since you'll have many documents which would benefit by sharing the font. But perhaps I'm overlooking arguments in favor of it being per-document? Happy to discuss this further and work out what's best.

Keavon: > This would be quite a large refactor.

We can avoid doing that refactor now if you'd prefer (let's just add a TODO comment and probably file an issue for it). Is there an intermediary state where this code can integrate the comments I gave (to generally clean up/generify the concept of the default font) without doing that full refactor?

Keavon (approved):
Excellent improvements!

PR #663, opened 2022-05-26
Fix viewport zoom/rotation while snapping to show snapped values

### LINKED ISSUE #643: Viewport zoom/rotation while snapping doesn't show the snapped values
In the viewport, the top right number inputs display the current precise values of the ongoing zoom/rotation, but they should show the snapped value which is what the viewport is rendering.

RahulHi:
The zoom and tilt inputs on the top right corner of the viewport previously did not display the correct snapped when snap zooming or snap tilting and instead displayed the precise zoom and tilt values. I changed the appropriate widget value's to use ``snapped_scale()`` and ``snapped_angle()`` to fix this.

Closes #643

### DIFF
--- a/editor/src/document/document_message_handler.rs
+++ b/editor/src/document/document_message_handler.rs
@@ -611,7 +611,7 @@ impl DocumentMessageHandler {
 				})),
 				WidgetHolder::new(Widget::NumberInput(NumberInput {
 					unit: "°".into(),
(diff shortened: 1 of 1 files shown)


Keavon:
Sure is nice when the bug fixes are simple! The less complexity, the more maintainable. Hopefully you got the chance to feel your way around the code a little bit during this endeavor!

Keavon (approved):

PR #664, opened 2022-06-05
Starter implementation of Bezier library and UI documentation

hannahli2010:
- Create constructors of the Bezier struct in library
- Use `wasm-bindgen` to create UI that will feature the different library functions
    - Currently displays manipulatable quadratic and cubic curves


bezier-rs/lib/src/lib.rs
```diff
@@ -1,4 +1,112 @@
...
+	Cubic,
+}
+
+/// Representation of a bezier curve with 2D points
+pub struct Bezier {
+	/// Vector containing the bezier points (represented as DVec2s)
+	points: [Option<DVec2>; 4],
+	/// The type of bezier curve
+	bezier_type: BezierType,
+}
```
0HyperCube, line 19: Would it be easier to have the points inside the enum to reduce the need to unwrap when using the points?
```rs
pub enum BezierSegement {
    Quadratic { start: DVec2, handle: DVec2, end: DVec2 },
    Cubic { start: DVec2, handle1: DVec2, handle2: DVec2, end: DVec2 },
}
```
This would additionally reduce the memory from 104 bytes to 72.

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.js
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	deselectPointHandler() {
+		if (this.dragIndex != null) {
+			this.points[this.dragIndex].selected = false;
+			this.ctx.clearRect(1, 1, this.canvas.width - 2, this.canvas.height - 2);
+			this.updateBezier();
+			this.dragIndex = null;
+		}
+	}
```
0HyperCube, line 64: Minor nit but it would be nice if a drag that started inside the canvas wasn't canceled as soon as it went out, you can perhaps use [setPointerCapture](https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture) on your canvas element on pointer down (you would have to use pointerdown rather than mousedown).

nupsi:
Should the interactive doc be made using TypeScript rathar than JavaScript to match the Graphite frontend?

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.js
hannahli2010: We decided to have it cancel once you move out of the canvas so that it isn't possible to "lose" a point (i.e by letting go of the point while off canvas).

bezier-rs/lib/src/lib.rs
hannahli2010: We refactored the code using this idea - do you have any thoughts on the new approach?

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.js
0HyperCube: Sounds reasonable

bezier-rs/lib/src/lib.rs
0HyperCube: Looks good - it is nice to not have to have unwrap all over the code.

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.js
```diff
@@ -0,0 +1,89 @@
...
+			this.bezier = this.wasm.WasmBezier.new_cubic(
+				this.points[0].x,
+				this.points[0].y,
+				this.points[1].x,
+				this.points[1].y,
+				this.points[2].x,
+				this.points[2].y,
+				this.points[3].x,
+				this.points[3].y
+			);
```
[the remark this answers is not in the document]
hannahli2010: There should be is a way to send them as a list of tuples - although I'm not sure if it has a performance impact or not. I'm going to leave this PR as is and add the change in a later one.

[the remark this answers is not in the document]
0HyperCube: Sounds good.

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -0,0 +1,110 @@
...
+		this.canvas = canvas;
+
+		this.canvas.width = 200;
+		this.canvas.height = 200;
+
+		const ctx = this.canvas.getContext("2d");
+		if (ctx == null) {
+			throw Error("Failed to create context");
+		}
+		this.ctx = ctx;
```
RobNadal, line 44: We can use the utility function here in the future

bezier-rs/docs/interactive-docs/src/components/Example.vue
```diff
@@ -0,0 +1,33 @@
...
+	</div>
+</template>
+
+<script lang="ts">
+import { defineComponent, PropType } from "vue";
+
+import { WasmBezierInstance } from "../utils/wasm-comm";
+
+import BezierDrawing from "./BezierDrawing";
+
```
RobNadal, line 14: Let's fix the import styling in a future branch

bezier-rs/docs/interactive-docs/src/components/ExamplePane.vue
```diff
@@ -0,0 +1,59 @@
...
+	</div>
+</template>
+
+<script lang="ts">
+import { defineComponent } from "vue";
+
+import { WasmBezierInstance } from "../utils/wasm-comm";
+
+import Example from "./Example.vue";
+// import wasm from "bezier-rs-wasm";
```
RobNadal, line 15: remove comment

PR #665, opened 2022-06-07
Fix Crash When Deselecting Layer While Editing Properties

### LINKED ISSUE #645: Crash when Properties panel changes before input widget is sent to backend
Select a layer, edit an input widget like a NumberInput, click in the viewport so it deselects the layer, this causes a crash because it's sending the widget value update to the backend which arrives after the widget ID has been disposed since the Properties panel isn't showing that widget anymore. Ideally, we'd apply the update before deselecting so the user's input isn't lost.

mfish33:
Closes #645


Keavon (approved):
Thanks for this!

PR #667, opened 2022-06-07
Implement viewport culling

### LINKED ISSUE #608: Viewport culling
Don't draw shapes that have their bounding box outside the viewport coordinates. So let's not render those SVG elements in Rust to save on SVG string-related performance on the Rust and Rust->JS end. This should actually be relatively easy since we already have functions for working with layer bounding boxes.

mfish33:
Closes #608

### DIFF
--- a/editor/src/document/document_message_handler.rs
+++ b/editor/src/document/document_message_handler.rs
@@ -1060,7 +1060,7 @@ impl MessageHandler<DocumentMessage, (&InputPreprocessorMessageHandler, &FontCac
 					false => file_name + file_suffix,
 				};
 
-				let rendered = self.graphene_document.render_root(self.view_mode, font_cache);
+				let rendered = self.graphene_document.render_root(self.view_mode, font_cache, None);
 				let document = format!(
 					r#"<svg xmlns="http://www.w3.org/2000/svg" viewBox="{} {} {} {}" width="{}px" height="{}">{}{}</svg>"#,
 					bbox[0].x, bbox[0].y, size.x, size.y, size.x, size.y, "\n", rendered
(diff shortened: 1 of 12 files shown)


editor/src/document/document_message_handler.rs
```diff
@@ -873,7 +873,8 @@ impl MessageHandler<DocumentMessage, (&InputPreprocessorMessageHandler, &FontCac
 			}
 			#[remain::unsorted]
 			Overlays(message) => {
-				self.overlays_message_handler.process_action(message, (self.overlays_visible, font_cache), responses);
+				self.overlays_message_handler.process_action(message, (self.overlays_visible, font_cache, ipp), responses);
+				// CAN THIS COMMENT BELOW BE REMOVED?
 				// responses.push_back(OverlaysMessage::RenderOverlays.into());
```
0HyperCube, line 878: Yes - I think it was from a restructuring of the overlay system a while ago.

editor/src/document/document_message_handler.rs
```diff
@@ -1100,7 +1101,8 @@ impl MessageHandler<DocumentMessage, (&InputPreprocessorMessageHandler, &FontCac
 				}
 			}
 			FolderChanged { affected_folder_path } => {
-				let _ = self.graphene_document.render_root(self.view_mode, font_cache);
+				// TODO: WHY IS IT RENDERING HERE? Can the line below be removed?
+				let _ = self.graphene_document.render_root(self.view_mode, font_cache, None);
```
0HyperCube, line 1105: I think it can be removed - it doesn't do anything except from updating any dirty layers svg cache (which would happen at some point anyway)

graphene/src/layers/folder_layer.rs
```diff
@@ -37,10 +37,13 @@ impl LayerData for FolderLayer {
...
 	fn bounding_box(&self, transform: glam::DAffine2, font_cache: &FontCache) -> Option<[DVec2; 2]> {
-		self.layers
+		let bounding_box = self
+			.layers
 			.iter()
 			.filter_map(|layer| layer.data.bounding_box(transform * layer.transform, font_cache))
-			.reduce(|a, b| [a[0].min(b[0]), a[1].max(b[1])])
+			.reduce(|a, b| [a[0].min(b[0]), a[1].max(b[1])]);
+
+		bounding_box
```
0HyperCube, line 46: Why do we need this temporary variable?

0HyperCube (approved):
Looks good

graphene/src/layers/folder_layer.rs
mfish33: Left over from debugging

Keavon (approved):

PR #672, opened 2022-06-09
Fix nudge and textbox creation

### LINKED ISSUE #660: Fix Interacting in GUI with text tool
When interacting with the web GUI with the text tool active, random text boxes spawn. This is most easily reproducible when dragging layers in the layer tree but can also happen when clicking in the properties panel.
BrokenGuiInteraction

This is related to how mouse down Rust events are sent when the bitflags from the mouse move event are changed. This is required to allow us to detect additional mouse presses when a mouse button is already down, however this behaviour should not be activated when no mouse buttons are yet pressed.

### LINKED ISSUE #669: Nudging layers isn't working
There seems to have been a regression. Can't use arrow keys to nudge layers anymore.

0HyperCube:
Closes #669 and closes #660

### DIFF
--- a/frontend/src/io-managers/input.ts
+++ b/frontend/src/io-managers/input.ts
@@ -13,7 +13,6 @@ type EventListenerTarget = {
 export function createInputManager(editor: Editor, container: HTMLElement, dialog: DialogState, document: PortfolioState, fullscreen: FullscreenState): () => void {
 	const app = window.document.querySelector("[data-app]") as HTMLElement | undefined;
 	app?.focus();
-	console.info(app);
 
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	const listeners: { target: EventListenerTarget; eventName: EventName; action: (event: any) => void; options?: boolean | AddEventListenerOptions }[] = [
@@ -40,10 +39,10 @@ export function createInputManager(editor: Editor, container: HTMLElement, dialo
 
 	let viewportPointerInteractionOngoing = false;
 	let textInput = undefined as undefined | HTMLDivElement;
-	let inCanvas = true;
+	let canvasFocused = true;
 
 	function blurApp(): void {
-		inCanvas = false;
+		canvasFocused = false;
 	}
 
 	// Keyboard events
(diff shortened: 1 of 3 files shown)


Keavon:
Just a little note for future reference, GitHub doesn't parse the "and" in your PR message so it only recognizes the first bug number for the "Closes" command. You have to write out "closes" for each number. I wish they would fix this!

frontend/src/io-managers/input.ts
```diff
@@ -31,6 +31,13 @@ export function createInputManager(editor: Editor, container: HTMLElement, dialo
 
 	let viewportPointerInteractionOngoing = false;
 	let textInput = undefined as undefined | HTMLDivElement;
+	let inCanvas = true;
+
+	const app = window.document.getElementById("app");
```
Keavon, line 36: Nitpick: we have a slight preference to use `document.querySelector("#app")` for consistency, since we don't use the older `getElementById` JS API anywhere else in the codebase. If we ever need to search for all the functions in the project that interact with the DOM, it's nice to have them only be the three: `querySelector`, `querySelectorAll`, and `closest`.

frontend/src/io-managers/input.ts
```diff
@@ -31,6 +31,13 @@ export function createInputManager(editor: Editor, container: HTMLElement, dialo
...
+	app?.focus();
+	app?.addEventListener("blur", () => {
```
Keavon, line 38: We need to clean this up when `unbindListeners` is called. (This occurs every time hot module replacement updates the app during development, and my previous failure to clean things up caused breakage.) Are you able to integrate this event listener with the block of listeners declared above (`const listeners = ...`)?

frontend/src/io-managers/input.ts
Keavon: Also, this is another case where it would be slightly preferable to select based on a data attribute not a class name or ID, just to be clean about our separation between ID/class/tag names being for CSS only and data attributes being for JS only.

Keavon (approved):
Lookin' good!

frontend/src/io-managers/input.ts
```diff
@@ -131,7 +151,7 @@ export function createInputManager(editor: Editor, container: HTMLElement, dialo
 
 		if (!inTextInput) {
 			if (textInput) editor.instance.on_change_text(textInputCleanup(textInput.innerText));
-			else if (inCanvas) viewportPointerInteractionOngoing = true;
+			else viewportPointerInteractionOngoing = inCanvas instanceof Element;
```
nupsi, line 154: I think the "inCanvas" naming is a bit confusing at the moment because it can mean two different things without a clear distinction. First its declared as a boolean on the top level and then as a element inside some functions. Maybe the local variable names should be changes to "canvasElement"? The confusion was the mainly the result of scrolling through the diff and not seeing the  full code.

0HyperCube: Good point - I've renamed the top level one to `canvasFocused` and the other one to `isTargetingCanvas`

Keavon: Those names are much clearer, thanks both of you!

PR #676, opened 2022-06-11
Bezier length implementation

hannahli2010:
Core changes
- Implement the length function for a bezier segment
- Add section for length in the interactive documentation and display the length value

Other changes:
- Refactor `WasmBezier` to take lists of points
- Add back `eslint` rule that was causing problems

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -1,4 +1,5 @@
 use glam::DVec2;
+use std::fmt::Write;
 
 pub enum BezierHandles {
 	Quadratic { handle: DVec2 },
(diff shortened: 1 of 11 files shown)


bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,86 @@ impl Bezier {
...
 	pub fn get_handle2(&self) -> Option<DVec2> {
 		match self.handles {
-			BezierHandles::Quadratic { handle: _ } => {
-				None
-			}
-			BezierHandles::Cubic { handle1: _, handle2 } => {
-				Some(handle2)
-			}
+			BezierHandles::Quadratic { handle: _ } => None,
+			BezierHandles::Cubic { handle1: _, handle2 } => Some(handle2),
```
nupsi, line 146: Unused fields can be also ignored with a `..` syntax which ignores all the fields that are not explicitly matched.
For example here
```
BezierHandles::Quadratic { handle: _ } => None,
BezierHandles::Cubic { handle1: _, handle2 } => Some(handle2),
```
could be written as
```
BezierHandles::Quadratic { .. } => None,
BezierHandles::Cubic { handle2, .. } => Some(handle2),
```

bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,86 @@ impl Bezier {
...
 	}
 
-	/// Return the length of the bezier curve
-	pub fn length() -> i32 {
-		0
-	}
-}
+	///  Calculate the x-value of a point on the curve based on the t-value provided
+	///  basis code based off of pseudocode found here: https://pomax.github.io/bezierinfo/#explanation
+	pub fn get_x_basis(&self, t: f64) -> f64 {
```
0HyperCube, line 159: Could we not make these two functions into one function that returns a `DVec2` to avoid duplication and then relevant code could do `point.x` and `point.y` when necessary?

bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,86 @@ impl Bezier {
...
+			let dx = ox - x;
+			let dy = oy - y;
+			clen += (dx * dx + dy * dy).sqrt();
+			// update ox, oy for next subdivision
+			ox = x;
+			oy = y;
+		}
 
-*/
+		clen
```
0HyperCube, line 218: `clen` means nothing to me - perhaps `total_length` or `length_subtotal` would be more informative.

bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,86 @@ impl Bezier {
...
+			// get end point of the subdivision
+			let x = self.get_x_basis(f64::from(i) * RATIO);
+			let y = self.get_y_basis(f64::from(i) * RATIO);
+			// calculate distance of subdivision
+			let dx = ox - x;
+			let dy = oy - y;
+			clen += (dx * dx + dy * dy).sqrt();
+			// update ox, oy for next subdivision
+			ox = x;
+			oy = y;
```
0HyperCube, line 215: Then this could be simplified to use points (e.g. `let mut old_positon: DVec2 = ...`) - you can also do things like `total_distance += (old_position - current_position).length();`

bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,86 @@ impl Bezier {
...
-	let sum = 0;
-	for (let i = 0, t; i < len; i++) {
-	  t = z * T[i] + z;
-	  sum += C[i] * this.arcfn(t, curve.derivative(t));
+		match self.handles {
+			BezierHandles::Quadratic { handle } => mt2 * self.start[1] + 2.0 * mt * t * handle[1] + t2 * self.end[1],
+			BezierHandles::Cubic { handle1, handle2 } => {
+				let t3 = t2 * t;
+				let mt3 = mt2 * mt;
+				mt3 * self.start[1] + 3.0 * mt2 * t * handle1[1] + 3.0 * mt * t2 * handle2[1] + t3 * self.end[1]
```
0HyperCube, line 186: Also instead of `some_dvec2[0]` and `some_dvec2[1]` you can do `some_dvec2.x` and `some_dvec2.y` which is probably easier to read.

bezier-rs/lib/src/lib.rs
```diff
@@ -122,57 +135,64 @@ impl Bezier {
...
-	pub fn length() -> i32 {
-		0
-	}
-}
+	///  Calculate the point on the curve based on the t-value provided
+	///  basis code based off of pseudocode found here: https://pomax.github.io/bezierinfo/#explanation
+	pub fn get_basis(&self, t: f64) -> DVec2 {
+		let t2 = t * t;
+		let mt = 1.0 - t;
+		let mt2 = mt * mt;
```
0HyperCube, line 162: Perhaps try and expand some of these abbreviations in line with graphite style guidelines.

0HyperCube (approved):
The improved use of Vec2 has definitely made it easier to read - thanks for doing that.

bezier-rs/lib/src/lib.rs
Keavon: (for reference: https://graphite.rs/contribute/#contributing-guide although I still need to expand that with more details)

hannahli2010: I'm not super sure what's the best way to expand these since they're mainly used to simplify the line of code that does the mathematical computation. Would it be better to use names such as:
- `t_squared` instead of `t2`
- `one_minus_t` instead of `mt`
- `one_minus_t_squared` instead of `mt2`
- etc?

Keavon: Those would be great names, yes :)

hannahli2010: I've updated the names - is this better?

Keavon: Sure, looks good!

PR #677, opened 2022-06-13
Add get and getLUT functions

hannahli2010:

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -172,9 +172,9 @@ impl Bezier {
 	}
 
 	/// Return a selection of equidistant points on the bezier curve
-	/// If no value is provided for `steps`, then the function will default `steps` to be 20
+	/// If no value is provided for `steps`, then the function will default `steps` to be 10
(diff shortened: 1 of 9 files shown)


bezier-rs/lib/src/lib.rs
```diff
@@ -171,28 +175,44 @@ impl Bezier {
 		}
 	}
 
+	///  Get the point on the curve based on the t-value provided, alias for compute
+	pub fn get(&self, t: f64) -> DVec2 {
+		self.compute(t)
+	}
```
0HyperCube, line 181: Why do you need an alias? This adds unnecessary complexity.

bezier-rs/lib/src/lib.rs
```diff
@@ -171,28 +175,44 @@ impl Bezier {
...
+
+	/// Return a selection of equidistant points on the bezier curve
+	/// If no value is provided for `steps`, then the function will default `steps` to be 10
+	pub fn get_lookup_table(&self, steps: Option<i32>) -> Vec<DVec2> {
```
0HyperCube, line 185: `get` implies that we are just retrieving something rather than computing it - perhaps `compute_lookup_table` would be more clear?

0HyperCube (approved):

PR #680, opened 2022-06-18
Bezier split and trim implementation

hannahli2010:
Implement the split and trim functions for a Bezier curve
- `split` returns the two Bezier curves that result from splitting the original at the point corresponding to `t`
- `trim` returns the single Bezier curve that results from taking the segment from `t1` to `t2` of the original curve
  - If `t1` > `t2`, the returned curve will travel in the opposite direction from the original
- Refactored the frontend to allow for variable amounts of sliders per example
- Add helper for drawing additional curves on the UI canvas

### DIFF
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -66,18 +66,18 @@ export default defineComponent({
 				{
 					id: 3,
 					name: "Compute",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const point = JSON.parse(bezier.compute(parseFloat(options)));
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const point = JSON.parse(bezier.compute(options.t));
 						drawPoint(getContextFromCanvas(canvas), point, 4, "Red");
 					},
 					template: markRaw(SliderExample),
-					templateOptions: tSliderOptions,
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
 				},
 				{
 					id: 4,
 					name: "Lookup Table",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const lookupPoints = bezier.compute_lookup_table(Number(options));
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const lookupPoints = bezier.compute_lookup_table(options.Steps);
 						lookupPoints.forEach((serialisedPoint, index) => {
 							if (index !== 0 && index !== lookupPoints.length - 1) {
 								drawPoint(getContextFromCanvas(canvas), JSON.parse(serialisedPoint), 3, "Red");
@@ -86,22 +86,25 @@ export default defineComponent({
 					},
 					template: markRaw(SliderExample),
 					templateOptions: {
-						min: 2,
-						max: 15,
-						step: 1,
-						default: 5,
-						variable: "Steps",
+						sliders: [
+							{
+								min: 2,
+								max: 15,
+								step: 1,
+								default: 5,
+								variable: "Steps",
+							},
+						],
 					},
 				},
 				{
 					id: 5,
 					name: "Derivative",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const t = parseFloat(options);
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
 						const context = getContextFromCanvas(canvas);
 
-						const intersection = JSON.parse(bezier.compute(t));
-						const derivative = JSON.parse(bezier.derivative(t));
+						const intersection = JSON.parse(bezier.compute(options.t));
+						const derivative = JSON.parse(bezier.derivative(options.t));
 						const curveFactor = bezier.get_points().length - 1;
 
 						const tangentStart = {
@@ -119,17 +122,16 @@ export default defineComponent({
 						drawPoint(context, tangentEnd, 3, "Red");
 					},
 					template: markRaw(SliderExample),
-					templateOptions: tSliderOptions,
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
 				},
 				{
 					id: 6,
 					name: "Normal",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const t = parseFloat(options);
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
 						const context = getContextFromCanvas(canvas);
 
-						const intersection = JSON.parse(bezier.compute(t));
-						const normal = JSON.parse(bezier.normal(t));
+						const intersection = JSON.parse(bezier.compute(options.t));
+						const normal = JSON.parse(bezier.normal(options.t));
 
 						const normalStart = {
 							x: intersection.x - normal.x * 20,
@@ -146,21 +148,48 @@ export default defineComponent({
 						drawPoint(context, normalEnd, 3, "Red");
 					},
 					template: markRaw(SliderExample),
-					templateOptions: tSliderOptions,
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
 				},
 				{
 					id: 6,
 					name: "Split",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const t = parseFloat(options);
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
 						const context = getContextFromCanvas(canvas);
-						const bezierPair = bezier.split(t);
+						const bezierPair = bezier.split(options.t);
 
-						drawBezierHelper(context, bezierPair.second, "orange");
-						drawBezierHelper(context, bezierPair.first, "Red");
+						drawBezierHelper(context, bezierPair.second, "orange", 3.5);
+						drawBezierHelper(context, bezierPair.first, "Red", 3.5);
 					},
 					template: markRaw(SliderExample),
-					templateOptions: tSliderOptions,
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
+				},
+				{
+					id: 7,
+					name: "Trim",
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const context = getContextFromCanvas(canvas);
+						const trimmedBezier = bezier.trim(options.t1, options.t2);
+						drawBezierHelper(context, trimmedBezier, "Red", 3.5);
+					},
+					template: markRaw(SliderExample),
+					templateOptions: {
+						sliders: [
+							{
+								variable: "t1",
+								min: 0,
+								max: 1,
+								step: 0.01,
+								default: 0.25,
+							},
+							{
+								variable: "t2",
+								min: 0,
+								max: 1,
+								step: 0.01,
+								default: 0.75,
+							},
+						],
+					},
 				},
 			],
 		};
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -13,7 +13,7 @@
 <script lang="ts">
 import { defineComponent, markRaw } from "vue";
 
-import { drawText, drawPoint, drawLine, getContextFromCanvas } from "@/utils/drawing";
+import { drawText, drawPoint, drawLine, getContextFromCanvas, drawBezierHelper } from "@/utils/drawing";
 import { WasmBezierInstance } from "@/utils/types";
 
 import ExamplePane from "@/components/ExamplePane.vue";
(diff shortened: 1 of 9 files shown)


bezier-rs/docs/interactive-docs/src/components/ExamplePane.vue
```diff
@@ -3,7 +3,7 @@
 		<h2 class="example_pane_header">{{ name }}</h2>
 		<div class="example_row">
 			<div v-for="example in exampleData" :key="example.id">
-				<component :is="template" :templateOptions="templateOptions" :title="example.title" :bezier="example.bezier" :callback="callback" />
+				<component :is="template" :templateOptions="{ ...templateOptions }" :title="example.title" :bezier="example.bezier" :callback="callback" />
```
Keavon, line 6: Why is this necessary?

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
```diff
@@ -35,9 +37,8 @@ export default defineComponent({
 		},
 	},
 	data() {
-		return {
-			value: this.templateOptions.default,
-		};
+		const sliders: SliderOption[] = this.templateOptions.sliders;
+		return Object.assign({}, ...sliders.map((s) => ({ [s.variable]: s.default })));
```
Keavon, line 41: Should this be `computed`, rather than data`?

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
```diff
@@ -1,15 +1,17 @@
 <template>
 	<div>
-		<Example :title="title" :bezier="bezier" :callback="callback" :options="value.toString()" />
-		<div class="slider_label">{{ templateOptions.variable }} = {{ value }}</div>
-		<input class="slider" v-model="value" type="range" :step="templateOptions.step" :min="templateOptions.min" :max="templateOptions.max" />
+		<Example :title="title" :bezier="bezier" :callback="callback" :options="$data" />
+		<div v-for="(slider, index) in templateOptions.sliders" :key="index">
+			<div class="slider_label">{{ slider.variable }} = {{ $data[slider.variable] }}</div>
+			<input class="slider" v-model.number="$data[slider.variable]" type="range" :step="slider.step" :min="slider.min" :max="slider.max" />
```
Keavon, line 6: I'd feel more comfortable if this `$data[slider.variable]` was done in a method or computed function.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -1,11 +1,6 @@
-import { Point } from "@/utils/types";
+import { Point, WasmBezierInstance } from "@/utils/types";
 
-const RADIUS_SIZE = {
-	large: 5,
-	small: 3,
-};
-
-export const getPointSizeByIndex = (index: number, numPoints: number): number => RADIUS_SIZE[index === 0 || index === numPoints - 1 ? "large" : "small"];
+export const getPointSizeByIndex = (index: number, numPoints: number, radius = 5): number => (index === 0 || index === numPoints - 1 ? radius : (radius * 2) / 3);
```
Keavon, line 3: Move `2 / 3` to a `const ?_FACTOR` (where `?` is a name you pick) variable.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -46,7 +41,17 @@ export const drawText = (ctx: CanvasRenderingContext2D, text: string, x: number,
...
+		ctx,
+		bezier.get_points().map((p) => JSON.parse(p)),
+		stroke,
+		null,
+		radius
+	);
+};
+
+export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], stroke = "black", dragIndex: number | null = null, radius = 5): void => {
 	/* Until a bezier representation is finalized, treat the points as follows
```
Keavon, line 55: Style: we don't use `/* */` comments.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -46,7 +41,17 @@ export const drawText = (ctx: CanvasRenderingContext2D, text: string, x: number,
...
+	drawBezier(
+		ctx,
+		bezier.get_points().map((p) => JSON.parse(p)),
+		stroke,
+		null,
+		radius
+	);
+};
+
+export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], stroke = "black", dragIndex: number | null = null, radius = 5): void => {
```
Keavon, line 54: I'd suggest putting `dragIndex` before or after the `stroke` and `radius` parameters because those two both deal with styling and should be contiguous. (If `dragIndex` is the index within `points`, those two parameters should also probably be contiguous.)

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -66,18 +66,18 @@ export default defineComponent({
...
-					templateOptions: tSliderOptions,
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
 				},
 				{
 					id: 4,
 					name: "Lookup Table",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: string): void => {
-						const lookupPoints = bezier.compute_lookup_table(Number(options));
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const lookupPoints = bezier.compute_lookup_table(options.Steps);
```
Keavon, line 80: The capital word `Steps` doesn't seem to match with any symbol in the codebase.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -146,7 +148,48 @@ export default defineComponent({
...
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
+				},
+				{
+					id: 6,
+					name: "Split",
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const context = getContextFromCanvas(canvas);
+						const bezierPair = bezier.split(options.t);
+
+						drawBezierHelper(context, bezierPair.second, "orange", 3.5);
```
Keavon, line 160: "orange" is inconsistently not capitalized, but these are just CSS variables, right? I recommend making them all lowercase.

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -9,12 +9,22 @@ struct Point {
 	y: f64,
 }
 
+/// Wrapper of the `Bezier` struct to be used in JS.
 #[wasm_bindgen]
-/// Wrapper of the `Bezier` struct to be used in JS
+#[derive(Clone)]
 pub struct WasmBezier {
```
Keavon, line 15: Have you considered using tuple syntax here instead of key-value pair syntax? `pub struct WasmBezier(Bezier);`

bezier-rs/lib/src/lib.rs
```diff
@@ -254,4 +256,53 @@ impl Bezier {
...
+					split_point,
+					t_squared * self.end - 2. * t * t_minus_one * handle_end + squared_t_minus_one * handle_start,
+					t * self.end - t_minus_one * handle_end,
+					self.end,
+				),
+			],
+		}
+	}
+
+	/// Returns the Bezier curve representing the sub-curve starting at the point corresponding to `t1` and ending add the point corresponding to `t2`
```
Keavon, line 291: "and ending add" -> "and ending at"

bezier-rs/lib/src/lib.rs
```diff
@@ -254,4 +256,53 @@ impl Bezier {
...
+	pub fn trim(&self, t1: f64, t2: f64) -> Bezier {
+		// Depending on the order of t1 and t2, determine which half of the split we need to keep
+		let t1_split_side = if t1 <= t2 { 1 } else { 0 };
+		let t2_split_side = if t1 <= t2 { 0 } else { 1 };
+		let bezier_starting_at_t1 = self.split(t1)[t1_split_side];
+		// Ajust the ratio t2 to its corresponding value on the new curve that was split on t1
```
Keavon, line 297: Adjust* (also put backticks around t1, t2, etc. throughout these comments please.

Keavon (approved):
Good to merge after code review.

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
RobNadal: We could create a function here using v-on="() => {update(variable.name, ...)}" but this is what v-model does under the hood already. Using an function here is also not ideal as the goal in this component is to allow the slider's variable names to be dynamic, which v-model handles very succinctly. I think the use of $data here though is unclean and we shouldn't be passing it directly to the child component, so instead I've created a sub-object within data to be modified and passed instead.

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
RobNadal: I made the change to tuple syntax. To support this, I've swapped all instances of `self.internal` to `self.0`. Is the best way of referencing the underlying Bezier object?

RobNadal:
This change previously introduced `WasmBezierPair` because we needed a work around to return multiple Bezier objects, and WASM was being uncooperative. This wasn't a great solution though, because we will need at some point to be able to return lists of Bezier objects where the length isn't fixed. We had the idea instead to simply return the list of points corresponding to each generated Bezier, as all we need to do anyway on the frontend is draw the bezier anyway.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -34,6 +34,9 @@ const testBezierLib = async () => {
 	});
 };
 
+const NON_INTERACTIVE_STROKE_1 = "red";
```
hannahli2010, line 37: It might be nice to change this to the following and then move it to the `drawing.ts` file:
```
const COLORS = {
  INTERACTIVE: {
    STROKE_1: "black",
    STROKE_2: "grey",
    SELECTED: "blue",
  },
  NON_INTERACTIVE: {
    STROKE_1: "red",
    STROKE_2: "orange",
  }
}
```
Using the word "color" here might be more insightful than "stroke" on its own.
Also, maybe we can use the color "grey" for the line draw from a point to its handler?
If you use this, make sure to replace instances of `"black"` and `"blue"` to use the new constant!

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
```diff
@@ -38,7 +38,9 @@ export default defineComponent({
 	},
 	data() {
 		const sliders: SliderOption[] = this.templateOptions.sliders;
-		return Object.assign({}, ...sliders.map((s) => ({ [s.variable]: s.default })));
+		return {
+			sliderData: Object.assign({}, ...sliders.map((s) => ({ [s.variable]: s.default }))),
```
Keavon, line 42: Instead of `Object.assign({}, ...X}`, can't you just do `{ ...X }`?

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -35,72 +24,68 @@ impl WasmBezier {
...
 		let points: [DVec2; 4] = js_points.into_serde().unwrap();
-		WasmBezier {
-			internal: Bezier::from_cubic_dvec2(points[0], points[1], points[2], points[3]),
-		}
+		WasmBezier(Bezier::from_cubic_dvec2(points[0], points[1], points[2], points[3]))
 	}
 
 	pub fn set_start(&mut self, x: f64, y: f64) {
-		self.internal.set_start(DVec2::from((x, y)));
+		self.0.set_start(DVec2::from((x, y)));
```
hannahli2010, line 37: Oh yeah, I'm wondering if we want to switch to using `DVec2::new(x, y)` instead of having to create a tuple i.e. this'll get rid of some brackets.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -41,23 +45,22 @@ export const drawText = (ctx: CanvasRenderingContext2D, text: string, x: number,
 	ctx.fillText(text, x, y);
 };
 
-export const drawBezierHelper = (ctx: CanvasRenderingContext2D, bezier: WasmBezierInstance, stroke = "black", radius = 5): void => {
+export const drawBezierHelper = (ctx: CanvasRenderingContext2D, bezier: WasmBezierInstance, stroke = "black", radius = DEFAULT_ENDPOINT_RADIUS): void => {
 	drawBezier(
 		ctx,
-		bezier.get_points().map((p) => JSON.parse(p)),
+		bezier.get_points().map((p: string) => JSON.parse(p)),
```
Keavon, line 51: Can you set this long line to a variable on the line above in order to make this function call not require so many lines?

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
Keavon: Yes, unfortunately there's no other way (besides destructuring into a variable assignment on a separate line, which is unnecessarily verbose in this case) besides the `.0` syntax in Rust. That is idiomatic.

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
RobNadal: The spread operator in your solution unpacks the list, not the maps therein. `sliders = [{variable: 't1', default: 0.5}, {variable: 't2', default: 0.2}]` produces `{0: {t1: 0.5}, 1: {t2: 0.2}`, but we want `{t1: 0.5, t2: 0.2}`

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
RobNadal: Fixed all occurrences of this issue in wasm/src/lib.rs and lib/src/lib.r

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -41,23 +45,22 @@ export const drawText = (ctx: CanvasRenderingContext2D, text: string, x: number,
...
 };
 
-export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], stroke = "black", dragIndex: number | null = null, radius = 5): void => {
-	/* Until a bezier representation is finalized, treat the points as follows
-		points[0] = start point
-		points[1] = handle start
-		points[2] = (optional) handle end
-		points[3] = end point
-	*/
+export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], stroke = "black", radius = DEFAULT_ENDPOINT_RADIUS, dragIndex: number | null = null): void => {
```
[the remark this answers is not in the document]
RobNadal: Thanks for the clarification! I've moved the parameters around as you suggested.

Androxium (approved):

PR #681, opened 2022-06-18
Menu bar definition in backend

### LINKED ISSUE #556: Move the menu bar entry definitions to the backend
Soft blocked on #653 (work on that before this).

The *File*, *Edit*, *Layer*, *Document*, *View*, *Help* menu and its sub-menus should be ported over to the new layout definition system so we can avoid all these extra functions in `api.rs` where the JS has to call the actions. It is currently defined in JS. This should hopefully mean that checkboxes should become functional. This will also make #531 easier/cleaner.

mfish33:
Closes #556

### DIFF
--- a/editor/src/layout/widgets.rs
+++ b/editor/src/layout/widgets.rs
@@ -81,6 +81,10 @@ impl MenuEntry {
 			on_update: WidgetCallback::new(callback),
 		}))
 	}
+
+	pub fn no_action() -> WidgetHolder {
+		MenuEntry::create_action(|_| Message::NoOp)
+	}
 }
 
 impl Default for MenuEntry {
--- a/editor/src/layout/widgets.rs
+++ b/editor/src/layout/widgets.rs
@@ -1,14 +1,14 @@
 use std::rc::Rc;
 
 use super::layout_message::LayoutTarget;
-use crate::message_prelude::*;
+use crate::{input::keyboard::Key, message_prelude::*};
 
 use derivative::*;
 use serde::{Deserialize, Serialize};
 
 pub trait PropertyHolder {
-	fn properties(&self) -> WidgetLayout {
-		WidgetLayout::default()
+	fn properties(&self) -> Layout {
+		Layout::WidgetLayout(WidgetLayout::default())
 	}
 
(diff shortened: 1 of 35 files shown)


Keavon (approved):

editor/src/layout/widgets.rs
```diff
@@ -425,3 +569,11 @@ pub struct TextLabel {
 	#[serde(rename = "tableAlign")]
 	pub table_align: bool,
 }
+
+#[derive(Clone, Serialize, Deserialize, Derivative, Default)]
+#[derivative(Debug, PartialEq)]
+pub struct Invisible {
```
Keavon, line 575: Please comment the intended purpose of this since it's non-obvious.

editor/src/layout/widgets.rs
```diff
@@ -22,6 +22,149 @@ pub trait PropertyHolder {
...
+				Some(&mut menu_entry.action)
+			}
+			None => None,
+		}
+	}
+}
+
 #[derive(Debug, Default, Clone, Serialize, Deserialize, PartialEq)]
 pub struct WidgetLayout {
 	pub layout: SubLayout,
```
Keavon, line 170: (This comment applies ~27 lines below the targeted line since GitHub doesn't let me comment on that.)

Are you able to address the `// TODO: Rename LayoutRow to something more generic` in this PR? It seems pretty related to the work you're doing and you are probably better able to figure out a name for this than me with your understanding of the full system design.

frontend/wasm/src/editor_api.rs
```diff
@@ -117,6 +117,9 @@ impl JsEditorHandle {
 
 		let message = MovementMessage::TranslateCanvas { delta: (0., 0.).into() };
 		self.dispatch(message);
+
+		let message = MenuBarMessage::SendLayout;
+		self.dispatch(message);
```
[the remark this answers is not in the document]
mfish33: The Document widgets also get initialized at the beginning `UpdateDocumentWidgets`

editor/src/layout/widgets.rs
mfish33: I renamed it to `LayoutGroup`. What do you think?

Keavon: Good name!

PR #684, opened 2022-06-20
Fix layer deletion bugs when tools are in use

### LINKED ISSUE #613: Stuck overlays in pen tool when layer deleted
To reproduce:
- Select pen tool
- Start drawing by clicking
- Draw a few points
- Press the delete key
- Note how the path disappears but the overlay does not until tools are switched

AlexiWolf:
This PR fixes a crash triggered when the user deletes the active layer (by pressing the `delete` key) while they are drawing with the `PenTool`.  ~~In addition, it fixes `LayerNotFound` errors that pop up in the log.~~  This PR is part of #613.

Changes Made:

- [x] Fixed crash when layer is deleted while the `PenTool` is in use.
- [x] Added crash test cases for all `ToolType`s that can have their layers deleted while in use.
- [x] Fixed buggy behavior in the `Spline` tool introduced by the crash fix.
- [x] Fixed buggy behavior in the `Pencil` tool introduced by the crash fix.
- [x] ~~Fixed `LayerNotFound` errors showing up in the log when a layer is deleted while using a tool.~~ We suspect this error will be resolved by future refactoring.  Since it does not appear to effect the behavior of the application, we're opting not to fix this for the time being.

### DIFF
--- a/frontend/wasm/src/editor_api.rs
+++ b/frontend/wasm/src/editor_api.rs
@@ -435,23 +435,3 @@ impl Drop for JsEditorHandle {
 		EDITOR_INSTANCES.with(|instances| instances.borrow_mut().remove(&self.editor_id));
 	}
 }
-
-#[cfg(test)]
-mod layer_delete_bug_tests {
-	use super::*;
-
-	use editor::misc::test_utils::EditorTestUtils;
-	use editor::viewport_tools::tool::ToolType;
-
-	#[test]
-	fn should_not_crash_when_layer_is_deleted_while_using_a_tool() {
-		editor::communication::set_uuid_seed(0);
-		let mut test_editor = Editor::new();
-
-		test_editor.select_tool(ToolType::Pen);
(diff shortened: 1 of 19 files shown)


frontend/wasm/src/editor_api.rs
```diff
@@ -435,3 +435,23 @@ impl Drop for JsEditorHandle {
...
+
+#[cfg(test)]
+mod layer_delete_bug_tests {
+	use super::*;
+
+	use editor::misc::test_utils::EditorTestUtils;
+	use editor::viewport_tools::tool::ToolType;
+
+	#[test]
+	fn should_not_crash_when_layer_is_deleted() {
```
Keavon, line 447: This should probably say something about "when using a tool", maybe `layer_deleted_while_using_tool` is a better name for it?

Keavon (approved):

Keavon (approved):

PR #685, opened 2022-06-23
Image and text bugfixes

### LINKED ISSUE #686: Exporting bitmap images does not appear to work
When trying to export a bitmap image as...

### An SVG
I get the following error
image 
``` XML Parsing Error: prefix not bound to a namespace
Location: file:///home/esmecat/Downloads/Untitled%20Document.svg
Line Number 5, Column 34:
1,-0,-0,1,575.7999877929688,93)"><image width="2560" height="1440" transform="matrix(1,0,0,1,-575.7999877929688,-93)" xlink:href="blob:https://editor.graphite.rs/23120a9c-1bd4-42a1-94f3-19c56bf1e99a" /></g></g>
---------------------------------^
```

### A PNG or JPG
Nothing happens, as shown below
<video src="https://user-images.githubusercontent.com/71032999/175361395-99cf8ed2-fe24-417c-929d-a57084994711.mp4">

### Additional context
This is the picture which I dragged into the canvas
Some random picture of mountains

0HyperCube:
Closes #686 

- [x] Export selection
- [x] Refactor render functions to take RenderData for the viewmode, font cache and culling bounds
- [x] Make `load_layer_resources` (called when loading a graphite file from autosave) load fonts as well as images - previously fonts did not show up.
- [x] `.jpeg` exports now have a white background
- [x] Bitmap images serialised with base64 in graphite file (large file size decrease) - this was discussed but I forgot to implement it
- [x] Bitmap images use `href` rather than the depricated `xlink:href`
- [x] Bitmap images can be embedded into the svg rather than as blob urls when exported (previously when you exported an svg with an embedded image it did not render as the blob url no longer existed)
- [x] Remove unnecessary `regenerate_path` from text layer


Keavon:
We seem to be having an offset, only at certain times, to the export bounds based on Selection. I have no idea what's going on here, but it seems to be related to the location of the shapes and the pan of the viewport. Here's my attempt to narrow down on some patterns of behavior: [video](https://files.keavon.com/-/HiddenWoefulBittern/capture_2_.mp4).

editor/src/dialog/dialogs/export_dialog.rs
```diff
@@ -61,7 +61,7 @@ impl PropertyHolder for Export {
 		];
 
 		let artboards = self.artboards.iter().map(|(&val, name)| (ExportBounds::Artboard(val), name.to_string()));
-		let mut export_area_options = vec![(ExportBounds::AllArtwork, "All Artwork".to_string())];
+		let mut export_area_options = vec![(ExportBounds::AllArtwork, "All Artwork".to_string()), (ExportBounds::Selection, "Selection".to_string())];
```
Keavon, line 64: The ideal behavior would be a contextual graying-out of the "Selection" entry when nothing is selected. That would require plumbing through the ability to gray out menu entries, but this might be a good opportunity to do that. A less-ideal alternative would be removing the "Selection" entry entirely, but add a TODO comment suggesting it be grayed out instead someday. Graying out provides a better UX because it allows users to know what would be available in other situations, and to avoid getting confused about how something isn't showing up like they remember seeing in the past without realizing it's contextually unavailable.

frontend/src/state-providers/portfolio.ts
```diff
@@ -40,7 +40,7 @@ export function createPortfolioState(editor: Editor) {
 		if (!context) return;
 
 		// Fill the canvas with white if jpeg (does not support transparency and defaults to black)
-		if (triggerRasterDownload.mime.endsWith("jpg")) {
+		if (triggerRasterDownload.mime.endsWith("jpeg")) {
```
Keavon, line 43: Ah, sorry for breaking that back when I made this change.

graphene/src/layers/layer_info.rs
```diff
@@ -70,7 +107,7 @@ pub trait LayerData {
 	///     </g>"
 	/// );
 	/// ```
-	fn render(&mut self, svg: &mut String, svg_defs: &mut String, transforms: &mut Vec<glam::DAffine2>, view_mode: ViewMode, font_cache: &FontCache, culling_bounds: Option<[DVec2; 2]>);
+	fn render(&mut self, svg: &mut String, svg_defs: &mut String, transforms: &mut Vec<glam::DAffine2>, render_data: RenderData);
```
Keavon, line 110: Good change, to move this to a ViewMode struct, thank you!

graphene/src/document.rs
```diff
@@ -375,6 +375,23 @@ impl Document {
...
+			let mut dirty = false;
+			for l in f.layers_mut() {
+				dirty = Self::mark_layer_type_as_dirty(l, data_type) || dirty;
+			}
+			root.cache_dirty = dirty;
+		}
+		root.cache_dirty
+	}
+
+	pub fn mark_all_layer_type_as_dirty(&mut self, data_type: LayerDataTypeDiscriminant) -> bool {
```
Keavon, line 391: "type" should probably be plural: "types". Also could you please comment these two functions? (Or at least the non-plural one.)

graphene/src/document.rs
```diff
@@ -375,6 +375,23 @@ impl Document {
 		Ok(())
 	}
 
+	fn mark_layer_type_as_dirty(root: &mut Layer, data_type: LayerDataTypeDiscriminant) -> bool {
+		if LayerDataTypeDiscriminant::from(&root.data) == data_type {
+			root.cache_dirty = true;
+		} else if let LayerDataType::Folder(f) = &mut root.data {
+			let mut dirty = false;
+			for l in f.layers_mut() {
```
Keavon, line 383: Slight preference for full words instead of single-letter variable names, although not a big deal in small-coverage areas like this. Feel free to disregard this if you disagree regarding cleanliness of the code.

editor/src/document/document_message_handler.rs
```diff
@@ -21,16 +21,16 @@ use graphene::color::Color;
...
+use graphene::layers::style::{Fill, RenderData, ViewMode};
+use graphene::layers::text_layer::{Font, FontCache};
 use graphene::{DocumentError, DocumentResponse, LayerId, Operation as DocumentOperation};
 
 use glam::{DAffine2, DVec2};
 use log::warn;
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
 use std::collections::VecDeque;
+use std::collections::{HashMap, HashSet};
```
Keavon, line 33: Combine these please.

Keavon (approved):
Excellent improvements! If you can figure out the bounds bug (and address the nit picks in code review), feel free to merge when you're ready with it.

Keavon:
It would also be great to have another option in the export dialog: Background. It could be set to Artboard Background, Transparent, Black, and White. Transparent would be grayed out if JPG is selected. SVG would need to add a colored rectangle behind everything with dimensions equal to the viewBox bounds (except when Transparent is chosen).

0HyperCube:
> We seem to be having an offset, only at certain times, to the export bounds based on Selection. I have no idea what's going on here, but it seems to be related to the location of the shapes and the pan of the viewport. Here's my attempt to narrow down on some patterns of behavior: [video](https://files.keavon.com/-/HiddenWoefulBittern/capture_2_.mp4).

I can't seem to replicate this - I've changed it so that all caches are marked as dirty on export which may help?

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -20,11 +20,11 @@
...
 					class="row"
-					:class="{ open: isEntryOpen(entry), active: entry.label === highlighted?.label }"
+					:class="{ open: isEntryOpen(entry), active: entry.label === highlighted?.label, disabled: entry.disabled }"
 					:style="{ height: virtualScrollingEntryHeight || '20px' }"
-					@click="() => onEntryClick(entry)"
-					@pointerenter="() => onEntryPointerEnter(entry)"
-					@pointerleave="() => onEntryPointerLeave(entry)"
+					@click="() => entry.disabled || onEntryClick(entry)"
+					@pointerenter="() => entry.disabled || onEntryPointerEnter(entry)"
+					@pointerleave="() => entry.disabled || onEntryPointerLeave(entry)"
```
Keavon, line 27: I think you want `!entry.disabled && ...` since, if `disabled` is true, you're giving the value of `true` to these event handlers which it might not know what to do with. I think it needs a falsy value, not `true`.

frontend/src/components/floating-menus/MenuList.vue
```diff
@@ -141,6 +141,18 @@
 					color: var(--color-f-white);
 				}
 			}
+
+			&.disabled {
+				background: var(--color-2-mildblack);
```
Keavon, line 146: Is the background necessary? Isn't it already that?

frontend/src/components/floating-menus/MenuList.vue
0HyperCube: I'm not sure why it matters, event handlers can have any return type and it is just discarded by the browser.

Keavon: I was just thinking it might subtly cause warnings from Vue, now or in the future.

frontend/src/components/floating-menus/MenuList.vue
0HyperCube: This prevents the background changing on hover.

Keavon: Ah, could you set it to `none` instead then? That would be slightly clearer and cleaner in regards to maintainability.

0HyperCube: Then it would have no background?

frontend/src/components/floating-menus/MenuList.vue
0HyperCube: That warning would be pretty useless and probably spam everyone's projects but I will change it if you think it would be easier to understand?

frontend/src/components/floating-menus/MenuList.vue
Keavon: Setting `none` would disable the color that's set for the entry's hover background, but non-hovered backgrounds normally have no background color. The parent menu provides the background, and it sets a 95% opacity. This means that setting the 2-mildblack background for the grayed out entry is 100% opacity and will appear subtly different in color. It would also be clearer to set the background to none on the `:hovered: pseudo-selector to make it more obvious what's being overridden, and to prevent a future change in selectivity resulting in this no longer overriding the `:hover` color.

```scss
&.disabled {
    &:hover {
        background: none;
    }

    ...
}
```

frontend/src/components/floating-menus/MenuList.vue
Keavon: I haven't tested if Vue gives a warning or not. But at least I think it's slightly better practice to make sure the handler gets either a callback or a falsy value, for code consistency with the rest of the codebase at least. Probably not a big deal though. Thanks.

PR #687, opened 2022-06-26
Implement functions to create a Bezier that goes through 3 specified points

hannahli2010:
Implement functions to create a Bezier that goes through 3 specified points
- `quadraticThroughPoints` creates a quadratic Bezier that passes through 3 points
  - starts at the first point, ends at the third
  - will go through second at the position corresponding to `t` on the curve
- `cubicThroughPoints` creates a cubicBezier that passes through 2 points
  - starts at the first point, ends at the third
  - will go through second at the position corresponding to `t` on the curve
  - `strut` is a parameter used to determine how "wide" the resulting cubic curve will be. Specifically, it represents the distance between `e1` and `e2` in the [projection identify section of Pomax's Bezier primer](https://pomax.github.io/bezierinfo/#abc)
- Add section for tests in `lib.rs`
- Create `utils.rs` for mathematical computation helpers
- Add UI interactive section
  - Adjust existing framework to handle manipulating the point on the curve instead of the handlers
  - Add two sliders to the cubic case
  - Refactor `drawBezier` functions to allow for customizing colors
     - If no color is specified for the handle or handle lines, they will default to the curve color
- Delete unnecessary `wasm-comm.ts` file (`types.ts` has the same exports)


bezier-rs/lib/src/lib.rs
```diff
@@ -70,18 +70,61 @@ impl Bezier {
 		}
 	}
 
+	// TODO: Maybe move to a utils file
+	fn compute_abc(start_point: DVec2, point_on_curve: DVec2, end_point: DVec2, t_term: f64, one_minus_t_term: f64) -> [DVec2; 3] {
```
RobNadal, line 74: I don't love the names t_term or one_minus_t_term. t_to_nth_power is more clear but also kinda gross. t_power might be a good middle ground.

bezier-rs/lib/src/lib.rs
```diff
@@ -70,18 +70,61 @@ impl Bezier {
...
+		let point_c_ratio = one_minus_t_term / (t_term + one_minus_t_term);
+		let c = point_c_ratio * start_point + (1. - point_c_ratio) * end_point;
+		let ab_bc_ratio = (t_term + one_minus_t_term - 1.).abs() / (t_term + one_minus_t_term);
+		let a = point_on_curve + (point_on_curve - c) / ab_bc_ratio;
+		[a, point_on_curve, c]
```
RobNadal, line 79: I think you could return a tuple `(a, points_on_curve, c)` here instead, and this would allow you to use destructuring syntax at the callsites (e.g. `let a, b, c = compute_abc(...)`), which is a bit cleaner

RobNadal (commented):
Something a little funky is happening on the UI when t = 0 or t = 1. The bezier snaps to a straight line, and the endpoint teleports to the (first) handle. If this is the expected behavior, we should change the slider settings to prevent them from choosing those values of t.

bezier-rs/lib/src/lib.rs
Keavon: You can still destructure with array syntax. Since these are sequential (i.e. the ordering is a relevant factor), an array seems slightly more appropriate, but the distinction is both subtle and opinionated. Either way is fine as far as I'm concerned.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -56,6 +63,27 @@ export default defineComponent({
...
+					createFromPoints: true,
+					template: markRaw(SliderExample),
+					templateOptions: { sliders: [{ ...tSliderOptions }] },
+					cubicOptions: {
+						sliders: [
+							{ ...tSliderOptions },
+							{
+								min: 0,
+								max: 100,
+								step: 10,
```
RobNadal, line 80: Could we decrease step size to make the animation a bit smoother?

hannahli2010:
> Something a little funky is happening on the UI when t = 0 or t = 1. The bezier snaps to a straight line, and the endpoint teleports to the (first) handle. If this is the expected behavior, we should change the slider settings to prevent them from choosing those values of t.

This happens because for `t=0` or `t=1`, by the function definition, we are trying to create a curve where the middle point is either at the beginning or the end of the curve. If the start/end point provided are not equal to that middle point, it is technically impossible to create such a curve! So I followed what bezier js does, and overwrite the corresponding start/end point to be equal to the middle point.

Since it's a unique edge case, I think it might be nice to include it in the interactive docs to demonstrate that it's expected behaviour. On the other hand, I also agree that the sudden snapping can seem weird. Let me know what you think with this additional context!

RobNadal:
> 

This makes sense, thanks! Please add a comment explaining this decision into the function.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -56,6 +63,34 @@ export default defineComponent({
 					callback: (): void => {},
 				},
 				{
+					id: 1,
```
RobNadal, line 66: remove ID from this array

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -56,6 +63,34 @@ export default defineComponent({
...
+								max: 100,
+								step: 5,
+								default: 10,
+								variable: "d1",
+							},
+						],
+					},
+				},
+				{
+					id: 2,
```
RobNadal, line 93: See above comment

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -22,10 +23,13 @@ class BezierDrawing {
 
 	options: Record<string, number>;
 
-	constructor(bezier: WasmBezierInstance, callback: BezierCallback, options: Record<string, number>) {
+	createFromPoints: boolean;
+
+	constructor(bezier: WasmBezierInstance, callback: BezierCallback, options: Record<string, number>, createFromPoints = false) {
```
RobNadal, line 28: could you rename this boolean to createThroughPoints?

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -105,7 +113,49 @@ class BezierDrawing {
...
+					  )
+					: WasmBezier.cubic_through_points(
+							this.points.map((p) => [p.x, p.y]),
+							this.options.t,
+							this.options.strut
+					  )
+			  )
+					.get_points()
+					.map((p) => JSON.parse(p))
+			: this.points;
```
Keavon, line 134: Please don't ever do this. Ternaries are for quick inline expressions.

bezier-rs/docs/interactive-docs/src/components/Example.vue
```diff
@@ -33,6 +32,10 @@ export default defineComponent({
 			type: Object as PropType<Record<string, number>>,
 			default: () => ({}),
 		},
+		createThroughPoints: {
+			type: Boolean,
```
Keavon, line 36: `as PropType<boolean>`

bezier-rs/docs/interactive-docs/src/components/ExamplePane.vue
```diff
@@ -39,33 +38,43 @@ export default defineComponent({
 			default: Example,
 		},
 		templateOptions: Object,
+		cubicOptions: {
+			type: Object,
+			default: null,
+		},
+		createThroughPoints: {
+			type: Boolean,
```
Keavon, line 46: Missing `PropType`s. Also `templateOptions` above is missing that too.

bezier-rs/docs/interactive-docs/src/components/SliderExample.vue
```diff
@@ -35,6 +34,10 @@ export default defineComponent({
 			type: Object,
 			default: () => ({}),
 		},
+		createThroughPoints: {
+			type: Boolean,
```
Keavon, line 38: PropType

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -94,10 +110,11 @@ export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], dragI
...
 	ctx.stroke();
 
-	drawLine(ctx, start, handleStart, strokeColor);
-	drawLine(ctx, end, handleEnd, strokeColor);
+	drawLine(ctx, start, handleStart, styleConfig.handleLineStrokeColor);
+	drawLine(ctx, end, handleEnd, styleConfig.handleLineStrokeColor);
 
 	points.forEach((point, index) => {
-		drawPoint(ctx, point, getPointSizeByIndex(index, points.length, radius), index === dragIndex ? COLORS.INTERACTIVE.SELECTED : strokeColor);
+		const strokeColor = index === 0 || index === points.length - 1 ? styleConfig.curveStrokeColor : styleConfig.handleStrokeColor;
```
Keavon, line 117: Please break out the `index === 0 || index === points.length - 1` into a named variable so the ternary is easier to read and so that boolean condition can be self-documenting with a name.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -55,6 +62,32 @@ export default defineComponent({
...
+								step: 0.01,
+								default: 0.5,
+								variable: "t",
+							},
+							{
+								min: 0,
+								max: 100,
+								step: 5,
+								default: 10,
+								variable: "strut",
```
Keavon, line 86: What does "strut" mean? The only definition I know for that word is "a rod or bar forming part of a framework and designed to resist compression" which presumably isn't the case here?

bezier-rs/lib/src/lib.rs
```diff
@@ -306,3 +330,39 @@ impl Bezier {
...
+	fn compare_points(p1: DVec2, p2: DVec2) -> bool {
+		DVec2::new(0.001, 0.001).cmpge(p1 - p2).all()
+	}
+
+	#[test]
+	fn quadratic_from_points() {
+		let p1 = DVec2::new(30., 50.);
+		let p2 = DVec2::new(140., 30.);
+		let p3 = DVec2::new(160., 170.);
+		let bezier = Bezier::quadratic_through_points(p1, p2, p3, 0.5);
```
Keavon, line 348: Either call this `bezier1` or consider using shadowing to call them all `bezier`. I also suggest separating these into blocks with blank lines, between the `let p1/2/3` lines and between each of the `let bezier` and `assert!` line groups.

bezier-rs/lib/src/utils.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Compute a, b and c for a quadratic curve that fits the start, end and point on curve at `t`.
+/// The definition for the a, b, c points are defined in [the projection identity section](https://pomax.github.io/bezierinfo/#abc) of Pomax's bezier curve primer.
+pub fn compute_abc_for_quadratic_through_points(start_point: DVec2, point_on_curve: DVec2, end_point: DVec2, t: f64) -> [DVec2; 3] {
+	let t_squared = t * t;
+	let one_minus_t = 1. - t;
+	let squared_one_minus_t = one_minus_t * one_minus_t;
+	compute_abc_through_points(start_point, point_on_curve, end_point, t_squared, squared_one_minus_t)
+}
+
+/// Compute a, b and c for a cubic curve that fits the start, end and point on curve at `t`.
```
Keavon, line 23: And here.

bezier-rs/docs/interactive-docs/src/App.vue
hannahli2010: I named this variable here to be consistent with the function parameter in the Rust code. The variable in the rust function is meant to represent the distance between `e1` and `e2` in [this section of the primer](https://pomax.github.io/bezierinfo/#abc) and on a high level, it can sort of be understood as controlling how wide the resulting cubic curve will be around point `t`. I don't really know what's the best name for this parameter - if you have a better suggestion, please let me know!

bezier-rs/lib/src/lib.rs
```diff
@@ -70,18 +72,40 @@ impl Bezier {
...
 
-	/// Create a cubic bezier curve that goes through 3 points. d1 represents the strut.
-	// #[inline]
-	pub fn cubic_from_points(p1: DVec2, p2: DVec2, p3: DVec2, _t: f64, _d1: f64) -> Self {
-		// TODO: Implement logic to get actual curve through the points
-		Bezier::from_quadratic_dvec2(p1, p2, p3)
+	/// Create a cubic bezier curve that goes through 3 points, where the middle point will be at the corresponding position `t` on the curve.
+	/// Note that when `t = 0` or `t = 1`, the expectation is that the `point_on_curve` should be equal to `start` and `end` respectively.
+	/// In these cases, if the provided values are not equal, this function will use the `point_on_curve` as the `start`/`end` instead.
+	/// * `strut` is a representation of the how wide the resulting curve will be by designating the distance between the `e1` and `e2` defined in [the projection identity section](https://pomax.github.io/bezierinfo/#abc) of Pomax's bezier curve primer.
```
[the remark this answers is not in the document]
hannahli2010: The asterisk represent a bullet list item. I chose to format the rustdoc message in this way because I wanted the explanation of the function parameter to be separate from the core function description. When it's yet another sentence in the paragraph, it's easier to miss. But there might be some better convention that I'm not aware of, so please let me know if there's a different format that you prefer.

image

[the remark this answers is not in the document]
Keavon: Ok, if you think it helps readability, that's good. But please use a dash instead of an asterisk since that's our preferred markdown format.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -55,6 +62,32 @@ export default defineComponent({
 					// eslint-disable-next-line
 					callback: (): void => {},
 				},
+				{
+					name: "Bezier from points",
```
Keavon, line 66: I think you want to call this "Bezier through points" and sort it to not be at the top.

hannahli2010: Good catch on the name!

I kept the function near the top because it's a variant of the constructor so it felt right to keep them together. It's also the order that is used by `bezierjs`, although we do not need to necessarily do the same thing. Do you have any particular preference for the order in general?

Keavon: I assumed you just had it at the top for convenience while testing. I trust your discretion.

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -105,7 +113,51 @@ class BezierDrawing {
...
+		const actualBezierPointLength = this.bezier.get_points().length;
+		let pointsToDraw = this.points;
+
+		const styleConfig = {
+			handleLineStrokeColor: COLORS.INTERACTIVE.STROKE_2,
+		};
+		let dragIndex = this.dragIndex;
+		if (this.createThroughPoints) {
+			if (actualBezierPointLength === 3) {
+				pointsToDraw = WasmBezier.quadratic_through_points(
```
Keavon, line 128: If you move these two arguments into variables defined above, can you get this to format itself onto one line?

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -105,7 +113,51 @@ class BezierDrawing {
...
+					this.options.t
+				)
+					.get_points()
+					.map((p) => JSON.parse(p));
+			} else {
+				pointsToDraw = WasmBezier.cubic_through_points(
+					this.points.map((p) => [p.x, p.y]),
+					this.options.t,
+					this.options.strut
+				)
```
Keavon, line 139: Here too.

Keavon (approved):

bezier-rs/docs/interactive-docs/src/App.vue
hannahli2010: Update: I've changed this variable name to `midpoint_separation` (suggested by Keavon)

PR #693, opened 2022-06-28
Implement Bezier local_extrema function

RobNadal:
Created local_extrema function, which returns two lists of t-values representing the local extrema of the x and y parametric curves respectively. The local_extrema are defined to be points at which the derivative of the curve is equal to zero

The math used to calculate extrema could return a value that is not within [0, 1]. Given this functionality may be useful in the future, we've elected to keep this behavior as a private function, and expose only the interface which first filters the points such that they lie within [0, 1].

### DIFF
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -228,7 +228,7 @@ export default defineComponent({
 				},
 				{
 					name: "Project",
-					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>, mouseLocation: Point | null): void => {
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>, mouseLocation?: Point): void => {
 						if (mouseLocation != null) {
 							const context = getContextFromCanvas(canvas);
 							const closestPoint = JSON.parse(bezier.project(mouseLocation.x, mouseLocation.y));
@@ -240,12 +240,12 @@ export default defineComponent({
 					name: "Local Extrema",
 					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance): void => {
 						const context = getContextFromCanvas(canvas);
-						const dimensionColours = [COLORS.NON_INTERACTIVE.STROKE_1, COLORS.NON_INTERACTIVE.STROKE_2];
+						const dimensionColors = [COLORS.NON_INTERACTIVE.STROKE_1, COLORS.NON_INTERACTIVE.STROKE_2];
 						const extrema: number[][] = JSON.parse(bezier.local_extrema());
 						extrema.forEach((tValues, index) => {
 							tValues.forEach((t) => {
 								const point = JSON.parse(bezier.compute(t));
-								drawPoint(context, point, 4, dimensionColours[index]);
+								drawPoint(context, point, 4, dimensionColors[index]);
 							});
 						});
 					},
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -240,8 +240,14 @@ export default defineComponent({
 					name: "Extrema",
 					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance): void => {
 						// const context = getContextFromCanvas(canvas);
-						const extrema = JSON.parse(bezier.extrema());
-						console.log(extrema);
+						const dimensionColours = ["red", "orange"];
+						const extrema: number[][] = JSON.parse(bezier.extrema());
+						extrema.forEach((tValues, index) => {
+							tValues.forEach((t) => {
+								const point = JSON.parse(bezier.compute(t));
+								drawPoint(getContextFromCanvas(canvas), point, 4, dimensionColours[index]);
+							});
+						});
 					},
 				},
 			],
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -239,13 +239,13 @@ export default defineComponent({
 				{
 					name: "Extrema",
 					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance): void => {
-						// const context = getContextFromCanvas(canvas);
-						const dimensionColours = ["red", "orange"];
-						const extrema: number[][] = JSON.parse(bezier.extrema());
+						const context = getContextFromCanvas(canvas);
+						const dimensionColours = [COLORS.NON_INTERACTIVE.STROKE_1, COLORS.NON_INTERACTIVE.STROKE_2];
+						const extrema: number[][] = JSON.parse(bezier.local_extrema());
 						extrema.forEach((tValues, index) => {
 							tValues.forEach((t) => {
 								const point = JSON.parse(bezier.compute(t));
-								drawPoint(getContextFromCanvas(canvas), point, 4, dimensionColours[index]);
+								drawPoint(context, point, 4, dimensionColours[index]);
 							});
 						});
 					},
--- a/bezier-rs/docs/interactive-docs/src/App.vue
+++ b/bezier-rs/docs/interactive-docs/src/App.vue
@@ -237,7 +237,7 @@ export default defineComponent({
 					},
 				},
 				{
-					name: "Extrema",
+					name: "Local Extrema",
 					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance): void => {
 						const context = getContextFromCanvas(canvas);
 						const dimensionColours = [COLORS.NON_INTERACTIVE.STROKE_1, COLORS.NON_INTERACTIVE.STROKE_2];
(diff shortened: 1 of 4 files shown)


bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,58 @@ impl Bezier {
...
+	/// The local extrema are defined to be points at which the derivative of the curve is equal to zero
+	fn _local_extrema(&self) -> [Vec<f64>; 2] {
+		let mut x_values = Vec::new();
+		let mut y_values = Vec::new();
+		match self.handles {
+			BezierHandles::Quadratic { handle } => {
+				let a = handle - self.start;
+				let b = self.end - handle;
+				// check for x extrema
+				if (b.x - a.x) != 0. {
```
hannahli2010, line 319: Very minor comment: you could do the subtraction as a `DVec2` operation and then re-use the results in 2 if statements

bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,58 @@ impl Bezier {
...
+					x_values.push(-a.x / (b.x - a.x));
```
hannahli2010, line 320: Minor comment: It probably won't make that much of a difference, but you can avoid recomputing the `b.x - a.x` by storing the difference earlier.

bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,58 @@ impl Bezier {
...
+				if (b.y - a.y) != 0. {
+					y_values.push(-a.y / (b.y - a.y));
+				}
+			}
+			BezierHandles::Cubic { handle_start, handle_end } => {
+				let a = 3. * (-self.start + 3. * handle_start - 3. * handle_end + self.end);
+				let b = 6. * (self.start - 2. * handle_start + handle_end);
+				let c = 3. * (handle_start - self.start);
+				// check for x extrema
+				let x_discr = b.x * b.x - 4. * a.x * c.x;
```
hannahli2010, line 332: Minor comment: Again, you could compute the discriminants in a `DVec2` and the use the `.x` and `.y` results for comparisons after.

Also use `discriminant` instead of `discr` for clarity.

bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,58 @@ impl Bezier {
...
+				if x_discr > 0. {
+					x_values.push((-b.x + x_discr.sqrt()) / (2. * a.x));
```
hannahli2010, line 334: Maybe include a comment explaining what these computations mean? If you move it a helper function, the function name might make things self explanatory.

bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,58 @@ impl Bezier {
...
+				} else if y_discr == 0. {
+					y_values.push(-b.y / (2. * a.y));
+				}
+			}
+		}
+		[x_values, y_values]
+	}
+
+	/// Returns two lists of t-values representing the local extrema of the x and y parametric curves respectively
+	/// The extrema returned are restricted to those which fall within [0, 1]
```
hannahli2010, line 353: Can you use the tick character to surround variable names and the `[0, 1]` in this `rustdoc` comment? Same with the comments description of `_local_extrema`.

hannahli2010 (changes requested):

bezier-rs/lib/src/lib.rs
```diff
@@ -305,4 +305,68 @@ impl Bezier {
...
+				[
+					Bezier::solve_quadratic(discriminant.x, two_times_a.x, b.x, c.x),
+					Bezier::solve_quadratic(discriminant.y, two_times_a.y, b.y, c.y),
+				]
+			}
+		}
+	}
+
+	/// Returns two lists of `t`-values representing the local extrema of the `x` and `y` parametric curves respectively
+	/// The extrema returned are restricted to those which fall within `[0, 1]`
```
Androxium, line 363: Could make the comment more descriptive like "points are filtered such that they fall within the range [0,1]".

bezier-rs/lib/src/lib.rs
```diff
@@ -331,6 +331,67 @@ impl Bezier {
 		};
 		bezier_starting_at_t1.split(adjusted_t2)[t2_split_side]
 	}
+
+	/// Find the roots of the linear equation ax + b
+	fn solve_linear(a: f64, b: f64) -> Vec<f64> {
```
hannahli2010, line 336: Now that https://github.com/GraphiteEditor/Graphite/pull/687 has been merged, you can moves this and `solve_quadratic` into the `utils.rs` file.

bezier-rs/lib/src/lib.rs
```diff
@@ -331,6 +331,67 @@ impl Bezier {
...
+		let mut roots = Vec::new();
+		if b != 0. {
+			roots.push(-b / a);
+		}
+		roots
+	}
+
+	/// Find the roots of the linear equation ax^2 + bx + c
+	fn solve_quadratic(discriminant: f64, two_times_a: f64, b: f64, c: f64) -> Vec<f64> {
```
hannahli2010, line 345: Can you include a note saying that `discriminant` and `two_times_a` are passed in as parameters so that the computation can be done before hand using the `DVec2`s?

bezier-rs/lib/src/lib.rs
```diff
@@ -331,6 +331,67 @@ impl Bezier {
...
+
+	/// Returns two lists of `t`-values representing the local extrema of the `x` and `y` parametric curves respectively
+	/// The local extrema are defined to be points at which the derivative of the curve is equal to zero
+	fn _local_extrema(&self) -> [Vec<f64>; 2] {
+		match self.handles {
+			BezierHandles::Quadratic { handle } => {
+				let a = handle - self.start;
+				let b = self.end - handle;
+				let b_minus_a = b - a;
+				[Bezier::solve_linear(b_minus_a.x, a.x), Bezier::solve_linear(b_minus_a.y, a.y)]
```
hannahli2010, line 369: Not an actual request, but I'm wondering if there's a way to iterate over the `x` and `y` parameters in the `DVec2` rather than needing to write them out like this. It might be possible since you can also refer to `.x` and `.y` using indices. It's not that big of a deal though.

I thought it might be helpful for extending 3D Beziers later on, but we'd need a new struct for that anyways so it doesn't really matter.

bezier-rs/lib/src/utils.rs
```diff
@@ -20,12 +20,39 @@ pub fn compute_abc_for_quadratic_through_points(start_point: DVec2, point_on_cur
...
 	let one_minus_t = 1. - t;
 	let cubed_one_minus_t = one_minus_t * one_minus_t * one_minus_t;
 
 	compute_abc_through_points(start_point, point_on_curve, end_point, t_cubed, cubed_one_minus_t)
 }
+
+/// Find the roots of the linear equation `ax + b`
+pub fn solve_linear(a: f64, b: f64) -> Vec<f64> {
+	let mut roots = Vec::new();
+	if b != 0. {
```
hannahli2010, line 36: Just want to confirm: is this check supposed to be for `a` or `b`?

hannahli2010 (commented):
Added a comment for one small thing. Other than that, LGTM!

hannahli2010 (approved):

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -226,6 +226,20 @@ export default defineComponent({
 						],
 					},
 				},
+				{
+					name: "Local Extrema",
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance): void => {
+						const context = getContextFromCanvas(canvas);
+						const dimensionColours = [COLORS.NON_INTERACTIVE.STROKE_1, COLORS.NON_INTERACTIVE.STROKE_2];
```
Keavon, line 233: Please spell it `Colors` since we have to be consistent and we arbitrarily decided on American English (well, since I'm American, but also because the Rust project uses American English).

bezier-rs/lib/src/lib.rs
```diff
@@ -331,6 +331,41 @@ impl Bezier {
 		};
 		bezier_starting_at_t1.split(adjusted_t2)[t2_split_side]
 	}
+
+	/// Returns two lists of `t`-values representing the local extrema of the `x` and `y` parametric curves respectively
```
Keavon, line 335: In doc comments and multi-sentence comments, please end sentences with periods.

bezier-rs/lib/src/lib.rs
```diff
@@ -331,6 +331,41 @@ impl Bezier {
...
+	/// The local extrema are defined to be points at which the derivative of the curve is equal to zero
+	fn _local_extrema(&self) -> [Vec<f64>; 2] {
```
Keavon, line 337: Underscore implies unused. Can you pick a more descriptive name for this?

Keavon: Or better yet, define this helper function as a closure within its using function. (Unless you use it in multiple functions throughout the library.)

Keavon (approved):

bezier-rs/lib/src/lib.rs
RobNadal: I'm just going to rename it, as I believe we may need to use it elsewhere.

PR #702, opened 2022-06-30
Welcome screen

### LINKED ISSUE #405: Welcome screen for empty tabs
Instead of opening a blank document, we should have a welcome screen (like VS Code's) which displays in place of the documents. It should show the Graphite logo, perhaps some relevant information about its alpha status and file format incompatibility, other informational messages, maybe links to tutorials or documentation in the future, and a big button to create a new document (with #523).

This should probably be, itself, a tab in the place of the document tabs. Since currently we can only display document tabs in the left column of the workspace, we may need to either hack something briefly or begin on #195. I think the best UX would be hardcoding the behavior where the welcome screen panel shows up in the place of where a document panel tab was closed if there are zero other places where document panels exist docked in the workspace (that way, the Properties and Layer Tree panel column doesn't take up the whole editor all of a sudden when closing document tabs).

mfish33:
Closes #405


Keavon:
Bug I just found:

1. Refresh the page to the welcome screen (no existing documents)
2. Open the New Document dialog
3. Uncheck *Infinite Canvas* and click OK to make the new document

The white rectangle for the canvas is stuck in the top left corner (if the default 1920x1080 would already fill the viewport at 100% scale at your monitor resolution and window size, refresh the page and this time set the document dimensions to something smaller).

The JS console displays this:  
image

The canvas can't be panned when it is frozen. Also, the displayed scale 0%, although entering a new value is possible and doing so unfreezes panning.

This doesn't happen for infinite canvas. It doesn't happen for any documents created after the first document is created upon loading the app (even if the first document is infinite canvas, even if the first document is closed to show the welcome screen again).

I don't have time to debug this today, but I wonder if you have any ideas @0HyperCube?

0HyperCube:
@Keavon This is due to the zoom document to fit having the viewport bounds as (0,0) and so doing a division by 0. The viewport bounds are requested to be updated but the dispatcher keeps the frontend message in a queue and handles the whole backend stack first. I added an exception in this case, so we immediately deal with the frontend message. I also simplified the new document dialog code to remove unnecessary messages (which were previously needed due to the message ordering).

editor/src/communication/dispatcher.rs
```diff
@@ -93,7 +93,7 @@ impl Dispatcher {
 				}
 				Frontend(message) => {
 					// Image and font loading should be immediately handled
-					if let FrontendMessage::UpdateImageData { .. } | FrontendMessage::TriggerFontLoad { .. } = message {
+					if let FrontendMessage::UpdateImageData { .. } | FrontendMessage::TriggerFontLoad { .. } | FrontendMessage::TriggerRefreshBoundsOfViewports = message {
 						self.responses.push(message);
 						return;
 					}
```
Keavon, line 99: If we put the `push` line before the `if` statement, with only the `return;` inside, does that keep the logic the same? Assuming so, that would remove a redundant line and help clarify the logic. (If that wouldn't be logically equivalent, then I'm very confused by how this works, what's the difference between the two `push` lines?)

Keavon:
> @Keavon This is due to the zoom document to fit having the viewport bounds as (0,0) and so doing a division by 0. The viewport bounds are requested to be updated but the dispatcher keeps the frontend message in a queue and handles the whole backend stack first. I added an exception in this case, so we immediately deal with the frontend message. I also simplified the new document dialog code to remove unnecessary messages (which were previously needed due to the message ordering).

Excellent sleuthing, quite an interesting reminder about the challenges we'll continue to face with message ordering where vigilance and testing will be required. I'm glad you also noticed that you could clean up the code now so that didn't become an unneeded relic, thank you for keeping the code lean and clean!

editor/src/communication/dispatcher.rs
0HyperCube: The current way keeps the borrow checker happy with only a very simple line duplicated.

PR #708, opened 2022-07-01
Implement function to find intersections between a Bezier and a linear line

hannahli2010:
This function is implemented on top of https://github.com/GraphiteEditor/Graphite/pull/693 because it can re-use the root finding helper functions.

Functions implemented:
- `rotate` and `translate`
  -  Needed for the `line_intersection` function
- `line_intersection`
  - Rotates and translates the Bezier and line such that the line is on top of the x-axis, then perform root finding to get intersection points
  - Root finding is done with the `y` coordinates as the variable, since we want to find when `x = 0`
- Add tests for cubic root finding

UI changes
- Add sections for `rotate` and `line_intersection`

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -1,3 +1,5 @@
+//! Bezier-rs: A Bezier Math Library for Rust
+
 use glam::{DMat2, DVec2};
 
 mod utils;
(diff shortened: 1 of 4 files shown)


Androxium (approved):

TrueDoctor (approved):

bezier-rs/lib/src/utils.rs
```diff
@@ -64,3 +66,138 @@ pub fn solve_quadratic(discriminant: f64, two_times_a: f64, b: f64, c: f64) -> V
...
+		// Refactor cubic to be of the form: a(t^3 + pt + q), derivation from: https://trans4mind.com/personal_development/mathematics/polynomials/cubicAlgebra.htm
+		let p = (3. * new_b - new_a * new_a) / 3.;
+		let q = (2. * new_a.powi(3) - 9. * new_a * new_b + 27. * new_c) / 27.;
+		let discriminant = (p / 3.).powi(3) + (q / 2.).powi(2);
+		solve_reformatted_cubic(discriminant, new_a, p, q)
+	}
+}
+
+/// Compare two `f64` numbers with a provided max absolute value difference.
+pub fn compare_f64(f1: f64, f2: f64, max_abs_diff: f64) -> bool {
```
TrueDoctor, line 140: would it make more sense to call this f64_approximately_equals to be consistent with the in range functions?

bezier-rs/lib/src/lib.rs
```diff
@@ -199,11 +201,9 @@ impl Bezier {
 		}
 	}
 
-	/// Calculate the point on the curve based on the `t`-value provided.
-	/// Basis code based off of pseudocode found here: <https://pomax.github.io/bezierinfo/#explanation>.
-	pub fn compute(&self, t: f64) -> DVec2 {
-		assert!((0.0..=1.0).contains(&t));
-
+	///  Calculate the point on the curve based on the `t`-value provided.
+	///  Basis code based off of pseudocode found here: <https://pomax.github.io/bezierinfo/#explanation>.
```
Keavon, line 205: Please revert this change which added another space after the `///` (unless there's a reason for it?).

bezier-rs/lib/src/lib.rs
```diff
@@ -218,6 +218,13 @@ impl Bezier {
 		}
 	}
 
+	///  Calculate the point on the curve based on the `t`-value provided.
+	///  Expects `t` to be within the inclusive range `[0, 1]`.
```
Keavon, line 222: Here too, remove double spaces.

bezier-rs/lib/src/lib.rs
```diff
@@ -453,15 +460,89 @@ impl Bezier {
...
+			}
+			BezierHandles::Cubic { handle_start, handle_end } => {
+				let transformed_handle_start = transformation_function(handle_start);
+				let transformed_handle_end = transformation_function(handle_end);
+				Bezier::from_cubic_dvec2(transformed_start, transformed_handle_start, transformed_handle_end, transformed_end)
+			}
+		}
+	}
+
+	/// Returns a Bezier curve that results from rotating the curve by the given angle (in radians).
```
Keavon, line 481: Please mention relative to what pivot (I assume the origin of the coordinate system that the Bezier is defined in?).

bezier-rs/lib/src/lib.rs
```diff
@@ -453,15 +460,89 @@ impl Bezier {
...
+		self.apply_transformation(&|point| rotation_matrix.mul_vec2(point))
+	}
+
+	/// Returns a Bezier curve that results from translating the curve by the given `DVec2`.
+	pub fn translate(&self, translation: DVec2) -> Bezier {
+		self.apply_transformation(&|point| point + translation)
+	}
+
+	/// Returns a list of points where the provided `line` intersects with the Bezier curve.
+	/// - `line` - Expected to be received in the format of `[start_point, end_point]`.
```
Keavon, line 493: I think a comment like "A line passing through \`[start_point, end_point]\`." would be clearer to remove any potential for confusion with a line segment. (Or is it actually a line segment?)

bezier-rs/lib/src/lib.rs
```diff
@@ -453,15 +460,89 @@ impl Bezier {
...
+
+		// Compute the roots of the resulting bezier curve
+		let list_intersection_t = match translated_bezier.handles {
+			BezierHandles::Quadratic { handle } => {
+				let a = translated_bezier.start.y - 2. * handle.y + translated_bezier.end.y;
+				let b = 2. * (handle.y - translated_bezier.start.y);
+				let c = translated_bezier.start.y;
+				let discriminant = b * b - 4. * a * c;
+				let two_times_a = 2. * a;
+				utils::solve_quadratic(discriminant, two_times_a, b, c)
```
Keavon, line 514: Just a style suggestion in general, please try and break logical groups of lines into their own sections with blank lines. For example the `let a/b/c` lines, `let discriminant/two_times_a`, and `utils::solve_quadratic` lines seem like good groups here.

An exception to this suggestion is above though, at the top of this function, where there's a comment referring to the `let slope` through `let rotated_line` lines. Even though it would be nice to break the `let slope/angle` and `let rotation_*` lines, the need to keep that comment together with its group is of more importance. I'm just mentioning this to provide some more context of how I like to think of the tradeoffs of when and when to not use blank lines to separate things. Use your discretion but let's also try and keep the code breathable instead of being too compact. So far I've seen a pretty good job being done already in this regard, so I'm just bringing it up because I noticed here where it could be done.

bezier-rs/lib/src/lib.rs
```diff
@@ -453,15 +460,89 @@ impl Bezier {
...
+		};
+		let min = line[0].min(line[1]);
+		let max = line[0].max(line[1]);
+		let max_abs_diff = 1e-4;
+
+		list_intersection_t
+			.iter()
+			.filter(|&&t| utils::f64_approximately_in_range(t, 0., 1., max_abs_diff))
+			.map(|&t| self.unrestricted_compute(t))
+			.filter(|&p| utils::dvec2_approximately_in_range(p, min, max, max_abs_diff).all())
```
Keavon, line 533: Does `p` mean `point`?

Keavon (approved):
LGTM!

bezier-rs/lib/src/lib.rs
hannahli2010: Not sure how that happened, but thanks for catching it!

bezier-rs/lib/src/lib.rs
hannahli2010: It is actually a line segment, I'll change the description to make that more clear.

bezier-rs/lib/src/lib.rs
hannahli2010: Yup! I'll change it to `point` instead of `p`.

PR #710, opened 2022-07-01
Fix problem with compiling bezier-rs-wasm

hannahli2010:
- Fix issue where `npm run serve` would occasionally fail with no description.
  - The watch list for the plugin was not pointing to the correct location
- Adjust config files to match that of `frontend`
  - Remove `jsconfig.json`
  - change configuration in `tsconfig.json`


Keavon (approved):

Keavon:
Glad you found the problem!

PR #711, opened 2022-07-03
Implement Bezier curve reduce function

RobNadal:
- Added `reduce` function, which splits a Bezier curve into scalable segments based on the criteria defined in the [bezier primer](https://pomax.github.io/bezierinfo/#offsetting)
  - The function first splits the curve by its local extrema. 
  - If the resulting subcurves are still not scalable, the algorithm naively attempts to split each subcurve in half, which is generally sufficient for generating scalable segments. 
  - If the naive split does generate scalable segments, the algorithm switches to a greedy sliding window strategy of trimming maximally large scalable segments off the start of the subcurve
- Added `drawHandles` argument to `bezierStyleConfig`, to allow handles to be omitted in a bezier drawing


bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -15,10 +15,15 @@ struct Point {
 pub struct WasmBezier(Bezier);
 
 /// Convert a `DVec2` into a `JsValue`.
-pub fn vec_to_point(p: &DVec2) -> JsValue {
+fn vec_to_point(p: &DVec2) -> JsValue {
 	JsValue::from_serde(&serde_json::to_string(&Point { x: p.x, y: p.y }).unwrap()).unwrap()
 }
 
+/// Convert a bezier to a list of points
```
hannahli2010, line 22: Add a period at the end of this comment.

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,91 @@ impl Bezier {
...
+			let angle_2 = (self.end - self.start).angle_between(handle_end - self.start);
+			if (angle_1 > 0. && angle_2 < 0.) || (angle_1 < 0. && angle_2 > 0.) {
+				return false;
+			}
+		}
+		// Verify the on-curve point for t = 0.5 occurs roughly in the center of the polygon.
+		let normal_0 = self.normal(0.);
+		let normal_1 = self.normal(1.);
+		let s = normal_0.x * normal_1.x + normal_0.y * normal_1.y;
+		f64::abs(f64::acos(s)) < std::f64::consts::PI / 3.
```
hannahli2010, line 474: You can probably move some of this to an import at the top of the file, or do `s.acos()` instead.

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,91 @@ impl Bezier {
...
+	}
+
+	/// Split the curve into a number of scalable subcurves. This function may introduce gaps if subsections of the curve are not reducable.
+	/// `step_size` dictates the granularity at which the function searches for reducable subcurves.
```
hannahli2010, line 478: For comments on function parameters, I've been adding them as a bullet list underneath the main description so that it's easier to locate. i.e. it would be:
``/// - `step_size` dictates...``

I think the sentence after this can be part of the same bullet point. Also don't forget to add a period at the end of it.

hannahli2010: Add a comment that states the default value, or don't make it optional (there's a few functions I did where I made the params mandatory instead of optional, but maybe I should change that...)

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,92 @@ impl Bezier {
...
+
+		// Split each subcurve such that each resulting segment is scalable.
+		let mut result: Vec<Bezier> = Vec::new();
+		subcurves.iter().for_each(|&subcurve| {
+			// Do no processing on the subcurve it is already scalable.
+			if subcurve.is_scalable() {
+				result.push(subcurve);
+				return;
+			}
+			// It is generally sufficient to split an unscalable subcurve at `t = 0.5` to generate two scalable segments.
```
jackiechen73, line 505: Where is this from? Maybe we should reference the source if we're going to state this (just for future reference sake)

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,92 @@ impl Bezier {
...
+			let angle_2 = (self.end - self.start).angle_between(handle_end - self.start);
+			if (angle_1 > 0. && angle_2 < 0.) || (angle_1 < 0. && angle_2 > 0.) {
+				return false;
+			}
+		}
+		// Verify the on-curve point for t = 0.5 occurs roughly in the center of the polygon.
+		let normal_0 = self.normal(0.);
+		let normal_1 = self.normal(1.);
+		let s = normal_0.x * normal_1.x + normal_0.y * normal_1.y;
+		s.acos() < std::f64::consts::PI / 3.
```
linda-zheng, line 474: how did you pick this threshold? maybe make it evident that it is a constant (i.e. put it into a constant) so that future developers can change the threshold if needed.

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,95 @@ impl Bezier {
...
+		let normal_1 = self.normal(1.);
+		let s = normal_0.x * normal_1.x + normal_0.y * normal_1.y;
+		s.acos() < Bezier::SCALABLE_CURVE_MAX_ANGLE
+	}
+
+	/// Split the curve into a number of scalable subcurves. This function may introduce gaps if subsections of the curve are not reducable.
+	/// The function takes the following parameter:
+	/// - `step_size` - Dictates the granularity at which the function searches for reducable subcurves. The default value is `0.01`.
+	///   A small granularity may increase the chance the function does not introduce gaps, but will increase computation time.
+	pub fn reduce_with_step_size(&self, step_size: f64) -> Vec<Bezier> {
```
Keavon, line 565: I think it would be better to just have one function, with the last argument an `Option`.

RobNadal: Luckily I added only that second function in the most recent commit, so I'll just revert it.

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+			let angle_1 = (self.end - self.start).angle_between(handle_start - self.start);
+			let angle_2 = (self.end - self.start).angle_between(handle_end - self.start);
+			if (angle_1 > 0. && angle_2 < 0.) || (angle_1 < 0. && angle_2 > 0.) {
+				return false;
+			}
+		}
+		// Verify the on-curve point for t = 0.5 occurs roughly in the center of the polygon.
+		let normal_0 = self.normal(0.);
+		let normal_1 = self.normal(1.);
+		let s = normal_0.x * normal_1.x + normal_0.y * normal_1.y;
```
Keavon, line 557: Use a full descriptive variable name for `s` please, I don't know what that stands for.

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+		s.acos() < Bezier::SCALABLE_CURVE_MAX_ANGLE
+	}
+
+	/// Split the curve into a number of scalable subcurves. This function may introduce gaps if subsections of the curve are not reducable.
```
Keavon, line 561: reducible* (here and two lines below)

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+		let step_size = step_size.unwrap_or(0.01);
+
+		let mut extrema: Vec<f64> = self.local_extrema().into_iter().flatten().collect::<Vec<f64>>();
+		extrema.append(&mut vec![0., 1.]);
+		extrema.dedup();
+		extrema.sort_by(|ex1, ex2| ex1.partial_cmp(ex2).unwrap());
+
+		// Split the curve on the extremas. Simplifies procedure for ensuring each curve can be scaled.
+		let mut subcurves: Vec<Bezier> = Vec::new();
+		let mut t1: f64 = extrema[0];
```
Keavon, line 575: The explicit type is unnecessary.

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+		for t2 in extrema.iter().skip(1) {
+			subcurves.push(self.trim(t1, *t2));
+			t1 = *t2;
+		}
+
+		// Split each subcurve such that each resulting segment is scalable.
+		let mut result: Vec<Bezier> = Vec::new();
+		subcurves.iter().for_each(|&subcurve| {
+			// Do no processing on the subcurve it is already scalable.
```
Keavon, line 584: `// Perform no processing on the subcurve if it's already scalable.`

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+			if subcurve.is_scalable() {
+				result.push(subcurve);
+				return;
+			}
+			// From the above primer, it is generally sufficient to split subcurves with no local extrema at `t = 0.5` to generate two scalable segments.
```
Keavon, line 589: `// According to <https://pomax.github.io/bezierinfo/#offsetting>, it is generally...`

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +538,91 @@ impl Bezier {
...
+			// Greedily iterate across the subcurve at intervals of size `step_size` to break up the curve into maximally large segments
+			let mut segment: Bezier;
+			let mut t1 = 0.;
+			let mut t2 = step_size;
+			while t2 <= 1. + step_size {
+				segment = subcurve.trim(t1, f64::min(t2, 1.));
+				if !segment.is_scalable() {
+					t2 -= step_size;
+
+					// If the previous step does not exist, the start of the subcurve is unreducable.
```
Keavon, line 606: un-reducible*

Keavon (approved):

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -20,6 +20,15 @@ export const isIndexFirstOrLast = (index: number, arrayLength: number): boolean
...
 export const getPointSizeByIndex = (index: number, numPoints: number, radius = DEFAULT_ENDPOINT_RADIUS): number => (isIndexFirstOrLast(index, numPoints) ? radius : radius * HANDLE_RADIUS_FACTOR);
 
+// Based on an `index`, generate a color from the light spectrum such that sequential indices generate sequential colors.
+// Frequency parameter is proportional to the rate at which the colors shift through the spectrum. Generated colours are periodic in `index`.
+export const getSpectrumColor = (index: number, frequency = 0.5): string => {
+	const red = (Math.round(Math.sin(frequency * index - 2) * 127) + 128).toString(16).padStart(2, "0");
+	const green = (Math.round(Math.sin(frequency * index - 4) * 127) + 128).toString(16).padStart(2, "0");
+	const blue = (Math.round(Math.sin(frequency * index - 6) * 127) + 128).toString(16).padStart(2, "0");
+	return `#${red}${green}${blue}`.toUpperCase();
+};
```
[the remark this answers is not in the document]
RobNadal: Where does the `hsv()` function originate from?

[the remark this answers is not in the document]
Keavon: It is CSS syntax, see that link.

bezier-rs/lib/src/lib.rs
```diff
@@ -33,6 +33,8 @@ pub struct Bezier {
 }
 
 impl Bezier {
+	const SCALABLE_CURVE_MAX_ANGLE: f64 = std::f64::consts::PI / 3.;
```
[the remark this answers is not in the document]
RobNadal: I'll modify the usages and any other consts introduced by this pr, but we should probably leave the modification of any other default values to #713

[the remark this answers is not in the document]
Keavon: Agreed. It sounds like the plan is to define the defaults in the `impl Default` for each options struct. So we won't have a `consts.rs` file. Do whatever, or nothing, to provide the least friction towards that goal.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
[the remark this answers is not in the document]
RobNadal: Ah my mistake, I didn't even notice the link. Very interesting tool, serves the use case perfectly so no need anymore for `getSpectrumColor `

PR #712, opened 2022-07-03
Implement function to find inflection points of a Bezier curve

linda-zheng:
Implemented `unrestricted_inflections` and `inflections`  functions in bezier-rs directory.
- `unrestricted_inflections` finds the t-values corresponding to the inflection points of the curve.
- `inflections` calls `unrestricted_inflections` and filters the t-values, only keeping the t-values between [0, 1].

 Added inflection section to the UI.


bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -107,4 +107,9 @@ impl WasmBezier {
 		let local_extrema = self.0.local_extrema();
 		JsValue::from_serde(&serde_json::to_string(&local_extrema).unwrap()).unwrap()
 	}
+
+	pub fn inflections(&self) -> JsValue {
+		let inflections = self.0.inflections();
+		JsValue::from_serde(&serde_json::to_string(&inflections).unwrap()).unwrap()
```
RobNadal, line 113: Could we create a template function that does the `serde` / `json` wrapping and then use it both here and in extrema?

RobNadal (changes requested):
I think we should test the alternative implementation for inflection which involves rotating the curve, to see if it's a cleaner solution. @hannahli2010 already has that functionality on the line-intersection branch, so it shouldnt be too difficult.

RobNadal (approved):

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +721,48 @@ impl Bezier {
...
+		match self.handles {
+			// There exists no inflection points for linear and quadratic beziers.
+			BezierHandles::Linear => Vec::new(),
+			BezierHandles::Quadratic { .. } => Vec::new(),
+			BezierHandles::Cubic { .. } => {
+				// Axis align the curve.
+				let translated_bezier = self.translate(-self.start);
+				let angle = translated_bezier.end.angle_between(DVec2::new(1., 0.));
+				let rotated_bezier = translated_bezier.rotate(angle);
+				match rotated_bezier.handles {
```
Keavon, line 738: Use `if let` instead of `match`.

Keavon: @linda-zheng thanks for the changes. For future reference, please click "Resolve conversation" for anything that's been taken care of so it simplifies the view of what's left to be done. Thanks for the 👍 though, I've gone ahead and clicked resolve on all the ones you gave that reaction for.

Keavon:
We have some merge conflicts as a result of the PR that merged today, and one unresolved comment from Rob. If we can resolve those issues, feel free to hit merge.

PR #713, opened 2022-07-03
Make certain Bezier function parameters optional and other refactors

hannahli2010:
Refactor some functions to use optional parameters with default values.

Additional changes:
- Add `start` command to `package.json`
- Remove facvicon and update `index.html`
- Various stylistic changes
- Replaced derivative with tangent in the UI
- Changed normal and tangent to not show the reversed vector
- Rename `intersection_line` to `intersect_line_segement`
- Update comments


RobNadal:
There's so many options for project it could be worth creating a version of the function that just calls `project` with all the defaults so that a user who doesn't care about the options doesn't haven't to pass in `None` so many times. I just have no idea what to call such a function. Maybe just append a preposition for the one with parameters and call it `project_by` or `project_with`, or something

hannahli2010:
> There's so many options for project it could be worth creating a version of the function that just calls `project` with all the defaults so that a user who doesn't care about the options doesn't haven't to pass in `None` so many times. I just have no idea what to call such a function. Maybe just append a preposition for the one with parameters and call it `project_by` or `project_with`, or something

@Keavon Do you have any recommendations or preferences for this?

Keavon:
> @Keavon Do you have any recommendations or preferences for this?

@0HyperCube @TrueDoctor what's the idiomatic way to do this?

Keavon (approved):
Looks okay to merge but let's wait to hear about the idiomatic way to do this.

TrueDoctor:
Afaik using optional parameters is fine and as long as you have a small number of commonly used variations, you can create functions in the form of `function_with_pivot`. As soon as that becomes cumbersome to do, you usually employ the builder pattern  to cut down on the exponential number of requiried functions. This can be made easier by using a proc macro: https://docs.rs/derive_builder/0.11.2/derive_builder/

otdavies:
> Afaik using optional parameters is fine and as long as you have a small number of commonly used variations, you can create functions in the form of `function_with_pivot`. As soon as that becomes cumbersome to do, you usually employ the builder pattern  to cut down on the exponential number of requiried functions. This can be made easier by using a proc macro: https://docs.rs/derive_builder/0.11.2/derive_builder/

I second the builder pattern here. I think situations where you have greater than 2 optional parameters would be better suited to the builder pattern.

hannahli2010:
> Afaik using optional parameters is fine and as long as you have a small number of commonly used variations, you can create functions in the form of `function_with_pivot`. As soon as that becomes cumbersome to do, you usually employ the builder pattern to cut down on the exponential number of requiried functions. This can be made easier by using a proc macro: https://docs.rs/derive_builder/0.11.2/derive_builder/

Just to confirm, would `function_with_pivot` be a wrapped version of the function that does not require the user to pass in a value for the optional pivot?

TrueDoctor:
> 
> > Afaik using optional parameters is fine and as long as you have a small number of commonly used variations, you can create functions in the form of `function_with_pivot`. As soon as that becomes cumbersome to do, you usually employ the builder pattern to cut down on the exponential number of requiried functions. This can be made easier by using a proc macro: https://docs.rs/derive_builder/0.11.2/derive_builder/
> 
> Just to confirm, @TrueDoctor would `function_with_pivot` be a version of the function that **does** or **does not** require the user to pass in a value for the optional pivot?

It would be a function that takes a non optional pivot

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -265,16 +256,17 @@ export default defineComponent({
...
 								variable: "angle",
-								min: -90,
-								max: 90,
-								step: 5,
-								default: 15,
+								min: 0,
+								max: 2,
+								step: 1 / 16,
+								default: 1 / 8,
+								unit: "pi",
```
Keavon, line 263: Use `π` instead please, and remove the preceding space.

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -13,6 +13,7 @@ export const COLORS = {
 	NON_INTERACTIVE: {
 		STROKE_1: "red",
 		STROKE_2: "orange",
+		STROKE_3: "green",
```
Keavon, line 16: So because the standard colors for the X, Y, Z axes are Red, Green, and Blue (RGB -> XYZ) in virtually all software, I'm worried about the maintainability of the Local Extrema example changing colors in the future if we change this color scheme defined here. I'd therefore suggest not using these color consts and hardcoding `"red"` and `"green"` for only the circles and text used in the Local Extrema example. (Then you can remove `STROKE_3` since it isn't used elsewhere yet.)

bezier-rs/lib/src/lib.rs
```diff
@@ -75,9 +95,11 @@ impl Bezier {
 	}
 
 	/// Create a quadratic bezier curve that goes through 3 points, where the middle point will be at the corresponding position `t` on the curve.
+	/// - `t` - A representation of how far along the curve the provide point should occur at. The default value is 0.5.
```
Keavon, line 98: provided*

bezier-rs/lib/src/lib.rs
```diff
@@ -89,17 +111,21 @@ impl Bezier {
 	}
 
 	/// Create a cubic bezier curve that goes through 3 points, where the middle point will be at the corresponding position `t` on the curve.
+	/// - `t` - A representation of how far along the curve the provide point should occur at. The default value is 0.5.
```
Keavon, line 114: provided*

bezier-rs/lib/src/lib.rs
```diff
@@ -89,17 +111,21 @@ impl Bezier {
...
 		if t == 0. {
 			return Bezier::from_cubic_dvec2(point_on_curve, point_on_curve, end, end);
 		}
 		if t == 1. {
 			return Bezier::from_cubic_dvec2(start, start, point_on_curve, point_on_curve);
 		}
-		let [a, b, _] = utils::compute_abc_for_cubic_through_points(start, point_on_curve, end, t);
+		let [a, b, c] = utils::compute_abc_for_cubic_through_points(start, point_on_curve, end, t);
+		let default_midpoint_separation = b.distance(c);
+		let midpoint_separation = midpoint_separation.unwrap_or(default_midpoint_separation);
```
Keavon, line 128: Avoid wasting computation on calculating `default_midpoint_separation` if it's only sometimes needed. You can use `unwrap_or_else` instead of `unwrap_or` to achieve that:
```rust
let midpoint_separation = midpoint_separation.unwrap_or_else(|| b.distance(c));
```

bezier-rs/lib/src/lib.rs
```diff
@@ -340,12 +367,14 @@ impl Bezier {
...
-	/// - `convergence_limit` - Controls the number of iterations needed to consider that minimum distance to have converged.
-	/// - `iteration_limit` - Controls the maximum total number of iterations to be used.
-	pub fn project(&self, point: DVec2, lut_size: i32, convergence_epsilon: f64, convergence_limit: i32, iteration_limit: i32) -> DVec2 {
+	/// Uses a searching algorithm akin to binary search that can be customized using the [ProjectionOptions] structure.
+	pub fn project(&self, point: DVec2, options: ProjectionOptions) -> DVec2 {
+		// Unwrap optional parameters
+		let lut_size = options.lut_size;
+		let convergence_epsilon = options.convergence_epsilon;
+		let convergence_limit = options.convergence_limit;
+		let iteration_limit = options.iteration_limit;
```
Keavon, line 376: Destructuring would probably be nicer:
```rust
let ProjectionOptions {
	lut_size,
	convergence_epsilon,
	convergence_limit,
	iteration_limit,
} = options;
```

bezier-rs/lib/src/lib.rs
```diff
@@ -540,10 +569,11 @@ impl Bezier {
 
 #[cfg(test)]
 mod tests {
-	use crate::utils;
-	use crate::Bezier;
 	use glam::DVec2;
 
+	use super::*;
+	use crate::utils;
```
Keavon, line 575: Please put this group above the `use glam::DVec2;` group since our ordering is *own crate* then *other first-party crates* then *std or third-party crates*.

bezier-rs/lib/Cargo.toml
```diff
@@ -12,3 +12,4 @@ license = "Apache-2.0"
 
 [dependencies]
 glam = { version = "0.17", features = ["serde"] }
+derive_builder = "0.11.2"
```
Keavon, line 15: I am open to being convinced otherwise, but I don't feel like the builder pattern is quite right for our needs (it seems to be generally the right idea to use another struct, but overkill to have this infrastructure surrounding the options struct). How about just defining the options struct (which derives `Default`) and just manually building it the normal way (like below)?

```rust
FooOptions::default()
```
or:
```rust
FooOptions {
    quick: 123,
    brown: 456,
    fox: 789,
    ..Default::default()
}
```

hannahli2010: That sounds good to me, but I'm no expert with Rust. @TrueDoctor @otdavies Do you two have any additional thoughts on this? I also might not be using the builder pattern correctly, so I'd appreciate your thoughts on this!

TrueDoctor: One advantage is that you don't have to provide your own default implementation by using the struct annotations.
The second advantage is that this works for fileds that don't implement default.
You can for example have a mix of required fields and optional fields which wouldn't be the case with the struct initializer.
You could of course provide your own default struct 
```rust
FooOptions {
 quick: 123,
 brown: 456,
 fox: 789,
  ..MyDefauilt }
```
but that would require the field types to implement Clone.
In this singular example the advantage of using the builder_derive crate is negligable but I thought about introducing this dep anyway because it might be useful in other areas of the project.

hannahli2010: @Keavon based on the additional comment, do you still prefer the struct implementation over builder? I don't have a particular preference as I agree that builder isn't super necessary here, but it may be nice to include the dependency now if we might use it in the future.

Keavon: Since this is a library used by others, is especially important to keep down on the dependencies. I think that need outweighs the slight benefits @TrueDoctor mentioned (but thanks for describing those!). Let's stick to the simple option and just implement default.

Keavon (approved):

Issue #694, opened 2022-06-30
Provide a way for multiple systems to subscribe to message broadcasts


Keavon:
Things like `DocumentIsDirty` shouldn't have to be carefully routed to its many concerned parties. Those parties should subscribe to a `DocumentIsDirty` event.

PR #714, opened 2022-07-03
Fix select tool deselect all

0HyperCube:
The select tool doesn't clean up its overlays when all layers were deselected - this is because #694 no longer calls `DocumentIsDirty` on selection change so we have to call the recreate overlays on `SelectionChanged` as well.


Keavon (approved):

TrueDoctor (approved):

Keavon:
Is this waiting on anything?

PR #715, opened 2022-07-04
Implement bezier library's De Casteljau points function

jackiechen73:
- Added bezier `hull(t)` function
    - The function returns a list of lists of points
    - Each list of points represents the points in an iteration of the hull
    - Each iteration is calculated using de Castelijau's algorithm https://pomax.github.io/bezierinfo/#decasteljau


bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,46 @@ impl Bezier {
 			.try_into()
 			.unwrap()
 	}
+
+	/// Returns a list of lists of points representing the hull points for all iterations at the point corresponding to `t` using de Casteljau's algorithm
+	/// the ith list represents the points of the ith iteration
+	/// https://pomax.github.io/bezierinfo/#decasteljau
```
hannahli2010, line 459: Can you change the last line to say something like: `More information on the algorithm can be found in the [de Casteljau section in Pomax's Bezier Primer](https://pomax.github.io/bezierinfo/#decasteljau).`

Also btw, for urls, you can add `<url-here>` around it in the markdown so that rustdoc renders it as a hyperlink.

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,46 @@ impl Bezier {
...
+	pub fn hull(&self, t: f64) -> Vec<Vec<DVec2>> {
+		let mut current_points = Vec::new();
+		let mut hull_points = Vec::new();
+
+		match self.handles {
+			BezierHandles::Quadratic { handle } => {
+				current_points.push(self.start);
```
hannahli2010, line 466: You can move the push for `start` and `end` outside of the match statement (`start` in front, and `end` after).

bezier-rs/lib/src/lib.rs
```diff
@@ -453,6 +453,46 @@ impl Bezier {
...
+				current_points.push(handle);
+				current_points.push(self.end);
+
+				hull_points.push(vec![self.start, handle, self.end]);
```
hannahli2010, line 470: You can move the `hull_points.push(...)` after the match statement and push a cloned version of `current_points`.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -250,6 +237,36 @@ export default defineComponent({
...
+								const prevPoint: Point = { x: iteration[index - 1][0], y: iteration[index - 1][1] };
+								drawLine(getContextFromCanvas(canvas), newPoint, prevPoint, COLORS.NON_INTERACTIVE.STROKE_1);
+							});
+						});
+					},
+					template: markRaw(SliderExample),
+					templateOptions: {
+						sliders: [
+							{
+								min: 0.01,
```
hannahli2010, line 261: Does `hull` have any weird behaviour at `t = 0` or `t = 1`? A lot of the points will simply be stacked on top of each other right? I think you can use the following instead (i.e. change to `min: 0` and `max: 1`):

```
templateOptions: { sliders: [tSliderOptions] }
```

Keavon:
I see I was requested to review, but it doesn't look like @hannahli2010's previous round of comments was addressed yet. I'll wait until those are resolved and then take a look, please ping me when that's ready.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -250,6 +237,34 @@ export default defineComponent({
...
+									return;
+								}
+								const prevPoint: Point = iteration[index - 1];
+								drawLine(getContextFromCanvas(canvas), point, prevPoint, COLORS.NON_INTERACTIVE.STROKE_1);
+							});
+						});
+					},
+					template: markRaw(SliderExample),
+					templateOptions: {
+						sliders: [
```
RobNadal, line 257: We created a default for `t` sliders since we use it a lot. Use this instead of creating your own.
`templateOptions: { sliders: [tSliderOptions] },`

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +536,40 @@ impl Bezier {
 			.filter(|&point| utils::dvec2_approximately_in_range(point, min, max, max_abs_diff).all())
 			.collect::<Vec<DVec2>>()
 	}
+
+	/// Returns a list of lists of points representing the hull points for all iterations at the point corresponding to `t` using de Casteljau's algorithm
+	/// the ith element of the list represents the points of the ith iteration
```
RobNadal, line 541: Capitalize the start of sentences and add periods at the end of comments. Also, it would likely be beneficial to add some explanation about what a hull point is.

bezier-rs/lib/src/lib.rs
```diff
@@ -536,6 +536,40 @@ impl Bezier {
...
+				current_points.push(handle_start);
+				current_points.push(handle_end);
+			}
+		}
+		current_points.push(self.end);
+		hull_points.push(current_points.clone());
+
+		while current_points.len() > 1 {
+			let mut next_points = Vec::new();
+			// iterate until one point is left, that point will be equal to compute(t)
```
RobNadal, line 561: Is it convention that the last point is not considered part of the hull?

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -116,4 +116,14 @@ impl WasmBezier {
...
 	}
+
+	pub fn hull(&self, t: f64) -> JsValue {
+		let hull = self
+			.0
+			.hull(t)
+			.iter()
+			.map(|level| level.iter().map(|&point| Point { x: point.x, y: point.y }).collect::<Vec<Point>>())
+			.collect::<Vec<Vec<Point>>>();
+		JsValue::from_serde(&serde_json::to_string(&hull).unwrap()).unwrap()
```
RobNadal, line 127: Rebase this branch off master, and then use `to_js_value(hull)` here. `to_js_value` was added in #711

bezier-rs/lib/src/lib.rs
Keavon: Yes, thank you (I still don't know what a hull is in this context so I haven't reviewed the PR yet because I don't know what it does).

Also to elaborate on the capitalization/period rules: for doc comments (with `///`), always use complete sentences. For single-sentence inline comments (`//`), it's okay to omit the period since there aren't multiple sentences and it's more of a note, but still capitalize the first letter. If there are multiple sentences in an inline comment, use sentences with periods just like doc comments.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -250,6 +237,34 @@ export default defineComponent({
...
+							{
+								min: 0,
+								max: 1,
+								step: 0.01,
+								default: 0.5,
+								variable: "t",
```
RobNadal, line 263: I think @hannahli2010's comment was to use `templateOptions: { sliders: [tSliderOptions] },` here instead.

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -107,4 +107,9 @@ impl WasmBezier {
 		let local_extrema = self.0.local_extrema();
 		JsValue::from_serde(&serde_json::to_string(&local_extrema).unwrap()).unwrap()
 	}
+
+	pub fn hull(&self, t: f64) -> JsValue {
+		let hull = self.0.hull(t);
```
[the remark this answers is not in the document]
jackiechen73: Changed!

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -250,6 +237,36 @@ export default defineComponent({
 						});
 					},
 				},
+				{
+					name: "Hull",
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
```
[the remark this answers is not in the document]
jackiechen73: For sure. This looks a bit odd when there the handles are on opposite sides of the curve in the cubic case but I think it gives a better representation of the hull

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -250,6 +237,36 @@ export default defineComponent({
...
+						const hullPoints: number[][][] = JSON.parse(bezier.hull(options.t));
+						hullPoints.forEach((iteration: number[][], iterationNumber: number) => {
+							iteration.forEach((point: number[], index) => {
+								const newPoint: Point = { x: point[0], y: point[1] };
```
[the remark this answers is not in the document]
jackiechen73: Changed!

bezier-rs/lib/src/lib.rs
jackiechen73: Changed!

bezier-rs/lib/src/lib.rs
jackiechen73: Ah, thanks for pointing that out ill change that

bezier-rs/docs/interactive-docs/src/App.vue
jackiechen73: Nope, I was just using slider options from other examples. Thanks for pointing that out ill change that

bezier-rs/lib/src/lib.rs
jackiechen73: Changed!

bezier-rs/lib/src/lib.rs
jackiechen73: When it gets down to one point, that point is already added to the hull at the end of the previous iteration.

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
jackiechen73: Added. Thanks!

bezier-rs/docs/interactive-docs/src/App.vue
jackiechen73: OH, I misunderstood thanks for pointing that out

bezier-rs/lib/src/lib.rs
```diff
@@ -631,6 +631,41 @@ impl Bezier {
 			.collect::<Vec<DVec2>>()
 	}
 
+	/// Returns a list of lists of points representing the hull points for all iterations at the point corresponding to `t` using de Casteljau's algorithm
```
Keavon, line 634: "representing the hull points" -> what is a hull? What are hull points? What does this function actually do? This nicely describes how to interpret the data it returns, but doesn't explain what the algorithm is even for.

bezier-rs/docs/interactive-docs/src/App.vue
```diff
@@ -281,6 +281,24 @@ export default defineComponent({
 						],
 					},
 				},
+				{
+					name: "Hull",
+					callback: (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>): void => {
+						const hullPoints: Point[][] = JSON.parse(bezier.hull(options.t));
+						hullPoints.forEach((iteration: Point[]) => {
+							iteration.forEach((point: Point, index) => {
```
Keavon, line 289: Please color code the output, I currently just see a lot of red lines and dots but don't know how that translates to the function's output. Based on how I think this works, you can probably try coloring each of these iterations differently the same way that the reduce function picks its colors. Consider also making the connecting lines a darker/lighter or more desaturated color to avoid distracting from the actual output of the function, which is the circles for the points.

bezier-rs/lib/src/lib.rs
hannahli2010: One main reason why we're implementing this function is that it is in bezierjs.

Another is that `hull` can be used as an alternate implementation for `split` and `trim`. An intuitive visual representation of how that would work is available [here](https://pomax.github.io/bezierinfo/#splitting).

To be honest though, I'm not sure whether using `hull` is more or less performant than our current more algebraic approach (which is described [here](https://pomax.github.io/bezierinfo/#matrixsplit)). I think it boils down to whether a recursive solution is better than something involving more complex computations (which is also possible to rewrite to matrix multiplications). @TrueDoctor , do you have any input on this topic?

Keavon:
I pushed some simplifications to the main algorithm to reduce complexity, lines of code, and heap allocations from unnecessary cloning. It is more idiomatic and readable now.

Reading through the primer and searching for the three occurrences of "hull", I am wondering if that is the correct word for this operation?

jackiechen73:
@Keavon Most instances when I search for "hull" mostly relate to bezier curves and their convex hull.
I'm not sure what practical use this would have.
as @hannahli2010 mentioned in https://github.com/GraphiteEditor/Graphite/pull/715#discussion_r917435162
it seems to be used for alternative implementations of `split` and `trim` but since we've already implemented those, I'm not sure if there is another reason to have this function for the user.

Thoughts?

RobNadal:
> Most instances when I search for "hull" mostly relate to bezier curves and their convex hull.

@hannahli2010 made the suggestion to rename this to `de_casteljau_points`, which I think would be a good idea. From my understanding, we need this function to implement `bounding_box`, so I think it is still necessary.

Keavon:
I didn't realize we hadn't already merged this. Doing that now.

Keavon (approved):

PR #717, opened 2022-07-06
Support Linear line segments, add linear section to interactive docs

hannahli2010:
- Add `Linear` as option in `BezierHandles` enum
- Support `Linear` line segments in the existing functions
- Rewrite `derivative` to return an optional `Bezier` (`Linear` ones return `None`)
- Display a section for linear lines in the interactive docs
  - Add functionality to customize which degree of curves are shown per example
  - Refactor `drawBezier` function to handle line segments

### DIFF
--- a/bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
+++ b/bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
@@ -6,15 +6,14 @@ import { BezierCallback, BezierPoint, BezierStyleConfig, Point, WasmBezierMutato
 // Offset to increase selectable range, used to make points easier to grab
 const FUDGE_FACTOR = 3;
 
-const MAP_LENGTH_TO_MUTATOR_INDEX: { [k: number]: number[] } = {
-	2: [0, 3],
-	3: [0, 1, 3],
-	4: [0, 1, 2, 3],
+// Given the number of points in the curve, map the index of a point to the correct mutator key
+const MAP_POINTS_TO_MUTATOR_BY_NUMBER_POINTS: { [k: number]: WasmBezierMutatorKey[] } = {
+	2: ["set_start", "set_end"],
+	3: ["set_start", "set_handle_start", "set_end"],
+	4: ["set_start", "set_handle_start", "set_handle_end", "set_end"],
 };
 
 class BezierDrawing {
-	static indexToMutator: WasmBezierMutatorKey[] = ["set_start", "set_handle_start", "set_handle_end", "set_end"];
-
 	points: BezierPoint[];
 
 	canvas: HTMLCanvasElement;
@@ -44,7 +43,7 @@ class BezierDrawing {
 				y: p.y,
 				r: getPointSizeByIndex(i, points.length),
 				selected: false,
-				mutator: BezierDrawing.indexToMutator[MAP_LENGTH_TO_MUTATOR_INDEX[points.length][i]],
+				mutator: MAP_POINTS_TO_MUTATOR_BY_NUMBER_POINTS[points.length][i],
 			}));
 
 		if (this.createThroughPoints && this.points.length === 4) {
--- a/bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
+++ b/bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
@@ -1,13 +1,12 @@
 import { WasmBezier } from "@/../wasm/pkg";
+
 import { COLORS, drawBezier, drawPoint, getContextFromCanvas, getPointSizeByIndex } from "@/utils/drawing";
 import { BezierCallback, BezierPoint, BezierStyleConfig, Point, WasmBezierMutatorKey, WasmBezierInstance } from "@/utils/types";
 
 // Offset to increase selectable range, used to make points easier to grab
 const FUDGE_FACTOR = 3;
 
-const MAP_LENGTH_TO_MUTATOR_INDEX: {
-	[k: number]: number[];
-} = {
+const MAP_LENGTH_TO_MUTATOR_INDEX: { [k: number]: number[] } = {
 	2: [0, 3],
 	3: [0, 1, 3],
 	4: [0, 1, 2, 3],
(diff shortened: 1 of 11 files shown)


Androxium (approved):

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -95,27 +109,24 @@ export const drawBezier = (ctx: CanvasRenderingContext2D, points: Point[], dragI
...
+		handleEnd = points[1];
+		end = handleEnd;
 	}
-	ctx.stroke();
 
-	if (styleConfig.drawHandles) {
-		drawLine(ctx, start, handleStart, styleConfig.handleLineStrokeColor);
-		drawLine(ctx, end, handleEnd, styleConfig.handleLineStrokeColor);
+	if (points.length === 2) {
+		drawLine(ctx, start, end, styleConfig.curveStrokeColor);
```
RobNadal, line 123: We should adjust the `linewidth` so that the thickness of the linear segment matches the cubic and quadratic beziers.

bezier-rs/lib/src/lib.rs
```diff
@@ -272,52 +307,62 @@ impl Bezier {
...
-			// update the prev point
-			prev_point = *curr_point;
-		}
+		match self.handles {
+			BezierHandles::Linear => self.start.distance(self.end),
+			_ => {
+				// Code example from <https://gamedev.stackexchange.com/questions/5373/moving-ships-between-two-planets-along-a-bezier-missing-some-equations-for-acce/5427#5427>.
+
+				// We will use an approximate approach where
+				// we split the curve into many subdivisions
```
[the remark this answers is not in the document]
hannahli2010: I split it up because the description is the beginning of the implementation, while the comment with the link is referring to the entire function. It didn't seem right to have the in the same section and hence making them appear as tightly coupled comments.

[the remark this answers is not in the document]
Keavon: Just make the two short comment lines into one. The previous comment with the link is fine as-is.

Keavon:
The regression is my fault, I'll fix it later today

hannahli2010:
It's all good! I have a fix and will push soon.

bezier-rs/lib/src/lib.rs
```diff
@@ -272,52 +307,62 @@ impl Bezier {
...
+				// and calculate the euclidean distance between the two endpoints of the subdivision
+
+				let lookup_table = self.compute_lookup_table(Some(LENGTH_SUBDIVISIONS));
```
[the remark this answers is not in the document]
hannahli2010: Hmm, would it be weird that the optional parameter has no effect on the linear case?

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
```diff
@@ -5,6 +5,14 @@ import { BezierCallback, BezierPoint, BezierStyleConfig, Point, WasmBezierMutato
 // Offset to increase selectable range, used to make points easier to grab
 const FUDGE_FACTOR = 3;
 
+const MAP_LENGTH_TO_MUTATOR_INDEX: {
```
Keavon, line 8: Based on its name and looking at it, I really don't have any idea what this does or means. Let's chat about that and pick a name that hopefully can convey more understanding.

bezier-rs/lib/src/lib.rs
```diff
@@ -207,26 +237,30 @@ impl Bezier {
...
 			BezierHandles::Quadratic { .. } => None,
 			BezierHandles::Cubic { handle_end, .. } => Some(handle_end),
 		}
 	}
 
 	/// Get the coordinates of all points in an array of 4 optional points.
+	/// For a linear segment, the order of the points will be: `start`, `end`. The third and fourth elements will be `None`.
 	/// For a quadratic segment, the order of the points will be: `start`, `handle`, `end`. The fourth element will be `None`.
 	/// For a cubic segment, the order of the points will be: `start`, `handle_start`, `handle_end`, `end`.
 	pub fn get_points(&self) -> [Option<DVec2>; 4] {
```
Keavon, line 261: Should this return a slice instead of a fixed-length array of Options?

bezier-rs/lib/src/lib.rs
```diff
@@ -271,53 +306,62 @@ impl Bezier {
...
-		let mut prev_point = lookup_table[0];
-		// calculate approximate distance between subdivision
-		for curr_point in lookup_table.iter().skip(1) {
-			// calculate distance of subdivision
-			approx_curve_length += (*curr_point - prev_point).length();
-			// update the prev point
-			prev_point = *curr_point;
-		}
+	/// - `num_subdivisions` - Number of subdivisions used to approximate the curve. The default value is 1000.
+	pub fn length(&self, num_subdivisions: Option<i32>) -> f64 {
```
Keavon, line 310: I wonder if we could use SIMD for this function?

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -78,7 +84,7 @@ impl WasmBezier {
 	}
 
 	pub fn length(&self) -> f64 {
-		self.0.length()
+		self.0.length(None)
 	}
 
 	pub fn compute(&self, t: f64) -> JsValue {
```
Keavon, line 90: What would you think about renaming `compute()` to `evaluate()`? To me, at least, it has a more intuitive meaning. But maybe that's just me.

bezier-rs/lib/src/lib.rs
```diff
@@ -522,6 +571,8 @@ impl Bezier {
 		self.apply_transformation(&|point| point + translation)
 	}
 
+	// TODO: Change this to `intersect(&self, other: &Bezier)` to also work on quadratic and cubic segments
+	// TODO: (or keep this and add two more functions that perform the logic, and make the `intersect` function call the correct one)
 	/// Returns a list of points where the provided line segment intersects with the Bezier curve.
 	/// - `line` - A line segment expected to be received in the format of `[start_point, end_point]`.
```
Keavon, line 577: Something should be written about the case where a segment is colinear with the line and thus has infinite intersection points. (I'd expect there should be zero in that case, if floating point imprecision doesn't make that too difficult.)

bezier-rs/lib/src/lib.rs
TrueDoctor: simd is a bit tricky because you always need a reference to the previous iteration.
Processing this in parallel would likely not yield great results due to the extra overhead.
But we can do lots by improving the algorithm used.
Raph Levien (kurbo author) has a great article on exactly that and has gotten into a lot more details + did benchmarks on this: https://raphlinus.github.io/curves/2018/12/28/bezier-arclength.html

bezier-rs/docs/interactive-docs/src/components/BezierDrawing.ts
hannahli2010: I refactor this logic a little and added a more descriptive comment. Does it help in understanding the purpose of the constant?

bezier-rs/lib/src/lib.rs
hannahli2010: I added a test to ensure that `solve_linear` returns 0 when the line is on the x-axis.

Keavon (approved):

PR #718, opened 2022-07-08
Implement Bezier scale and offset

hannahli2010:
Implement the scale and offset functions. Note that it is not possible to offset or scale all bezier curves. So:
- the `scale` function asserts that the curve passed in is scalable
- the `offset` function is a "hack" that finds the scalable sub-curves of the bezier (aka calls `reduce`) and then scales those segments
   - Since this function uses reduce to identify scalable sub-curves, the UI doc uses the same colouring logic for each scaled segment


bezier-rs/lib/src/lib.rs
```diff
@@ -721,18 +726,64 @@ impl Bezier {
 		});
 		result
 	}
+
+	/// Scale will translate a bezier curve a fixed distance away from its original position, and stretch/compress the transformed curve to match the translation ratio.
+	/// Scale takes the following parameters:
+	/// - `distance` - The distance away from the curve that the new one will be scaled to. Positive values will scale the curve in the same direction as the endpoint normals,
+	/// while negative values will scale in the opposite direction.
+	pub fn scale(&self, distance: f64) -> Bezier {
+		assert!(self.is_scalable());
```
linda-zheng, line 735: Maybe put a comment explaining why we need a simple curve. You could even put it in the function description.

bezier-rs/lib/src/lib.rs
```diff
@@ -721,18 +726,64 @@ impl Bezier {
...
-	use crate::utils;
 
 	use glam::DVec2;
 
 	fn compare_points(p1: DVec2, p2: DVec2) -> bool {
-		utils::dvec2_compare(p1, p2, MAX_ABSOLUTE_DIFFERENCE).all()
+		p1.abs_diff_eq(p2, MAX_ABSOLUTE_DIFFERENCE)
+	}
+
+	fn compare_vector_of_points(vec1: Vec<DVec2>, vec2: Vec<DVec2>) -> bool {
```
linda-zheng, line 778: Since these are helper functions, adding some comments about them might help (i.e. maybe explain that we tolerate some difference)

bezier-rs/lib/src/lib.rs
```diff
@@ -721,18 +726,64 @@ impl Bezier {
...
+		vec1.len() == vec2.len() && vec1.into_iter().zip(vec2.into_iter()).all(|(a, b)| a.abs_diff_eq(b, MAX_ABSOLUTE_DIFFERENCE))
+	}
+
+	fn compare_vector_of_beziers(beziers: Vec<Bezier>, expected_bezier_points: Vec<Vec<DVec2>>) -> bool {
```
linda-zheng, line 782: Same here. Probably want to mention something about MAX_ABSOLUTE_DIFFERENCE.

linda-zheng (approved):

bezier-rs/lib/src/lib.rs
```diff
@@ -688,6 +688,9 @@ impl Bezier {
 	/// 2. The on-curve point for `t = 0.5` must occur roughly in the center of the polygon defined by the curve's endpoint normals.
 	/// See [the offset section](https://pomax.github.io/bezierinfo/#offsetting) of Pomax's bezier curve primer for more details.
 	fn is_scalable(&self) -> bool {
+		if let BezierHandles::Linear = self.handles {
```
Keavon, line 691: Remove `let` and change `=` to `==`, there is no variable being bound here so this isn't a case for `if let`.

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
 	pub fn inflections(&self) -> Vec<f64> {
 		self.unrestricted_inflections().into_iter().filter(|&t| t > 0. && t < 1.).collect::<Vec<f64>>()
 	}
+
+	/// Scale will translate a bezier curve a fixed distance away from its original position, and stretch/compress the transformed curve to match the translation ratio.
+	/// Note that not all bezier curves are possible to scale, so this function asserts that the provided curve is scalable.
+	/// A proof for why this is true can be found in the [Curve offsetting section](https://pomax.github.io/bezierinfo/#offsetting) of Pomax's bezier curve primer.
+	/// `scale` takes the following parameters:
```
Keavon, line 844: Seems weird in my opinion to say "the following parameters" and then write out a list of only one item. I'd just remove the "`scale` takes the following parameters" line and just talk about what `distance` means in its own sentence.

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
...
+	/// - `distance` - The distance away from the curve that the new one will be scaled to. Positive values will scale the curve in the same direction as the endpoint normals,
+	/// while negative values will scale in the opposite direction.
+	fn scale(&self, distance: f64) -> Bezier {
+		assert!(self.is_scalable());
```
Keavon, line 848: As with all `assert!()` calls, please give an error message as the second argument to clarify to the user of the library why their code crashed.

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
...
+
+		let normal_start = self.normal(0.);
+		let normal_end = self.normal(1.);
+
+		// If normal unit vectors are equal, then the lines are parallel
+		if normal_start == normal_end {
```
Keavon, line 854: I foresee floating point error causing bugs here, do you agree? Probably should check if the two vectors are within some epsilon distance of each other (or that their components both are, to avoid the square root of a distance calculation).

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
...
+		self.apply_transformation(&|point| {
+			let mut direction_unit_vector = (intersection - point).normalize();
+			if should_flip_direction {
+				direction_unit_vector *= -1.;
+			}
+			point + distance * direction_unit_vector
+		})
+	}
+
+	/// Offset will get all the reducable subcurves, and for each subcurve, it will scale the subcurve a set distance away from the original curve.
```
Keavon, line 871: reducable -> reduceable

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
...
+	/// Note that not all bezier curves are possible to offset, so this function first reduces the curve to scalable segments and then offsets those segments.
+	/// A proof for why this is true can be found in the [Curve offsetting section](https://pomax.github.io/bezierinfo/#offsetting) of Pomax's bezier curve primer.
+	/// Offset takes the following parameter:
+	/// - `distance` - The distance away from the curve that the new one will be offset to. Positive values will offset the curve in the same direction as the endpoint normals,
+	/// while negative values will offset in the opposite direction.
+	pub fn offset(&self, distance: f64) -> Vec<Bezier> {
+		self.reduce(None).into_iter().map(|bezier| bezier.scale(distance)).collect()
```
Keavon, line 878: You should be able to avoid a second heap allocation from `.collect()` by mutating the values in the Vector provided by `.reduce()`. Try:
```rs
let mut reduced = self.reduce(None);
reduced.iter_mut().for_each(|bezier| *bezier = bezier.scale(distance));
reduced
```

bezier-rs/lib/src/lib.rs
```diff
@@ -832,18 +837,71 @@ impl Bezier {
...
 	use glam::DVec2;
 
+	// Compare points by allowing some maximum absolute difference to account for floating point errors
 	fn compare_points(p1: DVec2, p2: DVec2) -> bool {
-		utils::dvec2_compare(p1, p2, MAX_ABSOLUTE_DIFFERENCE).all()
+		p1.abs_diff_eq(p2, MAX_ABSOLUTE_DIFFERENCE)
+	}
+
+	// Compare vectors of points by allowing some maximum absolute difference to account for floating point errors
+	fn compare_vector_of_points(vec1: Vec<DVec2>, vec2: Vec<DVec2>) -> bool {
```
Keavon, line 895: I'm a little concerned about using the variable name `vec1`/`vec2` because it's ambiguous in this case if you mean the `Vec` or the `DVec2` (I wish this word wasn't so overloaded!). I think it's perfectly fine to just call it `a` and `b` instead.

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,29 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
 	}
 }
 
+/// Returns the intersection of two lines. The lines are given by a point on the line and its slope (represented by a vector).
+pub fn line_intersection(point1: DVec2, point1_slope_vector: DVec2, point2: DVec2, point2_slope_vector: DVec2) -> DVec2 {
+	assert!(point1_slope_vector.normalize() != point2_slope_vector.normalize());
+	/*
+		*/
```
Keavon, line 153: Stray comment.

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,29 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
...
+	if f64_compare(point1_slope_vector.x, 0., MAX_ABSOLUTE_DIFFERENCE) {
```
Keavon, line 154: It'd be good if you could comment in English what each of these three cases represent.

Keavon (changes requested):
Done reviewing, ping me again when the changes are done (and as always, feel free to discuss if you disagree with anything mentioned) and I'll be able to approve it quickly.

Keavon (approved):

PR #725, opened 2022-07-13
Implement curvature function in Bezier math library

jackiechen73:
- Added bezier `curvature(t)` function
  - returns the curvature at point t along the curve
  - added circle drawing helper function to interactive documentation

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -394,6 +394,26 @@ impl Bezier {
 		self.tangent(t).perp()
 	}
 
+	/// Returns the curvature k of a curve at given t-value along the curve
+	/// 1/k is the radius of a circle whose derivative is equal to the derivative of the curve at that point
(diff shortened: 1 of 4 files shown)


Androxium (approved):

bezier-rs/lib/src/lib.rs
```diff
@@ -394,6 +394,26 @@ impl Bezier {
 		self.tangent(t).perp()
 	}
 
+	/// Returns the curvature k of a curve at given t-value along the curve
+	/// 1/k is the radius of a circle whose derivative is equal to the derivative of the curve at that point
```
Keavon, line 398: Mentioning "k" just hurts readability. I'd suggest this comment instead:
```
/// Returns the curvature, a scalar value for the derivative at the given `t`-value along the curve.
/// Curvature is 1 over the radius of a circle with an equivalent derivative.
```

Keavon (approved):

PR #726, opened 2022-07-13
Implement bounding box function for bezier-rs

linda-zheng:
Added `bbox()` function in bezier-rs.
The function returns a list of points that represent the coordinates of the bounding box of the curve.

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -773,6 +773,27 @@ impl Bezier {
 		});
 		result
 	}
+
+	// Return list of points that represent the bounding box of the curve
+	pub fn bbox(&self) -> Vec<DVec2> {
+		// start by taking min/max of endpoints
+		let mut min_point = self.start.min(self.end);
+		let mut max_point = self.start.max(self.end);
+
+		// iterate through extrema points
+		let extrema = self.local_extrema();
+		for t_values in extrema {
+			for t in t_values {
+				let point = self.evaluate(t);
+				// update bounding box if new min/max is found
+				min_point = min_point.min(point);
+				max_point = max_point.max(point);
+			}
+		}
+
+		// Return the 4 points that make up the bounding box
+		return vec![min_point, DVec2::new(min_point.x, max_point.y), max_point, DVec2::new(max_point.x, min_point.y)];
+	}
 }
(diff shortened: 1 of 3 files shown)


RobNadal (approved):

jackiechen73 (approved):

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +721,27 @@ impl Bezier {
 		});
 		result
 	}
+
+	// Return list of points that represent the bounding box of the curve
```
Keavon, line 725: `///` not `//` for doc comments.

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +721,27 @@ impl Bezier {
...
+	pub fn bounding_box(&self) -> Vec<DVec2> {
+		// start by taking min/max of endpoints
```
Keavon, line 727: Please capitalize the first letter of comments.

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +721,27 @@ impl Bezier {
...
+		let mut min_point = self.start.min(self.end);
+		let mut max_point = self.start.max(self.end);
```
Keavon, line 729: I think `endpoints_min` and `endpoints_max` would be clearer names, thoughts?

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +721,27 @@ impl Bezier {
...
+			for t in t_values {
+				let point = self.evaluate(t);
+				// update bounding box if new min/max is found
+				min_point = min_point.min(point);
+				max_point = max_point.max(point);
+			}
+		}
+
+		// Return the 4 points that make up the bounding box
+		return vec![min_point, DVec2::new(min_point.x, max_point.y), max_point, DVec2::new(max_point.x, min_point.y)];
```
Keavon, line 743: It would be more useful to return the DVec2 for the min (top left) and max (bottom right) corners.

Also: please don't use a `Vector` here since we know the length, just return an array. And remove the `return` keyword and trailing semicolon.

Keavon (approved):
Approved after changes

PR #731, opened 2022-07-18
Implement arcs for Bezier math library

Androxium:
This PR adds the `arcs` function to `bezier-rs`. `arcs` will approximate a bezier curve with a series of circular arcs. At the moment, there are 3 implementations of `arcs`: 
1. The first implementation follows what is done in Bezier.js and is guaranteed to fine the largest good circular arc approximation, but will occasionally return the wrong arc when points on the bezier curve are almost collinear
2. The second implementation will split the bezier curve on it's extremas first before finding circular arc approximations, which solves the issue of the first implementation but no longer guarantees the largest arc approximation is found
3. The last option starts with the Bezier.js implementation and then switches to the second once an erroneous case is encountered. This option is called `Automatic` and is the default. The heuristic for recognizing an erroneous case is when the found circle sector has an angle greater than 180 degrees.

`arcs` takes 2 optional arguments:
- `error` : the error bound used for fitting the circular arc to the bezier curve segment
- `max_iterations` : the number of iterations of arc fitting done on each curve segment


bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,74 @@
...
+		ArcsOptions {
+			maximize_arcs: false,
+			error: 0.5,
+			max_iterations: 100,
+		}
+	}
+}
+
+#[derive(Copy, Clone)]
+pub struct CircleArc {
```
RobNadal, line 52: Add a rustdoc for this structure

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,53 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
 	}
 }
 
+/// Returns the intersection of two lines. The lines are given by a point on the line and its slope (represented by a vector).
+pub fn line_intersection(point1: DVec2, point1_slope_vector: DVec2, point2: DVec2, point2_slope_vector: DVec2) -> DVec2 {
+	assert!(point1_slope_vector.normalize() != point2_slope_vector.normalize());
+	/*
+		*/
```
RobNadal, line 153: Remove empty multiline comment

bezier-rs/docs/interactive-docs/src/utils/drawing.ts
```diff
@@ -73,6 +73,21 @@ export const drawCurve = (ctx: CanvasRenderingContext2D, points: Point[], stroke
...
+	ctx.fillStyle = fillColor;
+	ctx.lineWidth = 2;
+
+	const { center, radius, start_angle: startAngle, end_angle: endAngle } = circleSector;
+	ctx.beginPath();
+	ctx.moveTo(center.x, center.y);
+	ctx.arc(center.x, center.y, radius, startAngle, endAngle);
+	ctx.lineTo(center.x, center.y);
+	ctx.closePath();
+	// ctx.stroke();
```
RobNadal, line 87: remove comment

bezier-rs/docs/interactive-docs/src/utils/types.ts
```diff
@@ -11,7 +11,8 @@ export enum BezierCurveType {
 	Cubic = "Cubic",
 }
 
-export type BezierCallback = (canvas: HTMLCanvasElement, bezier: WasmBezierInstance, options: Record<string, number>, mouseLocation?: Point) => void;
+export type option = number | boolean;
```
RobNadal, line 14: I don't think this is used anymore

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -138,4 +146,23 @@ impl WasmBezier {
...
+					center: Point {
+						x: sector.center.x,
+						y: sector.center.y,
+					},
+					radius: sector.radius,
+					start_angle: sector.start_angle,
+					end_angle: sector.end_angle,
+				})
+			})
+			.collect::<Vec<JsValue>>()
```
RobNadal, line 166: You could collect it to `Vec<CircleSector>` first, and then convert it using `to_js_value`. The return type of the function would just be `JsValue`. Then you would only need to call `JSON.parse` once in `App.vue`, which I personally think is a bit cleaner, and more consistent with other placing in `wasm.rs` where we return lists (well, except `lookup_table`, which you could change at the same time 😉 )

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +699,141 @@ impl Bezier {
...
+		let mut iterations = 0;
+
+		let mut previous_arc = CircleArc::default();
+		let mut was_previous_good = false;
+
+		let mut arcs = Vec::new();
+
+		// Find the arc
+		'iterate_curve: while low < local_high {
+			'find_good_segment: loop {
```
RobNadal, line 748: I was told once that having to label your loops is an indication that your function might be too complex. You can always avoid having to do this by just creating a helper function, and then `return` becomes the equivalent of the outer `break`.

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,53 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
...
+		let m1 = point1_slope_vector.y / point1_slope_vector.x;
+		let b1 = point1.y - m1 * point1.x;
+		let m2 = point2_slope_vector.y / point2_slope_vector.x;
+		let b2 = point2.y - m2 * point2.x;
+		let intersection_x = (b2 - b1) / (m1 - m2);
+		DVec2::new(intersection_x, intersection_x * m1 + b1)
+	}
+}
+
+/// Check if 3 points are collinear
```
RobNadal, line 172: Add periods to rustdocs

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,53 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
...
+pub fn are_points_collinear(p1: DVec2, p2: DVec2, p3: DVec2) -> bool {
+	let matrix = DMat2::from_cols(p1 - p2, p2 - p3);
+	f64_compare(matrix.determinant() / 2., 0., MAX_ABSOLUTE_DIFFERENCE)
+}
+
+/// Compute the center of the circle that passes through all three provided points
+pub fn compute_circle_center_from_points(p1: DVec2, p2: DVec2, p3: DVec2) -> DVec2 {
+	// TODO something breaks here
```
RobNadal, line 180: Is this comment resolved?

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +699,141 @@ impl Bezier {
...
+		extrema.sort_by(|ex1, ex2| ex1.partial_cmp(ex2).unwrap());
+
+		let mut t1 = extrema[0];
+		extrema
+			.iter()
+			.skip(1)
+			.flat_map(|&t2| {
+				let arcs = self.arcs_helper(t1, t2, error, max_iterations);
+				println!("t1, t2: {}, {}", t1, t2);
+				t1 = t2;
```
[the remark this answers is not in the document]
hannahli2010: One of the main reasons I did it this way was to avoid having to push to a vector, but to be fair I'm not sure if that makes that much of a difference in performance.

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +699,141 @@ impl Bezier {
...
+	/// The algorithm can be customized using the [ArcsOptions] structure.
+	pub fn arcs(&self, arcs_options: ArcsOptions) -> Vec<CircleArc> {
+		let ArcsOptions { maximize_arcs, error, max_iterations } = arcs_options;
+		if maximize_arcs {
+			return self.arcs_helper(0., 1., error, max_iterations);
+		}
+		let mut extrema: Vec<f64> = self.local_extrema().into_iter().flatten().collect::<Vec<f64>>();
+		extrema.append(&mut vec![0., 1.]);
+		extrema.dedup();
+		extrema.sort_by(|ex1, ex2| ex1.partial_cmp(ex2).unwrap());
```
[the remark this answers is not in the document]
hannahli2010: I'm assuming you mean reduce? Also, what's the benefit of making the helper public? I.e. would users care about this functionality/when would they use it?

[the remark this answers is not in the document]
RobNadal: Oops, yes, I meant reduce. I remember that bezier.rs returned a third argument (in addition to the x and y extrema) which was a list of both x and y extrema with duplicates removed. Created a function to handle output of that form could be valuable. But, you could always just make it a private helper for now because making it public is easily something that could be done in the future if there a need.

bezier-rs/lib/src/lib.rs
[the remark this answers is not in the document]
RobNadal: Optimizing for performance is a valid enough reason, I say just leave it as is then.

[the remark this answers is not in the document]
hannahli2010: Another way to do this that avoids side effects is: 
```
extrema_t_list
  .iter()
  .take(extrema_t_list.len() - 1)
  .zip(extrema_t_list.iter().skip(1))
  .flat_map(|(&t1, &t2)| self.arcs_helper(t1, t2, error, max_iterations))
  .collect::<Vec<CircleArc>>()
```
Any thoughts on this?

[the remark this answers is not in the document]
hannahli2010: We found out about the `window` function and used that instead

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+/// Struct to represent optional parameters that can be passed to the `arcs` function.
+#[derive(Copy, Clone)]
+pub struct ArcsOptions {
+	/// Determines whether the algorithm tries to greedily maximize the approximated arcs or to prioritize correctness by first splitting the curve between its local extremas.
+	/// When maximizing the arcs, the algorithm may return incorrect arcs when the curve contains any small loops or segements that look like a very thin "U".
+	/// The default value is `false`.
+	pub maximize_arcs: bool,
```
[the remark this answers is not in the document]
hannahli2010: I named the struct `MaximizeArcs` with options `Automatic`, `On`, and `Off` based on the discussion we had with Keavon. Let me know if you want different names.

RobNadal (approved):
I really like the changes done here for commonizing the wasm return types.

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -1,25 +1,40 @@
...
 struct Point {
 	x: f64,
 	y: f64,
 }
 
+#[wasm_bindgen]
+pub enum WasmMaximizeArcs {
+	Automatic, // 0
+	On,        // 1
+	Off,       // 2
```
Keavon, line 27: Please use the "unit" feature to try and convey to the user what each of these mean. So perhaps the unit can be something like:
```
(0: auto, 1: on, 2: off)
```

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -32,6 +47,14 @@ fn to_js_value<T: Serialize>(data: T) -> JsValue {
 	JsValue::from_serde(&serde_json::to_string(&data).unwrap()).unwrap()
 }
 
+fn convert_wasm_mazimize_arcs(wasm_enum_value: WasmMaximizeArcs) -> MaximizeArcs {
```
Keavon, line 50: `mazimize` is a typo

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -123,23 +161,27 @@ impl WasmBezier {
...
 	pub fn inflections(&self) -> JsValue {
-		let inflections = self.0.inflections();
+		let inflections: Vec<f64> = self.0.inflections();
 		to_js_value(inflections)
 	}
 
+	/// The wrapped return type is `Vec<Vec<Point>>`.
 	pub fn de_casteljau_points(&self, t: f64) -> JsValue {
-		let hull = self
+		let hull: Vec<Vec<Point>> = self
```
Keavon, line 184: Here's another place where we can deprecate the old name `hull`.

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+/// Struct used to represent the different strategies for generating arc approximations.
+#[derive(Copy, Clone)]
+pub enum MaximizeArcs {
+	/// Start with the strategy of maximizing arcs and automatically switch when an erroneous case is encountered.
+	Automatic,
+	/// Use the strategy to maximize approximated arcs, despite potentially erroneous arcs.
+	On,
+	/// Use the strategy that prioritizes correctness over maximal arcs.
+	Off,
+}
```
Keavon, line 37: These comments are helpful although the concept of "maximizing arcs" doesn't really mean much to me without preexisting context. Any way to explain that aspect a little more clearly?

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Copy, Clone, PartialEq)]
+pub struct CircleArc {
+	// the center point of the circle
+	pub center: DVec2,
+	// the radius of the circle
+	pub radius: f64,
+	// the start angle of the circle sector in rad
+	pub start_angle: f64,
+	// the end angle of the circle sector in rad
+	pub end_angle: f64,
```
Keavon, line 77: Remember to use `///` and make these sentences (capital first letter, ending in a period).

bezier-rs/lib/src/utils.rs
```diff
@@ -181,6 +181,29 @@ pub fn line_intersection(point1: DVec2, point1_slope_vector: DVec2, point2: DVec
 	}
 }
 
+/// Check if 3 points are collinear.
+pub fn are_points_collinear(p1: DVec2, p2: DVec2, p3: DVec2) -> bool {
+	let matrix = DMat2::from_cols(p1 - p2, p2 - p3);
+	f64_compare(matrix.determinant() / 2., 0., MAX_ABSOLUTE_DIFFERENCE)
+}
+
+/// Compute the center of the circle that passes through all three provided points.
```
Keavon, line 190: Please explain what happens if the three points are colinear, I don't see how it's possible to fit a circle to it in that case.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+	/// This algorithm uses a method akin to binary search to find an arc that approximates a maximal segment of the curve.
+	/// Once a maximal arc has been found for a sub-segment of the curve, the algorithm continues by starting again at the end of the previous approximation.
+	/// More details can be found in the [Approximating a Bezier curve with circular arcs](https://pomax.github.io/bezierinfo/#arcapproximation) section of Pomax's bezier curve primer.
+	/// A caveat with this algorithm is that it is possible to find erroneous approximations in cases such as in a very narrow `U`.
+	/// - `stop_when_invalid`: Used to determine whether the algorithm should end early erroneous approximations are encountered.
+	///
+	/// Returns a tuple where the first element is the list of circular arcs and the second is the `t` value where the next segment should start from.
+	/// The second value will be `1.` except for when `stop_when_invalid` is true and an invalid approximation is encountered.
+	fn arcs_helper(&self, local_low: f64, local_high: f64, error: f64, max_iterations: i32, stop_when_invalid: bool) -> (Vec<CircleArc>, f64) {
```
Keavon, line 883: A more descriptive function name would be preferable.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+			MaximizeArcs::Automatic => {
+				let (auto_arcs, final_low_t) = self.arcs_helper(0., 1., error, max_iterations, true);
+				if final_low_t != 1. {
+					[
+						auto_arcs,
+						self.split(final_low_t)[1].arcs(ArcsOptions {
+							maximize_arcs: MaximizeArcs::Off,
+							error,
+							max_iterations,
+						}),
```
Keavon, line 858: Assign this to a variable, please, so the `[auto_arcs, your_new_variable].concat()` can be on a single line.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+				.collect::<Vec<CircleArc>>(),
+		}
+	}
+
+	/// Implements an algorithm that approximates a bezier curve with circular arcs.
+	/// This algorithm uses a method akin to binary search to find an arc that approximates a maximal segment of the curve.
+	/// Once a maximal arc has been found for a sub-segment of the curve, the algorithm continues by starting again at the end of the previous approximation.
+	/// More details can be found in the [Approximating a Bezier curve with circular arcs](https://pomax.github.io/bezierinfo/#arcapproximation) section of Pomax's bezier curve primer.
+	/// A caveat with this algorithm is that it is possible to find erroneous approximations in cases such as in a very narrow `U`.
+	/// - `stop_when_invalid`: Used to determine whether the algorithm should end early erroneous approximations are encountered.
```
Keavon, line 879: After "end early": missing the word "if" or "when". Also maybe change "end" to "terminate".

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+		let mut iterations = 0;
+		let mut previous_arc = CircleArc::default();
+		let mut was_previous_good = false;
+		let mut arcs = Vec::new();
+
+		// Outer loop to iterate over the curve
+		while low < local_high {
+			if iterations > max_iterations {
+				break;
+			}
```
Keavon, line 898: A regular `for` loop would be the cleaner way to do this: `for iterations in 0..max_iterations { ... }`

Keavon: Wait, I misread that, disregard.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+			// Inner loop to find the next maximal segment of the curve that can be approximated with a circular arc
+			loop {
+				if iterations > max_iterations {
+					break;
+				}
```
Keavon, line 903: I think you can combine this as `while iterations <= max_iterations {`? And then that would let you remove the `if iterations > max_iterations { break; }` on the previous few lines? Double check my reasoning. As it currently stands, this feels a bit messy, and I think that would help clean it up.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+				let p1 = self.evaluate(low);
+				let p2 = self.evaluate(middle);
+				let p3 = self.evaluate(high);
+
+				// If the segment is linear, move on to next segment
+				if utils::are_points_collinear(p1, p2, p3) {
+					previous_high = high;
+					low = high;
+					high = 1.;
+					middle = low + (high - low) / 2.;
```
Keavon, line 914: `middle = (low + high) / 2;`

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+				}
+
+				let new_arc = CircleArc {
+					center,
+					radius,
+					start_angle,
+					end_angle,
+				};
+
+				// Use points in between low, middle and high to evaluate how well the arc approximates the curve
```
Keavon, line 945: Nit: insert a comma avert "middle".

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+				let e1 = self.evaluate(low + (middle - low) / 2.);
+				let e2 = self.evaluate(middle + (high - middle) / 2.);
```
Keavon, line 947: Here too, simplify the math expressions.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+					if high == local_high {
+						// Found the final arc approximation
+						arcs.push(new_arc);
+						low = high;
+						break;
+					}
+					// If the approximation is good, expand the segment by half to try finding a larger good approximation
+					previous_high = high;
+					high = (high + (high - low) / 2.).min(local_high);
+					middle = low + (high - low) / 2.;
```
Keavon, line 968: And here.

bezier-rs/lib/src/lib.rs
```diff
@@ -853,6 +838,159 @@ impl Bezier {
...
+					end_angle,
+				};
+
+				// Use points in between low, middle and high to evaluate how well the arc approximates the curve
+				let e1 = self.evaluate(low + (middle - low) / 2.);
+				let e2 = self.evaluate(middle + (high - middle) / 2.);
+
+				// Iterate until we find the largest good approximation such that the next iteration is not a good approximation with an arc
+				if utils::f64_compare(radius, e1.distance(center), error) && utils::f64_compare(radius, e2.distance(center), error) {
+					// Check if the good approximation is actually valid: the sector angle cannot be larger than 180 degrees (aka 2*PI)
```
Keavon, line 951: 180 degrees is pi not 2*pi.

Keavon (changes requested):
First round done, ping me when changes are ready and I'll glance over it again for a final approval. Thanks for your patience.

bezier-rs/lib/src/lib.rs
hannahli2010: Oops, good eye.

bezier-rs/lib/src/lib.rs
Androxium: The reason we use `low + (high-low) / 2` instead of (low+high) / 2` is to avoid overlow in the case that `low+high` is larger than the size of a `f64` type.

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+#[derive(Copy, Clone)]
+pub struct ProjectionOptions {
+	/// Size of the lookup table for the initial passthrough. The default value is `20`.
+	pub lut_size: i32,
+	/// Difference used between floating point numbers to be considered as equal. The default value is `0.0001`
+	pub convergence_epsilon: f64,
+	/// Controls the number of iterations needed to consider that minimum distance to have converged. The default value is `3`.
+	pub convergence_limit: i32,
+	/// Controls the maximum total number of iterations to be used. The default value is `10`.
+	pub iteration_limit: i32,
```
Keavon, line 14: These `i32` types should be `usize` because they deal with counting. Negative numbers should not be allowed.

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+			convergence_epsilon: 1e-4,
+			convergence_limit: 3,
+			iteration_limit: 10,
+		}
+	}
+}
+
+/// Struct used to represent the different strategies for generating arc approximations.
+#[derive(Copy, Clone)]
+pub enum MaximizeArcs {
```
Keavon, line 30: How about calling this `Strategy`? And then the variants can be `Auto` and perhaps `FavorLargerArcs` and `FavorCorrectness`? The name "maximize arcs" on its own conveys very little understanding to me. When changing it here, remember to change it everywhere else (other functions, local variables within those other functions, comments, wasm wrapper, frontend, etc.)

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+	On,
+	/// Use the divide-and-conquer strategy that prioritizes correctness over maximal arcs.
+	Off,
+}
+
+/// Struct to represent optional parameters that can be passed to the `arcs` function.
+#[derive(Copy, Clone)]
+pub struct ArcsOptions {
+	/// Determines how the approximated arcs are computed.
+	/// When maximizing the arcs, the algorithm may return incorrect arcs when the curve contains any small loops or segements that look like a very thin "U".
```
Keavon, line 43: segements -> segments

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -185,12 +227,30 @@ impl WasmBezier {
...
+	/// The wrapped return type is `Vec<Vec<Point>>`.
 	pub fn offset(&self, distance: f64) -> JsValue {
 		let bezier_points: Vec<Vec<Point>> = self.0.offset(distance).into_iter().map(bezier_to_points).collect();
 		to_js_value(bezier_points)
 	}
 
-	pub fn curvature(&self, t: f64) -> f64 {
-		self.0.curvature(t)
+	/// The wrapped return type is `Vec<CircleSector>`.
+	pub fn arcs(&self, error: f64, max_iterations: i32, maximize_arcs: WasmMaximizeArcs) -> JsValue {
```
Keavon, line 237: `usize` not `i32`

bezier-rs/lib/src/lib.rs
```diff
@@ -923,6 +908,150 @@ impl Bezier {
 		self.reduced_curves_and_t_values(step_size).0
 	}
 
+	/// Approximate a bezier curve with circular arcs.
+	/// The algorithm can be customized using the [ArcsOptions] structure by setting the following fields within ArcOptions:
+	/// - maximize_arcs - [MaximizeArcs] - The default value is `Automatic`.,
+	/// - error: f64 - The error used for approximating the arc's fit. The default is `0.5`.
+	/// - max_iterations: i32 - The maximum number of segment iterations used as attempts for arc approximations. The default is `100`.
```
Keavon, line 915: `usize` not `i32`

bezier-rs/lib/src/lib.rs
```diff
@@ -923,6 +908,150 @@ impl Bezier {
...
+	/// This algorithm uses a method akin to binary search to find an arc that approximates a maximal segment of the curve.
+	/// Once a maximal arc has been found for a sub-segment of the curve, the algorithm continues by starting again at the end of the previous approximation.
+	/// More details can be found in the [Approximating a Bezier curve with circular arcs](https://pomax.github.io/bezierinfo/#arcapproximation) section of Pomax's bezier curve primer.
+	/// A caveat with this algorithm is that it is possible to find erroneous approximations in cases such as in a very narrow `U`.
+	/// - `stop_when_invalid`: Used to determine whether the algorithm should terminate early if erroneous approximations are encountered.
+	///
+	/// Returns a tuple where the first element is the list of circular arcs and the second is the `t` value where the next segment should start from.
+	/// The second value will be `1.` except for when `stop_when_invalid` is true and an invalid approximation is encountered.
+	fn approximate_curve_with_arcs(&self, local_low: f64, local_high: f64, error: f64, max_iterations: i32, stop_when_invalid: bool) -> (Vec<CircleArc>, f64) {
```
Keavon, line 950: `usize` not `i32`

bezier-rs/lib/src/lib.rs
Keavon: Hadn't thought of that, thanks for pointing it out. Although if `high` is a large number and `low` is a small (high negative) number, that minus sign flips the `low` to be positive and then it's adding two high numbers again and causing overflow within the parentheses. However, it's probably more likely that `low` and `high` will be similar, not wildly far apart, so your solution here is likely more robst. Is it worth the extra few clock cycles? Not sure. But I'd say keep it as is.

bezier-rs/lib/src/structs.rs
```diff
@@ -0,0 +1,95 @@
...
+	/// - `Automatic`: Maximize arcs until an erroneous approximation is found. Compute the arcs of the rest of the curve by first splitting on extremas to ensure no more erroneous cases are encountered.
+	/// - `On`: Maximize arcs using the original algorithm from the [Approximating a Bezier curve with circular arcs](https://pomax.github.io/bezierinfo/#arcapproximation) section of Pomax's bezier curve primer. Erroneous arcs are possible.
+	/// - `Off`: Prioritize correctness by first spliting the curve by its extremas and determine the arc approximation of each segment instead.
+	///
+	/// The default value is `Automatic`.
+	pub maximize_arcs: MaximizeArcs,
+	/// The error used for approximating the arc's fit. The default is `0.5`.
+	pub error: f64,
+	/// The maximum number of segment iterations used as attempts for arc approximations. The default is `100`.
+	pub max_iterations: i32,
```
Keavon, line 54: `usize` not `i32`

bezier-rs/docs/interactive-docs/wasm/src/lib.rs
```diff
@@ -90,26 +115,39 @@ impl WasmBezier {
...
 	pub fn evaluate(&self, t: f64) -> JsValue {
-		vec_to_point(&self.0.evaluate(t))
+		let point: Point = vec_to_point(&self.0.evaluate(t));
+		to_js_value(point)
 	}
 
-	pub fn compute_lookup_table(&self, steps: i32) -> Vec<JsValue> {
-		self.0.compute_lookup_table(Some(steps)).iter().map(vec_to_point).collect()
+	/// The wrapped return type is `Vec<Point>`.
+	pub fn compute_lookup_table(&self, steps: i32) -> JsValue {
```
Keavon, line 125: `usize` not `i32`. And change the `compute_lookup_table` implementation to only use `usize` also.

bezier-rs/lib/src/utils.rs
```diff
@@ -181,6 +181,31 @@ pub fn line_intersection(point1: DVec2, point1_slope_vector: DVec2, point2: DVec
...
 }
 
+/// Check if 3 points are collinear.
+pub fn are_points_collinear(p1: DVec2, p2: DVec2, p3: DVec2) -> bool {
+	let matrix = DMat2::from_cols(p1 - p2, p2 - p3);
+	f64_compare(matrix.determinant() / 2., 0., MAX_ABSOLUTE_DIFFERENCE)
+}
+
+/// Compute the center of the circle that passes through all three provided points. The provided points cannot be collinear.
+pub fn compute_circle_center_from_points(p1: DVec2, p2: DVec2, p3: DVec2) -> DVec2 {
```
Keavon, line 191: It would be preferable to return an `Option<DVec2>` so this function doesn't panic on certain inputs. Libraries should never panic.

bezier-rs/lib/src/lib.rs
```diff
@@ -852,10 +840,7 @@ impl Bezier {
 
 		let step_size = step_size.unwrap_or(DEFAULT_REDUCE_STEP_SIZE);
 
-		let mut extrema: Vec<f64> = self.local_extrema().into_iter().flatten().collect::<Vec<f64>>();
-		extrema.append(&mut vec![0., 1.]);
-		extrema.dedup();
-		extrema.sort_by(|ex1, ex2| ex1.partial_cmp(ex2).unwrap());
+		let extrema: Vec<f64> = self.get_extrema_t_list();
```
Keavon, line 843: Explicit type unnecessary.

bezier-rs/lib/src/lib.rs
```diff
@@ -923,6 +908,150 @@ impl Bezier {
 		self.reduced_curves_and_t_values(step_size).0
 	}
 
+	/// Approximate a bezier curve with circular arcs.
+	/// The algorithm can be customized using the [ArcsOptions] structure by setting the following fields within ArcOptions:
+	/// - maximize_arcs - [MaximizeArcs] - The default value is `Automatic`.,
+	/// - error: f64 - The error used for approximating the arc's fit. The default is `0.5`.
+	/// - max_iterations: i32 - The maximum number of segment iterations used as attempts for arc approximations. The default is `100`.
+	pub fn arcs(&self, arcs_options: ArcsOptions) -> Vec<CircleArc> {
+		let ArcsOptions { maximize_arcs, error, max_iterations } = arcs_options;
```
Keavon, line 917: Add a blank line below this for readability, thanks.

Keavon (changes requested):
Another review in. Should be quick to approve once these are fixed though!

bezier-rs/lib/src/lib.rs
hannahli2010: There's some places in `intersections_between_subcurves` where we use something akin to `(low + high)/2` rather than `low + (high-low)/2`. Should we change those or leave them be?

Keavon: @hannahli2010 Yeah, I had given the same comment to you about changing those to the form `(low + high)/2` since I didn't think about overflow/underflow. I'd say change them back to `low + (high - low) / 2` wherever there's the possibility that user input could give really high or low numbers, but for efficiency keep it with `(low + high) / 2` form in places where it's unlikely or impossible that user input would cause overflow/underflow. Robustness, then efficiency, is our priority. Just remember that both styles will overflow or underflow, it's just less likely with the `low + (high - low) / 2` form because they both have to be large numbers of opposite signs. So please just use your judgement about what seems possible or likely about the user input and be robust where necessary then efficient as a second priority.

bezier-rs/lib/src/lib.rs
```diff
@@ -721,6 +705,159 @@ impl Bezier {
 		});
 		result
 	}
+
+	/// Approximate a bezier curve with circular arcs.
+	/// The algorithm can be customized using the [ArcsOptions] structure.
```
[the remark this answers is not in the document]
hannahli2010: The `ArcsOptions` struct is also in the public interface and we already explain the options there. I'd rather not have duplicate comments describing the same thing (i.e. the defaults, etc), and it makes it especially bothersome when changes are made to the struct (as I'm doing right now).

In addition, the `arcs` rustdoc description and the function signature both provide links to the `ArcsOptions` section in the docs, so we're already providing an easy way to get to the detailed description. This is also what we did with `ProjectionOptions`.

bezier-rs/lib/src/lib.rs
hannahli2010: I think it should be fine to change because for `low` and `high`, we're iterating between the range of 0 and 1 (i.e. we're iterating over the `t` range of a bezier curve.

Keavon: Ah, I was confused about that. Then yes, definitely use `(low + high) / 2` when the range is bounded.

Keavon (approved):

PR #732, opened 2022-07-18
Improve behavior of occluded layer selection and dragging

### LINKED ISSUE #705: Alter behavior of dragging unselected layers through selected layers
Right now, when an unselected layer occludes a selected layer, the user is able to click and drag on the selected layer *through* the unselected layer and make it drag. The behavior should be altered so that it requires direct click targeting to be dragged, otherwise a new drag occurs on the unselected layer, thereby selecting it immediately and beginning to drag it.

dchiasson:
Layers or groups of layers can only be dragged if they are
not occluded at the point of dragging. Otherwise the layer(s)
are deselected and the occluding layer is selected instead.

Closes #705

### DIFF
--- a/editor/src/viewport_tools/tools/select_tool.rs
+++ b/editor/src/viewport_tools/tools/select_tool.rs
@@ -445,7 +445,7 @@ impl Fsm for SelectToolFsmState {
 						tool_data.layers_dragging = selected;
 
 						RotatingBounds
-					} else if selected.iter().any(|path| intersection.contains(path)) {
+					} else if intersection.last().map(|last| selected.contains(last)).unwrap_or(false) {
 						responses.push_back(DocumentMessage::StartTransaction.into());
 						tool_data.layers_dragging = selected;
 


dchiasson:
This demonstrates the desired behavior, but I'm pretty sure it's not the right way to do it in Rust. Any tips would be appreciated.

0HyperCube:
@dchiasson It is easy to understand and it works. If you really wanted to avoid the unwrap you could do something like `intersection.last().map(|last| selected.contains(last)).unwrap_or(false)` but that may be less readable.

Keavon (approved):
The behavior feels right based on my testing. Also thanks @0HyperCube for the tips on avoiding the `.unwrap()`. @dchiasson, I'll wait for you to update that if you prefer, or we can merge now if you don't want to opt for the suggested Rust syntax change.

Are you part of the Discord server for the project, by the way? I'd like to promote your role to Code Contributor once this is merged, if so. Welcome and thank you for this contribution!

dchiasson:
> Are you part of the Discord server for the project, by the way? I'd like to promote your role to Code Contributor once this is merged, if so. Welcome and thank you for this contribution!

Yeah, thanks! My server nickname is the same as my github id.

PR #735, opened 2022-07-22
Handle intersections between two bezier curves

hannahli2010:
Refactor `intersect_line_segments` to a new `intersections` function that takes in a `Bezier` as the parameter.
- Handle intersections between quadratics and cubics in addition to linear lines 
   - Uses an algorithm akin to binary search with a curve's bounding box to locate the intersection points

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -446,7 +446,7 @@ impl Bezier {
 		bezier_starting_at_t1.split(adjusted_t2)[t2_split_side]
 	}
 
-	/// Returns the t value that corresponds to the closest point on the curve to the provided point.
+	/// Returns the `t` value that corresponds to the closest point on the curve to the provided point.
 	/// Uses a searching algorithm akin to binary search that can be customized using the [ProjectionOptions] structure.
 	pub fn project(&self, point: DVec2, options: ProjectionOptions) -> f64 {
 		let ProjectionOptions {
@@ -604,13 +604,13 @@ impl Bezier {
 
 	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
 	/// `curve1_t_interval` is used to identify the t values of the original `curve1` that the current iteration is representing.
-	/// Note that the `t` interval for only the first curve is needed since we want to return `t` with repsect to it.
-	fn intersection_curves(curve1: &Bezier, curve1_t_interval: [f64; 2], curve2: &Bezier, intersections: &mut Vec<f64>, error: f64) {
+	/// Note that the `t` interval for only the first curve is needed since we want to return `t` with respect to it.
+	fn subcurve_intersection(curve1: &Bezier, original_curve1_t_interval: [f64; 2], curve2: &Bezier, error: f64) -> Vec<f64> {
 		let bounding_box1 = curve1.bounding_box();
 		let bounding_box2 = curve2.bounding_box();
 
 		// Determine the t interval of the original curve for the split curve1
-		let [curve1_start_t, curve1_end_t] = curve1_t_interval;
+		let [curve1_start_t, curve1_end_t] = original_curve1_t_interval;
 		let curve1_mid_t = curve1_start_t + (curve1_end_t - curve1_start_t) / 2.;
 
 		let error_threshold = DVec2::new(error, error);
@@ -620,8 +620,7 @@ impl Bezier {
 			// If bounding boxes are within the error threshold (i.e. are small enough), we have found an intersection
 			if (bounding_box1[1] - bounding_box1[0]).lt(&error_threshold) && (bounding_box2[1] - bounding_box2[0]).lt(&error_threshold) {
 				// Use the middle t value
-				intersections.push(curve1_mid_t);
-				return;
+				return vec![curve1_mid_t];
 			}
 
 			// Split curves in half and repeat with the combinations of the two halves of each curve
@@ -631,17 +630,22 @@ impl Bezier {
 			// Get the new t intervals for
 			let interval_1a = [curve1_start_t, curve1_mid_t];
 			let interval_1b = [curve1_mid_t, curve1_end_t];
-			Bezier::intersection_curves(&split1a, interval_1a, &split2a, intersections, error);
-			Bezier::intersection_curves(&split1a, interval_1a, &split2b, intersections, error);
-			Bezier::intersection_curves(&split1b, interval_1b, &split2a, intersections, error);
-			Bezier::intersection_curves(&split1b, interval_1b, &split2b, intersections, error);
+			[
+				Bezier::subcurve_intersection(&split1a, interval_1a, &split2a, error),
+				Bezier::subcurve_intersection(&split1a, interval_1a, &split2b, error),
+				Bezier::subcurve_intersection(&split1b, interval_1b, &split2a, error),
+				Bezier::subcurve_intersection(&split1b, interval_1b, &split2b, error),
+			]
+			.concat()
+		} else {
+			vec![]
 		}
 	}
 
 	// TODO: Use an `impl Iterator` return type instead of a `Vec`
 	/// Returns a list of `t` values that correspond to intersection points between the current bezier curve and the provided one. The returned `t` values are with respect to the current bezier, not the provided parameter.
 	/// If either curve is linear, then zero intersection points will be returned along colinear segments.
-	/// - `error`: For intersections with non-linear beziers, `error` defines the threshold for bounding boxes to be considered an intersection point.
+	/// - `error` - For intersections with non-linear beziers, `error` defines the threshold for bounding boxes to be considered an intersection point.
 	pub fn intersections(&self, curve: &Bezier, error: Option<f64>) -> Vec<f64> {
 		let error = error.unwrap_or(0.5);
 		if let BezierHandles::Linear {} = curve.handles {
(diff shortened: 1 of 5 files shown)


bezier-rs/lib/src/lib.rs
```diff
@@ -418,9 +418,9 @@ impl Bezier {
 		bezier_starting_at_t1.split(adjusted_t2)[t2_split_side]
 	}
 
-	/// Returns the closest point on the curve to the provided point.
+	/// Returns the t value that corresponds to the closest point on the curve to the provided point.
```
RobNadal, line 421: I think we've been putting ticks around t elsewhere.

bezier-rs/lib/src/lib.rs
```diff
@@ -574,61 +574,115 @@ impl Bezier {
...
-	pub fn intersect_line_segment(&self, line: [DVec2; 2]) -> Vec<DVec2> {
-		// Rotate the bezier and the line by the angle that the line makes with the x axis
-		let slope = line[1] - line[0];
-		let angle = slope.angle_between(DVec2::new(1., 0.));
-		let rotation_matrix = DMat2::from_angle(angle);
-		let rotated_bezier = self.apply_transformation(&|point| rotation_matrix.mul_vec2(point));
-		let rotated_line = [rotation_matrix.mul_vec2(line[0]), rotation_matrix.mul_vec2(line[1])];
+	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
+	/// `curve1_t_interval` is used to identify the t values of the original `curve1` that the current iteration is representing.
+	/// Note that the `t` interval for only the first curve is needed since we want to return `t` with repsect to it.
```
RobNadal, line 579: typo repsect -> respect

bezier-rs/lib/src/lib.rs
```diff
@@ -756,6 +810,10 @@ mod tests {
 		utils::dvec2_compare(p1, p2, MAX_ABSOLUTE_DIFFERENCE).all()
 	}
 
+	fn compare_vec_of_points(vec1: Vec<DVec2>, vec2: Vec<DVec2>, max_absolute_difference: f64) -> bool {
```
RobNadal, line 813: Its not really clear from the function signature what this function is doing. Add a comment or adjust the name

bezier-rs/lib/src/utils.rs
```diff
@@ -146,6 +146,14 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
 	}
 }
 
+/// Determine if two rectangles have any overlap. The rectangles are represnted by a pair of coordinates that designate the top left and bottom right corners (in a graphical coordinate system).
```
RobNadal, line 149: typo represnted  -> represented

bezier-rs/lib/src/lib.rs
```diff
@@ -574,61 +574,115 @@ impl Bezier {
...
-		// Rotate the bezier and the line by the angle that the line makes with the x axis
-		let slope = line[1] - line[0];
-		let angle = slope.angle_between(DVec2::new(1., 0.));
-		let rotation_matrix = DMat2::from_angle(angle);
-		let rotated_bezier = self.apply_transformation(&|point| rotation_matrix.mul_vec2(point));
-		let rotated_line = [rotation_matrix.mul_vec2(line[0]), rotation_matrix.mul_vec2(line[1])];
+	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
+	/// `curve1_t_interval` is used to identify the t values of the original `curve1` that the current iteration is representing.
+	/// Note that the `t` interval for only the first curve is needed since we want to return `t` with repsect to it.
+	fn intersection_curves(curve1: &Bezier, curve1_t_interval: [f64; 2], curve2: &Bezier, intersections: &mut Vec<f64>, error: f64) {
```
linda-zheng, line 580: I think the function name could be clearer. It would be nice if it could somewhat explain the purpose of curve1_t_interval. Maybe something like `subcurve_intersection`. Also, maybe you can rename `curve1_t_interval` to `original_curve1_t_interval` or something so that it's easier to read.

linda-zheng: Also, maybe we should specify in the function description that the "return value" is actually one of the parameters.

bezier-rs/lib/src/lib.rs
```diff
@@ -845,16 +903,30 @@ mod tests {
...
+		let intersections2 = bezier.intersections(&line2, None);
 		assert!(intersections2.len() == 2);
-		assert!(compare_points(intersections2[0], p1));
-		assert!(compare_points(intersections2[1], DVec2::new(85.84, 85.84)));
+		assert!(compare_points(bezier.evaluate(intersections2[0]), p1));
+		assert!(compare_points(bezier.evaluate(intersections2[1]), DVec2::new(85.84, 85.84)));
+	}
+
+	#[test]
+	fn intersect_curve() {
```
linda-zheng, line 920: not sure, but are we prefixing our tests with `test_`?

bezier-rs/lib/src/lib.rs
hannahli2010: I rewrote it to return a vector - any thoughts on the changes?

hannahli2010:
I realized when looking into self-intersection, that bezier.js uses a slightly different algorithm for curve intersection than what's described in the primer. The slightly different algorithm is necessary for self-intersection, but it doesn't seem to be the case for curve intersection.

I plan on leaving this PR as is and do some testing with the 2 options in the self-intersection branch (since I have to implement the alternate option for self-intersection anyways). If I choose to change curve intersection, then I'll do it in that branch.

bezier-rs/lib/src/lib.rs
```diff
@@ -602,61 +602,115 @@ impl Bezier {
...
-				let b = 3. * start_y - 6. * handle_start.y + 3. * handle_end.y;
-				let c = -3. * start_y + 3. * handle_start.y;
-				let d = start_y;
+	// TODO: Use an `impl Iterator` return type instead of a `Vec`
+	/// Returns a list of `t` values that correspond to intersection points between the current bezier curve and the provided one. The returned `t` values are with respect to the current bezier, not the provided parameter.
+	/// If either curve is linear, then zero intersection points will be returned along colinear segments.
+	/// - `error` - For intersections with non-linear beziers, `error` defines the threshold for bounding boxes to be considered an intersection point.
+	pub fn intersections(&self, curve: &Bezier, error: Option<f64>) -> Vec<f64> {
+		let error = error.unwrap_or(0.5);
+		if let BezierHandles::Linear {} = curve.handles {
```
RobNadal, line 651: curve.handles = BezierHandles::Linear

bezier-rs/lib/src/lib.rs
```diff
@@ -602,61 +602,115 @@ impl Bezier {
...
-
-		// Compute the roots of the resulting bezier curve
-		let list_intersection_t = match translated_bezier.handles {
-			BezierHandles::Linear => {
-				// If the transformed linear bezier is on the x-axis, `a` and `b` will both be zero and `solve_linear` will return no roots
-				let a = translated_bezier.end.y - translated_bezier.start.y;
-				let b = translated_bezier.start.y;
-				utils::solve_linear(a, b)
+	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
+	/// `curve1_t_interval` is used to identify the t values of the original `curve1` that the current iteration is representing.
```
Keavon, line 606: `curve1_t_interval` isn't the parameter name (anymore).

bezier-rs/lib/src/lib.rs
```diff
@@ -602,61 +602,115 @@ impl Bezier {
...
+	/// Note that the `t` interval for only the first curve is needed since we want to return `t` with respect to it.
+	fn intersections_between_subcurves(curve1: &Bezier, original_curve1_t_interval: [f64; 2], curve2: &Bezier, error: f64) -> Vec<f64> {
```
Keavon, line 608: Shouldn't this take `&self` instead of two `&Bezier`s? Then you can call `curve2` "`other`".
Please also document what `original_curve1_t_interval` (or what it ends up being called, perhaps "self" or "own" instead of "curve1" in the name) means/does.

bezier-rs/lib/src/lib.rs
```diff
@@ -602,61 +602,115 @@ impl Bezier {
...
-			BezierHandles::Quadratic { handle } => {
-				let a = translated_bezier.start.y - 2. * handle.y + translated_bezier.end.y;
-				let b = 2. * (handle.y - translated_bezier.start.y);
-				let c = translated_bezier.start.y;
 
-				let discriminant = b * b - 4. * a * c;
-				let two_times_a = 2. * a;
+			// Split curves in half and repeat with the combinations of the two halves of each curve
+			let [split1a, split1b] = curve1.split(0.5);
+			let [split2a, split2b] = curve2.split(0.5);
```
Keavon, line 628: Add underscores between `split` and `1/2``a/b`.

bezier-rs/lib/src/lib.rs
```diff
@@ -602,61 +602,115 @@ impl Bezier {
...
+
+			return list_intersection_t
+				.into_iter()
+				// Accept the t value if it is approximately in [0, 1] and if the coresponding coordinates are within the range of the linear line
+				.filter(|&t| {
+					utils::f64_approximately_in_range(t, 0., 1., MAX_ABSOLUTE_DIFFERENCE)
+						&& utils::dvec2_approximately_in_range(self.unrestricted_evaluate(t), min, max, MAX_ABSOLUTE_DIFFERENCE).all()
+				})
+				// Ensure the returned value is within the correct range
+				.map(|t| t.max(0.).min(1.))
```
Keavon, line 703: Use `.clamp()`

Keavon (approved):

PR #736, opened 2022-07-24
Add full support for Mac-specific keyboard layouts

### LINKED ISSUE #206: Add full support for Mac-specific keyboard layouts
Windows and Linux use the standard keyboard layout. Mac uses <kbd>Cmd</kbd> instead of <kbd>Ctrl</kbd>. For example, copy (<kbd>Ctrl</kbd><kbd>C</kbd>) is <kbd>Cmd</kbd><kbd>A</kbd> on a Mac. Mac also uses <kbd>Option</kbd> instead of <kbd>Alt</kbd>. We need two different versions of our key bindings depending on whether it's Mac or non-Mac.

This is probably an important feature before we launch at the end of Sprint 5 with version 0.1.

- [x] Rust (Input Preprocessor/Input Mapper)
- [x] Rust (menu bar)
- [x] Rust (tooltips)
- [x] Rust (hints) - temporary solution to override the displayed hint on Macs, eventually should be replaced with #741 (also related: #739)
- [x] TypeScript
- [x] Vue ([~~event modifiers~~](https://vuejs.org/guide/essentials/event-handling.html#event-modifiers), [~~key aliases~~](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers)~~, and~~ [system modifier keys](https://vuejs.org/guide/essentials/event-handling.html#key-modifiers) - ~~some progress in #419~~)
- [ ] Switch from event.key to event.code because Alt+key combinations transform the key to special characters that don't match **(saving for another PR)**

Keavon:
Closes #206


TrueDoctor:
Did not check the frontend code, but the rest all looks good!

TrueDoctor (approved):

PR #740, opened 2022-07-29
Implement self-intersection function for Beziers

hannahli2010:
Return any self-intersections for a given Bezier curve.

### DIFF
--- a/bezier-rs/lib/src/lib.rs
+++ b/bezier-rs/lib/src/lib.rs
@@ -638,8 +638,8 @@ impl Bezier {
 
 	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
 	/// - `self_original_t_interval` - Used to identify the `t` values of the original parent of `self` that the current iteration is representing.
-	/// Note that the `t` interval the other curve is not needed since we want to return `t` with respect to it.
-	fn intersections_between_subcurves(&self, self_original_t_interval: [f64; 2], other: &Bezier, error: f64) -> Vec<f64> {
+	/// - `other_original_t_interval` - Used to identify the `t` values of the original parent of `other` that the current iteration is representing.
+	fn intersections_between_subcurves(&self, self_original_t_interval: [f64; 2], other: &Bezier, other_original_t_interval: [f64; 2], error: f64) -> Vec<[f64; 2]> {
 		let bounding_box1 = self.bounding_box();
 		let bounding_box2 = other.bounding_box();
 
@@ -647,28 +647,34 @@ impl Bezier {
 		let [curve1_start_t, curve1_end_t] = self_original_t_interval;
 		let curve1_mid_t = curve1_start_t + (curve1_end_t - curve1_start_t) / 2.;
 
+		// Get the `t` interval of the original parent of `other` and determine the middle `t` value
+		let [curve2_start_t, curve2_end_t] = other_original_t_interval;
+		let curve2_mid_t = curve2_start_t + (curve2_end_t - curve2_start_t) / 2.;
+
 		let error_threshold = DVec2::new(error, error);
 
 		// Check if the bounding boxes overlap
 		if utils::do_rectangles_overlap(bounding_box1, bounding_box2) {
 			// If bounding boxes are within the error threshold (i.e. are small enough), we have found an intersection
 			if (bounding_box1[1] - bounding_box1[0]).lt(&error_threshold) && (bounding_box2[1] - bounding_box2[0]).lt(&error_threshold) {
-				// Use the middle t value
-				return vec![curve1_mid_t];
+				// Use the middle t value, return the corresponding `t` value for `self` and `other`
+				return vec![[curve1_mid_t, curve2_mid_t]];
 			}
 
 			// Split curves in half and repeat with the combinations of the two halves of each curve
 			let [split_1_a, split_1_b] = self.split(0.5);
 			let [split_2_a, split_2_b] = other.split(0.5);
 
-			// Get the new `t` intervals for the split halves of `self`
+			// Get the new `t` intervals for the split halves of `self` and `other`
 			let interval_1_a = [curve1_start_t, curve1_mid_t];
 			let interval_1_b = [curve1_mid_t, curve1_end_t];
+			let interval_2_a = [curve2_start_t, curve2_mid_t];
+			let interval_2_b = [curve2_mid_t, curve2_end_t];
 			[
-				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_a, error),
-				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_b, error),
-				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_a, error),
-				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_b, error),
+				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_a, interval_2_a, error),
+				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_b, interval_2_b, error),
+				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_a, interval_2_a, error),
+				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_b, interval_2_b, error),
 			]
 			.concat()
 		} else {
@@ -744,12 +750,11 @@ impl Bezier {
 		}
 
 		// Otherwise, use bounding box to determine intersections
-		self.intersections_between_subcurves([0., 1.], curve, error)
+		self.intersections_between_subcurves([0., 1.], curve, [0., 1.], error).iter().map(|t_values| t_values[0]).collect()
 	}
 
 	/// Helper function to compute intersections between lists of subcurves.
-	fn intersections_between_vectors_of_curves(subcurves_1: Vec<&(Bezier, [f64; 2])>, subcurves_2: Vec<&(Bezier, [f64; 2])>, error: f64) -> Vec<f64> {
-		// let list_2: Vec<(Bezier, [f64; 2])> = subcurves_2.collect();
+	fn intersections_between_vectors_of_curves(subcurves_1: Vec<&(Bezier, [f64; 2])>, subcurves_2: Vec<&(Bezier, [f64; 2])>, error: f64) -> Vec<[f64; 2]> {
 		let segment_pairs = subcurves_1.iter().flat_map(|&(curve_1, curve_1_t_pair)| {
 			subcurves_2.iter().filter_map(move |&(curve_2, curve_2_t_pair)| {
 				if utils::do_rectangles_overlap(curve_1.bounding_box(), curve_2.bounding_box()) {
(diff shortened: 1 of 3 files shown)


bezier-rs/lib/src/lib.rs
```diff
@@ -295,6 +302,13 @@ impl Bezier {
 		}
 	}
 
+	pub fn abs_diff_eq(&self, other: &Bezier, max_abs_diff: f64) -> bool {
```
Androxium, line 305: add a rustdoc comment here because it's a public funciton

RobNadal (approved):

bezier-rs/lib/src/lib.rs
```diff
@@ -624,37 +640,43 @@ impl Bezier {
 
 	/// Implementation of the algorithm to find curve intersections by iterating on bounding boxes.
 	/// - `self_original_t_interval` - Used to identify the `t` values of the original parent of `self` that the current iteration is representing.
-	/// Note that the `t` interval the other curve is not needed since we want to return `t` with respect to it.
-	fn intersections_between_subcurves(&self, self_original_t_interval: [f64; 2], other: &Bezier, error: f64) -> Vec<f64> {
+	/// - `other_original_t_interval` - Used to identify the `t` values of the original parent of `other` that the current iteration is representing.
```
Keavon, line 643: Can you look into using [Range](https://doc.rust-lang.org/std/ops/struct.Range.html) instead of an array of length 2? I *think* this is the right use for that.

bezier-rs/lib/src/lib.rs
```diff
@@ -729,8 +751,57 @@ impl Bezier {
 			return curve.intersections(self, Some(error));
 		}
 
+		// TODO: Considering using the `intersections_between_vectors_of_curves` helper function here
```
Keavon, line 754: Consider*?

bezier-rs/lib/src/lib.rs
```diff
@@ -624,37 +640,43 @@ impl Bezier {
...
-				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_b, error),
-				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_a, error),
-				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_b, error),
+				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_a, interval_2_a, error),
+				split_1_a.intersections_between_subcurves(interval_1_a, &split_2_b, interval_2_b, error),
+				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_a, interval_2_a, error),
+				split_1_b.intersections_between_subcurves(interval_1_b, &split_2_b, interval_2_b, error),
 			]
 			.concat()
 		} else {
```
Keavon, line 682: A few lines below (GitHub doesn't let me comment on it directly), the `fn intersections` has a parameter named `curve`, can you rename that `other` please?

bezier-rs/lib/src/lib.rs
```diff
@@ -729,8 +751,57 @@ impl Bezier {
 			return curve.intersections(self, Some(error));
 		}
 
+		// TODO: Considering using the `intersections_between_vectors_of_curves` helper function here
 		// Otherwise, use bounding box to determine intersections
-		self.intersections_between_subcurves([0., 1.], curve, error)
+		self.intersections_between_subcurves([0., 1.], curve, [0., 1.], error).iter().map(|t_values| t_values[0]).collect()
```
Keavon, line 756: Avoid `collect()`, you can just mutate the values provided by `intersections_between_subcurves()`.

bezier-rs/lib/src/lib.rs
```diff
@@ -729,8 +751,57 @@ impl Bezier {
...
+	}
+
+	/// Helper function to compute intersections between lists of subcurves.
+	/// This function uses the algorithm implemented in `intersections_between_subcurves`.
+	fn intersections_between_vectors_of_curves(subcurves_1: Vec<&(Bezier, [f64; 2])>, subcurves_2: Vec<&(Bezier, [f64; 2])>, error: f64) -> Vec<[f64; 2]> {
+		let segment_pairs = subcurves_1.iter().flat_map(|&(curve_1, curve_1_t_pair)| {
+			subcurves_2.iter().filter_map(move |&(curve_2, curve_2_t_pair)| {
+				if utils::do_rectangles_overlap(curve_1.bounding_box(), curve_2.bounding_box()) {
+					Some((curve_1, curve_1_t_pair, curve_2, curve_2_t_pair))
```
Keavon, line 765: I think it would be cleaner to change every `curve_1`/`curve_2` into `curve1` and `curve2` throughout this whole file. Meaning some, like `curve_1_a` and `curve_1_t_pair` would become `curve1_a` and `curve1_t_pair`. That'd make it consistent with other variables around here just called `curve1`.

bezier-rs/lib/src/lib.rs
```diff
@@ -780,14 +851,15 @@ impl Bezier {
...
 	}
 
-	/// Split the curve into a number of scalable subcurves. This function may introduce gaps if subsections of the curve are not reducible.
+	/// Returns a tuple of the scalable subcurves and the corresponding `t` values that were used to split the curve.
+	/// This function may introduce gaps if subsections of the curve are not reducible.
 	/// The function takes the following parameter:
 	/// - `step_size` - Dictates the granularity at which the function searches for reducible subcurves. The default value is `0.01`.
 	///   A small granularity may increase the chance the function does not introduce gaps, but will increase computation time.
-	pub fn reduce(&self, step_size: Option<f64>) -> Vec<Bezier> {
+	fn reduced_curves_and_t_values(&self, step_size: Option<f64>) -> (Vec<Bezier>, Vec<f64>) {
```
Keavon, line 859: Would it make sense to return a Vec of tuples rather than a tuple of Vecs? `-> Vec<(Bezier, f64)>`

hannahli2010: That wouldn't work because there's one more `f64` than `Beziers`, since each `Bezier` is the curve between every pair of `f64`s. I could do `-> Vec<Bezier, [f64; s]>`, but I'm not sure if the duplicate data is worth it or not.

Keavon: Ok, I wasn't sure if they were parallel Vectors or not. Fine as is.

bezier-rs/lib/src/lib.rs
```diff
@@ -729,8 +751,57 @@ impl Bezier {
...
+		// Second one needs to be a list because Iterator does not implement copy
+		let combined_list_2: Vec<(Bezier, [f64; 2])> = self_2.into_iter().zip(self_2_t_values.windows(2).map(|t_pair| [t_pair[0], t_pair[1]])).collect();
+
+		// Adjacent reduced curves cannot intersect
+		// So for each curve, look for intersections with every curve that is at least 2 indices away
+		combined_iterator_1
+			.take(num_curves - 2)
+			.enumerate()
+			.flat_map(|(index, (subcurve, t_pair))| {
+				Bezier::intersections_between_vectors_of_curves(vec![&(subcurve, t_pair)], combined_list_2.iter().skip(index + 2).collect::<Vec<&(Bezier, [f64; 2])>>(), error)
```
hannahli2010, line 802: @Keavon  I was wondering if there's a better way to get a slice over the range `[index+2, end]` rather than turning `combined_list_2` to an iterator and using `skip`?

Keavon: `[index+2..]`

bezier-rs/lib/src/lib.rs
hannahli2010: The return type from `intersections_between_subcurves` is a pair of 2 values. I only want to return one of them here, so I can't mutate the returned list.

bezier-rs/lib/src/lib.rs
```diff
@@ -709,8 +730,8 @@ impl Bezier {
 				}
 			};
 
-			let min = curve.start.min(curve.end);
-			let max = curve.start.max(curve.end);
+			let min = other.start.min(other.end);
+			let max = other.start.max(other.end);
 
 			return list_intersection_t
```
Keavon, line 736: Remove `return`. Also a todo list item for you: if you can try and investigate why Clippy isn't automatically suggesting these things for you, that'd be a helpful at-the-time-of-writing code reviewer for you. The main codebase is set up to do that but I'm not sure why it isn't working here in your library.

bezier-rs/lib/src/lib.rs
Keavon: Ah you're right, I didn't notice that.

bezier-rs/lib/src/lib.rs
hannahli2010: There `return` statement is necessary here because this is inside of an `if` block and instead having of an `else` block to handle the other case, I simply return early.

bezier-rs/lib/src/lib.rs
```diff
@@ -295,6 +303,15 @@ impl Bezier {
 		}
 	}
 
+	/// Returns true if the corresponding points of the two `Bezier`s are within the provided absolute value difference from each other.
+	/// The points considered includes the start, end and any relevant handles.
```
Keavon, line 307: Nit pick: add a comma after "end".

bezier-rs/lib/src/lib.rs
Keavon: Oh, my bad, somehow I didn't notice that this wasn't the end of the function. It looked like it when I was reading, and I even remember double checking. Strange, sorry!

Glad to hear Clippy is working also and that you're finding it useful.

bezier-rs/lib/src/lib.rs
```diff
@@ -624,37 +641,41 @@ impl Bezier {
...
-		let curve1_mid_t = curve1_start_t + (curve1_end_t - curve1_start_t) / 2.;
+		let Range { start: self_start_t, end: self_end_t } = self_original_t_interval;
+		let self_mid_t = self_start_t + (self_end_t - self_start_t) / 2.;
+
+		// Get the `t` interval of the original parent of `other` and determine the middle `t` value
+		let Range {
+			start: other_start_t,
+			end: other_end_t,
+		} = other_original_t_interval;
+		let other_mid_t = other_start_t + (other_end_t - other_start_t) / 2.;
```
Keavon, line 658: (both this one and the one a few lines above)
```rs
let other_mid_t = (other_end_t + other_start_t) / 2.;
```
is equivalent, and more efficient:
image

bezier-rs/lib/src/lib.rs
```diff
@@ -724,13 +745,55 @@ impl Bezier {
...
+	/// Helper function to compute intersections between lists of subcurves.
+	/// This function uses the algorithm implemented in `intersections_between_subcurves`.
+	fn intersections_between_vectors_of_curves(subcurves1: &[(Bezier, Range<f64>)], subcurves2: &[(Bezier, Range<f64>)], error: f64) -> Vec<[f64; 2]> {
+		let segment_pairs = subcurves1.iter().flat_map(move |(curve1, curve1_t_pair)| {
+			subcurves2.iter().filter_map(move |(curve2, curve2_t_pair)| {
+				if utils::do_rectangles_overlap(curve1.bounding_box(), curve2.bounding_box()) {
+					Some((curve1, curve1_t_pair, curve2, curve2_t_pair))
+				} else {
+					None
+				}
```
Keavon, line 762: It would be slightly more idiomatic to use `.then()` here, see https://stackoverflow.com/questions/54841351/how-do-i-idiomatically-convert-a-bool-to-an-option-or-result-in-rust

bezier-rs/lib/src/lib.rs
```diff
@@ -972,7 +1046,7 @@ mod tests {
 	}
 
 	// Compare vectors of beziers by allowing some maximum absolute difference between points to account for floating point errors
-	fn compare_vector_of_beziers(beziers: Vec<Bezier>, expected_bezier_points: Vec<Vec<DVec2>>) -> bool {
+	fn compare_vector_of_beziers(beziers: &[Bezier], expected_bezier_points: Vec<Vec<DVec2>>) -> bool {
```
Keavon, line 1049: Good change, thanks.

Keavon (approved):
Looks good! Just a few little things, then please go ahead and merge :)

bezier-rs/lib/src/lib.rs
hannahli2010: Credits go to Clippy :)

Keavon: Clippy is the MVP!

PR #745, opened 2022-08-06
Fix misc. typos

luzpaz:
Found via `codespell -q 3 -L complies,crate,doubleclick,ist,ser,vew`


Keavon (approved):
Thank you! At some point I should really figure out some proper infrastructural solution to prevent code from being merged with spelling errors. But we also have to whitelist a lot of things like compound words and abbreviations used by libraries. And I don't like waiting 5 minutes for a CI failure being a common outcome, so that leads me to think Git hooks or something else to warn the user instantly. But there's no nice and simple way, without depending on complex dependencies, to set up Git hooks for all contributors. I'm open to ideas if you have any! In the meantime, thanks again for this PR.

luzpaz:
@Keavon Check out https://github.com/codespell-project/actions-codespell

PR #751, opened 2022-08-17
Divide the large Bezier-rs implementation file into smaller ones

hannahli2010:
Refactor the `Bezier` implementation in `lib.rs` by splitting it into separate files and moving them into a `bezier/` folder.


0HyperCube (approved):
Looks good

libraries/bezier-rs/src/utils.rs
```diff
@@ -131,6 +129,7 @@ pub fn solve_cubic(a: f64, b: f64, c: f64, d: f64) -> Vec<f64> {
 		} else {
 			// If a is approximately 0, treat as a quadratic problem
 			let discriminant = c * c - 4. * b * d;
+			println!("{:?}", discriminant);
```
Keavon, line 132: Remember to remove println, which I assume is left over.

libraries/bezier-rs/src/utils.rs
```diff
@@ -271,6 +270,7 @@ mod tests {
 
 		// quadratic
 		let roots6 = solve_cubic(0., 3., 0., -3.);
+		println!("{:?}", roots6);
```
Keavon, line 273: Remember to remove println, which I assume is left over.

libraries/bezier-rs/src/lib.rs
```diff
@@ -1,1447 +1,9 @@
 //! Bezier-rs: A Bezier Math Library for Rust
 
+mod bezier;
```
Keavon, line 3: What do you think about calling this module "segment" instead of "bezier"?

libraries/bezier-rs/src/utils.rs
hannahli2010: Ah, thanks for catching that!

libraries/bezier-rs/src/bezier/mod.rs
```diff
@@ -0,0 +1,51 @@
...
+mod core;
+mod lookup;
+mod manipulators;
+mod solvers;
+mod structs;
+mod transform;
+
+use crate::consts::*;
+use crate::utils;
+pub use structs::*;
```
Keavon, line 13: Since this last one is `pub` unlike the previous two, please put a newline above it to separate the non-pub and pub ones.

libraries/bezier-rs/src/lib.rs
Keavon: We decided to defer this change because it would require renaming the `Bezier` and `BezierHandles` structs and those would necessitate lots of comment changes also. This should probably be done before publishing the library though.

Keavon (approved):

PR #753, opened 2022-08-18
Add path closing and segment extension to the Pen tool

### LINKED ISSUE #577: Closing shapes with the Pen tool
Click the start point to create a closed path.

### LINKED ISSUE #757: Extend path segments with the Pen tool

0HyperCube:
Closes #577 
Closes #757

When we get inserting points with path tool, we can become an actual vector editor
image


Keavon (approved):

Keavon:
Finally! 😁 

One very small thing to mention about the UX: it would be a tiny bit better if the close shape (`Z` command) could be part of the rendered shape as soon as the user clicks down on the starting anchor point (and maybe subsequently drags around the handle, or maybe just waits there without moving the mouse until releasing). Right now it only adds the `Z` when the user releases from the drag. This is very minor and not important enough to waste much time on yet, but if it's a fast change then it'd be a slight bit more polished that way. Feel free to merge now or after such a change, up to you!

0HyperCube:
That is somewhat hard to do because then a curve gets drawn with a start and end at the anchor and with the handle that you are dragging (but is then deleted).

editor/src/messages/tool/tool_messages/pen_tool.rs
```diff
@@ -236,114 +271,245 @@ impl Fsm for PenToolFsmState {
...
+						let subpath = document.graphene_document.layer(layer_path).ok().and_then(|layer| layer.as_subpath())?;
+
+						// Get the last manipulator group and the one previous to that
+						let mut manipulator_groups = subpath.manipulator_groups().enumerate();
+						let (&last_id, last_manipulator_group) = if tool_data.from_start { manipulator_groups.next()? } else { manipulator_groups.next_back()? };
+						let previous = if tool_data.from_start { manipulator_groups.next() } else { manipulator_groups.next_back() };
+
+						// Get the first manipulator group
+						let mut manipulator_groups = subpath.manipulator_groups().enumerate();
+						let (&first_id, first_manipulator_group) = if tool_data.from_start { manipulator_groups.next_back()? } else { manipulator_groups.next()? };
```
Keavon, line 296: I'm noticing we repeat this (and perhaps some of the other code blocks near this section) in a handful of places in this file. Do you see any way we can reduce the code duplication?

Keavon (approved):
Very neat trick to use a closure so you can use the `?` operator! I'll need to remember to use that myself. Check out my one comment and see if that's addressable or not, but besides that please go ahead and merge this as soon as you're ready! I'm delighted to have this very important feature, I can't wait to release it!

PR #754, opened 2022-08-18
Add point insertion to the Path tool

### LINKED ISSUE #544: Adding points to shapes with the Path Tool
The path tool needs the ability to create points.

When creating points we want to maintain the curve / shape of the Bezier, this resource should help: https://pomax.github.io/bezierinfo/ 

When deleting points we want to differentiate between dissolving and deleting entirely. Dissolving keeps the connections to nearby points, deleting deletes the edges joining them and the point. A modifier key might work here. If one had to be chosen to be implemented, dissolve is preferred.

The files to look at when working on this are: shape_editor.rs,  vector_shape.rs, vector_anchor.rs

Things to keep in mind: Remember to cleanup the overlays by calling "remove_overlay(..)" or similar when you delete an anchor.

Additional context: The path tools life cycle is a little strange, whenever you "select" a shape it recreates the VectorShape for that shape. You may need to recreate the VectorShape each time you create or delete a point currently. The path tool does not support editing the VectorShape without recreating it, for example update_shape assumes the shape is of the same number of PathElements. If VectorShape needs to be more dynamic, you'll need to do some refactoring. I'd start with the constructor and break out how it loads in the shape. Obviously recreating the VectorShape each selection is a stop-gap, not ideal.

See Pen tool for some "hacks" to get around these limitations for now.

0HyperCube:
Closes #544

I integrated `beziers-rs` library as well seeing as it contains the functionality we need.

A `bezier_iter()` function on a subpath allows you to iterate over `BezierId`s which contain `bezier_rs`'s bézier segments along with the ids of the manipulator groups for the start and end points, and, for quadratics, the mid point.

I also replaced kurbo's bounding box with simply finding the extremes of these segments.

For operations applied to the whole path like booleans we will probably convert the entire path to the `bezier_rs` representation and then back, however this will probably be quite rare.

We've almost got a functioning path and pen tool now :)


graphene/src/layers/vector/subpath.rs
```diff
@@ -334,15 +334,14 @@ impl Subpath {
 		&mut self.0
 	}
 
-	// ** INTERFACE WITH KURBO **
-
-	// TODO Implement our own a local bounding box calculation
 	/// Return the bounding box of the shape
-	pub fn bounding_box(&self) -> Rect {
-		<&Self as Into<BezPath>>::into(self).bounding_box()
+	pub fn bounding_box(&self) -> Option<[DVec2; 2]> {
```
hannahli2010, line 338: We should definitely add this function to the `bezier-rs` implementation of `Subpath`. I'm wondering if it'd be beneficial to do it as part of this PR.

If you were to add this code into `libraries/bezier-rs/subpath/solvers.rs` and use the `bezier-rs` `Subpath` instead, I think you'd be able to avoid needing `PathIter` for this part of the code.

Something similar can be said for the projection part in `closest_segement`.

0HyperCube: Heap allocating the `bezier_rs` subpath again based on the graphite subpath every time we want the bounding box (every frame when you drag the select cage) is probably quite expensive, especially for text.

Keavon:
I love that double clicking works while in the Path tool to add an anchor point. That's something Illustrator doesn't do, but should. Illustrator does it by left clicking using the Pen tool, which is a little weird if you think about it but it's something people get used to. For discoverability purposes, would it be difficult to also add that behavior also to the Pen tool so single-clicking along an edge of any existing path also inserts a point?

If the user has a previously-created layer selected, hovering somewhere along the curve (within the snapping distance const) would draw an overlay dot there along the curve. If the layer is *not* selected, that isn't the case and clicking along a path would just start a new path.

If the user is actively drawing a path by extending it with more segments, it would ideally work by temporarily hiding the currently-being-drawn segment preview and showing an overlay dot along the path at the point where clicking would insert the point; then after inserting the point, the user can move the mouse around and continue seeing the currently-being-drawn segment preview from where they left off. (In other words, inserting a point along the path temporarily switches to supporting that goal instead of the task of drawing additional segments.) Handling the case in the previous paragraph is more important than the one described in this paragraph (consider this one an eventually-nice-to-have stretch goal, the previous one is more important for users familiar with the Illustrator control scheme to figure out how point insertion works).

It's great to see you were able to integrate the bezier-rs library and found its functions useful! That means this initiative is working which I'm glad to see. It also seems that the point insertion is very accurate, I don't see any changes to the shape of the path which is excellent. Real vector editor, here we come! :)

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -259,6 +259,73 @@ impl ShapeEditor {
 		result
 	}
 
+	/// Find the path segement we have clicked upon
+	///
+	/// Returns the [`BezierId`] and t as an f64
+	fn closest_segement(&self, document: &Document, layer_path: &[LayerId], pos: glam::DVec2, tolerance: f64) -> Option<(BezierId, f64)> {
```
nupsi, line 265: s**ege**ment is used in here and in comments in `graphene/src/layers/vector/subpath.rs` is this a typo for segment, or does it mean something special?

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -259,6 +259,73 @@ impl ShapeEditor {
...
+				}
+			}
+		}
+		result
+	}
+
+	/// Find the path segement we have clicked upon
+	///
+	/// Returns the [`BezierId`] and t as an f64
+	pub fn split(&self, document: &Document, pos: glam::DVec2, tolerance: f64, responses: &mut VecDeque<Message>) {
```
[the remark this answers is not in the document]
0HyperCube: Good spot

editor/src/messages/tool/common_functionality/shape_editor.rs
0HyperCube: Just a typo

0HyperCube:
> I love that double clicking works while in the Path tool to add an anchor point

Seems to be pretty standard to me (in krita and inkscape).

> If the user has a previously-created layer selected, hovering somewhere along the curve (within the snapping distance const) would draw an overlay dot there along the curve. 

Ok - should it change to the path tool when a point is inserted?

> If the user is actively drawing a path by extending it with more segments, it would ideally work by temporarily hiding the currently-being-drawn segment preview and showing an overlay dot along the path at the point where clicking would insert the point

What happens if the user wants to place an anchor so that it is ontop of an existing piece of path?

0HyperCube:
I modified a lot of the pen tool in #753 to remove the nested `if let Some`s and to support adding to the start of the path if you extend from the start so I might wait until that is merged. Does this block this PR as neither inkscape or krita seem to have this functionality?

Keavon:
> Ok - should it change to the path tool when a point is inserted?

Nope, it stays on the Pen tool and just adds the point to the selected layer, that's all. Which means the Pen tool has two functions when a layer is selected: clicking along a path inserts a point (this pPR) and clicking on an endpoint continues extending segments (#753). There's actually one more we'll want to implement soon also: clicking an anchor point deletes the anchor point (so you can insert, delete, and extend the endpoints with the Pen tool, which is pretty much all you need to make arbitrary modifications to a curved shape).

> What happens if the user wants to place an anchor so that it is ontop of an existing piece of path?

Fair enough, that's a good point. Maybe disregard that idea then, because I see the value in what you described.

> Does this block this PR as neither inkscape or krita seem to have this functionality?

As is always the case, further polishing should never be considered a blocker; incremental PRs are a good idea. It's always your call if you want to integrate code review comments in the same PR or a later one (or disregard them entirely, that's entirely ok too). I'm in agreement that it's probably wise to keep these PRs small and sequence them so we don't have unnecessary merge conflicts.

graphene/src/layers/vector/subpath.rs
Keavon: Is there any chance we could work towards a goal of making the bezier-rs library all stack-based? Or is there no way to achieve that because of some fundamental contradiction with the problem it solves?

Keavon (approved):
LGTM!

graphene/src/layers/vector/subpath.rs
0HyperCube: The way I'm using it here (with the `PathIter`) is all stack based - this bounding_box function will not heap allocate anything.

PR #758, opened 2022-08-24
Refactor bezier-rs' `to_svg` functionality

hannahli2010:
- Update the `to_svg` implementation for the `Bezier` struct
- Re-organise the `bezier-rs-demo` wasm module in preparation for the future refactor (switching to return svgs)


libraries/bezier-rs/src/bezier/core.rs
```diff
@@ -141,13 +142,34 @@ impl Bezier {
...
+			self.start.x, self.start.y, anchor_arguments, self.end.x, self.end.y, anchor_arguments
+		);
+
+		let handle_point_arguments = options.formatted_handle_point_arguments();
+		let handle_circles = match self.handles {
+			BezierHandles::Linear => None,
+			BezierHandles::Quadratic { handle } => Some(format!(r#"<circle cx="{}" cy="{}" {}/>"#, handle.x, handle.y, handle_point_arguments)),
+			BezierHandles::Cubic { handle_start, handle_end } => Some(format!(
+				r#"<circle cx="{}" cy="{}" {}/><circle cx="{}" cy="{}" {}/>"#,
+				handle_start.x, handle_start.y, handle_point_arguments, handle_end.x, handle_end.y, handle_point_arguments
```
RobNadal, line 159: Here and elsewhere, when possible just stick variables directly into the `{}`, I think it makes it more readable. So in this case, handle_point_arguments could go in the string directly.

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -0,0 +1,259 @@
...
+
+	/// The wrapped return type is `Vec<Point>`.
+	pub fn get_points(&self) -> JsValue {
+		let points: Vec<Point> = self.0.get_points().map(|point| vec_to_point(&point)).collect();
+		to_js_value(points)
+	}
+
+	// pub fn to_svg(&self) -> String {
+	// 	self.0.to_svg()
+	// }
```
RobNadal, line 112: remove extraneous comment

libraries/bezier-rs/src/bezier/core.rs
hannahli2010: I think I got all the locations that can be changed, but I would appreciate another quick look!

libraries/bezier-rs/src/bezier/core.rs
```diff
@@ -141,13 +142,34 @@ impl Bezier {
...
+			r#"<path d="{SVG_ARG_MOVE}{} {} {}" {}/><path d="{}" {}/>{}{}"#,
 			self.start.x,
 			self.start.y,
-			self.svg_curve_argument()
+			self.svg_curve_argument(),
+			options.formatted_curve_arguments(),
+			self.svg_handle_line_argument().unwrap_or_else(|| "".to_string()),
+			options.formatted_handle_line_arguments(),
+			anchor_circles,
+			handle_circles.unwrap_or_else(|| "".to_string()),
```
0HyperCube, line 172: Perhaps this function could take a mutable string and use `write!` macro to avoid excess heap allocations (like formatting anchor_circles` and `handle_circles` and then formatting them again)?

libraries/bezier-rs/src/svg.rs
```diff
@@ -0,0 +1,74 @@
...
+	/// Stroke color outlining circles marking anchors on the `Subpath`. Defaulted to `black`.
+	pub anchor_stroke_color: String,
+	/// Stroke width outlining circles marking anchors on the `Subpath`. Defaulted to `2.`.
+	pub anchor_stroke_width: f64,
+	/// Radius of the circles marking anchors on the `Subpath`. Defaulted to `4.`.
+	pub anchor_radius: f64,
+	/// Fill color of the circles marking anchors on the `Subpath`. Defaulted to `white`.
+	pub anchor_fill: String,
+	/// Color of the line segments connecting anchors to handle points. Defaulted to `gray`.
+	pub handle_line_stroke_color: String,
```
0HyperCube, line 16: Could we make some of these `String`s `&'static str`s to reduce heap allocations? I don't really see a case where these should be generated at runtime.

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -0,0 +1,256 @@
...
+struct Point {
+	x: f64,
+	y: f64,
+}
+
+#[wasm_bindgen]
+pub enum WasmMaximizeArcs {
+	Automatic, // 0
+	On,        // 1
+	Off,       // 2
```
0HyperCube, line 27: If you want to be explicit in the ordering you can assign explicit integer tag values:
```rs
enum MyEnum {
    A = 123,
    B = 456,
}
```

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -0,0 +1,256 @@
...
+
+#[derive(Serialize, Deserialize)]
+struct CircleSector {
+	center: Point,
+	radius: f64,
+	#[serde(rename = "startAngle")]
+	start_angle: f64,
+	#[serde(rename = "endAngle")]
+	end_angle: f64,
+}
```
0HyperCube, line 15: Perhaps a clearer name for this would be `Arc`?

Keavon (commented):
I agree with Hypercube's comments and didn't notice anything else to add. Will approve after those are resolved 👍

website/other/bezier-rs-demos/wasm/src/bezier.rs
Keavon: The 0/1/2 comments are in reference to the possible integer values of the range slider widget to which each of these correspond, so I don't think that's necessary unless I misunderstand your intent.

website/other/bezier-rs-demos/wasm/src/bezier.rs
Keavon: I'm happy with either. I agree "Arc" might be the more official name, but I do think "circle sector" may be more specific and less overloaded. To me at least, I get a more intuitive understanding from the current name. I don't mind either way.

website/other/bezier-rs-demos/wasm/src/bezier.rs
hannahli2010: The ordering for this enum does not really matter when using the library. The ordering is only for the UI slider representation of the 3 options which is unique to the `bezier-rs-demos` web app.

0HyperCube: OK, I misunderstood what this was signifying.

libraries/bezier-rs/src/svg.rs
hannahli2010: I can't say for other people's use cases for the `to_svg` function, but we're currently trying to replace all our HTML canvas usages to this svg handler. For the most part, we're mainly using colours like red, black, gray, etc. But there's a few `bezier-rs` demos that use something akin to `hsl(${40 * idx}, 100%, 50%)` to create a rainbow effect with variable number of colours. 

Do you have any recommendations in mind for how to replicate this if we change the `String`s to `&'static str`s?

libraries/bezier-rs/src/bezier/core.rs
hannahli2010: I've made what I think is the change you requested - could you confirm that I did what you expected?

libraries/bezier-rs/src/svg.rs
0HyperCube: Actually perhaps if you provide a public function to write just the string inside the `d="..."` attribute then this means that graphite can apply whatever styles we like (e.g. a dash array) without increasing the scope of the library beyond bezier curves. Then this struct isn't so performance critical.

Keavon: I like this plan best.

libraries/bezier-rs/src/bezier/core.rs
```diff
@@ -142,14 +144,36 @@ impl Bezier {
...
 
-	/// Convert `Bezier` to SVG `path`.
-	pub fn to_svg(&self) -> String {
-		format!(
-			r#"<path d="{SVG_ARG_MOVE}{} {} {}" stroke="black" fill="none"/>"#,
+	/// Returns an SVG representation of the `Bezier`.
+	pub fn to_svg(&self, svg: &mut String, options: ToSVGOptions) {
+		let _ = write!(
+			svg,
+			r#"<path d="{SVG_ARG_MOVE}{} {} {}" {}/><path d="{}" {}/>"#,
```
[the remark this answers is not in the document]
0HyperCube: `SVG_ARG_MOVE` is just the 'M' character, then it writes the start X then the start Y and then the rest of the curve.

libraries/bezier-rs/src/svg.rs
hannahli2010: In this case, would it be better to split the `to_svg` function up into different ones that draw:

1. The curve itself
2. The curve's endpoints
4. The curve's handles
4. The lines from the endpoints to the handles

In this way, each function will have only one string parameter to write (i.e. the `"d=..."` attribute of a specific path/shape/HTML entity, assuming that endpoints and such should have the same styling). Additionally, the users have the flexibility to choose which of those features they want drawn in the SVG without needing to overcomplicate function parameters/options.

0HyperCube: Sounds good.

hannahli2010: I've changed the `Bezier` code to no longer use this struct. The `Subpath` logic still uses `ToSVGOption` at the moment, but I'll work on that after this refactor is finalized.

0HyperCube (approved):
Looks good 👍🏽

PR #760, opened 2022-08-27
Use .then() instead of .after() for sequencing nodes

0HyperCube:
I think it provides a clearer ordering but I don't really mind sticking with .after().


Keavon (approved):
Hundreds of times more readable and more feasible to wrap one's mind around using `then`, thank you.

0HyperCube:
Fixed @Keavon. I didn't see the `AfterBox` and `AfterRef` traits below the `After` trait.

Keavon (approved):

PR #762, opened 2022-08-27
Add Alt-dragging to duplicate layers

### LINKED ISSUE #766: Alt-drag should duplicate layers
Allow duplication by dragging while holding Alt, since we have a keyboard hint that doesn't say "(Coming Soon)" saying this is possible

0HyperCube:
Closes #766
> Allow duplication by dragging while holding Alt, since we have a keyboard hint that doesn't say "(Coming Soon)" saying this is possible

Notes:
- Duplicated layers go to the top of the folder of the folder that the previously selected layers were in.
- You drag the duplicated layers and not the originals.
- Dragging a layer as well as its parent duplicates just the parent.

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -517,7 +517,7 @@ impl Fsm for SelectToolFsmState {
 					if input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_none() {
 						tool_data.start_duplicates(document, responses);
 					} else if !input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_some() {
-						tool_data.remove_duplicates(responses);
+						tool_data.stop_duplicates(responses);
 					}
 
 					let closest_move = tool_data.snap_manager.snap_layers(responses, document, snap, mouse_delta);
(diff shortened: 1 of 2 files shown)


Keavon:
We'd like it so holding <kbd>Alt</kbd>, and then clicking and dragging an unselected layer, should perform the operation. Not only if it's already selected. Thanks :)

Keavon:
It would also be great (although I realize it would be a little harder to implement) if you can randomly add or remove the <kbd>Alt</kbd> key at any time before or during the drag makes it so the layer is duplicated or not duplicated.

- If you're holding the <kbd>Alt</kbd> key at any time: you are dragging the copy
- If you're not holding the <kbd>Alt</kbd> key at any time: you are dragging the original

This does mean that the duplicate layer(s) has to be created or deleted potentially as many times as you press and release the key.

0HyperCube:
@Keavon These are both added now.

Keavon:
> @Keavon These are both added now.

Nice! This works perfectly. 👌

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -509,6 +514,12 @@ impl Fsm for SelectToolFsmState {
 						.flat_map(snapping::expand_bounds)
 						.collect();
 
+					if input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_none() {
+						tool_data.duplicate(document, responses);
+					} else if !input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_some() {
+						tool_data.undo_duplicate(responses);
```
Keavon, line 520: How about calling them `create_duplicate()` and `remove_duplicate()`?

Keavon (approved):
I'm surprised this can be implemented with less than 100 lines of code added. It's very clean!

PR #763, opened 2022-08-27
Fix fast drag with path tool eating handles

0HyperCube:
Fixes:
> If you drag an anchor fast enough towards a handle that is attached to it the handle will disappear.

I removed the absolute position (which is sent through a surprisingly large number of files) so points are now just dragged by a delta. Assuming the input system isn't broken, then all the deltas should add up to exactly the absolute position (the floating point error won't be very much unless you are extremely zoomed in and extremely far away from the origin).


Keavon (approved):
What a bizarre bug this was. Nice job even discovering it. Fix looks good!

PR #772, opened 2022-08-29
Add a pivot widget to selected layer(s) to control the origin(s)

### LINKED ISSUE #525: Setting pivot of rotation/scale in the transform cage
It would be useful to provide a center of rotation. 

This can be especially useful for creating gauges etc. The center of rotation should be remembered (eg. saved with object).

Is this functionality planned, or is it implemented and did I miss it?

0HyperCube:
Closes #525

Pivot is viewable and draggable in select tool and can be controlled with the tiny grid of squares. Functionality is roughly the same as in Inkscape (except I didn't add shift clicking the pivot to reset it as you can click the dot in the grid).

### DIFF
--- a/editor/src/messages/tool/common_functionality/pivot.rs
+++ b/editor/src/messages/tool/common_functionality/pivot.rs
@@ -9,6 +9,9 @@ use graphene::{layers::text_layer::FontCache, LayerId};
 use glam::{DAffine2, DVec2};
 use std::collections::VecDeque;
 
+const PIVOT_WIDTH: f64 = 10.;
+const PIVOT_SIZE: f64 = 40.;
+
 #[derive(Clone, Debug)]
 pub struct Pivot {
 	/// Pivot between (0,0) and (1,1)
@@ -95,9 +98,6 @@ impl Pivot {
 			}
 		};
 
-		const PIVOT_WIDTH: f64 = 10.;
-		const PIVOT_SIZE: f64 = 40.;
-
 		let layer_paths = [vec![generate_uuid()], vec![generate_uuid()]];
 		for index in 0..=1 {
 			responses.push_back(
@@ -152,5 +152,9 @@ impl Pivot {
 
 	pub fn set_normalised_position(&self, position: DVec2, document: &DocumentMessageHandler, font_cache: &FontCache, responses: &mut VecDeque<Message>) {
 		self.set_viewport_position(self.transform_from_normalized.transform_point2(position), document, font_cache, responses);
+	}
+
+	pub fn is_over(&self, mouse: DVec2) -> bool {
+		self.pivot.filter(|&pivot| mouse.distance_squared(pivot) < (PIVOT_SIZE / 2.).powi(2)).is_some()
 	}
 }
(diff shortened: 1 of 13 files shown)


Keavon:
Working nicely! Just a few notes:
- [ ] I've noticed that layers besides the basic shapes (Pen tool drawings, text, booleaned shapes, etc.) are missing the crosshair (except when multiple layers are selected, then it shows the average position)
- [ ] Can you please make the crosshair into a 9x9 circle with a 1px inner blue stroke and 7x7 of filled white, with another 3x3 blue filled circle inside that?
  image
  image
- [ ] When hovering/dragging the pivot, let's use the `cursor: move;` CSS cursor property.
  image

0HyperCube:
> I've noticed that layers besides the basic shapes (Pen tool drawings, text, booleaned shapes, etc.) are missing the crosshair (except when multiple layers are selected, then it shows the average position)

I accidentally multiplied some matrices in the wrong order.

> Can you please make the crosshair into a 9x9 circle with a 1px inner blue stroke and 7x7 of filled white, with another 3x3 blue filled circle inside that?

We don't really have inner stroke so feel free to mess with `consts.rs` to change that to your liking @Keavon.

> When hovering/dragging the pivot, let's use the cursor: move; CSS cursor property.

OK - I've added the new cursor.

Keavon:
> We don't really have inner stroke so feel free to mess with `consts.rs` to change that to your liking @Keavon.
Ah true, in fact the entire SVG spec (to my great annoyance) doesn't support path alignment. (SVG 2 will add that, *someday*). But in this case 9x9 outer diameter and 7x7 inner diameter means we can just use 8x8 so it stretches half a pixel outwards and half a pixel inwards on both sides, adding 0.5 on both sides equaling 1 pixel on the outside and 1 pixel on the inside. So I've updated the consts.rs file to 8x8. With that done, this PR is ready for prime time! It's actually a really big quality of life improvement. I'll just code review it now.

Keavon:
Okay, after a bit more QA testing of the UX, there's just one other relatively specific issue. If you set the pivot to an edge, and drag the opposite edge while holding <kbd>Alt</kbd>, it appears to scale at twice the speed. The desired behavior would keep that edge aligned with the mouse along its axis. [Video](https://files.keavon.com/-/BrightOblongAndeancockoftherock/capture_3_.mp4)

editor/src/messages/tool/common_functionality/pivot.rs
```diff
@@ -0,0 +1,176 @@
...
+		let bounds_transform = DAffine2::from_translation(min) * DAffine2::from_scale(max - min);
+		let layer_transform = document.graphene_document.multiply_transforms(layer_path).unwrap_or(DAffine2::IDENTITY);
+		layer_transform * bounds_transform
+	}
+
+	/// Recomputes the pivot position and transform.
+	fn recalculate_pivot(&mut self, document: &DocumentMessageHandler, font_cache: &FontCache) {
+		let mut layers = document.selected_visible_layers();
+		if let Some(first) = layers.next() {
+			let selected_layers_count = layers.count() + 1;
```
Keavon, line 54: Could you please comment above this line why we add 1 to the count?

Keavon (approved):

Issue #759, opened 2022-08-27
Draw an overlay for the origin point of the selected layer(s)


Keavon:

dchiasson:
I will look into this one.

Keavon:
Obsoleted by #525.

Issue #722, opened 2022-07-09
Improve Path tool layer and anchor point selection behavior


Keavon:
- [x] Deselects the shape layer (from the layer panel) when clicking into empty space
- [x] Selects all the shape's layer (in the layer panel) and selects all the anchor points when clicking in the filled area of the shape
- [x] Selects the targeted shape when dragging (regardless of whether or not it was previously a selected layer or whether it had any or some or all anchor points selected) and then drags the whole shape similar to the Select tool
- [x] Selecting an individual already-selected anchor point deselects all but the clicked one

noblehelm:
Can I try to work on this?

Keavon:
@noblehelm Thanks for mentioning that some of this is already implemented! I went and checked off the ones that are implemented, and clarified a few of them. Those remaining are still needed, so you're welcome to work on this and I can help clarify anything if you need help with it. I saw you also commented on the history system, I'll just say that is of greater importance to usability but it's up to you which you'd like to work on.

noblehelm:
For the 4th task, is there a way to differentiate from a simple click and a _click and hold_? Because I think that would probably get in the way of selecting multiple anchors with Shift and releasing Shift to drag them.

noblehelm:
One caveat is that the InputEventHandler treats KeyUp(Lmb) as DragStop always for the Path tool, according to the images below:
![image](https://user-images.githubusercontent.com/5214283/195427365-140f1d55-94d3-4b06-a363-dec3ed885cf6.png)
![image](https://user-images.githubusercontent.com/5214283/195427424-6b8ea4a3-0d14-4876-a80e-7be3f06eb479.png)

So, should I separate into two different INputEvents, one for the already present DragStop, and another for a SimpleClick? If so, how can I separate it? Should I treat a mouse cursor movement as <= 10px in between MouseDown and MouseUp as a single click, and anything else as DragStart/DragStop?

ChrisMend19:
Test comment for assignment for this issue

OllieDolan:
Working on issue

PR #773, opened 2022-08-30
Add overlay for origin point

dchiasson:
Closes: #759

Closes #

### DIFF
--- a/editor/src/messages/tool/common_functionality/transformation_cage.rs
+++ b/editor/src/messages/tool/common_functionality/transformation_cage.rs
@@ -10,6 +10,7 @@ use graphene::LayerId;
 use graphene::Operation;
 
 use glam::{DAffine2, DVec2};
+use std::f64::consts::PI;
 
 /// Contains the edges that are being dragged along with the original bounds.
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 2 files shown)


dchiasson:
This is a draft of the origin overlay #759, which may be irrelevant due to #772.

The current origin overlay isn't quite centered over the origin and visually conflicts with the transformation handles, but I though I would make sure this change is still necessary before fixing it.

Keavon:
We just now merged #722 and it does, I regret to say, obsolete this feature. A few days ago I expected that they were entirely separate features, but in the past couple days we ended up converging the features for what hopefully provides a better UX instead of having both an origin and a separate pivot concept.

I do hope that you were able to find the learning process beneficial from this so it wasn't entirely a waste of time. I can still code review it for you if you'd like in order to get feedback on what you implemented and learn from that, just let me know if you'd want that. I'll help you find something else that is less likely to be overlapped this time by ongoing work, but which should still be at a good introductory-ish level. Thanks for your patience and sorry if this wasn't well communicated before you put in the work.

dchiasson:
No problem, I agree that the new pivot feature is a great user experience. Don't worry about reviewing these changes. I didn't finish them and it was still beneficial learning. Feel free to suggest another suitable task or I will just browse through the open issues.

PR #777, opened 2022-09-09
Fix crashes related to 0-scale shapes

### LINKED ISSUE #699: Modifying the transform of a zero-area layer causes a crash
There are multiple symptoms:

- Dragging a transform cage breaks when a layer's bounding box has 0 area (width or height is 0), achieved by scaling with, e.g., <kbd>S</kbd> <kbd>Y</kbd> <kbd>0</kbd>. It makes the shape disappear, and then trying to draw other shapes makes the editor crash.
- In the Properties panel, after setting a layer's X or Y scale or dimensions to 0, setting it back to a non-zero value crashes the editor. Even rotating it causes the crash. Also the rotation sometimes shows "NaN"`.

The expected result is that scaling along the axis which was previously squashed to 0 should instead just do nothing, meaning that the user drags along that axis and it simply doesn't end up changing. Typing a new value in the Properties panel should just display 0 again for the Dimensions, and either 0 for the scale or whatever value the user types which ends up doing nothing. Rotating should work, though, and then scaling along an axis would mean that the infinitely-thin sliver of a layer becomes a rotated, sheared sliver.

Bobox214:
Aims to Close #699

### DIFF
--- a/editor/src/messages/tool/common_functionality/transformation_cage.rs
+++ b/editor/src/messages/tool/common_functionality/transformation_cage.rs
@@ -130,7 +130,14 @@ impl SelectedEdges {
 
 	/// Calculates the required scaling to resize the bounding box
 	pub fn bounds_to_scale_transform(&self, position: DVec2, size: DVec2) -> (DAffine2, DVec2) {
-		let enlargement_factor = size / (self.bounds[1] - self.bounds[0]);
+		let mut enlargement_factor = size / (self.bounds[1] - self.bounds[0]);
+		if enlargement_factor.x.is_nan() {
+			enlargement_factor.x = 0.;
+		}
+		if enlargement_factor.y.is_nan() {
+			enlargement_factor.y = 0.;
+		}
+		log::info!("enlargement factor {:?}, -> {:?}", enlargement_factor, DAffine2::from_scale(enlargement_factor));
 		let mut pivot = (self.bounds[0] * enlargement_factor - position) / (enlargement_factor - DVec2::splat(1.));
 		if pivot.x.is_nan() {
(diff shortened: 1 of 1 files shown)


Bobox214:
This PR is now ready to review.

Keavon:
The code looks clean, good job on that. It looks like all the bugs are solved except one: when a layer is scaled to 0 width and 0 height, it displays a rotation of `NaN°` and typing a value in crashes it. The expected behavior would preserve the preexisting rotation, or it would also be fine to display `0°` as a fallback if that's not possible to derive from the scaled-to-zero matrix.

Bobox214:
If the scale X and scale Y are both 0, the rotation matrix of the transform becomes [ 0 0 , 0 0 ] and the rotation information is lost.
So you can't preserve the rotation, and you cannot  set a new rotation value.
I will compute a value of `0 in this case in place of NaN ( even if NaN is really the meaningfull mathematical value here, but probably less user-friendly).

But in any case  it should effectively not crash, I'll fix that.

Keavon:
Ok, so we basically want to always display 0° and let the user type any new value but ignore it and set it back to 0. Thanks!

Bobox214:
> Ok, so we basically want to always display 0° and let the user type any new value but ignore it and set it back to 0. Thanks!

Yes, this is what I have just added in the latest commit.

Keavon (approved):
Alright, behavior seems good from my testing! I've given my approval but I'd ideally like to wait until @0HyperCube can take a look at it as well. If that doesn't work with his schedule in the next day or two, we can go ahead and merge with just my review.

0HyperCube (approved):
Looks good

PR #779, opened 2022-09-13
Bezier-rs: continue converting demos from canvas to SVG

hannahli2010:


website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -136,75 +138,219 @@ impl WasmBezier {
...
 
-	pub fn length(&self) -> f64 {
-		self.0.length(None)
+	pub fn to_svg(&self) -> String {
+		wrap_svg_tag(self.get_bezier_path())
+	}
+
+	pub fn length(&self) -> String {
+		let bezier = self.get_bezier_path();
+		wrap_svg_tag(format!("{bezier}{}", draw_text(format!("Length: {:.2}", self.0.length(None)), 5., 193., BLACK)))
```
Androxium, line 150: Would it be beneficial to assign these constants to a `const` variable to add extra context to what they mean/why they're chosen?

Keavon (approved):

PR #783, opened 2022-09-30
Convert reduce() and offset() Bezier-rs wasm demo functions to render as SVG

Androxium:
This PR converts the `reduce` and `offest` functions in the wasm intermediary to return an SVG for rendering on the interactive docs.

### DIFF
--- a/website/other/bezier-rs-demos/wasm/src/bezier.rs
+++ b/website/other/bezier-rs-demos/wasm/src/bezier.rs
@@ -397,6 +397,7 @@ impl WasmBezier {
 	}
 
 	pub fn reduce(&self) -> String {
+		let empty_string = String::new();
 		let original_curve_svg = self.get_bezier_path();
 		let bezier_curves_svg: String = self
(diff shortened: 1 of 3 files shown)


website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -396,15 +396,48 @@ impl WasmBezier {
...
+			.0
+			.reduce(None)
+			.iter()
+			.enumerate()
+			.map(|(idx, bezier_curve)| {
+				let mut curve_svg = String::new();
+				bezier_curve.to_svg(
+					&mut curve_svg,
+					CURVE_ATTRIBUTES.to_string().replace(BLACK, &format!("hsl({}, 100%, 50%)", (40 * idx))),
+					String::new(),
```
hannahli2010, line 411: Can we create an empty string once, and pass that in three times instead?

Keavon (approved):
LGTM

PR #784, opened 2022-10-07
Integrate Stable Diffusion with the Imaginate layer

### LINKED ISSUE #797: [Crash Report] When loading auto-saved documents, an image in a non-active tab can't find its layer
**Describe the Crash**
When reloading the page, the auto-save system reloads all the saved documents consecutively. If the first tab's document has a layer, it sends a FrontendMessage which will then asynchronously generate a blob URL and eventually call back to the backend with that URL to assign to its layer in that document. But by then, the document loading has moved onto the next document and it has made that one be active. It is then unable to find the Image layer in the first document for some reason.

**Steps To Reproduce**
Describe precisely how the crash occurred, step by step, starting with a new editor window.
1. Open the Graphite Editor at https://editor.graphite.rs
2. Paste an image into the document
3. Make a new document
4. Draw a shape or something
5. Reload the page

**Browser and OS**
Chrome 105, Windows 10 or 11

**Stack Trace**
Copied from the crash dialog in the Graphite Editor:

```
panicked at 'Blob url for invalid layer: LayerNotFound([])', graphene/src/document.rs:757:51

Error
    at https://dev.graphite.rs/js/app.9f6d9d9c.js:9:995
    at Object.n [as handleJsMessage] (https://dev.graphite.rs/js/app.9f6d9d9c.js:1:8158)
    at https://dev.graphite.rs/js/app.9f6d9d9c.js:1:8656
    at https://dev.graphite.rs/js/123.33d206ae.js:1:13830
    at N (https://dev.graphite.rs/js/123.33d206ae.js:1:5550)
    at Cn (https://dev.graphite.rs/js/123.33d206ae.js:1:13796)
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[2323]:0x1c801a
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[912]:0x1741bf
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[699]:0x158434
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[1887]:0x1bb875
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[1764]:0x1b6229
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[1974]:0x1bed08
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[2530]:0x1cbea4
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[2538]:0x1cc07b
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[2057]:0x1c1618
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[53]:0x11a64
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[54]:0x1819c
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[58]:0x2810a
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[131]:0x9d3ac
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[905]:0x173599
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[872]:0x16f5bb
    at https://dev.graphite.rs/cd8ddf81a0c422b7.module.wasm:wasm-function[1979]:0x1bef68
    at Proxy.setImageBlobUrl (https://dev.graphite.rs/js/123.33d206ae.js:1:11040)
    at Proxy.<anonymous> (https://dev.graphite.rs/js/app.9f6d9d9c.js:1:532)
    at https://dev.graphite.rs/js/app.9f6d9d9c.js:1:9069
```

### LINKED ISSUE #798: Add the Imaginate layer to synthesize images as adjustment layers

Keavon:
Closes #798

MVP in this PR:

- [x] Integrate the AI Artist tool shelf icon with other raster icons
- [x]  Tooltips
- [x]  Persist preferences to browser storage
- [x]  Garbage collect images
- [x] Add a refresh button for the network status (right of text)
- [x] Fix crash when changing documents while generating
- [x]  Crash on mousemove when loading page
- [x]  Bug where 0% doesn't show
- [x] Save images in layer with document
- [x] Copy pasting don't preserve the image
- [x] Alt-drag duplication clears the other copy when regenerating
- [x] Generate with new random seed button
- [x] See if there's any messages that can be deleted because they are just pass-throughs
- [x] Only request the next image after each one arrives
- [x] Update connection status based on other network requests in case things change
- [x] Abort the polling request when terminating
- [x] Handle the user terminating immediately after starting generation
- [x] Fix bug where it checks the default server hostname before it reads from IndexedDB preferences (racy)
- [x] Crash when assigning a new blob URL to a nonexistent layer when loading multiple autosaved documents (fixes #797)
- [x] Add a Sampling Method dropdown selection box
- [x] Confirm API is up-to-update
- [x] Icons
- [x] Rename feature to "Imaginate"

To be done later:

- [ ] If the network request fails when pressing Generate, show the user a dialog with info
- [ ] Handle rotated AI Artist layers
- [ ] Move percentage and generating status to LayerInfo not Layer
- [ ] Range slider
- [ ] Download button for image from AI Artist layer
- [ ] Convert to image layer button
- [ ] Generation history settings explorer
- [ ] Flickering workaround
- [ ] Boolean images?


otdavies:
Looking very good - works great.

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -568,6 +562,13 @@ impl Fsm for SelectToolFsmState {
 						);
 					}
 					tool_data.drag_current = mouse_position + closest_move;
+
+					if input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_none() {
+						tool_data.start_duplicates(document, responses);
+					} else if !input.keyboard.get(duplicate as usize) && tool_data.not_duplicated_layers.is_some() {
+						tool_data.stop_duplicates(responses);
+					}
+
```
Keavon, line 571: @0HyperCube do you foresee any problem with moving this down like I did here? It was needed to fix a message ordering issue with the overlays. I haven't noticed any problems but haven't QA tested it extensively. Thanks.

mfish33 (approved):

0HyperCube:
> Flickering workaround

I believe the flickering is due to the web browser rendering the rest of the svg before it has decoded the embedded PNG.

Keavon:
@0HyperCube I came to the same conclusion. I have a few ideas that may or may not work:

- Use an HTML `` tag inside a `<foreignObject>` which provides more options that could possibly help us:
    - Use the [`decoding` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-decoding) to force it to block the SVG rendering, if that's how that works (and functions as expected in a foreignObject)
    - Use the [`srcset` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-srcset) to provide a fallback blob URL of the previous image (with a delay to the disposal of the blob URL, perhaps just a simple setTimeout would be fine in the disposal code)
- Render the document SVG with two `<image>` tags instead of the current one, with the fallback behind the new image
- Before swapping in the new SVG for the document, preload the images outside the DOM and only do the SVG replacement once they are all ready
    - Perhaps this can be easily achieved by preloading it when the JS generates the blob URL, before sending the freshly created URL back to the backend (this would be the simplest option if it works)

Keavon:
@0HyperCube Alright, it turns out this was nice and easy since JS provides a [handy promise-based API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decode) specifically for this purpose. See my implementation: https://github.com/GraphiteEditor/Graphite/commit/3fe552ff20a9ed28477802d2f82c9dbfe6acdce8

PR #786, opened 2022-10-08
Break apart and improve the JS for rasterizing SVGs and downloading files

Keavon:
It now uses a blob URL instead of a data URL in the download process, which is cleaner and better performance.


TrueDoctor:
I will trust you that this just works :tm:

TrueDoctor (approved):

PR #791, opened 2022-10-11
Bezier-rs: Convert bezier-rs intersection functions to SVG

jackiechen73:
This PR converts all the `intersection` functions in the wasm representation to return an SVG tag instead of JS points.


Androxium (approved):

website/other/bezier-rs-demos/src/App.vue
```diff
@@ -288,6 +288,93 @@ export default defineComponent({
...
 					},
 				},
+				{
+					name: "Intersect (Line Segment)",
+					callback: (bezier: WasmBezierInstance): string => {
+						const line = [
+							{ x: 150, y: 150 },
+							{ x: 20, y: 20 },
+						];
+						const mappedLine = line.map((p) => [p.x, p.y]);
```
RobNadal, line 298: I don't really see the point of this - why can't you just define the line as 
```
const line = [
    [150, 150],
    [20, 20],
];
```
to begin with (here and below)?

website/other/bezier-rs-demos/src/App.vue
```diff
@@ -288,6 +288,93 @@ export default defineComponent({
...
+					},
+					exampleOptions: {
+						[BezierCurveType.Quadratic]: {
+							sliderOptions: [
+								{
+									variable: "error",
+									min: 0.1,
+									max: 2,
+									step: 0.1,
+									default: 0.5,
```
RobNadal, line 321: Could you make this slider option a default like we did for `tSliderOption` to reduce duplication? Maybe something like `errorSliderOption`?

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -372,28 +372,102 @@ impl WasmBezier {
...
+		);
+
+		let intersections_svg = self
+			.intersect(&line, None)
+			.iter()
+			.map(|intersection_t| {
+				let point = &self.0.evaluate(*intersection_t);
+				draw_circle(point.x, point.y, 4., RED, 1.5, WHITE)
+			})
+			.fold(String::new(), |acc, item| format!("{acc}{item}"));
```
RobNadal, line 398: Can we just use the `.concat()` function here? (here and below)

Keavon:
If the code review comments above have been resolved, please mark them as such (otherwise resolve them, or point out if I should give my opinion on any of them if things are uncertain). Once that's done, I can come back and give this an actual code review. Thanks! And sorry for the long delay, October is a crazy month.

Keavon (approved):

PR #792, opened 2022-10-12
Bezier-rs: Add a pivot point to the 'rotate' function

hannahli2010:
Add a function that rotates the bezier around a given point
 - Changed the `rotate` demo to rotate around the centre of the image

image

### DIFF
--- a/libraries/bezier-rs/src/bezier/transform.rs
+++ b/libraries/bezier-rs/src/bezier/transform.rs
@@ -90,6 +90,12 @@ impl Bezier {
 		self.apply_transformation(&|point| rotation_matrix.mul_vec2(point))
 	}
 
+	/// Returns a Bezier curve that results from rotating the curve around the provided point by the given angle (in radians).
+	pub fn rotate_about_point(&self, angle: f64, pivot: DVec2) -> Bezier {
+		let rotation_matrix = DMat2::from_angle(angle);
+		self.apply_transformation(&|point| rotation_matrix.mul_vec2(point - pivot) + pivot)
+	}
+
 	/// Returns a Bezier curve that results from translating the curve by the given `DVec2`.
 	pub fn translate(&self, translation: DVec2) -> Bezier {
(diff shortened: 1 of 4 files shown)


libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -90,6 +90,12 @@ impl Bezier {
 		self.apply_transformation(&|point| rotation_matrix.mul_vec2(point))
 	}
 
+	/// Returns a Bezier curve that results from rotating the curve around the provided point by the given angle (in radians).
+	pub fn rotate_about_point(&self, angle: f64, origin: DVec2) -> Bezier {
```
RobNadal, line 94: I always felt that the origin was the point (0, 0). How do we feel about renaming this argument `pivot`?

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -365,7 +365,7 @@ impl WasmBezier {
 	}
 
 	pub fn rotate(&self, angle: f64) -> WasmBezier {
-		WasmBezier(self.0.rotate(angle))
+		WasmBezier(self.0.rotate_about_point(angle, DVec2::new(100., 100.)))
```
RobNadal, line 368: It doesn't make a significant difference because these functions aren't user-facing, but I don't like the second argument being hard-coded here. I think we should be defining the point around which to rotate in `App.vue`

RobNadal:
We should also be adding to the `rotate` visualization the point around which we are rotating, but if you do it here #793 will just have to rewrite it. Whichever one of you merges second should add this.

Keavon:
If this is ready for me to review, please request me as a reviewer and I can take a look. (I'm currently just going through the open PRs and figuring out which of them are in a state that's waiting on my review, and commenting on the ones that possibly might not be yet.)

Keavon (approved):
LGTM

PR #793, opened 2022-10-13
Bezier-rs: Convert 'rotate' and 'de casteljau' wasm functions from canvas to SVG rendering

linda-zheng:
Refactor the `rotate` and `de_casteljau_points` wasm functions to output an svg (in the form of a string) for the interactive bezier-rs demo.


website/other/bezier-rs-demos/src/App.vue
```diff
@@ -288,52 +288,35 @@ export default defineComponent({
...
-								}
-
-								if (index !== 0) {
-									const prevPoint: Point = iteration[index - 1];
-									drawLine(getContextFromCanvas(canvas), point, prevPoint, colorLight);
-								}
-							});
-						});
+					name: "Rotate",
+					callback: (bezier: WasmBezierInstance, options: Record<string, number>): string => bezier.rotate(options.angle),
```
hannahli2010, line 293: I think you need to multiply by `Math.PI` here

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -353,19 +348,53 @@ impl WasmBezier {
...
+							draw_circle(point.x, point.y, 3., &color_light, 1.5, WHITE),
+							if index != 0 {
+								let prev_point = points[index - 1];
+								draw_line(prev_point.x, prev_point.y, point.x, point.y, &color_light, 1.5)
+							} else {
+								String::new()
+							}
+						)
+					})
+					.fold("".to_string(), |acc, point_svg| acc + &point_svg);
```
RobNadal, line 376: Use `String::new()` for consistency and unabbreviate `acc` here and for the line below. Alternatively, could we just use `.concat()`?

RobNadal:
We should also be adding to the `rotate` svg the point around which we are rotating, but you'd need #792 to do it. Whichever one of you merges first should add this.

Keavon:
If this is ready for me to review, please request me as a reviewer and I can take a look. (I'm currently just going through the open PRs and figuring out which of them are in a state that's waiting on my review, and commenting on the ones that possibly might not be yet.)

website/other/bezier-rs-demos/wasm/src/bezier.rs
linda-zheng: I think it makes sense to use this because it is more consistent with the existing code (i.e. local extrema, inflections, compute lookup table)

PR #794, opened 2022-10-13
Disallow zero-area artboards

### LINKED ISSUE #704: Zero-area artboards are confusing and impossible to deal with
If the user creates a zero-area artboard, such as by simply clicking with the Artboard tool in some empty space, there's no way to notice this exists and it causes confusion because deleting all apparent artboards fails to restore the white infinite canvas. There's no way to select them.

- We should prevent artboards being less than a pixel in either dimension.
- We should draw some kind of border around the outside of all artboards while using the tool which act as a click target.

And we should eventually have some way to view a list of artboards, give them names, and delete them from the list. (But this would probably be a separate issue that may be best implemented once the node system exists.)

Kit-p:
This PR ensures that adding and resizing artboards will not result in zero width or zero height.

Regarding adding borders to all artboards when the artboard tool is selected, I believe a separate PR is more suitable. When this PR is merged, a new issue should be created.

This is my first time writing Rust, any feedback is welcome!

Closes #704.


Keavon:
We should ideally not create an artboard if you click and release without moving the mouse by a pixel, rather than causing that click-without-drag to make a 1x1 artboard. From my quick look at the code, I assume it does the latter (but I haven't had the chance to test it yet). If that's the case, is that something you'd be able to explore? Nice job on your first PR and welcome to the wonderful world of Rust programming! :)

Kit-p:
@Keavon Interesting, didn't think about that.
I believe that the original logic is to create a 0x0 artboard on click and resize when drag. If we don't want any artboard unless user does drag, maybe I should only add an artboard when user starts dragging?

Keavon:
That's probably a good way to do it. But just beware of the case where the user drags the cursor back to the exact point where the drag began. In that case, you can take your current approach of enforcing a minimum size of 1x1. Or you could reject the artboard in that case. From a UX perspective, both options have pros and cons and I think either is equally fine.

Kit-p:
@Keavon Changed as requested, please review again, thank you!

editor/src/messages/tool/tool_messages/artboard_tool.rs
```diff
@@ -226,20 +226,7 @@ impl Fsm for ArtboardToolFsmState {
...
-
-							responses.push_back(
-								ArtboardMessage::AddArtboard {
-									id: Some(id),
-									position: (0., 0.),
-									size: (0., 0.),
-								}
-								.into(),
-							);
+							tool_data.selected_board = None;
```
Keavon, line 229: For clarity, please use the full name "artboard" to hedge against anyone in the future mistaking the meaning for something else.

0HyperCube (commented):
I can still create a 0 x 0 area art board if I drag but then move back to the original position (which is easy if you zoom in a lot as sizes are rounded to integer coordinates.

editor/src/messages/tool/tool_messages/artboard_tool.rs
0HyperCube: This was introduced before your MR @Kit-p however it would still be great if you could quickly rename this for clarity.

Keavon: Oh I overlooked that fact, thanks @0HyperCube for pointing out that @Kit-p didn't choose that name to begin with. Yes, please do rename it for clarity while you're at it, thank you!

Kit-p: No problem, renamed.

Kit-p:
> I can still create a 0 x 0 area art board if I drag but then move back to the original position (which is easy if you zoom in a lot as sizes are rounded to integer coordinates.

@0HyperCube I tried reproducing but I can't, would you mind recording a short video showcasing it?

Kit-p:
@0HyperCube For your reference, this is what happened on my side.

https://user-images.githubusercontent.com/65013409/195959628-06ed34e4-e14e-4561-8dd1-7f903c89ad78.mp4

0HyperCube:
Sorry for the late response @Kit-p. I think perhaps I mistakenly didn't apply the full diff last time. The artboard tool now works as expected, however when resizing a rectangle we get some odd behaviour (attached).
Recording 2022-10-21 at 15 06 00

Kit-p:
@0HyperCube
Thank you for the review, that is indeed a bug introduced by my changes. Will look into it asap!

Kit-p:
@0HyperCube The problem has been fixed, I would appreciate another review, thank you!

0HyperCube (approved):
Looks good to me 👍🏽

PR #801, opened 2022-10-20
Remake Eyedropper tool to sample pixel colors from viewport canvas

### LINKED ISSUE #545: Pixel-based color sampling for the Eyedropper tool
Since there's no way to obtain the color of a pixel from a rendered web page, the Eyedropper tool currently just grabs the fill color of the shape that's clicked on. It doesn't know if you are clicking on the stroke of a shape instead. Or if you are clicking on somewhere along a gradient where you'd expect to pick the interpolated color. (For gradients, we're just using the color on one end of the gradient.) Picking the exact color of the pixel under the mouse cursor would be a more useful behavior. Given the above-mentioned issue, there are two approaches with pros and cons:

- Use the [EyeDropper](https://developer.mozilla.org/en-US/docs/Web/API/EyeDropper) API which is supported on Chromium-based browsers.
  Pro: works even outside the viewport or web browser entirely
  Con: limited browser support (no Firefox and Safari)
- Rasterize the document and render it to an off-screen canvas then sample the corresponding pixel colors from that, using the same code for exporting artwork as PNG/JPG (see #629). It'd have to re-render every time the canvas changes, in case the user pans/scrolls/zooms while picking colors. Be careful of possible DPI scaling-related gotchas. This can be done by rasterizing the canvas with 1x1 dimensions and just taking the color of that one pixel.
  Pro: works in all browsers
  Con: possibly worse performance

The second option is probably a better long-term solution.

Keavon:
Closes #545

### DIFF
--- a/editor/src/messages/frontend/frontend_message.rs
+++ b/editor/src/messages/frontend/frontend_message.rs
@@ -157,6 +157,18 @@ pub enum FrontendMessage {
 		size: (f64, f64),
 		multiplier: (f64, f64),
 	},
+	UpdateEyedropperSamplingState {
+		#[serde(rename = "mousePosition")]
+		mouse_position: Option<(f64, f64)>,
+		#[serde(rename = "primaryColor")]
+		primary_color: String,
+		#[serde(rename = "secondaryColor")]
+		secondary_color: String,
+		#[serde(rename = "samplingPrimaryOrSecondary")]
+		sampling_primary_or_secondary: String,
+		#[serde(rename = "setColorChoice")]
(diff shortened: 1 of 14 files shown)


otdavies (approved):
Skimmed these changes, they look good to me, nothing jumped out at me as overly concerning. My only sticking point was your comment regarding unusual behavior with strange DPI, but that can likely be resolved in a later PR. Looks good, operates great.

Keavon:
Add an option to hit the Escape key to cancel the action of setting a color.

PR #807, opened 2022-10-22
Fix image export bounds calculation for selected layers

0HyperCube:
Fixes the export bounds for selection or all artwork being incorrect when the viewport was transformed.


Keavon (approved):
Nice and straightforward, glad it's elegant and you didn't run into the double-mutable-borrow thing I was having trouble with in a related context a few days ago. Thanks!

Keavon:
Oh, hmm, maybe this wasn't actually the entire problem. I'm just seeing a black canvas when I export.

0HyperCube:
@Keavon Looks like I had the perfect test case at the origin so I didn't notice this :/.

I forgot to scale the offset transform correctly so you were just seeing an area very far away from your art boards.

Keavon:
All working now! Thank you for the fixes 😀

Issue #603, opened 2022-04-21
Improve the color picker floating menu


Keavon:
It should have inputs for the hex code, numerical RGBA and HSV values, a button to remove the color (set it to `None`), and a color picker button (see #545). The hex and remove-the-color features should also replace the button and text input fields in the ColorInput widget, making it one wide color area.

![image](https://user-images.githubusercontent.com/4388688/197609186-7be29eae-472d-486c-b1e7-b38708e95579.png)

PR #809, opened 2022-10-23
Refactor color picker floating menu

Keavon:
Part of #603


frontend/src/components/widgets/inputs/ColorInput.vue
```diff
@@ -98,16 +95,17 @@ export default defineComponent({
...
-			const a = parseInt(this.value.slice(6, 8), 16);
-			return { r, g, b, a: a / 255 };
+		color(): Color {
+			if (!this.value) return new Color(0, 0, 0, 1);
+
+			const r = parseInt(this.value.slice(0, 2), 16) / 255;
+			const g = parseInt(this.value.slice(2, 4), 16) / 255;
+			const b = parseInt(this.value.slice(4, 6), 16) / 255;
+			const a = parseInt(this.value.slice(6, 8), 16) / 255;
+
```
mfish33, line 105: don't know if this is cleaner but you could do

```ts
const [r, g, b, a] = this.value.match(/.{1,2}/g).map(byte => parseInt(byte, 16))
```
the match(/.{1,2}/g) chunks the string into pairs https://stackoverflow.com/questions/7033639/split-large-string-in-n-size-chunks-in-javascript

You also have to make the if statement above check for non conforming lengths

mfish33 (approved):
Just one suggestion but not blocking at all

frontend/src/components/widgets/inputs/ColorInput.vue
Keavon: I think the clarity of the current solution is worth the three lines, but I appreciate the cleverness and that's a new thing I just learned about regex, thanks. I'll add the length check guard to the if statement as you suggested.

PR #828, opened 2022-10-26
Bezier-rs: Add Euclidean parameterization to Bezier::evaluate

hannahli2010:
- Add enum to determine whether evaluate will use parametric or euclidean interpretation of the passed parameter
- Implement euclidean logic for evaluate
- Add toggle for switching between the two in the UI
- Update styling of the UI


Androxium:
I understand that for subpaths, it would make sense that we'd want to differentiate between parametric t vs eucliean t, but I'm not sure what the use case for having a euclidean t for bezier curves would be. Not sure what the best solution would be, but I thought that the `bezier` library would use only parametric t and the subpath library would do extra calculations to accomodate parametric t and euclidean t.

RobNadal:
> I understand that for subpaths, it would make sense that we'd want to differentiate between parametric t vs eucliean t, but I'm not sure what the use case for having a euclidean t for bezier curves would be. 

@Androxium I could be mistaken, but I remember discussing with @Keavon who mentioned that he felt that this would be useful. Regardless, I don't think its a bad idea for the library to exhibit consistency between the interfaces.

Androxium (approved):

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -21,11 +23,43 @@ impl Bezier {
 		}
 	}
 
+	pub(crate) fn unrestricted_euclidean_evaluate(&self, d: f64, error: f64) -> DVec2 {
```
linda-zheng, line 26: Add a description for the function. 

It's not clear to me whether d is the absolute distance, or the relative distance (i.e. relative to the total length). Maybe clarify this in your description.

linda-zheng (approved):

Keavon:
I might need to wait until Tuesday or Wednesday to finish reviewing this.

Keavon (approved):

PR #853, opened 2022-11-14
Bezier-rs: Remove unused legacy drawing components and structs

RobNadal:
- Removed components related to drawing on canvas. The usage of these components has been entirely replaced using SVG drawing on the WASM side.
- Removed Point abstraction from Wasm and Vue layers. This abstraction was originally created to help with drawing on the JS side, but we predominantly points as arrays of numbers elsewhere in the front-end code


Androxium (approved):

website/other/bezier-rs-demos/src/App.vue
```diff
@@ -233,8 +220,8 @@ export default defineComponent({
 				},
 				{
 					name: "Project",
-					callback: (bezier: WasmBezierInstance, _: Record<string, number>, mouseLocation: Point): string =>
-						mouseLocation ? bezier.project(mouseLocation.x, mouseLocation.y) : bezier.to_svg(),
+					callback: (bezier: WasmBezierInstance, _: Record<string, number>, mouseLocation: number[]): string =>
```
Keavon, line 223: Replace `mouseLocation: number[]` with `[x, y]: [number, number]`

website/other/bezier-rs-demos/wasm/src/bezier.rs
```diff
@@ -15,11 +15,6 @@ struct CircleSector {
 }
 
 #[derive(Serialize, Deserialize)]
```
Keavon, line 17: Forgot to delete this line, unless you intend for it to be added to the enum below.

Keavon (approved):
Good to go after the two comment fixes.

website/other/bezier-rs-demos/wasm/src/bezier.rs
RobNadal: Great catch, good eye!

website/other/bezier-rs-demos/src/App.vue
RobNadal: This destructing notation seems to cause issues when we send an undefined mouseLocation to the project function, ie. on the initial render. Do you know of a good workaround @Keavon, or can we just leave it as is?

Keavon: We shouldn't be sending `undefined` if the declared type is `number[]` and if it's being read as `mouseLocation[0]` and `mouseLocation[1]` which would throw an exception for reading properties of `undefined`.

Keavon (approved):

PR #854, opened 2022-11-14
Bezier-rs: Convert to_svg used in subpaths to match style used in beziers

RobNadal:
- Converted the `Subpath::to_svg` to use the same arguments and style as `Bezier::to_svg`.
- Removed some unnecessary string creations in `wasm/src/Bezier.rs`


Androxium (approved):

Keavon (approved):
LGTM, glad to be removing lines of code!

PR #869, opened 2022-11-26
Add some additional image effect nodes

### LINKED ISSUE #811: Switch to an attribute-based procedural vector format

0HyperCube:
Add some basic vector and raster nodes.

Also some node utilities such as deleting nodes as well as a better node list selector on right click.

Colour Adjustment Nodes:
- [x] Brightness and Contrast
- [x] Hue and Saturation
- [x] Grey scale (although perhaps use a [better / user definable brightness algorithm][brightness])
- [x] Invert
- [x] Posterize - you can really do this however you want although here is a [simple algorithm for it][Posterize]
- [x] Exposure - [overview of in photoshop][exposure]
- [ ] Vibrancy which is a [weird photoshop thing apparently][vibrancy]

Blocked histogram rendering ([how do we determine brightness][brightness]):
- Levels which allows setting of new range and gamma value (where gamma is just when you take a channel between 0 and 1 and raise it to the power of something)
- Curves which allows the user to create a new curve to remap channels or brightness - we also need some curve editor)

[brightness]: https://stackoverflow.com/questions/596216/formula-to-determine-perceived-brightness-of-rgb-color#answer-56678483
[vibrancy]: https://stackoverflow.com/questions/33966121/what-is-the-algorithm-for-vibrance-filters
[exposure]: https://stackoverflow.com/questions/12166117/what-is-the-math-behind-exposure-adjustment-on-photoshop
[Posterize]: http://www.axiomx.com/posterize.htm


0HyperCube:
Colour Adjustment Nodes:
- [x] Brightness and Contrast
- [x] Hue and Saturation
- [x] Grey scale (although perhaps use a [better / user definable brightness algorithm][brightness])
- [x] Invert
- [ ] Posterize - you can really do this however you want although here is a [simple algorithm for it][Posterize]
- [ ] Exposure - [overview of in photoshop][exposure]
- [ ] Vibrancy which is a [weird photoshop thing apparently][vibrancy]

Blocked histogram rendering ([how do we determine brightness][brightness]):
- Levels which allows setting of new range and gamma value (where gamma is just when you take a channel between 0 and 1 and raise it to the power of something)
- Curves which allows the user to create a new curve to remap channels or brightness - we also need some curve editor)

[brightness]: https://stackoverflow.com/questions/596216/formula-to-determine-perceived-brightness-of-rgb-color#answer-56678483
[vibrancy]: https://stackoverflow.com/questions/33966121/what-is-the-algorithm-for-vibrance-filters
[exposure]: https://stackoverflow.com/questions/12166117/what-is-the-math-behind-exposure-adjustment-on-photoshop
[Posterize]: https://github.com/gautamsreekumar/posterize#algorithm

frontend/src/App.vue
```diff
@@ -43,10 +43,10 @@
 	--color-data-general-rgb: 197, 197, 197;
 	--color-data-general-dim: #767676;
 	--color-data-general-dim-rgb: 118, 118, 118;
-	--color-data-vector: #65bbe5;
-	--color-data-vector-rgb: 101, 187, 229;
-	--color-data-vector-dim: #4b778c;
-	--color-data-vector-dim-rgb: 75, 119, 140;
```
Keavon, line 49: Is there another way that can avoid this change? It breaks the naming convention for the high-level data types people will know by name (raster, vector, bitmap, text, etc.) and also breaks the coloring used in the tool shelf (referenced in the `.svg` files).

0HyperCube: Raster and bitmap are the same thing aren't they? You can name the things whatever you like.

0HyperCube: Vector is also a confusing name as it represents both a mathematical vector (magnitude and direction) as well as a method of defining the geometry of an image.

0HyperCube: I'll rename this back to vector so we can merge this PR.

Keavon (approved):

PR #876, opened 2022-12-07
Bezier-rs: Insert function for subpath

hannahli2010:

### DIFF
--- a/libraries/bezier-rs/src/subpath/manipulators.rs
+++ b/libraries/bezier-rs/src/subpath/manipulators.rs
@@ -4,7 +4,7 @@ use crate::ComputeType;
 impl Subpath {
 	/// Calculate the point on the subpath based on the parametric `t`-value provided.
 	/// Expects `t` to be within the inclusive range `[0, 1]`.
-	pub fn add_manipulator_group(&self, t: ComputeType) {
+	pub fn add_manipulator_group(&mut self, t: ComputeType) {
 		match t {
 			ComputeType::Parametric(t) => {
 				assert!((0.0..=1.).contains(&t));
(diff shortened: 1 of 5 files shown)


RobNadal:
If we don't support euclidean parameterization yet for this function we should disable it in the App.vue

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -0,0 +1,114 @@
...
+			ComputeType::Parametric(t) => {
+				assert!((0.0..=1.).contains(&t));
+
+				let number_of_curves = self.len_segments() as f64;
+				let scaled_t = t * number_of_curves;
+
+				let target_curve_index = scaled_t.floor() as i32;
+				let target_curve_t = scaled_t % 1.;
+
+				if target_curve_t == 0. || target_curve_t == 1. {
```
RobNadal, line 18: Do you think we should check these conditions to some epsilon rather than strict equality?

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use glam::DVec2;
+
+	#[test]
+	fn add_manipulator_group() {
```
RobNadal, line 50: Could you break down these tests further and give them more descriptive names based on the conditions they are verifying?

hannahli2010: I added comments in between each section as descriptions. Is that alright or do you want them to be individual functions?

RobNadal: I usually leave a single test case per test because its more informative. That way if an early test fails, you still know the results of the later tests. As a compromise could we at least separate out the open / closed test cases? Also there's some grammatical issues in a few of the comments.

RobNadal:
Good to open for Graphite developer review once the test nit is resolved.

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -0,0 +1,121 @@
...
+
+		subpath.closed = true;
+
+		// Add manipulator group to the final segment which closes the subpath
+		let location4 = subpath.evaluate(ComputeType::Parametric(0.9));
+		let split_pair4 = subpath.iter().nth(4).unwrap().split((0.9 * 5.) % 1.);
+		subpath.add_manipulator_group(ComputeType::Parametric(0.9));
+		assert_eq!(subpath.manipulator_groups[5].anchor, location4);
+		assert_eq!(split_pair4[0], subpath.iter().nth(4).unwrap());
+		assert_eq!(split_pair4[1], subpath.iter().nth(5).unwrap());
```
RobNadal, line 111: Might be a good idea to assert that the subpath is still closed.

RobNadal (approved):

Keavon:
Please call this "split" for consistency with the segment API. I didn't know what "add manipulator group" meant until I tried out the demo.

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -0,0 +1,147 @@
+use super::*;
+use crate::{consts::MAX_ABSOLUTE_DIFFERENCE, utils::f64_compare, ComputeType};
```
Keavon, line 2: Split this into three import lines.

website/other/bezier-rs-demos/src/App.vue
```diff
@@ -601,6 +601,13 @@ export default defineComponent({
 					name: "Constructor",
 					callback: (subpath: WasmSubpathInstance): string => subpath.to_svg(),
 				},
+				{
+					name: "Add manipulator group",
+					callback: (subpath: WasmSubpathInstance, options: Record<string, number>, _: undefined, computeType: ComputeType): string =>
+						subpath.add_manipulator_group(options.computeArgument, computeType),
+					sliderOptions: [{ ...tSliderOptions, variable: "computeArgument" }],
+					// chooseComputeType: true, // TODO: Uncomment this after implementing the Euclidean version
```
Keavon, line 609: Please put this TODO comment on the preceding line.

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -0,0 +1,147 @@
...
+
+impl Subpath {
+	/// Calculate the point on the subpath based on the parametric `t`-value provided.
```
Keavon, line 5: This sounds inaccurate (it doesn't just calculate a point). It inserts an anchor, and handles, at a certain point along the subpath where each constituent segment accounts for an equal proportion of the 0..1 interval.

RobNadal:
> Please call this "split" for consistency with the segment API. I didn't know what "add manipulator group" meant until I tried out the demo.

Although this functionality seems similar to `split`, we wanted to reserve the keyword `split` for different Subpath functionality, that is, splitting an open Subpath into two distinct Subpaths or opening a closed Subpath at a particular point. We are not particularly attached to the name `add_manipulator_group` though if you have any suggestions.

Keavon:
Rename to `insert` please.

Keavon (approved):

PR #877, opened 2022-12-10
Fix inability to expose the Add node's parameter

0HyperCube:
The number input for the add properties didn't show up as it was expecting an f32 and got an f64


Keavon (approved):

Keavon:
Glad to see this 67 LOC reduction!

PR #887, opened 2022-12-17
Quantization node

TrueDoctor:
This PR introduces a node to perform image compression with high retention of visual details


Keavon:
I don't see the node in the Add Nodes menu, or is that not the intention here?

node-graph/gcore/src/raster.rs
```diff
@@ -53,6 +53,32 @@ impl<N: Node<(), Output = f32> + Copy> BrightenColorNode<N> {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+pub struct GammaColorNode<N: Node<(), Output = f32>>(N);
+
+impl<N: Node<(), Output = f32>> Node<Color> for GammaColorNode<N> {
+	type Output = Color;
+	fn eval(self, color: Color) -> Color {
+		let gamma = self.0.eval(());
+		let per_channel = |col: f32| col.powf(gamma);
```
0HyperCube, line 63: The current GammaNode for an image actually does the inverse of this (gamma correction?) - which approach doe we want @Keavon?

node-graph/gcore/src/raster.rs
```diff
@@ -53,6 +53,32 @@ impl<N: Node<(), Output = f32> + Copy> BrightenColorNode<N> {
...
+		Color::from_rgbaf32_unchecked(per_channel(color.r()), per_channel(color.g()), per_channel(color.b()), color.a())
+	}
+}
+impl<N: Node<(), Output = f32> + Copy> Node<Color> for &GammaColorNode<N> {
+	type Output = Color;
+	fn eval(self, color: Color) -> Color {
+		let gamma = self.0.eval(());
+		let per_channel = |col: f32| col.powf(gamma);
+		Color::from_rgbaf32_unchecked(per_channel(color.r()), per_channel(color.g()), per_channel(color.b()), color.a())
```
0HyperCube, line 72: Could we use the `node_fn` macro to avoid this code duplication?

node-graph/gstd/src/quantization.rs
```diff
@@ -0,0 +1,93 @@
+use graphene_core::raster::{Color, Image};
+use graphene_core::Node;
+
+pub struct GenerateQuantizationNode<N: Node<(), Output = u32>, M: Node<(), Output = u32>>(N, M);
```
0HyperCube, line 4: It would be great if you could write a comment explaining what quantization is and what it achieves.

node-graph/gstd/src/quantization.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl<N: Node<(), Output = u32>, M: Node<(), Output = u32>> GenerateQuantizationNode<N, M> {
+	pub fn new(node: N, index: M) -> Self {
+		Self(node, index)
+	}
+}
+
+fn generate_quantization_fn(samples: u32, function: u32, input: Image) -> Image {
+	let max_energy = 16380.;
```
0HyperCube, line 33: Magic constant? Perhaps add a comment explaining where this number comes from.

node-graph/gstd/src/quantization.rs
```diff
@@ -0,0 +1,93 @@
...
+	let data: Vec<f64> = input
+		.data
+		.iter()
+		.map(|x| vec![x.r() as f64, x.g() as f64, x.b() as f64])
+		.reduce(|mut acc, x| {
+			acc.extend_from_slice(&x);
+			acc
+		})
```
0HyperCube, line 41: Wouldn't a flat_map and collect make more sense.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -199,6 +213,12 @@ pub fn posterize_properties(document_node: &DocumentNode, node_id: NodeId, _cont
 	vec![LayoutGroup::Row { widgets: value }]
 }
 
+pub fn quantize_properties(document_node: &DocumentNode, node_id: NodeId, _context: &mut NodePropertiesContext) -> Vec<LayoutGroup> {
+	let value = number_widget(document_node, node_id, 1, "Levels", NumberInput::new().min(1.).max(1000.).int(), true);
+	let index = number_widget(document_node, node_id, 1, "Fit Fn Index", NumberInput::new().min(0.).max(2.).int(), true);
```
0HyperCube, line 218: Would a dropdown be more suited for this?

0HyperCube:
@Keavon The node is `#[cfg(feature = "quantization")]` which is not enabled by default.

node-graph/interpreted-executor/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "MIT OR Apache-2.0"
 default = []
 serde = ["dep:serde", "graphene-std/serde", "glam/serde"]
 gpu = ["graphene-std/gpu"]
+quantization = ["graphene-std/quantization"]
```
0HyperCube, line 11: How come this is not enabled by default?

0HyperCube (approved):
Code looks good, although I haven't yet tried it out.

node-graph/gcore/src/raster.rs
Keavon: We flow linear colors through our node graph, it's up to any particular node to gamma or inverse gamma the color if necessary (like when importing or exporting an image). Besides that, I don't have enough context or understanding to comment about this part of the code. Happy to discuss it though if it's important.

0HyperCube: The gamma node that already exists in the node graph actually does gamma correction. Do we want to change this?

Keavon: It's a useful node to have, so there's no need to change it at the moment.

TrueDoctor: We should at some point transition the node that works on images to use a per channel node internally.
We could also have the node system use the correct implementation based on the input type (Image or Color)

node-graph/gcore/src/raster.rs
TrueDoctor: Yes, will fix

node-graph/gstd/src/quantization.rs
TrueDoctor: good catch

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
TrueDoctor: yes but this is mostly temporary

node-graph/interpreted-executor/Cargo.toml
TrueDoctor: It pulls in extra dependencies which I didn't want to slow down compile time by default.
+ this is mostly experimental still I just wanted to include this in the main repo to reduce the number of diverging branches

Issue #881, opened 2022-12-12
Node graph editing UX improvements


Keavon:
- [x] Dragging existing flow connections out of their input ports to be moved elsewhere or dropped
- [x] Drag around multiple selected nodes at once
- [x] Copy/pasting and duplicating (Ctrl+D) nodes with their parameters kept
- [x] Drag a node onto a link to automatically wire it up through its primary input/output
- [x] Add a hide button and toggle hotkey to mute/disable a node to bypass its primary input/output data flow
- [x] Add a preview button to each node to temporarily supplant the connection to the Output node
- [ ] https://github.com/GraphiteEditor/Graphite/pull/905#issuecomment-1364268788

Keavon:
Reopening till the UX improvements suggested in https://github.com/GraphiteEditor/Graphite/pull/905#issuecomment-1364268788 are merged.

PR #894, opened 2022-12-21
Drag out links to disconnect nodes

0HyperCube:
Drag an input to disconnect. Part of #881.


Keavon (approved):
Fantastic, this works beautifully. Code looks idiomatic as well.

PR #897, opened 2022-12-21
Bezier-rs: Added function to find intersections between a bezier and a rectangle

RobNadal:
- Added function to find intersections between a Bezier and a rectangle

image

- Added function which returns true if the bezier is contained entirely within a rectangel


Keavon (approved):
Looks great, thank you!

PR #898, opened 2022-12-22
Better downcast panics

0HyperCube:
Downcast panics now contain the actual type you failed to downcast as well as your target - meaning that debugging panics is easier. Embedding all the type names can be disabled by disabling the "log-bad-types" feature.


TrueDoctor (approved):
Looks good

PR #901, opened 2022-12-22
Add drag to connect nodes

0HyperCube:
Part of #881


frontend/src/components/panels/NodeGraph.vue
```diff
@@ -632,6 +647,38 @@ export default defineComponent({
 					}
 				}
 				this.editor.instance.moveSelectedNodes(this.draggingNodes.roundX, this.draggingNodes.roundY);
+
+				if (this.selected.length === 1) {
+					const id = this.selected[0];
+					const selectedNode = containerBounds.querySelector(` [data-node="${String(id)}"]`);
```
Keavon, line 653: Nit: extraneous space at the start of the template string.

While you're here, would you mind giving the code a little breathing room by spacing out the related blocks of lines?

Also while you're working in this event handler, would you mind adding something for closing the Add Nodes menu if you click out of it while it's open? Thanks :)

frontend/wasm/src/editor_api.rs
```diff
@@ -594,6 +594,18 @@ impl JsEditorHandle {
...
+	/// Check for intersections between the curve and a rectangle defined by opposite corners
+	#[wasm_bindgen(js_name = rectangleIntersects)]
+	pub fn rectangle_intersects(&self, bezier_x: Vec<f64>, bezier_y: Vec<f64>, top: f64, left: f64, bottom: f64, right: f64) -> bool {
+		let bezier = bezier_rs::Bezier::from_cubic_dvec2(
+			(bezier_x[0], bezier_y[0]).into(),
+			(bezier_x[1], bezier_y[1]).into(),
+			(bezier_x[2], bezier_y[2]).into(),
+			(bezier_x[3], bezier_y[3]).into(),
+		);
+		!bezier.rectangle_intersections((left, top).into(), (right, bottom).into()).is_empty()
```
Keavon, line 606: It's probably safe to also check if the connector curve is entirely within a node (which, once we change the design to make nodes bigger, or perhaps if you have a wide node with a long name and a lot of exposed inputs, this will matter more): `is_contained_within()`

Keavon (approved):
Going to bed now, you're welcome to merge before or after addressing the other comments I left. G'night!

node-graph/graph-craft/src/document.rs
```diff
@@ -180,6 +180,19 @@ impl NodeNetwork {
 			.collect();
 	}
 
+	/// Collect a hashmap of nodes with a list of the nodes that use it as input
```
Keavon, line 183: Does outward mean downstream? That's how I interpret it but I want to make sure that's correct. Maybe if you can put "(downstream/rightward)" in the comment, that'd help make it clearer. And consider, depending if you agree with me about its greater clarity, renaming the function from `outwards` to `downstream`.

0HyperCube: It is a map of each node to a list of the nodes that use it as input.

0HyperCube: This is explained in the comment linked.

Keavon: I see, I think I misinterpreted the intent by assuming something about it that wasn't relevant. Looks good to me! Ready to merge.

PR #906, opened 2022-12-23
Avoid deprecated into_serde()

0HyperCube:
Avoids the warning about `.into_serde()` being deprecated.


Keavon (approved):
Thank you for taking the time to deal with this annoying warning :)

PR #907, opened 2022-12-23
Refactor widget callbacks as a macro

mfish33:
I think this cleans up some of the syntaxes around widget callbacks. I have switched all uses of WidgetCallbacks to use the macro instead of the new function. I am not married to this but think that when we go to a more performant widget system this pattern will become more common.


mfish33:
There are some small build issues. Working on fixing them but the general idea is there

Keavon:
I personally don't see how this is very different, either a function call or a function-looking macro call. My only concern is that rustfmt doesn't handle macros, so that's a QoL downgrade in that specific respect. @0HyperCube @TrueDoctor do you have thoughts?

TrueDoctor:
> I personally don't see how this is very different, either a function call or a function-looking macro call. My only concern is that rustfmt doesn't handle macros, so that's a QoL downgrade in that specific respect. @0HyperCube @TrueDoctor do you have thoughts?

I think this allows us to remove some of the 
```rust
let data_move_1 = data.clone();
let data_move_2 = data.clone();
```
But I'd have to double check if we can't find a different way around that

Keavon:
Since we didn't really come to a conclusion about whether this improves things or not, and since there are a bunch of merge conflicts, I'm going to close this for now. Feel free to bring it up again and we can discuss the pros and cons more.

PR #910, opened 2022-12-24
Diff simple layout changes to avoid excessive DOM replacements

0HyperCube:
The Rust widgets are diffed and only the changed ones are updated. This probably won't speed things up that much but it reduces the amount of data put through serde each frame.

### DIFF
--- a/editor/src/messages/layout/layout_message_handler.rs
+++ b/editor/src/messages/layout/layout_message_handler.rs
@@ -37,10 +37,9 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
 									// Return if this is the correct ID
 									if widget.widget_id == id {
 										path.push(index);
-										return Some(WidgetDiff {
-											path,
-											new_val: DiffUpdate::Widget(widget.clone()),
-										});
+										let new_val = DiffUpdate::Widget(widget.clone());
+										let widget_path = path;
+										return Some(WidgetDiff { widget_path, new_val });
 									}
 								}
 							}
--- a/editor/src/messages/layout/layout_message_handler.rs
+++ b/editor/src/messages/layout/layout_message_handler.rs
@@ -37,9 +37,9 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
 									// Return if this is the correct ID
 									if widget.widget_id == id {
 										path.push(index);
-										let new_val = DiffUpdate::Widget(widget.clone());
+										let new_value = DiffUpdate::Widget(widget.clone());
 										let widget_path = path;
-										return Some(WidgetDiff { widget_path, new_val });
+										return Some(WidgetDiff { widget_path, new_value });
 									}
 								}
 							}
@@ -272,7 +272,7 @@ impl LayoutMessageHandler {
 	/// Send a diff to the frontend based on the layout target.
 	#[remain::check]
 	fn send_diff(&self, mut diff: Vec<WidgetDiff>, layout_target: LayoutTarget, responses: &mut VecDeque<Message>, action_input_mapping: &impl Fn(&MessageDiscriminant) -> Vec<KeysGroup>) {
-		diff.iter_mut().for_each(|diff| diff.new_val.apply_shortcut(action_input_mapping));
+		diff.iter_mut().for_each(|diff| diff.new_value.apply_shortcut(action_input_mapping));
 
 		info!("{layout_target:?} diff {diff:#?}");
 
--- a/editor/src/messages/layout/layout_message_handler.rs
+++ b/editor/src/messages/layout/layout_message_handler.rs
@@ -1,4 +1,4 @@
-use super::utility_types::layout_widget::{LayoutGroup, WidgetDiff};
+use super::utility_types::layout_widget::{LayoutGroup, WidgetDiff, WidgetHolder};
 use super::utility_types::misc::LayoutTarget;
 use crate::messages::input_mapper::utility_types::input_keyboard::KeysGroup;
 use crate::messages::layout::utility_types::layout_widget::{DiffUpdate, Widget};
@@ -17,6 +17,32 @@ pub struct LayoutMessageHandler {
 	layouts: [Layout; LayoutTarget::LayoutTargetLength as usize],
 }
 
+impl LayoutMessageHandler {
+	/// Get the widget path for the widget with the specified id
+	fn get_widget_path(widget_layout: &WidgetLayout, id: u64) -> Option<(&WidgetHolder, Vec<usize>)> {
+		let mut stack = widget_layout.layout.iter().enumerate().map(|(index, val)| (vec![index], val)).collect::<Vec<_>>();
+		while let Some((mut widget_path, group)) = stack.pop() {
+			match group {
+				// Check if any of the widgets in the current column or row have the correct id
+				LayoutGroup::Column { widgets } | LayoutGroup::Row { widgets } => {
+					for (index, widget) in widgets.iter().enumerate() {
+						// Return if this is the correct ID
+						if widget.widget_id == id {
+							widget_path.push(index);
+							return Some((widget, widget_path));
+						}
+					}
+				}
+				// A section contains more LayoutGroups which we add to the stack.
+				LayoutGroup::Section { layout, .. } => {
+					stack.extend(layout.iter().enumerate().map(|(index, val)| ([widget_path.as_slice(), &[index]].concat(), val)));
+				}
+			}
+		}
+		None
+	}
+}
+
 impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage, F> for LayoutMessageHandler {
 	#[remain::check]
 	fn process_message(&mut self, message: LayoutMessage, data: F, responses: &mut std::collections::VecDeque<Message>) {
@@ -26,35 +52,14 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
 		#[remain::sorted]
 		match message {
 			RefreshLayout { layout_target, dirty_id } => {
-				// Create a widget update diff for the relevant id
-				let create_widget_diff = |widget_layout: &WidgetLayout, id: u64| {
-					let mut stack = widget_layout.layout.iter().enumerate().map(|(index, val)| (vec![index], val)).collect::<Vec<_>>();
-					while let Some((mut path, group)) = stack.pop() {
-						match group {
-							// Check if any of the widgets in the current column or row have the correct id
-							LayoutGroup::Column { widgets } | LayoutGroup::Row { widgets } => {
-								for (index, widget) in widgets.iter().enumerate() {
-									// Return if this is the correct ID
-									if widget.widget_id == id {
-										path.push(index);
-										let new_value = DiffUpdate::Widget(widget.clone());
-										let widget_path = path;
-										return Some(WidgetDiff { widget_path, new_value });
-									}
-								}
-							}
-							// A section contains more LayoutGroups which we add to the stack.
-							LayoutGroup::Section { layout, .. } => {
-								stack.extend(layout.iter().enumerate().map(|(index, val)| ([path.as_slice(), &[index]].concat(), val)));
-							}
-						}
-					}
-					None
-				};
 				// Find the updated diff based on the specified layout target
 				let Some(diff) = (match &self.layouts[layout_target as usize] {
 					Layout::MenuLayout(_) => return,
-					Layout::WidgetLayout(layout) => create_widget_diff(layout, dirty_id),
+					Layout::WidgetLayout(layout) => Self::get_widget_path(layout, dirty_id).map(|(widget, widget_path)| {
+						// Create a widget update diff for the relevant id
+						let new_value = DiffUpdate::Widget(widget.clone());
+						WidgetDiff { widget_path, new_value }
+					}),
 				}) else {
 					return;
 				};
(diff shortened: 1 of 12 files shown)


frontend/src/wasm-communication/messages.ts
```diff
@@ -1188,13 +1190,46 @@ export type WidgetLayout = {
...
 export function defaultWidgetLayout(): WidgetLayout {
 	return {
 		layoutTarget: undefined,
 		layout: [],
 	};
 }
 
+// Updates a widget layout based on a list of updates, returning the new layout
+export function patchWidgetLayout(layout: WidgetLayout, updates: WidgetDiffUpdate): void {
+	layout.layoutTarget = updates.layoutTarget;
```
mfish33, line 1209: If anything I think this should be an asset. I do not think there is a case where the update and the base do not have the same `layoutTarget`

frontend/src/wasm-communication/messages.ts
```diff
@@ -1188,13 +1190,46 @@ export type WidgetLayout = {
...
+			else if ("rowWidgets" in targetLayout) targetLayout = targetLayout.rowWidgets[index];
+			else if ("layout" in targetLayout) targetLayout = targetLayout.layout[index];
+			// eslint-disable-next-line no-console
+			else if (targetLayout instanceof Widget) console.error("Tried to index widget");
+			else if ("action" in targetLayout) targetLayout = targetLayout.children ? targetLayout.children[index] : undefined;
+			else targetLayout = targetLayout[index];
+		});
+
+		// Some odd code required to update the objects (assigning things doesn't work?)
+		if (targetLayout !== undefined && "length" in targetLayout) targetLayout.length = 0;
```
mfish33, line 1226: These three lines seem sketchy. Do you mind adding some comments on why exactly they are needed?

editor/src/messages/layout/utility_types/layout_widget.rs
```diff
@@ -34,85 +34,103 @@ pub enum Layout {
...
+	SubLayout(SubLayout),
+	#[serde(rename = "layoutGroup")]
+	LayoutGroup(LayoutGroup),
+	#[serde(rename = "widget")]
+	Widget(WidgetHolder),
+}
+
+#[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub struct WidgetDiff {
+	pub path: Vec<usize>,
```
Keavon, line 49: Would `widget_path` be less ambiguous? We're trying to reduce the highly-overloaded nature of the word `path` throughout our codebase.

frontend/src/wasm-communication/messages.ts
```diff
@@ -1214,115 +1249,127 @@ export function isWidgetSection(layoutRow: LayoutGroup): layoutRow is WidgetSect
...
+	});
+	return diff;
+}
 
-		if (layoutType.row) {
-			const rowWidgets = hoistWidgetHolders(layoutType.row.rowWidgets);
+// Unpacking a layout group
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+function createLayoutGroup(newVal: any): LayoutGroup {
+	if (newVal.column) {
```
mfish33, line 1272: Do not really like the name `newVal` here. Can we use something more descriptive here and above as well?

frontend/src/wasm-communication/messages.ts
```diff
@@ -1188,13 +1190,46 @@ export type WidgetLayout = {
...
 	};
 }
 
+// Updates a widget layout based on a list of updates, returning the new layout
+export function patchWidgetLayout(layout: WidgetLayout, updates: WidgetDiffUpdate): void {
+	layout.layoutTarget = updates.layoutTarget;
+
+	updates.diff.forEach((update) => {
+		// Find the object where the diff applies to
+		let targetLayout = layout.layout as LayoutGroup[] | LayoutGroup | Widget | MenuBarEntry[] | MenuBarEntry | undefined;
```
Keavon, line 1213: I suggest assigning `LayoutGroup[] | LayoutGroup | Widget | MenuBarEntry[]` to a named type, and using it above in your `type WidgetDiff = ...` declaration also.

frontend/src/wasm-communication/messages.ts
```diff
@@ -1188,13 +1190,46 @@ export type WidgetLayout = {
...
+		update.path.forEach((index) => {
```
Keavon, line 1214: It looks like the goal of this loop is to find a single entry in the array, is that correct? Would it, thus, be desirable to break from the loop once it's found? You'd need to probably use a regular old loop in that case in order to break. Or do you intend to continue until the end of the loop every time?

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -24,14 +25,44 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
 		#[remain::sorted]
 		match message {
-			RefreshLayout { layout_target } => {
-				self.send_layout(layout_target, responses, &action_input_mapping);
-			}
-			SendLayout { layout, layout_target } => {
-				self.layouts[layout_target as usize] = layout;
-
-				self.send_layout(layout_target, responses, &action_input_mapping);
+			RefreshLayout { layout_target, dirty_id } => {
```
mfish33, line 28: I think this should be broken up into two functions. I think a function should be added to get a path from a layout given a widget_id that is impl'd onto the struct. The function signature could be

```rust
fn get_path(&self, widget_id:u64) -> Option<(Path, &Widget)>
```

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -208,55 +239,57 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
 }
 
 impl LayoutMessageHandler {
+	/// Diff the update and send to the frontend where necessary
+	fn send_layout(&mut self, layout_target: LayoutTarget, new_layout: Layout, responses: &mut VecDeque<Message>, action_input_mapping: &impl Fn(&MessageDiscriminant) -> Vec<KeysGroup>) {
+		// We don't diff the menu bar layout yet.
+		if layout_target == LayoutTarget::MenuBar {
```
mfish33, line 245: I would not check layout target but rather by layout type. `MenuLayout` vs `WidgetLayout`

```rust
pub enum Layout {
	WidgetLayout(WidgetLayout),
	MenuLayout(MenuLayout),
}
```

editor/src/messages/layout/utility_types/layout_widget.rs
```diff
@@ -129,6 +147,21 @@ impl Layout {
 			Layout::WidgetLayout(widget_layout) => Box::new(widget_layout.iter_mut()),
 		}
 	}
+
+	/// Diffing updates self (where self is old) based on new, updating the list of modifications as it does so.
+	pub fn diff(&mut self, new: Self, path: &mut Vec<usize>, widget_diffs: &mut Vec<WidgetDiff>) {
+		match (self, new) {
+			(Self::WidgetLayout(s), Self::WidgetLayout(new)) => return s.diff(new, path, widget_diffs),
```
Keavon, line 154: Remove `return`

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -208,55 +239,57 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+				FrontendMessage::UpdateMenuBarLayout {
+					layout_target,
+					layout: self.layouts[layout_target as usize].clone().unwrap_menu_layout(action_input_mapping).layout,
+				}
+				.into(),
+			);
+			return;
+		}
+
+		let mut widget_diffs = Vec::new();
```
mfish33, line 263: I think that we should only do diffs for when we `RefreshLayout`. I think we need a hybrid system where a diff or the entire layout can be sent. Every time something on the backend sends a layout all of the `widget_ids` are changed. I think we can do something to compare shapes and update on fine-grained details. However, I think that this is out of scope for just the initial work

frontend/src/wasm-communication/messages.ts
0HyperCube: Well they don't on the first update.

frontend/src/wasm-communication/messages.ts
0HyperCube: > Some odd code required to update the objects (assigning things doesn't work?)

Already commented.

frontend/src/wasm-communication/messages.ts
0HyperCube: The path is a list of indexes i.e. the second widget in the third group in the first section would be a widgetPath of `[0, 2, 1]` We always want to go through each item of the path and sample the layoutTarget at that index.

frontend/src/wasm-communication/messages.ts
0HyperCube: I don't really know why javascript likes this to be honest. Something to do with how the implicit pointers are done.

frontend/src/wasm-communication/messages.ts
mfish33: I wrote this before seeing that the system is diff only. I think a hybrid system is necessary so that diffs only patch existing layouts

frontend/src/wasm-communication/messages.ts
mfish33: Some comments on what you are expecting the shapes of the objects would be helpful or maybe some links to stack overflow. I have never seen something needed like this in all my web experiance

editor/src/messages/layout/layout_message_handler.rs
0HyperCube: > I think we need a hybrid system where a diff or the entire layout can be sent. 

We currently send always the diff and there are no issues? Or am I misunderstanding?

frontend/src/wasm-communication/messages.ts
0HyperCube: It is probably possible to avoid this but I don't know enough js to do so.

frontend/src/wasm-communication/messages.ts
0HyperCube: Currently on start we just send a diff with the full layout and a path of [] - which isn't too much overhead for a more consistent system.

editor/src/messages/layout/layout_message_handler.rs
mfish33: I think always sending the diff is less efficient in many cases than what we currently are doing. This is because even if nothing changes in the layout all of the `widget_ids` change. I think the system would be simpler if diffs were only of individual widgets.

mfish33: To add, it does not seem like this affected the memory usage of UI elements in preliminary testing on the preview

0HyperCube: `RefreshLayout` I renamed to `ResendActiveWidget` as that is what it does now.

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -208,55 +239,57 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+	}
+
+	/// Send a diff to the frontend based on the layout target.
 	#[remain::check]
-	fn send_layout(&self, layout_target: LayoutTarget, responses: &mut VecDeque<Message>, action_input_mapping: &impl Fn(&MessageDiscriminant) -> Vec<KeysGroup>) {
-		let layout = &self.layouts[layout_target as usize];
+	fn send_diff(&self, mut diff: Vec<WidgetDiff>, layout_target: LayoutTarget, responses: &mut VecDeque<Message>, action_input_mapping: &impl Fn(&MessageDiscriminant) -> Vec<KeysGroup>) {
+		diff.iter_mut().for_each(|diff| diff.new_val.apply_shortcut(action_input_mapping));
+
+		info!("{layout_target:?} diff {diff:#?}");
```
Keavon, line 278: Please remove this log, or switch it to `trace!()`. In the future I also recommend debug logging with `debug!()` which, just like `info!()`, is always enabled regardless of debug verbosity level, and it's easier to remember that we are supposed to remove `debug!()` statements before committing. We use `info!()` for things like printing the message tree or serialized document format, when the user specifically requests something be shown and the log verbosity level is irrelevant in that. Thanks :)

editor/src/messages/layout/layout_message_handler.rs
0HyperCube: > Every time something on the backend sends a layout all of the widget_ids are changed.

No - a new layout is generated with new ids but we don't actually change the ids of the cached layout unless part of it has changed.

editor/src/messages/layout/layout_message_handler.rs
0HyperCube: `debug!` is too many keystrokes :/.

Keavon: Haha, fair enough :)

Keavon: And at least now I'll probably remember exactly what `info!()` is so I can keep an eye out during code review for that.

editor/src/messages/layout/layout_message_handler.rs
0HyperCube: > This is because even if nothing changes in the layout all of the widget_ids change.

Well a new layout is generated with new ids, but they do not get updated in the cached layout or in the frontend so same on update time unless the widget has actually changed.

0HyperCube: > To add, it does not seem like this affected the memory usage of UI elements in preliminary testing on the preview

This was not intended to reduce memory usage, only to reduce the amount of time per frame serialising data with serde.

Keavon (approved):

mfish33 (approved):

PR #913, opened 2022-12-26
Ability to debounce inputs going to the backend 

mfish33:
This helps out with widget memory usage since it cuts down on the number of messages being sent back and forth. Currently, it is only being applied to the`NumberInput` but more inputs could be added easily


Keavon:
This looks like a great idea but we can't merge it until successfully debugging and fixing the big memory/perf leak in the app, since this makes it harder to effectively reproduce.

Keavon (approved):

Keavon:
I merged this anyways since we probably won't make any more insightful progress in the short term towards figuring out the elusive memory/perf leak, or if it even exists at all anymore (perhaps it might have truly been fixed).

Keavon:
Closes #974

PR #914, opened 2022-12-26
Bezier-rs: Add project function for Subpaths

RobNadal:
- Added naive (brute force) project function to subpath interface
- Added update on mouse move option to subpath components


libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -7,13 +9,34 @@ impl Subpath {
 	pub fn length(&self, num_subdivisions: Option<usize>) -> f64 {
 		self.iter().fold(0., |accumulator, bezier| accumulator + bezier.length(num_subdivisions))
 	}
+
+	/// Returns the segment index and `t` value that corresponds to the closest point on the curve to the provided point.
+	/// Uses a searching algorithm akin to binary search that can be customized using the [ProjectionOptions] structure.
+	pub fn project(&self, point: DVec2, options: ProjectionOptions) -> (usize, f64) {
+		if self.is_empty() {
+			panic!("Can not project to an empty Subpath")
+		}
```
Androxium, line 18: Should we add this guard to all of our subpath functions?

libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -7,13 +9,34 @@ impl Subpath {
...
+		let (index, (_, project_t)) = self
+			.iter()
+			.map(|bezier| {
+				let project_t = bezier.project(point, options);
+				(bezier.evaluate(ComputeType::Parametric(project_t)).distance(point), project_t)
+			})
+			.enumerate()
+			.fold(
+				(0, (f64::INFINITY, 0.)), // If the Subpath contains only a single manipulator group, returns (0, 0.)
+				|(i1, (d1, t1)), (i2, (d2, t2))| if d1 < d2 { (i1, (d1, t1)) } else { (i2, (d2, t2)) },
```
Androxium, line 30: Nit: what does `i1` and `d1` represent? Is that like `index1` and `distance1` for the subpath segment and project distance for that current loop iteration?

Androxium (approved):

libraries/bezier-rs/src/subpath/lookup.rs
RobNadal: Length and intersection are the only other subpath functions, and I don't believe we need to worry about this edge case for either. Length just returns 0, and intersection correctly returns an empty list

libraries/bezier-rs/src/subpath/lookup.rs
RobNadal: Yah so I wanted to just use the `min` function, but it doesn't work properly for floats, so I had to use `fold` instead. So technically `(i1, (d1, t1))` is the accumulator, but for the purpose of emulating the `min` function its not really accumulating anything, so I think the naming convention makes more sense by styling it as I would a comparison function. I really don't wanna write it out as `(index1, (distance1, tvalue1)), (index2, (distance2, tvalue2))` thats so gross

libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -7,13 +9,34 @@ impl Subpath {
 	pub fn length(&self, num_subdivisions: Option<usize>) -> f64 {
 		self.iter().fold(0., |accumulator, bezier| accumulator + bezier.length(num_subdivisions))
 	}
+
+	/// Returns the segment index and `t` value that corresponds to the closest point on the curve to the provided point.
+	/// Uses a searching algorithm akin to binary search that can be customized using the [ProjectionOptions] structure.
+	pub fn project(&self, point: DVec2, options: ProjectionOptions) -> (usize, f64) {
+		if self.is_empty() {
+			panic!("Can not project to an empty Subpath")
```
Keavon, line 17: @0HyperCube @TrueDoctor thoughts on whether this should return `Option`, `Result`, or panic?

Keavon (approved):
Looks good from my perspective but let's wait to hear what @0HyperCube and @TrueDoctor have to say before merging.

libraries/bezier-rs/src/subpath/lookup.rs
TrueDoctor: I'd favor returning an Option because empty is afaict a valid state for Subpath.

libraries/bezier-rs/src/subpath/lookup.rs
TrueDoctor: You can still use `min_by(|a, b|  a.total_cmp(&b))` which would probably improve code readability

RobNadal:
@Keavon and @TrueDoctor I've resolved all comments, let me know if we're good to merge.

PR #915, opened 2022-12-26
Bezier-rs: Add self_intersection to subpath

Androxium:
This PR adds a naive brute-force implementation of the `self_intersections` function to the `Subpath` library. This PR also updates the function description of the `intersections` function to include the `minimum_separation` variable.


libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -31,8 +34,10 @@ impl Subpath {
 
 	/// Calculates the intersection points the subpath has with a given line and returns a list of parameteric `t`-values.
 	/// This function expects the following:
-	/// - other: a [Bezier] curve to check intersections against
-	/// - error: an optional f64 value to provide an error bound
+	/// - `other`: a [Bezier] curve to check intersections against
+	/// - `error`: an optional f64 value to provide an error bound
+	/// - minimum_seperation: the minimum difference 2 adjacent `t`-values must have, when comparing adjacent `t`-values, if 2 `t`-values do not satisfy the
```
RobNadal, line 39: itty bitty nit: can you put ticks around this argument like the two above? Also, you could probably split this sentence at the if so it flows a bit nicer (here and below)

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -58,6 +63,32 @@ impl Subpath {
...
+	/// Returns a list of `t` values that correspond to the self intersection points of the subpath. For each intersection point, the returned `t` value is the smaller of the two that correspond to the point.
+	/// - `error` - For intersections with non-linear beziers, `error` defines the threshold for bounding boxes to be considered an intersection point.
+	/// - `minimum_seperation`: the minimum difference 2 adjacent `t`-values must have, when comparing adjacent t-values, if 2 `t`-values do not satisfy the
+	/// condition, the function takes the larger `t`-value
+	pub fn self_intersections(&self, error: Option<f64>, minimum_seperation: Option<f64>) -> Vec<f64> {
+		let mut intersections_vec = Vec::new();
+		let n = self.len_segments();
+		// TODO: optimization opportunity - this for-loop currently compares all intersections with all curve-segments in the subpath collection
+		for i in 0..n {
+			let other = self.iter().nth(i).unwrap();
```
RobNadal, line 76: It feels a bit counter-intuitive to how iterators are usually used to be doing a for-loop over `i` and then selecting the `i`th element in the iterator. Could you try using `foreach` instead here, and in the loop below? If you still want the indices, you can use `enumerate` method beforehand as well. The iterator `skip` method would also probably work well for how you only iterate from `i` onwards in the second loop structure (e.g. `self.iter().skip(i).foreach(...)`

website/other/bezier-rs-demos/wasm/src/subpath.rs
```diff
@@ -154,4 +154,19 @@ impl WasmSubpath {
...
+
+	pub fn self_intersections(&self) -> String {
+		let subpath_svg = self.to_default_svg();
+		let self_intersections_svg = self
+			.0
+			.self_intersections(None, None)
+			.iter()
+			.map(|intersection_t| {
+				let point = &self.0.evaluate(ComputeType::Parametric(*intersection_t));
+				draw_circle(*point, 4., RED, 1.5, WHITE)
```
RobNadal, line 166: Lots of reference shenanigans going on here. Could you verify that all the `*`s and `&`s are necessary in this lambda?

RobNadal:
Hmm, so I'm seeing two issues unfortunately. 
image

First, the endpoints are occasionally being counted as intersections. I'm not 100% sure why, but just wiggling the endpoints around on the visualization and you'll see that they occasionally flash red.

Second, I'm seeing this rust error (in the above screenshot) which is causing the visualization to crash. The conditions for the crash seem unclear, but it's very easy to trigger - just try and make any of the segments a straight line.

Androxium:
For the endpoints showing up as an intersection point, I need to manually filter them out because the way they are currently stored, the endpoint of 1 bezier curve will intersect with the startpoint of the next bezier curve in the subpath. I can solve this by increasing the minimum separation `t`-values must satisfy as a stopgap solution.

For the recursive structure error, I think it's originating from the cubic bezier itself. It's reproducable if you make a cubic bezier a straight line. Will investigate

RobNadal:
Unfortunately, I'm still seeing flickering on the endpoints. I'm not sure if modifying the minimum separation argument will work, because that ensures there won't be two points at a particular distance from one another - but I don't think thats the issue, because I don't know when there will be two intersections at the endpoint. I'm thinking we should filter intersections which occur at within some epsilon from t=0, and have that be a configurable argument or something. Otherwise, we should add an argument called `include_endpoints` to the intersections function, which by default is `true`.

libraries/bezier-rs/src/consts.rs
```diff
@@ -9,7 +9,7 @@ pub const NUM_DISTANCES: usize = 5;
 /// Maximum allowed angle that the normal of the `start` or `end` point can make with the normal of the corresponding handle for a curve to be considered scalable/simple.
 pub const SCALABLE_CURVE_MAX_ENDPOINT_NORMAL_ANGLE: f64 = std::f64::consts::PI / 3.;
 /// Minimum allowable separation between adjacent `t` values when calculating curve intersections
-pub const MIN_SEPERATION_VALUE: f64 = 5. * 1e-3;
+pub const MIN_SEPERATION_VALUE: f64 = 5. * 1e-2;
```
RobNadal, line 12: Do we still need to change this?

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -32,8 +34,10 @@ impl Subpath {
 
 	/// Calculates the intersection points the subpath has with a given line and returns a list of parameteric `t`-values.
 	/// This function expects the following:
-	/// - other: a [Bezier] curve to check intersections against
-	/// - error: an optional f64 value to provide an error bound
+	/// - `other`: a [Bezier] curve to check intersections against
+	/// - `error`: an optional f64 value to provide an error bound
+	/// - `minimum_seperation`: the minimum difference 2 adjacent `t`-values must have when comparing adjacent `t`-values in sorted order.
+	/// If the comparison condition is not satisfied, the function takes the larger `t`-value of the 2
```
RobNadal, line 40: Can you replace the instances of `2` with `two`, here and in the comment below

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +63,34 @@ impl Subpath {
...
+		// TODO: optimization opportunity - this for-loop currently compares all intersections with all curve-segments in the subpath collection
+		self.iter().enumerate().for_each(|(i, other)| {
+			intersections_vec.extend(other.self_intersections(error).iter().map(|value| (value[0] + (i as f64)) / (n as f64)));
+			self.iter().enumerate().skip(i + 1).for_each(|(j, curve)| {
+				intersections_vec.extend(
+					curve
+						.intersections(&other, error, minimum_seperation)
+						.iter()
+						.filter(|&value| {
+							let err = error.unwrap_or(MAX_ABSOLUTE_DIFFERENCE);
```
RobNadal, line 85: Move this to the top of the function

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +63,34 @@ impl Subpath {
...
+							value > &err && (1. - value) > err
+						})
+						.map(|value| (value + (j as f64)) / (n as f64)),
```
RobNadal, line 88: Could you simplify this to return tuples of `(curve_index, t_value)`? I haven't been able to make the ComputeType changes for subpaths, but I think within our own code local t values are so much easier to work with than the global t value. 

This will make it more annoying to draw the points in the wasm folder, but I've done the same thing in Project, so I think returning  `(curve_index, t_value)` is more consistent.

website/other/bezier-rs-demos/src/features/subpath-features.ts
```diff
@@ -1,6 +1,22 @@
...
+	default: 0.02,
+};
+
+const tMinimumSeperationOptions = {
+	variable: "minimum_seperation",
+	min: 0.001,
+	max: 0.25,
+	step: 0.001,
+	default: 0.05,
+};
```
RobNadal, line 18: Could you move these option definitions to the `utils/options.ts` file and give them more descriptive names such as `intersectionErrorOptions`

RobNadal:
Tested it locally and flickering is gone, even for low error. Great job @Androxium!

libraries/bezier-rs/src/subpath/solvers.rs
Androxium: Just to be clear, you want to change the return of `intersections` and `self_intersections` to be `Vec<(usize, f64)>`, correct?

RobNadal: Yes please, thanks!

website/other/bezier-rs-demos/src/utils/options.ts
```diff
@@ -21,3 +21,19 @@ export const tMinimumSeperationOptions = {
...
+export const intersectionErrorOptions = {
+	variable: "error",
+	min: 0.001,
+	max: 0.525,
+	step: 0.0025,
+	default: 0.02,
+};
+
+export const intersectionMinimumSeperationOptions = {
+	variable: "minimum_seperation",
```
RobNadal, line 34: Small typo: separation vs. seperation. Might be easiest to just ctrl-f for seperation and replace all instances, because I've noticed it a few other places.

RobNadal (approved):

Androxium:
Rebase work onto this new PR https://github.com/GraphiteEditor/Graphite/pull/1035

PR #917, opened 2022-12-27
Make copying/duplicating nodes not preserve the incoming connection

0HyperCube:
Remove wires to nodes outside of copy, as mentioned in the discord todo list.


Keavon:
<kbd>Ctrl</kbd><kbd>D</kbd> doesn't seem to be working anymore as far as I can tell.

0HyperCube:
Looks like I cleared the selected nodes list before duplicating the nodes 🙃.

Keavon:
Would it be possible (read: easy) to, when pasting a node, check if that space is occupied and shift it down similar to Ctrl+D in that circumstance? Or maybe the easier route of just applying the offset to all pastes would be a reasonable stopgap. I realize the contracts my original spec that Ctrl+D offsets while Ctrl+V doesn't, but now that I've tried it out, I am concerned about overlapping nodes causing confusion.

Keavon:
Works perfect! 😄

Keavon (approved):

PR #923, opened 2022-12-28
Add a 'Preserve Aspect Ratio' checkbox to Properties panel

### LINKED ISSUE #919: Add a "lock aspect ratio" checkbox to a layer's transform Scale/Dimensions
This checkbox should go in the Assists column of the Scale row. Reference this part of the UI mockup:

image

For now, it can just live in the Scale row since we don't have any way to span multiple rows yet.

It should be checked by default. When checked, any value typed into the Scale or Dimensions in X/Y or W/H should proportionally scale (according to the current aspect ratio) the other X/Y or W/H value.

The 12px icon is called "Link".

0HyperCube:
Closes #919

Also properties panel updates are now centred around the pivot.

### DIFF
--- a/document-legacy/src/document.rs
+++ b/document-legacy/src/document.rs
@@ -628,6 +628,12 @@ impl Document {
 				}
 				Some(vec![LayerChanged { path: layer_path.clone() }])
 			}
+			Operation::SetLayerPreserveAspect { layer_path, preserve_aspect } => {
+				if let Ok(layer) = self.layer_mut(&layer_path) {
+					layer.preserve_aspect = preserve_aspect;
+				}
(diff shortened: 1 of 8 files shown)


Keavon (approved):

mariusa:
Hi, I tested Preserve aspect ratio at https://editor.graphite.rs/
but it doesn't have any effect: when dragging a shape by lateral/up/down markers, it's still resized only on 1 dimension, instead of keeping aspect ratio.
Screenshot from 2023-01-10 19-15-48

Keavon:
You have to hold Shift. This checkbox only affects the Scale and Dimensions properties.

mariusa:
Thanks for clarification, I see now.

PR #924, opened 2022-12-28
Only open the node graph panel after drawing a new frame to avoid layout shift interruption

0HyperCube:
The node graph window no longer appears whilst drawing the node graph. Drag start positions are also now based on document coordinates.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -3,8 +3,9 @@ use crate::messages::layout::utility_types::widgets::button_widgets::{Breadcrumb
 use crate::messages::prelude::*;
 
 use document_legacy::document::Document;
-use document_legacy::layers::layer_info::LayerDataType;
+use document_legacy::layers::layer_info::{LayerDataType, LayerDataTypeDiscriminant};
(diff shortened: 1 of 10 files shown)


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -101,6 +102,8 @@ pub struct NodeGraphMessageHandler {
 	pub selected_nodes: Vec<graph_craft::document::NodeId>,
 	#[serde(skip)]
 	pub widgets: [LayoutGroup; 2],
+	#[serde(skip)]
+	pub is_drawing: bool,
```
Keavon, line 106: Based on my read of the code, this is only updated when the Imaginate or Node Graph Frame tools send the update messages-- and other tools don't update it. To avoid confusion, I'd suggest adding a comment to this bool explaining what it does and how it's only used by certain tools that need it like Imaginate and Node Graph Frame.

0HyperCube: I renamed it to `is_drawing_node_graph_frame`

Keavon: Perfect

Keavon (approved):

PR #926, opened 2022-12-28
Add a button to refresh the input

0HyperCube:
As discussed in discord, the input node now has a button to refresh itself.


Keavon:
Can you please add an empty label so the left the of the button aligns the with other label left edges?

Keavon (approved):

PR #927, opened 2022-12-28
Gradient tool improvements

0HyperCube:
- Existing gradient stop points are reused (from discord todo list)
- Undo works on gradient tool modifications (not properties panel yet)
- Double click on the line to insert a stop (like inkscape)

### DIFF
--- a/document-legacy/src/color.rs
+++ b/document-legacy/src/color.rs
@@ -205,4 +205,17 @@ impl Color {
 
 		Some(Color::from_rgb8(r, g, b))
 	}
+
+	/// Linearly interpolates between two colors based on t.
+	///
+	/// T must be between 0 and 1.
+	pub fn lerp(self, other: Color, t: f32) -> Option<Self> {
+		assert!((0. ..=1.).contains(&t));
+		Color::from_rgbaf32(
+			self.red + ((other.red - self.red) * t),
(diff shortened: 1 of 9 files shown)


Keavon:
Can you make it so the Linear/Radial buttons in the tool's options bar syncs with the choice for the active layer in the Properties panel?

- With no layer selected that has a gradient, that dictates the type of the new gradient
- With a layer selected that has a gradient, changing it in either location changes the active gradient
- When you switch to a different active layer that has no gradient from one that did, it should probably switch back to what was most recently set on the tool options without a layer selected containing a gradient (in other words: with a layer with a gradient selected, that overrides the tool's own linear/radial setting until deselected)

Keavon:
I like being able to double click to add a gradient stop. But the click target area is really small right now, can that be increased a bit? I'd also love to be able to double click an existing stop point to remove it.

Keavon:
Undo is working great, that's very nice to see. Now I just gotta make a list of *all* the other places undo isn't working 😜.

Keavon (approved):
Point deletion handles very nicely! I love how you can delete an endpoint to shorten the gradient line and you can even delete an end of a pair to make it a solid color. Great attention to detail.

0HyperCube:
Just copied inkscape for the behaviour there.

PR #903, opened 2022-12-22
Restructure GPU compilation execution pipeline

TrueDoctor:
Closes #810

Remove the requirement to compile graphite with nightly rust to compile gpu code.
This also modularizes the execution model to add a wgpu runner.
Furthermore, this pr will add a rudimentary compilation server to allow the compilation and running of gpu shaders from the web editor

PR #935, opened 2022-12-31
Fix cargo-about license acceptance

0HyperCube:
Fix build failing due to cryptography libraries introduced by #903


Keavon:
Because I didn't want to have several commits in the history that failed to build, I amended the original commit that broke the build with your file change. Thank you for figuring out how to edit `about.toml` (True Doctor and me both researched it but weren't able to solve it in the time we had available.)

PR #937, opened 2023-01-02
Don't show the blue outline when bloburl is set

0HyperCube:
Don't show the blue outline around node graph frames when the blob url is set. This can make the frame invisibe.


Keavon (approved):

Keavon:
Thanks

PR #939, opened 2023-01-02
Fix viewport culling with nested layers

0HyperCube:
If the layer is outside of the viewport then we don't render it and mark the cache as dirty, however if a child of a folder is not rendered then the parent folder may not be marked as dirty, meaning that we do not check again if the layer is inside the viewport.

Steps to reproduce:
- Draw two rectangles next to each other
- Place them in a folder
- Move the folder off the screen
- Move the document so you can see the folder again, noticing that one of the rectangles has disappeared

This is because the folder is using a cached value where one of the rectangles has been culled, and it has not been marked as dirty.

### DIFF
--- a/document-legacy/src/layers/folder_layer.rs
+++ b/document-legacy/src/layers/folder_layer.rs
@@ -22,13 +22,13 @@ pub struct FolderLayer {
 
 impl LayerData for FolderLayer {
 	fn render(&mut self, svg: &mut String, svg_defs: &mut String, transforms: &mut Vec<glam::DAffine2>, render_data: RenderData) -> bool {
-		let mut failed = false;
+		let mut any_child_requires_refresh = false;
 		for layer in &mut self.layers {
-			let (svg_value, success) = layer.render(transforms, svg_defs, render_data);
+			let (svg_value, requires_refresh) = layer.render(transforms, svg_defs, render_data);
 			*svg += svg_value;
-			failed = failed || (!success);
+			any_child_requires_refresh = any_child_requires_refresh || requires_refresh;
 		}
-		!failed
+		!any_child_requires_refresh
 	}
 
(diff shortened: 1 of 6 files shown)


document-legacy/src/layers/layer_info.rs
```diff
@@ -119,7 +119,7 @@ impl<'a> TryFrom<&'a Layer> for &'a Subpath {
 
 /// Defines shared behavior for every layer type.
 pub trait LayerData {
-	/// Render the layer as an SVG tag to a given string.
+	/// Render the layer as an SVG tag to a given string, returning if rendering fully successful.
```
Keavon, line 122: I'm having trouble understanding this added part for the return value. Would you mind fixing the grammar and elaborating a bit on what it means to have rendered "fully" successfully? I can't tell if "fully" means the full area so no parts were culled, or if it means that some parts may have failed somehow, or what else that might mean.

document-legacy/src/layers/layer_info.rs
```diff
@@ -321,15 +324,17 @@ impl Layer {
 					transforms.pop();
 					self.cache.clear();
 					self.cache_dirty = true;
-					return "";
+					return ("", false);
 				}
 			}
 		}
 
+		let mut success = true;
```
Keavon, line 332: Here also, just the word "success" doesn't clearly indicate the precise meaning here. It's not obvious to me what it means for the render to be successful, or what the implications are in either case.

Keavon (approved):
The fix works. So please consider my comments and then merge while I'm asleep. Good night!

document-legacy/src/layers/folder_layer.rs
```diff
@@ -22,13 +22,13 @@ pub struct FolderLayer {
 
 impl LayerData for FolderLayer {
 	fn render(&mut self, svg: &mut String, svg_defs: &mut String, transforms: &mut Vec<glam::DAffine2>, render_data: RenderData) -> bool {
-		let mut failed = false;
+		let mut any_child_requires_refresh = false;
```
Keavon, line 25: So much clearer, thank you for the better name!

PR #941, opened 2023-01-02
Use builder pattern for widgets

0HyperCube:
Builder pattern for widgets e.g.
```rs
DropdownEntryData::new(mode.to_string())
	.on_update(update_value(move |_| TaggedValue::ImaginateMaskStartingFill(mode), node_id, mask_fill_index)),
```

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
@@ -17,7 +17,7 @@ pub fn string_properties(text: impl Into<String>) -> Vec<LayoutGroup> {
 	vec![LayoutGroup::Row { widgets: vec![widget] }]
 }
 
-fn update_value<T, F: Fn(&T) -> TaggedValue + 'static + Send + Sync>(value: F, node_id: NodeId, input_index: usize) -> impl Fn(&T) -> Message + 'static + Send + Sync {
+fn update_value<T>(value: impl Fn(&T) -> TaggedValue + 'static + Send + Sync, node_id: NodeId, input_index: usize) -> impl Fn(&T) -> Message + 'static + Send + Sync {
 	move |input_value: &T| {
 		NodeGraphMessage::SetInputValue {
 			node_id,
@@ -72,8 +72,7 @@ fn text_widget(document_node: &DocumentNode, node_id: NodeId, index: usize, name
 	{
 		widgets.extend_from_slice(&[
 			WidgetHolder::unrelated_separator(),
-			TextInput::new()
-				.value(x.clone())
+			TextInput::new(x.clone())
 				.on_update(update_value(|x: &TextInput| TaggedValue::String(x.value.clone()), node_id, index))
 				.widget_holder(),
 		])
@@ -91,8 +90,7 @@ fn text_area_widget(document_node: &DocumentNode, node_id: NodeId, index: usize,
 	{
 		widgets.extend_from_slice(&[
 			WidgetHolder::unrelated_separator(),
-			TextAreaInput::new()
-				.value(x.clone())
+			TextAreaInput::new(x.clone())
 				.on_update(update_value(|x: &TextAreaInput| TaggedValue::String(x.value.clone()), node_id, index))
 				.widget_holder(),
 		])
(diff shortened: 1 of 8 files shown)


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -161,12 +153,11 @@ fn number_widget(document_node: &DocumentNode, node_id: NodeId, index: usize, na
...
 pub fn input_properties(_document_node: &DocumentNode, _node_id: NodeId, _context: &mut NodePropertiesContext) -> Vec<LayoutGroup> {
 	let information = WidgetHolder::text_widget("The graph's input is the artwork under the frame layer");
-	let refresh_button = WidgetHolder::new(Widget::TextButton(TextButton {
-		label: "Refresh Input".to_string(),
-		tooltip: "Refresh the artwork under the frame".to_string(),
-		on_update: WidgetCallback::new(|_| DocumentMessage::NodeGraphFrameGenerate.into()),
-		..Default::default()
-	}));
+	let refresh_button = TextButton::new()
+		.label("Refresh Input")
```
Keavon, line 157: Can a TextButton's label be part of its `new()` call? That seems inherent to its purpose, not optional.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -254,21 +247,19 @@ pub fn _transform_properties(document_node: &DocumentNode, node_id: NodeId, _con
...
 				WidgetHolder::unrelated_separator(),
-				WidgetHolder::new(Widget::NumberInput(NumberInput {
-					value: Some(vec2.x),
-					label: "X".into(),
-					unit: " px".into(),
-					on_update: update_value(move |number_input: &NumberInput| TaggedValue::DVec2(DVec2::new(number_input.value.unwrap(), vec2.y)), node_id, index),
-					..NumberInput::default()
-				})),
+				NumberInput::new()
+					.value(Some(vec2.x))
```
Keavon, line 251: Perhaps also `value` should be part of the `new()` constructor for NumberInputs? (Or all *Inputs).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -377,23 +363,21 @@ pub fn imaginate_properties(document_node: &DocumentNode, node_id: NodeId, conte
...
 			WidgetHolder::text_widget("Server"),
 			WidgetHolder::unrelated_separator(),
-			WidgetHolder::new(Widget::IconButton(IconButton {
-				size: 24,
-				icon: "Settings".into(),
-				tooltip: "Preferences: Imaginate".into(),
-				on_update: WidgetCallback::new(|_| DialogMessage::RequestPreferencesDialog.into()),
-				..Default::default()
-			})),
+			IconButton::new()
```
Keavon, line 366: And IconButton should probably take its icon name and size since those are mandatory. Likewise for IconLabel.

Keavon (approved):
This is definitely cleaner! I did a project search for `WidgetHolder` and it looks like we have many more files still using the old system. I assume this is just a beginning, and we can transition more files over time, correct? That seems like a good approach to me, if that's what you intended.

0HyperCube:
I don't really like the builder pattern, it just seems like a more convoluted method of setting fields than struct assignment.

0HyperCube:
> This is definitely cleaner! I did a project search for `WidgetHolder` and it looks like we have many more files still using the old system. I assume this is just a beginning, and we can transition more files over time, correct? That seems like a good approach to me, if that's what you intended.

Everything is becoming a node, we can cleanup the widgets (such as reducing the large amounts of code duplication) as we move things to be nodes.

Keavon:
> I don't really like the builder pattern, it just seems like a more convoluted method of setting fields than struct assignment.

Is that in response to my comments on the code, pertaining to using mandatory arguments in the `new()` call? If you don't care for that stylistically, feel free to disregard my comments about that and merge this now.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
0HyperCube: Fair enough, I'll change this.

PR #946, opened 2023-01-03
Change Tauri folder structure for 'tauri build' command

locriacyber:


Keavon:
Hi, what problem is this meant to solve? I'm pretty sure this undoes the changes we previously made to fix an existing problem (although my memory isn't fully certain of the details). Could you provide an explanation for what you're trying to do here? Thanks.

locriacyber:
when I run `tauri build`, the build files are generated in `dist`. But tauri bundles assets in `dist-tauri` (which is empty).

That creates an app that is empty as well.

Keavon:
Could I please have you read over the transcript for this conversation from a couple weeks ago, that was when we created the `tauri-dist` folder? https://discord.com/channels/731730685944922173/860830913390575646/1055404968175222784 I don't recall all the details and I don't have time to dig in right now (it's too late and I need sleep). Hopefully that conversation gives some context. And then if you can update me on the outcome, I'd appreciate it. Thanks!

locriacyber:
> Could I please have you read over the transcript for this conversation from a couple weeks ago, that was when we created the `tauri-dist` folder? https://discord.com/channels/731730685944922173/860830913390575646/1055404968175222784 I don't recall all the details and I don't have time to dig in right now (it's too late and I need sleep). Hopefully that conversation gives some context. And then if you can update me on the outcome, I'd appreciate it. Thanks!

This link (discard channel) is not public.

Keavon:
Please first join through https://discord.graphite.rs

locriacyber:
basically, tauri wasn't able to build because no `dist/` (likely `npm run build` is not called).

I don't think is was related, since `tauri build` now runs `npm run build` automatically

TrueDoctor:
> when I run `tauri build`, the build files are generated in `dist`. But tauri bundles assets in `dist-tauri` (which is empty).
> 
> That creates an app that is empty as well.
> 
> I also made so that the developer tool won't pop up on release mode.

That's actually the default behavior, we force enable the debug console by enabling a tauri feature, we could just disable that

PR #947, opened 2023-01-06
Remove direct dependency on ttf_parser

locriacyber:
Because ttf_parser updated, rustybuzz was using an older version.


Keavon:
It looks like tests failed, could you please fix that?

0HyperCube (approved):

PR #949, opened 2023-01-08
Groundwork for integrating Specta

Brendonovich:
Specta is a library for generating TypeScript bindings to your Rust code, initially developed for `rspc`. After some Discord conversations, it's been decided that Specta will likely replace the existing manual approach to defining backend types in TypeScript. This PR does a couple of things but none of them change existing functionality:

- Adds `specta::Type` implementations for a whole lot of types (should this be behind `cfg_attr` for crates like `graph-craft`?)
- Introduces a `Uuid` type which is just a u64 that becomes a `string` when serialized

A custom branch of Specta is being used right now to

1. Implement types for `glam`
2. Allow lossy type generation for numbers, this will not be the case once `Uuid` is adopted though

A lot more work on the frontend side will need to be done to get Specta fully integrated. If a svelte refactor is committed to, Specta integration may come after that.

Oh also workspace dependencies are really neat and I'd recommend you use them more.

(I don't think there's an issue for this)


editor/src/lib.rs
```diff
@@ -12,3 +12,9 @@ pub mod dispatcher;
 pub mod messages;
 pub mod test_utils;
 pub mod utility_traits;
+
+#[cfg(test)]
+#[test]
+fn export_types() {
+	specta::export::ts("../types.ts");
```
0HyperCube, line 19: Do we want to `#[ignore]` this text or is it fast enough?

Also we perhaps want to clearly document that you can run this test to regenerate the typescript.

editor/src/messages/layout/utility_types/layout_widget.rs
```diff
@@ -10,6 +10,7 @@ use crate::messages::layout::utility_types::misc::LayoutTarget;
 use crate::messages::prelude::*;
 
 use serde::{Deserialize, Serialize};
+use specta::Type;
```
0HyperCube, line 13: Is this import necessary?

0HyperCube (approved):
Looks good, thanks for going through all of those structs we have around.

editor/src/lib.rs
Brendonovich: I think it's ok to keep around, it takes almost no time on my machine. I'll add some documentation for it.
image

editor/src/messages/layout/utility_types/layout_widget.rs
Brendonovich: Nah it's a mistake

Brendonovich:
Before this gets merged it's probably worth deciding where the TS file should be generated to and how it will be imported.
Imo it should go to `editor/types.ts`, since if you use an npm workspace and create an `editor/package.json`, you could import the types with something like `import { FrontendMessage } from "@graphite/editor"`.

Keavon:
When we figure out what's going on with Svelte, that may necessitate an overhaul of the build system for the way in which wasm-bindgen generates its bindings, and this generates its types in an analogous way. The current build system in the Svelte branch isn't adequate, so if that proceeds, we'll need to do some work to stitch that together and this can be a consideration in that.

oscartbeaumont:
Another consideration to this PR is that Specta will store all the type information in the impls for the `Type` trait which may or may not be included in the final app despite serving no value at that point.

It's probably worth doing `#[cfg_attr(test, derive(specta::Type))]` to make sure all the type information is stripped from production builds. With that, you might also be able to get away with making `specta` a `dev-dependencies` so that its entire existence is stripped although I haven't tried this myself so your mileage may vary (this assumes Rust bundles unused dependencies which I could be wrong but I think it does by default).

If you have LTO enabled it should theoretically do this for you but less code for Rust to pass is always gonna be better.

document-legacy/src/color.rs
```diff
@@ -1,11 +1,12 @@
 use serde::{Deserialize, Serialize};
+use specta::Type;
```
Keavon, line 2: For consistency with everywhere else, let's not `use` it here.

editor/src/messages/preferences/preferences_message_handler.rs
```diff
@@ -1,8 +1,9 @@
 use crate::messages::prelude::*;
 
 use serde::{Deserialize, Serialize};
+use specta::Type;
```
Keavon, line 4: Likewise, here too.

editor/src/messages/message.rs
```diff
@@ -62,3 +62,7 @@ impl Message {
 		s.finish()
 	}
 }
+
+#[derive(specta::Type)]
+#[specta(inline, remote = "MessageDiscriminant")]
+pub struct MessageDiscriminantDef(u8);
```
Keavon, line 68: Could I have you please comment what this is/how it works/what it does?

Keavon (approved):
🚀

PR #953, opened 2023-01-09
Bezier-rs: Add bounding box, extrema, and inflection functions for subpath

linda-zheng:
Added `bounding_box`, `local_extrema` and `inflections` functions for the subpath class in bezier-rs.
Note, the current functions do not consider the points in between two bezier curves. We may want to consider these points for local extrema and inflection.

### DIFF
--- a/libraries/bezier-rs/src/subpath/solvers.rs
+++ b/libraries/bezier-rs/src/subpath/solvers.rs
@@ -63,37 +63,22 @@ impl Subpath {
 	/// The list of `t`-values returned are filtered such that they fall within the range `[0, 1]`.
 	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#subpath/local-extrema/solo" title="Local Extrema Demo"></iframe>
 	pub fn local_extrema(&self) -> [Vec<f64>; 2] {
-		let (mut x_extremas, mut y_extremas) = (Vec::new(), Vec::new());
 		let number_of_curves = self.len_segments() as f64;
 
-		for (index, bezier) in self.iter().enumerate() {
-			// Convert t-values of bezier curve to t-values of subpath
-			let extremas = bezier.local_extrema();
-			x_extremas.extend(extremas[0].iter().map(|t| ((index as f64) + t) / number_of_curves).collect::<Vec<f64>>());
-			y_extremas.extend(extremas[1].iter().map(|t| ((index as f64) + t) / number_of_curves).collect::<Vec<f64>>());
-		}
-
 		// TODO: Consider the shared point between adjacent beziers.
-		[x_extremas, y_extremas]
+		self.iter().enumerate().fold([Vec::new(), Vec::new()], |mut acc, elem| {
+			let extremas = elem.1.local_extrema();
+			// Convert t-values of bezier curve to t-values of subpath
+			acc[0].extend(extremas[0].iter().map(|t| ((elem.0 as f64) + t) / number_of_curves).collect::<Vec<f64>>());
+			acc[1].extend(extremas[1].iter().map(|t| ((elem.0 as f64) + t) / number_of_curves).collect::<Vec<f64>>());
+			acc
+		})
 	}
 
 	/// Return the min and max corners that represent the bounding box of the subpath.
 	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#subpath/bounding-box/solo" title="Bounding Box Demo"></iframe>
 	pub fn bounding_box(&self) -> Option<[DVec2; 2]> {
-		let bounding_boxes = self.iter().map(|bezier| bezier.bounding_box()).collect::<Vec<[DVec2; 2]>>();
-		if bounding_boxes.is_empty() {
-			// There is no bounding box for empty subpath.
-			None
-		} else {
-			let mut min_point = bounding_boxes[0][0];
-			let mut max_point = bounding_boxes[0][1];
-			for bounding_box in bounding_boxes {
-				// We take min of mins and max of maxes.
-				min_point = min_point.min(bounding_box[0]);
-				max_point = max_point.max(bounding_box[1]);
-			}
-			Some([min_point, max_point])
-		}
+		self.iter().map(|bezier| bezier.bounding_box()).reduce(|bbox1, bbox2| [bbox1[0].min(bbox2[0]), bbox1[1].max(bbox2[1])])
 	}
 
 	/// Returns list of `t`-values representing the inflection points of the subpath.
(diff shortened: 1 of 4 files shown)


libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -58,6 +58,64 @@ impl Subpath {
...
+		for (index, bezier) in self.iter().enumerate() {
+			// Convert t-values of bezier curve to t-values of subpath
+			let extremas = bezier.local_extrema();
+			x_extremas.extend(extremas[0].iter().map(|t| ((index as f64) + t) / number_of_curves)
+					.collect::<Vec<f64>>());
+			y_extremas.extend(extremas[1].iter().map(|t| ((index as f64) + t) / number_of_curves)
+					.collect::<Vec<f64>>());
+		}
+
+		// TODO: Consider the point between bezier curves.
```
hannahli2010, line 95: This might be more clear if you change the wording to: "Consider the shared point between adjacent beziers"

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -58,6 +58,64 @@ impl Subpath {
...
+				bezier
+					.inflections()
+					.into_iter()
+					// Convert t-values of bezier curve to t-values of subpath
+					.map(|t| ((index as f64) + t) / number_of_curves)
+					.collect::<Vec<f64>>()
+			})
+			.collect();
+		
+		// TODO: Consider the point between bezier curves.
```
hannahli2010, line 116: Same here as the previous comment

hannahli2010 (commented):
The UI demo looks good! After you fix the rust formatting, go ahead and open the PR.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -58,6 +58,64 @@ impl Subpath {
...
+
+	/// Return the min and max corners that represent the bounding box of the subpath.
+	pub fn bounding_box(&self) -> Option<[DVec2; 2]> {
+		let bounding_boxes = self.iter().map(|bezier| bezier.bounding_box()).collect::<Vec<[DVec2; 2]>>();
+		if bounding_boxes.is_empty() {
+			// There is no bounding box for empty subpath.
+			None
+		} else {
+			let mut min_points = bounding_boxes[0][0];
+			let mut max_points = bounding_boxes[0][1];
```
RobNadal, line 70: Why is this variable plural?

RobNadal (commented):
Rebase off master and fix merge conflicts (move the feature definitions to features folder)

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +59,62 @@ impl Subpath {
...
+		} else {
+			let mut min_point = bounding_boxes[0][0];
+			let mut max_point = bounding_boxes[0][1];
+			for bounding_box in bounding_boxes {
+				// We take min of mins and max of maxes.
+				min_point = min_point.min(bounding_box[0]);
+				max_point = max_point.max(bounding_box[1]);
+			}
+			Some([min_point, max_point])
+		}
```
Keavon, line 78: The iterator `fold()` method would be a more idiomatic approach.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +59,62 @@ impl Subpath {
...
+		let inflection_t_values: Vec<f64> = self
+			.iter()
+			.enumerate()
+			.flat_map(|(index, bezier)| {
+				bezier
+					.inflections()
+					.into_iter()
+					// Convert t-values of bezier curve to t-values of subpath
+					.map(|t| ((index as f64) + t) / number_of_curves)
+					.collect::<Vec<f64>>()
```
Keavon, line 111: @0HyperCube can this inner `collect` be avoided?

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +59,62 @@ impl Subpath {
 
 		intersection_t_values
 	}
+
+	/// Return the min and max corners that represent the bounding box of the subpath.
+	pub fn bounding_box(&self) -> Option<[DVec2; 2]> {
+		let bounding_boxes = self.iter().map(|bezier| bezier.bounding_box()).collect::<Vec<[DVec2; 2]>>();
+		if bounding_boxes.is_empty() {
+			// There is no bounding box for empty subpath.
+			None
```
Keavon, line 68: @0HyperCube do you think this should return `None` in this case, or `[(0, 0), (0, 0)]`?

Keavon (changes requested):
See code review. Thanks.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -59,6 +59,62 @@ impl Subpath {
...
+				max_point = max_point.max(bounding_box[1]);
+			}
+			Some([min_point, max_point])
+		}
+	}
+
+	/// Returns two lists of `t`-values representing the local extrema of the `x` and `y` parametric subpaths respectively.
+	/// The list of `t`-values returned are filtered such that they fall within the range `[0, 1]`.
+	pub fn local_extrema(&self) -> [Vec<f64>; 2] {
+		let (mut x_extremas, mut y_extremas) = (Vec::new(), Vec::new());
```
Keavon, line 84: Consider also use `.fold()` here too, but since this doesn't have the emptiness check if/else like above, it's less important to change this one.

libraries/bezier-rs/src/subpath/solvers.rs
0HyperCube: It doesn't really matter, I personally prefer `None` as it makes it clear to users of this function that there may not always be a bounding box (they can call `.unwrap_or_default()` if they want your functionality)

Keavon: Sounds good, thanks for your insight.

libraries/bezier-rs/src/subpath/solvers.rs
0HyperCube: Yes, you should be able to just return the iterator I think.

Keavon: I tried just deleting that last line (`.collect...`) but it didn't want to compile.

0HyperCube: The map statement has to `move` the index so that it is not dropped (but this is fine as usize is copy) - `.map(move |t| ((index as f64) + t) / number_of_curves)`

Keavon: Oh, I briefly tried putting a `move` in but I put it in the wrong place. Thanks.

Keavon:
Besides my comments, this also needs to have its conflicts resolved. Thanks.

Keavon:
After this is rebased, I can review it one last time and probably merge this time around.

Keavon (approved):

PR #955, opened 2023-01-10
Add self hosted runner to ci.yml

TrueDoctor:
Use Action runner on my server


TrueDoctor:
I think we might have some non deterministic testing errors exposed by the faster ci, thats something we should fix anyway, not sure if we have an issue for that, might be a good first contribution

Keavon:
I have definitely encountered them on occasion.

Keavon (approved):

PR #963, opened 2023-01-13
Selecting an individual anchor point deselects each of the other points

### LINKED ISSUE #722: Improve Path tool layer and anchor point selection behavior

ChrisMend19:
Improve Path tool layer and anchor point selection behavior. Closes #722 
- Selecting an individual already-selected anchor point deselects all but the clicked one


Keavon:
It is no longer possible to drag multiple selected shapes.

Existing desired behavior in master:

https://user-images.githubusercontent.com/4388688/215388479-9c963930-3497-4331-bae4-f829f482bb66.mp4

New undesired behavior currently in this branch:

https://user-images.githubusercontent.com/4388688/215388641-c6a09a8c-6eff-498a-939a-cb6c02568b67.mp4

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -201,12 +205,19 @@ impl Fsm for PathToolFsmState {
 							if toggle_add_to_selection {
 								responses.push_back(DocumentMessage::AddSelectedLayers { additional_layers: intersection }.into());
 							} else {
+								// Selects the topmost layer when selecting intersecting shapes
```
Keavon, line 208: This doesn't seem to be true. When I test it with two intersecting shapes, both end up being selected.

Question for @TrueDoctor: is it even necessary to pick the top one, or would the occlusion normally be taken into account? (In other words, for selecting shapes with the Select tool, does that have to be manually filtered for what's on top like the code here?)

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -187,7 +190,8 @@ impl Fsm for PathToolFsmState {
 						let include_handles = tool_data.shape_editor.selected_layers_ref();
 						tool_data.snap_manager.add_all_document_handles(document, input, &include_handles, &[], &selected_points.points);
 
-						tool_data.drag_start_pos = input.mouse.position - selected_points.offset;
+						tool_data.drag_start_pos = input.mouse.position;
+						tool_data.previous_mouse_position = input.mouse.position - selected_points.offset;
```
Keavon, line 194: Is `previous_mouse_position` an accurate and descriptive name for this? I'm having trouble seeing how the current mouse position minus the selected points offset is the "previous mouse position". To me, that implies the position where the mouse was located at the moment the drag began-- except that's what the previous line does when it sets `drag_start_pos`. Is there a better name for `previous_mouse_position`?

editor/src/messages/tool/tool_messages/path_tool.rs
TrueDoctor: > This doesn't seem to be true. When I test it with two intersecting shapes, both end up being selected. Please fix that so only the top one is selected.
This is probably caused by the SetSelectedLayers message on line 210 I would guess

As for your question, I'm not sure about that would have to look that up

TrueDoctor (approved):
LGTM

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -187,6 +188,11 @@ impl ShapeEditor {
 		self.iter(document).flat_map(|shape| shape.manipulator_groups().iter())
 	}
 
+	// Sets the selected points to all points for the corresponding intersection
+	pub fn select_all_anchors(&self, responses: &mut VecDeque<Message>, itersections: Vec<u64>) {
```
0HyperCube, line 192: Note that anchors are not the same as manipulator points - anchors are just the square widgets whereas manipulator points include both the circles and the squares.

editor/src/messages/tool/tool_messages/path_tool.rs
0HyperCube: It is the mouse position subtract any snapping offset (which is usually nothing) - so I would say `previous_mouse_position` is a reasonable name.

0HyperCube (approved):
Looks good.

editor/src/messages/tool/common_functionality/shape_editor.rs
ChrisMend19: Thanks for clarifying. The name uses anchors, however the body of the function calls the operation SelectAllManipulatorPoints. Would this need to be changed to some other operation or function that selects all anchors since the issue's description mentions anchors not manipulator points

Keavon: Just from testing this branch's behavior, it looks like when you use the Path tool to click the filled region of a shape, only the anchors do get selected, not the handles. So from a behavior point of view, does the correct thing, assuming I understand correctly what's being talked about here. If any names of functions or messages don't reflect their functionality, please feel empowered to rename those. (In general, as you run across code, always don't hesitate to improve code quality along the way, as long as it's not massively disruptive amounting to a full refactor itself— a separate branch for really big changes would be preferred, while the same branch is fine for smaller changes.)

ChrisMend19: Ok thanks! I believe it's just a renaming of the operation I mentioned above. I made it at the start of this branch.

Keavon (approved):
QA testing looks good.

PR #967, opened 2023-01-14
Bezier-rs: Updated Bezier function signatures to accept TValue

RobNadal:
Updated Bezier function signatures to take in a `ComputeType`. Affected Functions:
- Evaluate
- tangent
- curvature
- normal
- split
- trim
- De Casteljau Points


libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -45,28 +22,54 @@ impl Bezier {
...
 				assert!((0.0..=1.).contains(&t));
-				self.unrestricted_parametric_evaluate(t)
+				t
 			}
-			ComputeType::Euclidean(t) => {
-				assert!((0.0..=1.).contains(&t));
-				self.unrestricted_euclidean_evaluate(t, 0.0001)
+			ComputeType::Euclidean(distance) => {
+				assert!((0.0..=1.).contains(&distance));
+				self.euclidean_to_parametric(distance, 0.0001)
```
hannahli2010, line 38: Maybe assign this error value as a constant.

website/other/bezier-rs-demos/src/features/bezier-features.ts
```diff
@@ -111,54 +111,58 @@ const bezierFeatures = {
...
 			Quadratic: {
-				sliderOptions: [tSliderOptions],
+				sliderOptions: [cSliderOptions],
 			},
 		},
+		chooseComputeType: true,
 	},
 	Trim: {
-		callback: (bezier: WasmBezierInstance, options: Record<string, number>): string => bezier.trim(options.t1, options.t2),
+		callback: (bezier: WasmBezierInstance, options: Record<string, number>, _: undefined, computeType: ComputeType): string => bezier.trim(options.c2, options.c1, computeType),
```
hannahli2010, line 153: Is there a reason why `c2` is passed before `c1`?

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -45,28 +22,54 @@ impl Bezier {
...
+			BezierHandles::Quadratic { handle } => squared_one_minus_t * self.start + 2. * one_minus_t * t * handle + t_squared * self.end,
+			BezierHandles::Cubic { handle_start, handle_end } => {
+				let t_cubed = t_squared * t;
+				let cubed_one_minus_t = squared_one_minus_t * one_minus_t;
+				cubed_one_minus_t * self.start + 3. * squared_one_minus_t * t * handle_start + 3. * one_minus_t * t_squared * handle_end + t_cubed * self.end
 			}
 		}
 	}
 
+	/// Calculate the point on the curve based on the `t`-value provided.
```
hannahli2010, line 66: Change comment to `c`

hannahli2010 (commented):
Can you remove the comment on line 248 in `bezier/solvers.rs`?

website/other/bezier-rs-demos/src/features/bezier-features.ts
RobNadal: good catch

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -4,39 +4,16 @@ use super::*;
...
-	}
-
-	/// Calculate the point along the curve that is a factor of `d` away from the start.
-	pub(crate) fn unrestricted_euclidean_evaluate(&self, d: f64, error: f64) -> DVec2 {
-		if let BezierHandles::Linear = self.handles {
-			return self.unrestricted_parametric_evaluate(d);
-		}
-
+	/// Convert a euclidean distance along the `Bezier` to a parametric `t`-value.
+	pub fn euclidean_to_parametric(&self, d: f64, error: f64) -> f64 {
```
Keavon, line 8: Does "euclidean distance" range from 0 to the arc length of the curve, or from 0 to 1? If it's 0 to 1, please call this a t-value not a d-value. "d" would imply actual distance (arc length) not a time ranging from 0 to 1, calculated in either euclidean or parametric space depending on context.

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -45,28 +22,54 @@ impl Bezier {
...
+
+		mid
 	}
 
-	/// Calculate the point on the curve based on the `t`-value provided.
-	/// Expects `t` to be within the inclusive range `[0, 1]`.
-	pub fn evaluate(&self, t: ComputeType) -> DVec2 {
-		match t {
+	/// Convert a [ComputeType] to a parametric `t`-value.
+	pub(crate) fn compute_type_to_parametric(&self, c: ComputeType) -> f64 {
```
Keavon, line 30: Rename `c` to `compute_type` since we avoid cryptic variable names.

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -45,28 +22,54 @@ impl Bezier {
...
-				self.unrestricted_euclidean_evaluate(t, epsilon)
+			ComputeType::EuclideanWithinError { distance, epsilon } => {
+				assert!((0.0..=1.).contains(&distance));
+				self.euclidean_to_parametric(distance, epsilon)
 			}
 		}
 	}
 
+	/// Calculate the point on the curve based on the `t`-value provided.
+	pub(crate) fn unrestricted_parametric_evaluate(&self, t: f64) -> DVec2 {
```
Keavon, line 48: This name that appears in many functions, "unrestricted", doesn't have a clear meaning to me. Can you suggest some alternatives in this comment's discussion so we can settle on a better name throughout the code?

RobNadal: We use `unrestricted` for functions which may perform some computation or return some value in which the `t`-values used are not bounded by [0., 1.]. We could convert this naming convention to `unbounded` if that is preferred,

libraries/bezier-rs/src/bezier/lookup.rs
RobNadal: Our library and Bezier.js always use the argument `t` to represent specifically the parametric case, so I'm not fond of using `t` or `ComputeT` since `ComputeType` has the Euclidean variant which means something very different. I've added documentation now to the ComputeType enum which I hope makes the distinction more clear.

In general, I don't love the name `ComputeType` either, but we called it this because Parametric and Euclidean are each methods to **compute** a point along the curve. We could go a different route and name the Enum / instances something like `PointRef`  / `p` since the data essentially references a point on the curve, if you like that better, or anything you had in mind.

libraries/bezier-rs/src/bezier/lookup.rs
RobNadal: I'm worried that naming the Euclidean argument to `t` we conflate it with the parametric concept of `t` we've been using elsewhere. You're right though that distance was the wrong name, because the variable is representation a ratio of the total distance along the curve - for that reason, I'd lean towards calling it `r`, `ratio` or `distance_ratio`.

RobNadal: For this function alone, I've designated the variable as `ratio`, with an appropriate comment. Elsewhere I've used `t` for consistency as per our discussion on discord. Let me know if that work.

libraries/bezier-rs/src/bezier/lookup.rs
RobNadal: I've renamed this Enum to `TValue`. This has the unfortunate side effect of bloating the MR somewhat, as many variables in the frontend and wasm layers had to be renamed appropriately.

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -49,11 +50,12 @@ impl Bezier {
 		}
 	}
 
-	/// Returns the Bezier curve representing the sub-curve starting at the point corresponding to `t1` and ending at the point corresponding to `t2`.
-	pub fn trim(&self, t1: f64, t2: f64) -> Bezier {
+	/// Returns the Bezier curve representing the sub-curve starting at the point `c1` and ending at the point `c2` along the curve.
```
Keavon, line 53: Please specify in this doc comment if it's okay to give c1 and c2 in any order, or if one has to be bigger than the other.

Keavon (approved):

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -1,72 +1,75 @@
...
+			BezierHandles::Cubic { handle_start, handle_end } => {
+				let t_cubed = t_squared * t;
+				let cubed_one_minus_t = squared_one_minus_t * one_minus_t;
+				cubed_one_minus_t * self.start + 3. * squared_one_minus_t * t * handle_start + 3. * one_minus_t * t_squared * handle_end + t_cubed * self.end
 			}
 		}
 	}
 
+	/// Calculate the coordinates of the point `t` along the curve.
+	/// Expects `t` or `distance` argument enclosed within `c` to be within the inclusive range `[0, 1]`.
```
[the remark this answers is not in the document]
RobNadal: Good catch. This is a remnant from a previous comment that didn't get removed properly.

PR #968, opened 2023-01-14
Bezier-rs: Add hash URLs for individual API examples

hannahli2010:
Handle displaying a single demo at a time using `hash`

Ex. `#bezier_intersectQuadratic_solo` will render only the example pane showing the "Intersect (Quadratic)" example for beziers

### DIFF
--- a/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
+++ b/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
@@ -46,10 +46,10 @@ class BezierDemoPane extends HTMLElement implements DemoPane {
 	computeType!: ComputeType;
 
 	connectedCallback(): void {
-		this.id = `${Math.random()}`.substring(2);
 		this.computeType = "Parametric";
 
 		this.key = (this.getAttribute("name") || "") as BezierFeatureKey;
+		this.id = `bezier_${this.key}`;
 		this.name = bezierFeatures[this.key].name;
 		this.demoOptions = JSON.parse(this.getAttribute("demoOptions") || "[]");
 		this.triggerOnMouseMove = this.getAttribute("triggerOnMouseMove") === "true";
--- a/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
+++ b/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
@@ -1,4 +1,4 @@
-import { BezierFeatureName } from "@/features/bezier-features";
+import bezierFeatures, { BezierFeatureKey } from "@/features/bezier-features";
 import { renderDemoPane } from "@/utils/render";
 import { BezierCurveType, BEZIER_CURVE_TYPE, ComputeType, BezierDemoOptions, SliderOption, Demo, DemoPane, BezierDemoArgs } from "@/utils/types";
 
@@ -28,7 +28,9 @@ const demoDefaults = {
 
 class BezierDemoPane extends HTMLElement implements DemoPane {
 	// Props
-	name!: BezierFeatureName;
+	key!: BezierFeatureKey;
+
+	name!: string;
 
 	demoOptions!: BezierDemoOptions;
 
@@ -47,7 +49,8 @@ class BezierDemoPane extends HTMLElement implements DemoPane {
 		this.id = `${Math.random()}`.substring(2);
 		this.computeType = "Parametric";
 
-		this.name = (this.getAttribute("name") || "") as BezierFeatureName;
+		this.key = (this.getAttribute("name") || "") as BezierFeatureKey;
+		this.name = bezierFeatures[this.key].name;
 		this.demoOptions = JSON.parse(this.getAttribute("demoOptions") || "[]");
 		this.triggerOnMouseMove = this.getAttribute("triggerOnMouseMove") === "true";
 		this.chooseComputeType = this.getAttribute("chooseComputeType") === "true";
@@ -74,7 +77,7 @@ class BezierDemoPane extends HTMLElement implements DemoPane {
 		const bezierDemo = document.createElement("bezier-demo");
 		bezierDemo.setAttribute("title", demo.title);
 		bezierDemo.setAttribute("points", JSON.stringify(demo.points));
-		bezierDemo.setAttribute("name", this.name);
+		bezierDemo.setAttribute("key", this.key);
 		bezierDemo.setAttribute("sliderOptions", JSON.stringify(demo.sliderOptions));
 		bezierDemo.setAttribute("triggerOnMouseMove", String(this.triggerOnMouseMove));
 		bezierDemo.setAttribute("computetype", this.computeType);
--- a/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
+++ b/website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
@@ -49,7 +49,7 @@ class BezierDemoPane extends HTMLElement implements DemoPane {
 		this.computeType = "Parametric";
 
 		this.key = (this.getAttribute("name") || "") as BezierFeatureKey;
-		this.id = `bezier_${this.key}`;
+		this.id = `bezier/${this.key}`;
 		this.name = bezierFeatures[this.key].name;
 		this.demoOptions = JSON.parse(this.getAttribute("demoOptions") || "[]");
 		this.triggerOnMouseMove = this.getAttribute("triggerOnMouseMove") === "true";
--- a/website/other/bezier-rs-demos/src/main.ts
+++ b/website/other/bezier-rs-demos/src/main.ts
@@ -48,7 +48,7 @@ function renderSubpathPane(featureName: SubpathFeatureKey, container: HTMLElemen
 
 function isUrlSolo(url: string): boolean {
 	const hash = url.split("#")?.[1];
-	const splitHash = hash?.split("_");
+	const splitHash = hash?.split("/");
 	return splitHash?.length === 3 && splitHash?.[2] === "solo";
 }
 
@@ -56,7 +56,7 @@ window.addEventListener("hashchange", (e: Event): void => {
 	const hashChangeEvent = e as HashChangeEvent;
 	const isOldHashSolo = isUrlSolo(hashChangeEvent.oldURL);
 	const isNewHashSolo = isUrlSolo(hashChangeEvent.newURL);
-	const target = document.querySelector(window.location.hash);
+	const target = document.getElementById(window.location.hash.substring(1));
 	// Determine whether the page needs to recompute which examples to show
 	if (!target || isOldHashSolo !== isNewHashSolo) {
 		renderExamples();
@@ -65,7 +65,7 @@ window.addEventListener("hashchange", (e: Event): void => {
 
 function renderExamples(): void {
 	const hash = window.location.hash;
-	const splitHash = hash.split("_");
+	const splitHash = hash.split("/");
 
 	// Determine which examples to render based on hash
 	if (splitHash[0] === "#bezier" && splitHash[1] in bezierFeatures && splitHash[2] === "solo") {
@@ -98,7 +98,7 @@ function renderExamples(): void {
 
 	// Scroll to specified hash if it exists
 	if (hash) {
-		const target = document.querySelector(hash);
+		const target = document.getElementById(hash.substring(1));
 		if (target) {
 			target.scrollIntoView();
 		}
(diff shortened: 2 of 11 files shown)


website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,69 @@ declare global {
...
+}
+
+const pathname = window.location.pathname;
+const splitPathName = pathname.split("/");
+
+// Render based on pathname
+if (splitPathName[1] === "bezier" && splitPathName[2] in bezierFeatures) {
+	window.document.body.innerHTML = `
+  <div id="bezier-demos"></div>
+  `.trim();
```
RobNadal, line 56: Why do we need to trim this? Can we just stick this on a single line? (Here and below)

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,69 @@ declare global {
...
+	renderBezierPane(splitPathName[2] as BezierFeatureKey, document.getElementById("bezier-demos"));
+} else if (splitPathName[1] === "subpath" && splitPathName[2] in subpathFeatures) {
+	window.document.body.innerHTML = `
+  <div id="subpath-demos"></div>
+  `.trim();
+	renderSubpathPane(splitPathName[2] as SubpathFeatureKey, document.getElementById("subpath-demos"));
+} else if (pathname !== "/") {
+	window.location.pathname = "/";
```
RobNadal, line 64: What's the purpose / reasoning behind this case?

RobNadal (approved):
Works really well - awesome job @hannahli2010!

website/other/bezier-rs-demos/src/main.ts
hannahli2010: No particular reason tbh, I can remove that.

website/other/bezier-rs-demos/src/main.ts
hannahli2010: I think prettier was giving me trouble but I'll try again

hannahli2010: When it's one line, eslint changes it to either `"<div·id=\"subpath-demos\"></div>"` or `'<div·id="subpath-demos"></div>'` and it is mad at both.

Keavon: There's a fix for that in the main project's eslintrc file, try and copy that over.

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,63 @@ declare global {
...
 	demo.setAttribute("sliderOptions", JSON.stringify(feature.sliderOptions || []));
 	demo.setAttribute("triggerOnMouseMove", String(feature.triggerOnMouseMove));
 	demo.setAttribute("chooseComputeType", String(feature.chooseComputeType));
-	subpathDemos?.append(demo);
-});
+	container?.append(demo);
+}
+
+const pathname = window.location.pathname;
+const splitPathName = pathname.split("/");
```
Keavon, line 50: You want to use `window.location.hash`.

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,63 @@ declare global {
...
+  <p>
+    This is the interactive documentation for the <a href="https://crates.io/crates/bezier-rs"><b>Bezier-rs</b></a> library. View the
+    <a href="https://docs.rs/bezier-rs/latest/bezier_rs">crate documentation</a>
+    for detailed function descriptions and API usage. Click and drag on the endpoints of the demo curves to visualize the various Bezier utilities and functions.
+  </p>
+  
+  <h2>Beziers</h2>
+  <div id="bezier-demos"></div>
+  <h2>Subpaths</h2>
+  <div id="subpath-demos"></div>
```
Keavon, line 71: Feature request in to be added to the demos: include a `#` link after the name header for each demo so it can be clicked to change the URL hash fragment to that. Similar to many blogs that have that for different headings within an article. (In the respective demo file(s), not these four highlighted lines, of course.)

website/other/bezier-rs-demos/src/main.ts
hannahli2010: To implement the behaviour that I desire here, I intend to use `pathname` not `hash`. This part of the code is used to determine which of the examples to render, whereas `hash` will be used to jump to one of the rendered example.

To further explain the behaviour here: This will parse a url (ex. `http://localhost:8080/bezier/trim`) and then render only the specified example (ex. the bezier trim). This is intended to facilitate embedding a single example in an iframe for the rust docs.

image

Keavon: I'm confused about the difference between the two things you're referring to. In your "bezier trim" example, what is the "bezier" part?

website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
```diff
@@ -46,10 +46,10 @@ class BezierDemoPane extends HTMLElement implements DemoPane {
 	computeType!: ComputeType;
 
 	connectedCallback(): void {
-		this.id = `${Math.random()}`.substring(2);
 		this.computeType = "Parametric";
 
 		this.key = (this.getAttribute("name") || "") as BezierFeatureKey;
+		this.id = `bezier_${this.key}`;
```
Keavon, line 52: Is this unique across the three: linear, quadratic, cubic?

website/other/bezier-rs-demos/src/main.ts
hannahli2010: "bezier" is used to differentiate between showing the beizer vs subpath versions of a function

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -77,3 +77,10 @@ if (splitPathName[1] === "bezier" && splitPathName[2] in bezierFeatures) {
 	(Object.keys(bezierFeatures) as BezierFeatureKey[]).forEach((feature) => renderBezierPane(feature, bezierDemos));
 	(Object.keys(subpathFeatures) as SubpathFeatureKey[]).forEach((feature) => renderSubpathPane(feature, subpathDemos));
 }
+
+// Scroll to specified hash if it exists
+const hash = window.location.hash;
+const target = document.querySelector(hash);
+if (target) {
+	target.scrollIntoView();
+}
```
Keavon, line 86: The browser already does this for you. But we want to hide all the others, not scroll to the specific one, if a valid hash fragment is present. And that should be achieved with CSS: https://developer.mozilla.org/en-US/docs/Web/CSS/:target

website/other/bezier-rs-demos/src/components/BezierDemoPane.ts
Keavon: Oh, this is `BezierDemoPane` not `BezierDemo`.

hannahli2010: Yup! The key is different for each (`intersectLinear`, `intersectQuadratic`, and `intersectCubic`).

website/other/bezier-rs-demos/src/main.ts
Keavon: I see. Those will both have to be part of the same hash fragment then. Routing based on the actual URL requires special treatment from the web server, which we can't do for the deployed website. It just so happens that Vue/Webpack is letting you visit the page (I'm not actually sure why it's configured to let you do that), but that's not true in production once the site is built and not served through Webpack's dev server.

hannahli2010: My intentions are that `localhost:8080/` will render all the examples for both beziers and subpaths. On this page, the hash will be used to jump to each example (ex. `localhost:8080/#bezier_trim`). If we want the page to render only one example, this can be done by going to the corresponding url with the following pattern: `localhost:8080/:bezierOrSubpath/:exampleKey` where `beizerOrSubpath` is either "bezier" or "subpath, and `exampleKey` is the corresponding example key in the `bezierFeatures` and `subpathFeatures` files.

hannahli2010: Based on your other comment, there seems to be a difference between what I intended vs what you were requesting.

Are you sure that you want a url with a hash to hide all the other examples? It seems like common practice for this is to scroll to the entity, while keeping the rest of the page the same.

Keavon: Yeah, see above about why we can't do that, sorry. If you want to make a routing scheme, you can implement that with the hash fragment only. Perhaps something like `#bezier/trim` and `#bezier/trim/solo` could be used for normal navigation vs. soloing a single entry for the iframe. You'll need your own logic to achieve that latter case by checking the end of the hash fragment and hiding all the others in that case. I suggested using CSS's `:target` selector earlier but for performance reasons I'd actually suggest filtering the list down to the one so we aren't initializing then hiding all the others unnecessarily, when separate instances of the page will be open dozens of times on the Rustdoc page inside many iframes.

website/other/bezier-rs-demos/src/main.ts
Keavon: Disregard the idea about using CSS for performance concerns.

website/other/bezier-rs-demos/src/main.ts
hannahli2010: Ah I see. I'll make that change!

Also oops, I didn't see your previous comment when writing mine - I hadn't refreshed the page yet.

website/other/bezier-rs-demos/src/main.ts
hannahli2010: The browser doesn't seem to be doing the scroll for me when going to a url with the hash directly and including this if statement solves it for me. I'm guessing that this might be because the examples are conditionally render.

Keavon: Try copying the URL, closing the tab, then opening a new tab and pasting the URL. I believe browsers don't re-scroll each time you hit enter in the URL bar or reload the page, it only scrolls when the hash changes or the page first loads.

hannahli2010: It wasn't working when the page first loads for me.

Keavon: Likely necessary because those elements aren't in the DOM yet.

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,92 @@ declare global {
...
 	demo.setAttribute("sliderOptions", JSON.stringify(feature.sliderOptions || []));
 	demo.setAttribute("triggerOnMouseMove", String(feature.triggerOnMouseMove));
 	demo.setAttribute("chooseComputeType", String(feature.chooseComputeType));
-	subpathDemos?.append(demo);
+	container?.append(demo);
+}
+
+function isUrlSolo(url: string): boolean {
+	const hash = url.split("#")?.[1];
+	const splitHash = hash?.split("_");
```
Keavon, line 51: Please use a slash instead of an underscore. Also, instead of the camelCase IDs, it would be nice to use kebab-case. That probably means changing the keys in you bezier-features.ts and subpath-features.ts.

website/other/bezier-rs-demos/src/main.ts
```diff
@@ -32,31 +17,92 @@ declare global {
...
+      This is the interactive documentation for the <a href="https://crates.io/crates/bezier-rs"><b>Bezier-rs</b></a> library. View the
+      <a href="https://docs.rs/bezier-rs/latest/bezier_rs">crate documentation</a>
+      for detailed function descriptions and API usage. Click and drag on the endpoints of the demo curves to visualize the various Bezier utilities and functions.
+    </p>
+    
+    <h2>Beziers</h2>
+    <div id="bezier-demos"></div>
+    <h2>Subpaths</h2>
+    <div id="subpath-demos"></div>
+    `.trim();
```
Keavon, line 90: This should use tabs not spaces to indent.

website/other/bezier-rs-demos/src/style.css
```diff
@@ -1,8 +1,11 @@
 html,
 body {
-	height: 100%;
+	height: auto;
```
Keavon, line 3: You can just remove this height line.

Keavon (approved):

website/other/bezier-rs-demos/src/main.ts
hannahli2010: The current url format is: `<beizer | subpath>_<idInCamelCase>` with an optional `<_solo>` at the end. If we change everything to be separated by a hyphen, it'll be harder to see the id in the url. For example: `bezier-bezier-through-points` looks a little weird since the word bezier is repeated. I just want to confirm that this is ok.

hannahli2010: Oh oops, sorry i misread that!

PR #977, opened 2023-01-18
Incremental compilation and stable node IDs

### LINKED ISSUE #996: Incremental compilation

TrueDoctor:
Restructures the node storage to a tree based representation that allows for incremental compilation via the usage of stable node ids

Closes #996


document-legacy/src/operation.rs
```diff
@@ -274,23 +274,375 @@ pub enum Operation {
...
+			SetManipulatorHandleMirroring { layer_path, id, mirror_angle } => {
+				52.hash(state);
+				layer_path.hash(state);
+				id.hash(state);
+				mirror_angle.hash(state);
+			}
+			SetSelectedHandleMirroring { layer_path, toggle_angle } => {
+				53.hash(state);
+				layer_path.hash(state);
+				toggle_angle.hash(state);
```
0HyperCube, line 644: The large amount of boilerplate is perhaps something we should file an issue for?

TrueDoctor: I think for now I'll just remove that commit and put it into its own pr

node-graph/graph-craft/src/proto.rs
```diff
@@ -378,8 +425,30 @@ mod test {
...
+		let ids: Vec<_> = construction_network.nodes.iter().map(|(id, _)| *id).collect();
+		assert_eq!(
+			ids,
+			vec![
+				17495035641492238530,
+				14931179783740213471,
+				2268573767208263092,
+				14616574692620381527,
+				12110007198416821768,
+				11185814750012198757
```
0HyperCube, line 445: > The internal algorithm is not specified, and so it and its hashes should not be relied upon over releases.

From std::collections::hash_map::DefaultHasher docs. Perhaps we could test another way to avoid the test breaking in new rust versions.

0HyperCube (commented):
It is great to see the node graph code becoming simpler :)

node-graph/graph-craft/src/proto.rs
TrueDoctor: That is true but the test is not that helpful anyway because I never actually checked the ids are correct, it is mainly there to detect changes. Not sure how to best test the desired behavior. I guess we could compare the nodes for different node networks to check if it behaves correctly

Keavon (approved):

PR #979, opened 2023-01-19
Update UI when deleting last point of shape with Path tool

auto-kad:
The Operation::DeleteSelectedManipulatorPoints re-routes the deletion logic to Operation::DeleteLayer when you delete the last point in a shape, which updates the Layer Tree structure in the frontend. After it's done deleting it emits a new DocumentResponse, DeletedSelectedManipulatorPoints, and when handled by process_message() in document_message_handler.rs, clears the Properties panel.

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -143,7 +143,7 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
 							}
 							DocumentResponse::DocumentChanged => responses.push_back(RenderDocument.into()),
 							DocumentResponse::DeletedSelectedManipulatorPoints => {
-								// clear Properties panel after deleting all points
+								// Clear Properties panel after deleting all points
 								responses.push_back(
 									FrontendMessage::UpdatePropertyPanelOptionsLayout {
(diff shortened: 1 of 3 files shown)


editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -143,6 +143,29 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+									FrontendMessage::UpdatePropertyPanelSectionsLayout {
+										layout_target: LayoutTarget::PropertiesSections,
+										diff: vec![WidgetDiff {
+											widget_path: vec![],
+											new_value: DiffUpdate::SubLayout(vec![]),
+										}],
+									}
+									.into(),
+								);
+							}
```
0HyperCube, line 168: This doesn't update the backend's widget state, which may cause bugs with the widget diffing. The standard way to clear a panel is with the `SendLayout` message, although this isn't currently very obvious due to lack of documentation (which we are aiming to improve).
```rs
responses.push_back(
	LayoutMessage::SendLayout {
		layout: Layout::WidgetLayout(WidgetLayout::new(vec![])),
		layout_target: LayoutTarget::[your layout target],
	}
	.into(),
);
```

0HyperCube (approved):
Thanks for the contribution, the code looks good.

PR #981, opened 2023-01-19
Replace Webpack with Vite as the frontend build system

locriacyber:
Use vite instead of webpack. Including changes to satisfy vite and browser's es module loader.


Keavon:
Superseded by the https://github.com/GraphiteEditor/Graphite/commits/vite-old branch to be used as a reference. Now that the switch to Svelte is live, we'll be following a guide and template for a fresh approach to the migration from Webpack to Vite.

PR #982, opened 2023-01-19
Svelte: Fix part of MenuBarInput

locriacyber:


frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
```diff
@@ -35,15 +35,13 @@
...
+			entries = entries;
+		}
 		else throw new Error("The menu bar floating menu has no associated ref");
 	}
 
-	function unFocusEntry(menuListEntry: MenuListEntry, e: FocusEvent) {
-		const blurTarget = (e.target as HTMLElement | undefined)?.closest("[data-menu-bar-input]");
-		if (blurTarget !== self && menuListEntry.ref) menuListEntry.ref.isOpen = false;
-	}
-
```
Keavon, line 46: Could you please explain your thoughts behind removing this? Why are you calling it dead code?

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
```diff
@@ -76,11 +74,10 @@
 </script>
 
 <div class="menu-bar-input" bind:this={self} data-menu-bar-input>
-	{#each entries as entry, index (index)}
+	{#each entries as entry, index}
```
Keavon, line 77: So I can learn Svelte better, is this redundant or is there some other reason not to do this? I'm used to keys being mandatory in Vue.

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
```diff
@@ -95,7 +92,7 @@
 				{/if}
 			</div>
 			{#if entry.children && entry.children.length > 0}
-				<MenuList open={entry.ref?.menuIsOpen() || false} entries={entry.children || []} direction="Bottom" minWidth={240} drawIcon={true} bind:this={entry.ref} />
+				<MenuList on:open={(e) => entry.ref.isOpen=e.detail} open={entry.ref?.isOpen || false} entries={entry.children || []} direction="Bottom" minWidth={240} drawIcon={true} bind:this={entry.ref} />
```
Keavon, line 95: Did `menuIsOpen()` not work? And how is it possible to directly get the `.isOpen` value from the component? I thought the component has to be configured to compile with a getter/setter for that to work.

Could you please fix the formatting by surrounding the `=` with spaces? Thanks.

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
locriacyber: There is no point of using a key in svelte. There is no vdom, so the whole item would get replaced.

Use (keyed) only if the item inside #each won't change, but only the props to the components.

There is no point of using array index as key at all.

In svelte, only use keyed each loop if performance is really bad.

Keavon: I see, thanks for the explanation. I'll clean up all the cases where we use an index then over the coming days.

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
locriacyber: It just works like this. The component object is behind a Proxy, so get/set will trigger update on that `export let`

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
Keavon: On this highlighted line, you can remove the `, index` part also since it's not used within the HTML, right?

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
locriacyber: It's not working. After removing this, it works as intended.

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
locriacyber: yes

locriacyber: You don't need to clean up other cases. It's a waste of time with no clear benefit.

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
Keavon: I did some further digging in the Vue version and, as far as I can tell, this does seem to be dead code

frontend-svelte/src/components/widgets/inputs/MenuBarInput.svelte
Keavon: I did some further digging in the Vue version and, as far as I can tell, this does seem to be dead code.

Keavon (approved):

PR #990, opened 2023-01-26
Add scaling by nudging with the Alt key

### LINKED ISSUE #930: Alt+Arrows should nudge to resize the bottom/right edges of layers
Currently we can nudge with the arrow keys, which translates the entire layer by 1px or 10px (while holding <kbd>Shift</kbd>). This feature calls for nudging the bottom or right edge, resulting in resizing the layer (pivoting around the top-left corner) instead of translating it, when the user is holding the <kbd>Alt</kbd> modifier in addition to tapping the arrows. Furthermore, adding also the <kbd>Ctrl</kbd> modifier should make it instead scale the top or left sides (pivoting around the bottom-right corner).

Further details (original issue description):

Similar to arrow keys (and the <kbd>Shift</kbd> modifier) to nudge the position of selected layer(s), this feature should "nudge" (by resizing) the right and bottom sides of the selected layer(s). Left and right arrows resize the layer(s) by 1px on the right edge. Up and down arrows resize the layer(s) by 1px on the bottom edge. The <kbd>Shift</kbd> makes it 10px. ~~This should happen in layer space, not viewport space, so regardless of viewport zoom the user should expect to see the dimensions in the Properties panel change by 1px or 10px in W or H. If the layer is rotated, it should scale along the axis of the layer, so the X and Y would change as expected but the W and H would not necessarily change by 1px/10px.~~ The bounding box should be resized, meaning if the layer is rotated, it can become skewed. Assuming it doesn't conflict with any other keys, <kbd>Ctrl</kbd> would be nice as a modifier that changes it to the Left/Top instead of Bottom/Right.

Sahra-Zhou:
- Added Alt+arrow key mapping
- Resizing feature has not been implemented yet
- Closes #930


Sahra-Zhou:
Closes #930

editor/src/messages/input_mapper/default_mapping.rs
```diff
@@ -57,6 +57,7 @@ pub fn default_mapping() -> Mapping {
 		entry!(KeyDown(Rmb); action_dispatch=SelectToolMessage::Abort),
 		entry!(KeyDown(Escape); action_dispatch=SelectToolMessage::Abort),
 		//
+
```
Keavon, line 60: Remove extraneous line addition

editor/src/messages/input_mapper/default_mapping.rs
```diff
@@ -233,31 +234,33 @@ pub fn default_mapping() -> Mapping {
...
+		entry!(KeyDown(ArrowLeft); modifiers=[ArrowUp], action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: -NUDGE_AMOUNT, delta_y: -NUDGE_AMOUNT, alt: Alt, ctrl: Control }),
+		entry!(KeyDown(ArrowLeft); modifiers=[ArrowDown], action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: -NUDGE_AMOUNT, delta_y: NUDGE_AMOUNT, alt: Alt, ctrl: Control }),
+		entry!(KeyDown(ArrowLeft); action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: -NUDGE_AMOUNT, delta_y: 0., alt: Alt, ctrl: Control }),
+		entry!(KeyDown(ArrowRight); modifiers=[ArrowUp], action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: NUDGE_AMOUNT, delta_y: -NUDGE_AMOUNT, alt: Alt, ctrl: Control }),
+		entry!(KeyDown(ArrowRight); modifiers=[ArrowDown], action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: NUDGE_AMOUNT, delta_y: NUDGE_AMOUNT, alt: Alt, ctrl: Control }),
+		entry!(KeyDown(ArrowRight); action_dispatch=DocumentMessage::NudgeResizeSelectedLayer { delta_x: NUDGE_AMOUNT, delta_y: 0., alt: Alt, ctrl: Control }),
 		//
+		// alt+arrow key to resize the bottom/right edges
+
```
Keavon, line 263: Remove extraneous additions

editor/src/messages/portfolio/document/document_message.rs
```diff
@@ -111,9 +112,11 @@ pub enum DocumentMessage {
 		layer_path: Vec<LayerId>,
 		node_path: Vec<NodeId>,
 	},
-	NudgeSelectedLayers {
+	NudgeResizeSelectedLayer {
 		delta_x: f64,
 		delta_y: f64,
+		alt: Key,
+		ctrl: Key,
```
Keavon, line 119: Rename `alt` to `resize` and `ctrl` to `resize_opposite_corner`.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -11,6 +11,7 @@ use crate::messages::layout::utility_types::widgets::button_widgets::{IconButton
 use crate::messages::layout::utility_types::widgets::input_widgets::{
 	DropdownEntryData, DropdownInput, NumberInput, NumberInputIncrementBehavior, NumberInputMode, OptionalInput, RadioEntryData, RadioInput,
 };
+
```
Keavon, line 14: Remove extraneous line addition

editor/src/messages/portfolio/document/document_message.rs
```diff
@@ -111,9 +112,11 @@ pub enum DocumentMessage {
 		layer_path: Vec<LayerId>,
 		node_path: Vec<NodeId>,
 	},
-	NudgeSelectedLayers {
+	NudgeResizeSelectedLayer {
```
Keavon, line 115: You can undo this rename, `NudgeSelectedLayers` is fine for this feature.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,13 +552,40 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
 				self.backup(responses);
+				let keyboard = &ipp.keyboard;
+				let is_scale = keyboard.key(alt);
+				let is_bottom_right = !keyboard.key(ctrl);
 				for path in self.selected_layers().map(|path| path.to_vec()) {
-					let operation = DocumentOperation::TransformLayerInViewport {
-						path,
-						transform: DAffine2::from_translation((delta_x, delta_y).into()).to_cols_array(),
-					};
+					let mut transform_matrix = DAffine2::from_translation((delta_x, delta_y).into()).to_cols_array();
```
Keavon, line 561: Mutating this variable can be avoided. And it's inefficient to assume this is a translation before overwriting it with a scale. I'd suggest using the `else` case instead.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,13 +552,40 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+								DAffine2::from_translation((-existing_bounds[1].x, -existing_bounds[1].y).into())
+							};
+							let scale = DAffine2::from_scale((scale_size_x, scale_size_y).into());
+							let offset_back = if is_bottom_right {
+								DAffine2::from_translation((existing_bounds[0].x, existing_bounds[0].y).into())
+							} else {
+								DAffine2::from_translation((existing_bounds[1].x, existing_bounds[1].y).into())
+							};
+							transform_matrix = (offset_back * scale * offset).to_cols_array();
+						}
```
Keavon, line 586: I'm finding this is extremely difficult to read. Could you please start by adding some breathing room with blank lines in all the places they can be used to provide separation between different concepts?

And then here are some general tips you can aim to employ which can reduce the complexity of this code:

- Avoid repeating the `if is_bottom_right` test four separate times.
- By returning a tuple from inside an if statement, then destructuring it outside, you can handle the X and Y values together.
- There is a high degree of code duplication. In your `let scale_size_x = if ... else ...;` and `let scale_size_y = if ... else ...;` expressions, the **only** difference is the `-` and `+` character between the if and else conditions. You should only need to write out that long math expression once and feed a variable that determined the `-` or `+` value.
- `DAffine2::from_scale((scale_size_x, scale_size_y).into());` is equivalent to `DAffine2::from_translation(-existing_bounds[0])`.
- Destructure `existing_bounds` into `[existing_top_left, existing_bottom_right]` for clarity so you can avoid writing `[0]` and `[1]`.
- Reverse the order of `let offset = ...` and `let scale = ...` so you are handling scale separately from offset instead of interweaving those concepts.
- Those are all the tips I can name for now, but try your very best to use equivalent logic to make this code as simple and readable as possible, please.

Keavon (changes requested):
The feature is looking good! There is a bug in an edge case: if you try to resize it down to zero, it behaves strangely. Please fix this by preventing the scaling of the layer to be less than 1px wide or 1px tall. If the nudge would cause it to be less than 1px (or 0 or even a negative number), reset it back to 1px wide or tall.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+				let is_bottom_right = !keyboard.key(resize_opposite_corner);
 				for path in self.selected_layers().map(|path| path.to_vec()) {
-					let operation = DocumentOperation::TransformLayerInViewport {
-						path,
-						transform: DAffine2::from_translation((delta_x, delta_y).into()).to_cols_array(),
-					};
-					responses.push_back(operation.into());
+					if is_scale {
+						if let Ok(Some(existing_bounds)) = self.document_legacy.viewport_bounding_box(&path, &render_data) {
+							let [existing_top_left, existing_bottom_right] = [existing_bounds[0], existing_bounds[1]];
```
Keavon, line 567: `[existing_bounds[0], existing_bounds[1]]` is redundant, you're converting `existing_bounds` into itself.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+							let new_width = if is_bottom_right { width + delta_x } else { width - delta_x };
+							let new_height = if is_bottom_right { height + delta_y } else { height - delta_y };
+							let scale_size_x = if new_width > 1. { new_width / width } else { 1. / width };
+							let scale_size_y = if new_height > 1. { new_height / height } else { 1. / height };
+							let offset = if is_bottom_right {
+								DAffine2::from_translation(-existing_top_left)
+							} else {
+								DAffine2::from_translation(-existing_bottom_right)
+							};
+							let scale = DAffine2::from_scale((scale_size_x, scale_size_y).into());
```
Keavon, line 579: Requesting again that `offset` and `scale` are swapped so you aren't interweaving steps.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
 				for path in self.selected_layers().map(|path| path.to_vec()) {
-					let operation = DocumentOperation::TransformLayerInViewport {
-						path,
-						transform: DAffine2::from_translation((delta_x, delta_y).into()).to_cols_array(),
-					};
-					responses.push_back(operation.into());
+					if is_scale {
+						if let Ok(Some(existing_bounds)) = self.document_legacy.viewport_bounding_box(&path, &render_data) {
+							let [existing_top_left, existing_bottom_right] = [existing_bounds[0], existing_bounds[1]];
+							let width = existing_bottom_right.x - existing_top_left.x;
```
Keavon, line 568: Please break apart separate logical groupings of lines with blank lines, it's very hard to read without some breathing room.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+							let height = existing_bottom_right.y - existing_top_left.y;
+							let new_width = if is_bottom_right { width + delta_x } else { width - delta_x };
```
Keavon, line 570: For each of these, all you need to do is factor out the if statement into `let sign = if is_bottom_right { 1 } else { -1 };`

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+							};
+							let scale = DAffine2::from_scale((scale_size_x, scale_size_y).into());
+							let offset_back = if is_bottom_right {
+								DAffine2::from_translation(existing_top_left)
+							} else {
+								DAffine2::from_translation(existing_bottom_right)
+							};
+							let operation = DocumentOperation::TransformLayerInViewport {
+								path,
+								transform: (offset_back * scale * offset).to_cols_array(),
```
Keavon, line 587: Replace `offset_back` with `offset.inverse()`.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -551,14 +551,50 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
+						if let Ok(Some(existing_bounds)) = self.document_legacy.viewport_bounding_box(&path, &render_data) {
+							let [existing_top_left, existing_bottom_right] = [existing_bounds[0], existing_bounds[1]];
+							let width = existing_bottom_right.x - existing_top_left.x;
+							let height = existing_bottom_right.y - existing_top_left.y;
+							let new_width = if is_bottom_right { width + delta_x } else { width - delta_x };
+							let new_height = if is_bottom_right { height + delta_y } else { height - delta_y };
+							let scale_size_x = if new_width > 1. { new_width / width } else { 1. / width };
+							let scale_size_y = if new_height > 1. { new_height / height } else { 1. / height };
+							let offset = if is_bottom_right {
+								DAffine2::from_translation(-existing_top_left)
```
Keavon, line 575: Move the `DAffine2::from_translation()` wrapper outside the if statement.

Keavon (changes requested):
I've added more tips to simplify. But this isn't an exhaustive list. Please keep thinking about every way you can possibly make it even simpler and try those, since it still has a sizable chunk of complexity that can be cut out. This is a good exercise to learn Rust in the process.

Keavon (approved):

PR #993, opened 2023-01-27
Reduce tool code duplication and boilerplate

0HyperCube:
- Hints now don't take up hundreds of lines.
- Widgets are updated using new helper methods
- Dispatching tool fsm state changes is not duplicated.


Keavon (approved):
Awesome improvement. Burninate the boilerplate!

PR #1001, opened 2023-01-30
Update Imaginate to output bitmap data to the graph via Image Frame node

### LINKED ISSUE #992: Add an Image Frame node and make the graph output vector data
Now that images are nodes, the next highest priority item is making bitmap data into vector data with an Image Frame node (that creates an `<image />` tag), and making the whole node graph export vector data rather than bitmap data. (Only one vector element can output for now.)

The graph's Input node has a secondary output value called "Transform" which exposes the value of the layer's Transform. Here's what the default graphs look like for an Image and an Imaginate:

image

The Transform input to the Imaginate node, if hooked up, allows the node to automatically pick its image generation dimensions (by rounding the Transform input's scale to the nearest 64px). It is also used to set the transform of the frame used in rasterization of a referenced mask layer.

If the Transform is not provided, the user must enter the image generation Dimensions field (its Edit checkbox becomes disabled so the W and H fields are set to editable regardless of that checkbox state). We should also invert that checkbox so instead of "enable the checkbox to manually edit", it's "enable the checkbox to to use the auto value". So "auto" is the default (checked), opposite of how it is now. I just need to find an icon that conveys "auto" instead of "edit" (the current ✏️ icon).

0HyperCube:
Closes #992

The bulk of this PR involves allowing nodes to have multiple outputs. This required some frontend Vue modifications as well as some work with the backend, ensuring all node inputs and network outputs have an index. A document graph rewriting step called `duplicate_outputs` was added to flatten the multiple outputs.

The node graph now outputs an ImageFrame struct that contains the image data and a transform. Further information can be easily added.  The transform is the local transform of the layer, and the resulting transform is applied to the layer. This means that if you disconnect the transform, the layer will resize to have an identity transform.

Modifications to the imaginate tool have not yet been added.

### DIFF
--- a/editor/src/lib.rs
+++ b/editor/src/lib.rs
@@ -11,5 +11,6 @@ pub mod application;
 pub mod consts;
 pub mod dispatcher;
 pub mod messages;
+pub mod node_graph_executor;
 pub mod test_utils;
 pub mod utility_traits;
--- a/editor/src/messages/portfolio/document/document_message.rs
+++ b/editor/src/messages/portfolio/document/document_message.rs
@@ -107,6 +107,7 @@ pub enum DocumentMessage {
 	},
 	NodeGraphFrameImaginateRandom {
 		imaginate_node: Vec<NodeId>,
+		then_generate: bool,
 	},
 	NodeGraphFrameImaginateTerminate {
 		layer_path: Vec<LayerId>,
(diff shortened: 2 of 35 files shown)


Keavon:
This seems to be crashing when I import an image (dragging in a PNG file to the viewport). Are you seeing that on your end?

Keavon (approved):

PR #1004, opened 2023-02-01
Fix text overflow in text/radio buttons

locriacyber:
[use-gradient.webm](https://user-images.githubusercontent.com/74560659/216037658-25ff3bb7-9a46-44fa-bb18-c996ed0a8306.webm)

### DIFF
--- a/frontend-svelte/src/components/widgets/buttons/TextButton.svelte
+++ b/frontend-svelte/src/components/widgets/buttons/TextButton.svelte
@@ -42,6 +42,7 @@
 		justify-content: center;
 		align-items: center;
 		flex: 0 0 auto;
+		width: 0;
 		height: 24px;
 		margin: 0;
 		padding: 0 8px;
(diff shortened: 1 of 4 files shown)


frontend-svelte/src/components/widgets/buttons/TextButton.svelte
```diff
@@ -38,6 +38,12 @@
 
 <style lang="scss" global>
 	.text-button {
+		span {
+			overflow: hidden;
+			text-overflow: ellipsis;
+		}
+		width: 0;
```
Keavon, line 45: Move down to the line above `height: 24px;`

frontend-svelte/src/components/widgets/buttons/TextButton.svelte
```diff
@@ -38,6 +38,12 @@
 
 <style lang="scss" global>
 	.text-button {
+		span {
+			overflow: hidden;
+			text-overflow: ellipsis;
+		}
```
Keavon, line 44: Move down to be just above the `.icon-label {` line (with one blank line above and below as standard).

Keavon (commented):
I still have to test it out in the frontend to make sure the UX is good. Please replicate changes to the Vue version. Thanks!!

Keavon (approved):

PR #1006, opened 2023-02-03
Bezier-rs: Add trim for Subpath

hannahli2010:
- Implement trim for Subpath
  - Used same colour styling as the UI for split


libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -87,11 +94,178 @@ impl Subpath {
...
+			// Add a duplicate manipulator group so that the returned Subpath is still a valid Bezier
+			let mut point = new_manipulator_groups[0];
+			if point.in_handle.is_some() {
+				point.in_handle = Some(point.anchor);
+			}
+			if point.out_handle.is_some() {
+				point.out_handle = Some(point.anchor);
+			}
+			new_manipulator_groups = vec![point, point];
+		}
```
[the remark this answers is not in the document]
hannahli2010: Should any handles on the point be removed?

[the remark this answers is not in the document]
RobNadal: I wouldn't expect there to be any handles, so you could remove it I think.

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -75,13 +86,184 @@ impl Subpath {
...
+	/// Returns a [Subpath] with a reversed winding order.
+	pub fn reverse(&self) -> Subpath {
+		Subpath {
+			manipulator_groups: Subpath::reverse_manipulator_groups(&self.manipulator_groups),
+			closed: self.closed,
+		}
+	}
+
+	/// Returns an open [Subpath] that results from trimming the original Subpath between the points corresponding to `t1` and `t2`.
+	/// If `t1` > `t2`, then behavior for computing the resulting Subpath will differ depending on whether the original Subpath is open or closed:
```
Keavon, line 112: I'm thinking this should be kept consistent to avoid confusion, so the Subpath returned by this function will always have the same winding order as the original. That means, in the case when `t1` > `t2`, swapping `t1` and `t2` so it behaves the same as if the user called this function with the arguments reversed. Then change the last line of these doc comments to say that, more generally, if the reverse direction is desired from the winding order of the original curve, to call `Subpath::reverse`.

(Please also read over this function's doc comments, which I updated for clarity, to confirm my claims are accurate.)

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -75,13 +86,184 @@ impl Subpath {
...
+			let mut point = new_manipulator_groups[0];
+			point.in_handle = None;
+			point.out_handle = None;
+			return Subpath {
+				manipulator_groups: vec![point],
+				closed: false,
+			};
+		}
+
+		let len_new_manip_groups = new_manipulator_groups.len();
```
Keavon, line 228: The awesome comments above seem to have stopped here, would you mind completing them through the end of the function?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -75,13 +86,184 @@ impl Subpath {
...
+		Subpath {
+			manipulator_groups: Subpath::reverse_manipulator_groups(&self.manipulator_groups),
+			closed: self.closed,
+		}
+	}
+
+	/// Returns an open [Subpath] that results from trimming the original Subpath between the points corresponding to `t1` and `t2`.
+	/// If `t1` > `t2`, then behavior for computing the resulting Subpath will differ depending on whether the original Subpath is open or closed:
+	/// - If the original Subpath was open, an open subpath from `t1` to `t2` will be returned (with a winding order reverse that of the original Subpath).
+	/// - If the original Subpath was closed, an open subpath from `t1` to `t2` that crosses the break between 1 and 0 will be returned (winding in the same direction as the original Subpath from `t1` to `t = 1 = 0` to `t2`).
```
Keavon, line 114: Upon reading your tests, I see you wisely distinguished between closed subpaths which have their start and end points at the same, and different, locations. I hadn't considered that when I wrote this `t = 1 = 0` part of this comment, which I assume may be incorrect due to that fact. Please update this for me so it's correct again, assuming I did actually state an inaccuracy. Thanks.

Keavon (changes requested):
Good to go after resolving the request. Loving the very clean code!

libraries/bezier-rs/src/subpath/transform.rs
hannahli2010: The current behavior for the open subpath case is consistent with the trim behavior for a `Bezier`. Should we change it to match what you're suggesting here as well?

libraries/bezier-rs/src/subpath/transform.rs
Keavon: I mean you have tests for closed subpaths where the start and end points are located in the same position. And others where they are located in different positions (but it's still closed).

libraries/bezier-rs/src/subpath/transform.rs
Keavon: I think that's wise. In both a segment and a subpath, treat it agnostic of argument order. People can call reverse if they want it reversed. This results in the least unexpected behavior.

libraries/bezier-rs/src/subpath/transform.rs
hannahli2010: Which tests are you referring to?

Keavon: Never mind, maybe I misread the tests. But even if I misread them, they still reminded me of the two cases where a closed subpath might, or might not, have endpoints at the same position. Please make sure you handle/account for that (and have tests for it). See https://files.keavon.com/-/SandyFancyYak/capture.png (note: each row has identical geometry on each side, with only the closed flag changed per side).

hannahli2010: I want to make sure we're on the same page about the behavior for these cases.

In our code, a subpath has an associated list of manipulator groups. When closing one, we simply assume that there is an additional bezier between the first and last elements of the manipulator group list.

In the first row of your above image, the open subpath has a start and end point at the exact same location. If the closed flag is enabled for this subpath, our code will keep the "duplicate" manipulator group at the end of the list, and treat it as if there's a bezier that travels between them. I.e. there's an extra bezier which is essentially a single point.

If you wanted the closed code to not have that additional bezier, the user would have to manually remove the extra manipulator group at the end.

Is this what you would expect?
Should we add a `close_subpath` function that will both enable the `closed` flag and remove the extra manipulator group if necessary?

libraries/bezier-rs/src/subpath/transform.rs
hannahli2010: Just to confirm: For a open subpath + a single segment, we perform the trim agnostic of argument order. Does this apply for a closed subpath as well? Or do we return the curve that crosses the break when `t1 > t2`?

RobNadal: We need to provide some mechanism to trim closed subpaths across the break. I think the current implementation for closed subpaths respects the 'winding' order that @Keavon described

libraries/bezier-rs/src/subpath/transform.rs
Keavon: I confused myself, thank you for breaking this down to clarify my mental misconceptions. Everything you described is correct. Your `close_subpath` idea would indeed be a useful function doing as you described. So with that in mind, I think we can disregard anything I asked about in this whole code review comment's thread (but just confirm the doc comment is accurate after my possibly flawed changes). Thanks.

Keavon: I confused myself, thank you for breaking this down to clarify my mental misconceptions. Everything you described is correct. Your `close_subpath` idea would indeed be a useful function doing as you described. So with that in mind, I think we can disregard anything I asked about in this whole code review comment's thread (but just confirm the doc comment is accurate after my possibly flawed changes). Thanks.

Keavon: I confused myself, thank you for breaking this down to clarify my mental misconceptions. Everything you described is correct. Your `close_subpath` idea would indeed be a useful function doing as you described. So with that in mind, I think we can disregard anything I asked about in this whole code review comment's thread (but just confirm the doc comment is accurate after my possibly flawed changes). Thanks.

libraries/bezier-rs/src/subpath/transform.rs
Keavon: Oh yes, argument order would matter for closed subpaths so it decides which way to go around the closed shapes. But in all cases, it would follow the existing winding order, that's what I meant when saying it was agnostic of argument order (which is an accurate description for bezier segments, and for open subpaths).

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -78,13 +89,190 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+		let are_arguments_reversed = t1_curve_index > t2_curve_index || (t1_curve_index == t2_curve_index && t1_curve_t > t2_curve_t);
+		if !self.closed && are_arguments_reversed {
+			(t1_curve_index, t2_curve_index) = (t2_curve_index, t1_curve_index);
+			(t1_curve_t, t2_curve_t) = (t2_curve_t, t1_curve_t);
+		}
+
+		// Get a new list from the manipulator groups that will be trimmed at the ends to form the resulting subpath.
+		// The list will contain enough manipulator groups such that the later code simply needs to trim the first and last bezier segments
+		// and then update the values of the corresponding first and last manipulator groups accordingly.
+		let mut cloned_manipulator_groups = self.manipulator_groups.clone();
```
hannahli2010, line 155: Hmm, does this need to include a call to `ManipulatorGroupId::new()` for each of these cloned manipulator groups?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -78,13 +89,190 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+			closed: self.closed,
+		}
+	}
+
+	/// Returns an open [Subpath] that results from trimming the original Subpath between the points corresponding to `t1` and `t2`.
+	/// If `t1` > `t2`, then behavior for computing the resulting Subpath will differ depending on whether the original Subpath is open or closed:
+	/// - If the original Subpath was open, an open subpath from `t2` to `t1` will be returned (with winding in the same direction as the original Subpath).
+	/// - If the original Subpath was closed, an open subpath from `t1` to `t2` that crosses the break between 1 and 0 will be returned (winding in the same direction as the original Subpath from `t1` to `t = 1 = 0` to `t2`).
+	///
+	/// If a path winding in the reverse direction is desired, call `trim` on the `Subpath` returned from `Subpath::reverse`.
```
Keavon, line 120:
```suggestion
	/// Returns an open [Subpath] that results from trimming the original Subpath between the points corresponding to `t1` and `t2`, maintaining the winding order of the original.
	/// If the original Subpath is closed, the order of arguments does matter.
	/// The resulting Subpath will wind from the given `t1` to `t2`.
	/// That means, if the value of `t1` > `t2`, it will cross the break between endpoints from `t1` to `t = 1 = 0` to `t2`.
	/// If a path winding in the reverse direction is desired, call `trim` on the `Subpath` returned from `Subpath::reverse`.
```

Keavon (approved):

libraries/bezier-rs/src/subpath/transform.rs
0HyperCube: No, it doesn't need to regenerate the id, as the id only has to be unique in the subpath.

Depending on how the vector nodes work, we may also want to avoiding regenerating the id of manipulator groups that are unchanged by the splitting operation (for example in `reverse_manipulator_groups` above).

This would allow the node network `Move Manipulator Points By Id --> Split Subpath` to be rearranged into `Split Subpath --> Move Manipulator Points By Id` which would give the same results when the moving and splitting modified different manipulator groups. I'm unsure if this is needed though @Keavon and @otdavies?

If we do not care about the consistency of ids, then we can regenerate them as part of the clone implementation on the `ManipulatorGroupId` type.

hannahli2010: Based on the discussion on discord, I'm assuming we do not need to regenerate these ids and will merge the PR as it is now.

PR #1009, opened 2023-02-05
Restore handle length when Shift is released while dragging opposite handle with Path tool

### LINKED ISSUE #818: Restore length of handle after releasing Shift for mirroring handle lengths with Path tool
Right now with the Path tool, if you hold <kbd>Shift</kbd> it will mirror the length of the handle on the opposite side of the handle you're currently dragging. But when you release the key, it leaves it at the current length. To prevent an accidental loss of information, it should snap back to the length right before the key was first held down.

It would also be nice to have the Shift key create the second handle on the opposite side if you are dragging a handle on an anchor with only one handle. Releasing Shift should restore it back to a single-handle anchor.

ShouvikGhosh2048:
Closes #818


Keavon:
It looks like your approach is to store the original length on the `ManipulatorGroup`'s `editor_state` field (a `ManipulatorGroupEditorState` struct). Above that field, we have a comment:
> // TODO: Remove this from Graphene, editor state should be stored in the frontend if possible.

(Graphene is the node graph data backend)

Basically we have some leftover code for that one existing boolean, which I apologize for probably looking like this would make for a nice solution. (And it is! It's just that we're unfortunately needing to deprecate that approach since it goes against our separation of concerns.)

Since the Path tool is manipulating the shape, it is the job of the Path tool to store the state for the length. So you will actually want to move your `previous_opposing_handle_length: Option<f64>,` to `struct PathToolData` in `path_tool.rs`. Then you can also nicely avoid needing the code which plumbs through a new set of messages to reach the document and its vector data.

Sorry for the confusion! Hopefully you find that this ends up being a more straightforward solution once the change is made.

And while you are at it, feel free (if you wish) to also move the `mirror_angle_between_handles: bool,` field out of `struct ManipulatorGroupEditorState` so that struct can be deleted as the TODO comment requests.

Keavon:
For this to work with multiple selected handles, you probably need to store it in some kind of collection (vec, hashmap, or something like that).

Keavon:
Can I have you please comment on #818 so I can assign that issue to you? Thanks.

otdavies:
@ShouvikGhosh2048 If you need any guidance on this, feel free to reach out :)

ShouvikGhosh2048:
I made the shift from `ManipulatorGroup`'s `editor_state` to `PathToolData`. I wanted to know if my current approach of passing data works / any improvements I can make.

I'll add comments to the functions and shift the `mirror_angle_between_handles` before the final review.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -110,6 +111,7 @@ struct PathToolData {
 
 	drag_start_pos: DVec2,
 	alt_debounce: bool,
+	opposing_handle_lengths: Option<HashMap<Vec<LayerId>, HashMap<u64, f64>>>,
```
Keavon, line 114: This looks correct now, that's the right place to store this state.

document-legacy/src/operation.rs
```diff
@@ -138,6 +139,7 @@ pub enum Operation {
 		layer_path: Vec<LayerId>,
 		delta: (f64, f64),
 		mirror_distance: bool,
+		reset_opposing_handle_lengths: Option<HashMap<u64, f64>>,
```
Keavon, line 142: But I don't think you should need to plumb any new arguments through in the operations. Can't you just call one (or perhaps multiple) existing operations to set the positions of the handle(s) to the intended locations, just like if the user had dragged them?

Keavon:
> I'll add comments to the functions and shift the `mirror_angle_between_handles` before the final review.

Comments would be good, thanks! I realized that asking you to also refactor out `mirror_angle_between_handles` would be unwise since that will be more complex (we'll have to find a robust algorithm to derive its value in the editor). Since this code will be replaced in a few months with the new vector data representation used in the node graph, let's keep `mirror_angle_between_handles` where it is now.

Keavon:
CC @otdavies.

document-legacy/src/operation.rs
otdavies: I agree with Keavon here. I would suggest creating a new message type that triggers the handle position reset.

Keavon: I wasn't suggesting making a new type of operation. I was suggesting using the existing choice of operations that are already used to drag handles around with the user's cursor and use that to set the position to the coordinates determined within the logic of the Path tool code.

Keavon: Ah, I hadn't checked that. Can you make a new operation which specifies which points (based on index or something)? Because soon, we'll be replacing the entire document backend with a more data-centric approach so highly specialized messages like the reset one you added won't make sense in that data-centric implementation.

ShouvikGhosh2048: Should I make a new operation or replace/modify `MoveSelectedManipulatorPoints`?

Keavon: Probably a new one.

ShouvikGhosh2048: I'm sorry, there's a `MoveManipulatorPoint` message that can be used. I don't have to make a new message.

Keavon: Awesome! What's the status of this now? Is it functional and I should test it out and then give it another code review? If so, please mark it as not a draft ("ready for review" button).

ShouvikGhosh2048: I have a few things to check, once done I'll mark it for review.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -140,6 +142,7 @@ impl Fsm for PathToolFsmState {
 						tool_data.overlay_renderer.render_subpath_overlays(&document.document_legacy, layer_path.to_vec(), responses);
 					}
 
+					tool_data.opposing_handle_lengths = None;
```
ShouvikGhosh2048, line 145: I reset the `opposing_handle_lengths` for multiple messages (like when the selected points change). I'm not 100% sure whether I've covered all the correct messages.

node-graph/gcore/src/vector/manipulator_group.rs
```diff
@@ -146,6 +146,7 @@ impl ManipulatorGroup {
 
 		// If the anchor isn't selected, but both handles are, drag only handles
 		if self.both_handles_selected() {
+			self.editor_state.mirror_angle_between_handles = false;
```
ShouvikGhosh2048, line 149: This line is to cancel handle angle mirroring when both handles are dragged without the anchor - is this line enough or do I have to make other changes?

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -206,6 +206,114 @@ impl ShapeEditor {
...
+			.filter_map(|(path, shape)| shape.as_subpath().map(|subpath| (path, subpath)))
+			.filter_map(|(path, shape)| opposing_handle_lengths.get(path).map(|layer_opposing_handle_lengths| (path, shape, layer_opposing_handle_lengths)))
+			.flat_map(|(path, shape, layer_opposing_handle_lengths)| {
+				shape
+					.manipulator_groups()
+					.enumerate()
+					.map(move |(id, manipulator_group)| (path, layer_opposing_handle_lengths, id, manipulator_group))
+			})
+			.for_each(|(path, layer_opposing_handle_lengths, id, manipulator_group)| {
+				if !manipulator_group.editor_state.mirror_angle_between_handles {
```
ShouvikGhosh2048, line 266: So I've added checks to recheck whether we have to mirror angles and whether the anchor is selected - should I keep them?

node-graph/gcore/src/vector/manipulator_group.rs
otdavies: Does the behavior of the tool reflect the change correctly? If so it is likely all that needs to be done. Test selecting both handles & the anchor and dragging as well. That is another situation to consider.

otdavies (commented):
A few more comments added - when you get a chance @ShouvikGhosh2048

node-graph/gcore/src/vector/subpath.rs
```diff
@@ -289,6 +296,11 @@ impl Subpath {
 		self.manipulator_groups_mut().iter_mut().filter(|manipulator_group| manipulator_group.any_points_selected())
 	}
 
+	/// Return all the selected [ManipulatorPoint]s by mutable reference along with id
+	pub fn selected_manipulator_groups_any_points_mut_with_id(&mut self) -> impl Iterator<Item = (&u64, &mut ManipulatorGroup)> {
```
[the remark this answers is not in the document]
ShouvikGhosh2048: I've removed this function in the last few commits.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -193,16 +221,31 @@ impl ShapeEditor {
...
-		for layer_path in &self.selected_layers {
-			responses.push_back(
-				DocumentMessage::MoveSelectedManipulatorPoints {
-					layer_path: layer_path.clone(),
-					delta: (delta.x, delta.y),
-					mirror_distance,
-				}
-				.into(),
-			);
+	pub fn move_selected_points(&self, delta: DVec2, mirror_distance: bool, reset_opposing_handle_lengths: Option<HashMap<Vec<LayerId>, HashMap<u64, f64>>>, responses: &mut VecDeque<Message>) {
```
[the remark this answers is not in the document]
ShouvikGhosh2048: I have changed this function - I now pass `opposing_handle_lengths` to the `reset_opposing_handle_lengths` function which uses the `MoveManipulatorPoint` message.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -206,6 +206,114 @@ impl ShapeEditor {
...
+			.iter()
+			.filter_map(|path| document.layer(path).ok().map(|layer| (path, layer)))
+			.filter_map(|(path, shape)| shape.as_subpath().map(|subpath| (path, subpath)))
+			.map(|(path, shape)| {
+				let opposing_handle_lengths = shape
+					.manipulator_groups()
+					.enumerate()
+					.filter_map(|(id, manipulator_group)| {
+						// We will keep track of the opposing handle length when:
+						// i) Both handles exist and exactly one is selected.
```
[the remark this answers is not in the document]
ShouvikGhosh2048: The comment refers to a single manipulator group (an anchor and the handles attached to it): we track the opposing handle length for a manipulator group if - 1) it has both handles and exactly one handle is selected, 2) the anchor is not selected and 3) we have to mirror the angle between the handles.

The Vec<LayerId> refers to a layer path. The u64 refers to the id of a manipulator group within the layer. We can select multiple manipulator groups within a single layer, and we can do this for multiple different layers at the same time.

editor/src/messages/tool/common_functionality/shape_editor.rs
ShouvikGhosh2048: As mentioned in the `opposing_handle_lengths` function, we track the opposing length when certain conditions are met - like `mirror_angle_between_handles` being true.

The reason I had the check was because I was unsure if these properties could change without updating `opposing_handle_lengths` in `tool_data`. These might be unnecessary.

editor/src/messages/tool/common_functionality/shape_editor.rs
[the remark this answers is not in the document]
otdavies: Ah, this makes more sense then. Carry on!

otdavies (approved):
As long as this behaves correctly when @Keavon tests it, I am happy.

editor/src/messages/tool/common_functionality/shape_editor.rs
0HyperCube: This check probably shouldn't fail but it is fine to leave it in anyway just to make sure.

editor/src/messages/tool/tool_messages/path_tool.rs
0HyperCube: Looks fine to me.

0HyperCube (approved):
Looks good :)

Keavon (approved):
Works great!

PR #1010, opened 2023-02-06
Replace operation pseudo hash with safe variant

TrueDoctor:


Keavon:
Status of this @TrueDoctor?

TrueDoctor:
This is a pretty ugly soulution and probably not worth to rebase

PR #1027, opened 2023-02-14
Bezier-rs: Add SubpathTValue and euclidean parameterization for subpaths

RobNadal:


libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -10,6 +12,65 @@ impl Subpath {
...
+			let length_ratio = length / total_length;
+			if accumulator <= global_t && global_t <= accumulator + length_ratio {
+				return (index, (global_t - accumulator) / length_ratio);
+			}
+			accumulator += length_ratio;
+		}
+		(0, 0.)
+	}
+
+	/// Convert a [SubpathTValue] to a parametric `(segment_index, t)` tuple.
```
hannahli2010, line 30: Can you add to this comment to assert that the returned value will not refer to a non-existent segment index?

libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -34,6 +93,8 @@ impl Subpath {
 
 #[cfg(test)]
 mod tests {
+	use crate::{consts::MAX_ABSOLUTE_DIFFERENCE, utils::f64_compare};
```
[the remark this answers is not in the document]
Keavon: Yes, and breaking it up (when `::` would appear within the `{}`) is part of our style guide.

libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -10,6 +12,70 @@ impl Subpath {
...
+			accumulator += length_ratio;
+		}
+		(0, 0.)
+	}
+
+	/// Convert a [SubpathTValue] to a parametric `(segment_index, t)` tuple.
+	/// - Asserts that `t` values contained within the `SubpathTValue` argument lie between 0 and 1.
+	/// - If the argument is a variant containing a `segment_index`, asserts that the index references a valid segment on the curve.
+	pub(crate) fn t_value_to_parametric(&self, t: SubpathTValue) -> (usize, f64) {
+		assert!(self.len() > 1);
```
Keavon, line 34: Why would we want to panic if the subpath is composed of just one segment?

libraries/bezier-rs/src/subpath/lookup.rs
```diff
@@ -10,6 +12,70 @@ impl Subpath {
...
+			}
+			accumulator += length_ratio;
+		}
+		(0, 0.)
+	}
+
+	/// Convert a [SubpathTValue] to a parametric `(segment_index, t)` tuple.
+	/// - Asserts that `t` values contained within the `SubpathTValue` argument lie between 0 and 1.
+	/// - If the argument is a variant containing a `segment_index`, asserts that the index references a valid segment on the curve.
+	pub(crate) fn t_value_to_parametric(&self, t: SubpathTValue) -> (usize, f64) {
```
Keavon, line 33: This should probably return an `Option` so we don't panic if you call it with invalid inputs. That means changing your `assert`s and `unwrap`s to return `None`.

libraries/bezier-rs/src/subpath/lookup.rs
RobNadal: `self.len()` is actually just a shortcut for `self.manipulator_groups.len`, so this is actually checking to make sure the number of segments is non-zero. `self.len_segments()` counts the number segments. I use that function instead to make this clearer.

libraries/bezier-rs/src/subpath/lookup.rs
RobNadal: We included the same assert logic in the `Bezier` version of `t_value_to_parametric` as well. We put in these assertions since otherwise we'd either need to place these assertions in every function that requires a tvalue, or have all those functions return optionals, which we thought cluttered the interface somewhat.

I'd like to keep the bezier and subpath versions consistent, so removing the assertions and converting everything to use optionals will be a more significant change. @Keavon should we do this as part of this MR or a separate one?

Keavon: I hadn't considered how every function would have to return an Option, good point. Let's defer on this.

Keavon (approved):

PR #1029, opened 2023-02-14
Add Minimum and Maximum option to Threshold node

isiko:
Adds options to the Threshold node that calculates the threshold based on the min / max channel of the Color. This might be useful later when splitting Images into separate Images for each Channel, as then the other modes might behave weirdly.
(I do not know if this is wanted, I just added it because I thought it might come in handy later, feel free to just close this PR)


Keavon:
We might remove it later if it ends up not being useful, but for now it doesn't hurt.

Keavon (approved):

PR #1033, opened 2023-02-16
Change tauri build directory back to `../dist`

TrueDoctor:
Reimplements #946

This does change the tauri build folder back to `../dist` but ensures that the dist file is included in git

Closes #946


isiko (approved):
Nice :+1: lgtm

PR #1035, opened 2023-02-16
Bezier-rs: Add self_intersection to subpath

Androxium:
This PR adds a naive brute-force implementation of the self_intersections function to the Subpath library. This PR also updates the function description of the intersections function to include the minimum_separation variable.

This PR is a rebase of the work done in #915.

### DIFF
--- a/libraries/bezier-rs/src/subpath/solvers.rs
+++ b/libraries/bezier-rs/src/subpath/solvers.rs
@@ -10,7 +10,14 @@ impl Subpath {
 	/// Expects `t` to be within the inclusive range `[0, 1]`.
 	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#subpath/evaluate/solo" title="Evaluate Demo"></iframe>
 	pub fn evaluate(&self, t: SubpathTValue) -> DVec2 {
-		let (segment_index, t) = self.t_value_to_parametric(t);
+		let (segment_index, t) = match t {
+			SubpathTValue::GlobalEuclidean(_) => todo!(),
+			SubpathTValue::Parametric { segment_index, t: segment_t } => (segment_index, segment_t),
+			SubpathTValue::GlobalParametric(_) => self.t_value_to_parametric(t),
+			SubpathTValue::Euclidean { segment_index, t } => todo!(),
+			SubpathTValue::EuclideanWithinError { segment_index, t, error } => todo!(),
+			SubpathTValue::GlobalEuclideanWithinError { t, error } => todo!(),
+		};
 		self.get_segment(segment_index).unwrap().evaluate(TValue::Parametric(t))
 	}
 
--- a/libraries/bezier-rs/src/subpath/solvers.rs
+++ b/libraries/bezier-rs/src/subpath/solvers.rs
@@ -1,5 +1,5 @@
 use super::*;
-use crate::consts::{MAX_ABSOLUTE_DIFFERENCE, MIN_SEPERATION_VALUE};
+use crate::consts::MAX_ABSOLUTE_DIFFERENCE;
 use crate::utils::SubpathTValue;
 use crate::TValue;
 
@@ -28,14 +28,6 @@ impl Subpath {
 			.enumerate()
 			.flat_map(|(index, bezier)| bezier.intersections(other, error, minimum_seperation).into_iter().map(|t| (index, t)).collect::<Vec<(usize, f64)>>())
 			.collect();
-
-		intersection_t_values.iter().fold(Vec::<(usize, f64)>::new(), |mut accumulator, (index, t)| {
-			if !accumulator.is_empty() && (accumulator.last().unwrap().1 - t).abs() < minimum_seperation.unwrap_or(MIN_SEPERATION_VALUE) {
-				accumulator.pop();
-			}
-			accumulator.push((*index, *t));
-			accumulator
-		});
 
 		intersection_t_values
 	}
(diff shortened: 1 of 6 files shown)


libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -10,35 +10,37 @@ impl Subpath {
...
 	pub fn evaluate(&self, t: SubpathTValue) -> DVec2 {
-		let (segment_index, t) = self.t_value_to_parametric(t);
+		let (segment_index, t) = match t {
+			SubpathTValue::GlobalEuclidean(_) => todo!(),
+			SubpathTValue::Parametric { segment_index, t: segment_t } => (segment_index, segment_t),
+			SubpathTValue::GlobalParametric(_) => self.t_value_to_parametric(t),
+			SubpathTValue::Euclidean { segment_index, t } => todo!(),
+			SubpathTValue::EuclideanWithinError { segment_index, t, error } => todo!(),
+			SubpathTValue::GlobalEuclideanWithinError { t, error } => todo!(),
+		};
```
RobNadal, line 20: This whole match statement can be removed

RobNadal, line 20: This whole match statement can be removed

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -52,7 +54,32 @@ impl Subpath {
...
 	}
 
-	/// Returns a normalized unit vector representing the direction of the normal on the subpath based on the parametric `t`-value provided.
+	/// Returns a list of `t` values that correspond to the self intersection points of the subpath. For each intersection point, the returned `t` value is the smaller of the two that correspond to the point.
+	/// - `error` - For intersections with non-linear beziers, `error` defines the threshold for bounding boxes to be considered an intersection point.
+	/// - `minimum_seperation`: the minimum difference two adjacent `t`-values must have when comparing adjacent `t`-values in sorted order.
+	/// If the comparison condition is not satisfied, the function takes the larger `t`-value of the two
+	///
+	/// **NOTE**: if an intersection were to occur within an `error` distance away from an anchor point, the algorithm will filter that intersection out.
+	// TODO: add iframe to rustdocs
```
RobNadal, line 63: You can add the iframe as part of this review, just follow the format of the others

website/other/bezier-rs-demos/src/features/subpath-features.ts
```diff
@@ -40,32 +40,52 @@ const subpathFeatures = {
 		sliderOptions: [tSliderOptions],
 		chooseTVariant: true,
 	},
-	"intersect-linear": {
+	"Intersect (Line Segment)": {
```
RobNadal, line 43: The shorthand name of the feature should not be modified (here, and for the other intersect features)

libraries/bezier-rs/src/subpath/solvers.rs
Androxium: By removed, do you mean replace all the different match branches with 1 default match branch `_ => todo!(),`? The reason I changed this assignment to a match statement is because `intersection` and `self-intersection` return `(segment_idx: usize, t_value: f64)` while other subpath functions return `GlobalParametric(t_value)`.

RobNadal: I don't see why the match statement is necessary. `self.t_value_to_parametric(t)` will do the conversion from the different variants to a `(segment, index)` tuple, so there is no need to replicate within `evaluate`

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -16,29 +16,24 @@ impl Subpath {
...
-		intersection_t_values.iter().fold(Vec::new(), |mut accumulator, t| {
-			if !accumulator.is_empty() && (accumulator.last().unwrap() - t).abs() < minimum_seperation.unwrap_or(MIN_SEPERATION_VALUE) {
+		intersection_t_values.iter().fold(Vec::<(usize, f64)>::new(), |mut accumulator, (index, t)| {
+			if !accumulator.is_empty() && (accumulator.last().unwrap().1 - t).abs() < minimum_seperation.unwrap_or(MIN_SEPERATION_VALUE) {
 				accumulator.pop();
 			}
-			accumulator.push(*t);
+			accumulator.push((*index, *t));
 			accumulator
 		});
```
RobNadal, line 38: Thinking about this a bit more, I'm not sure that the fold here is necessary.
1. Doesn't fold return a value? We're not actually capturing the return value of the fold, so I'm not confident that this line is actually doing anything. I don't think fold actually modifies things in place
2. The fold behavior is basically removing points that are within `MIN_SEPERATION_VALUE` of each other. But, doesn't Bezier::intersection already do that? The only points that wouldn't be `MIN_SEPERATION_VALUE` sequential are those that might be on different segments, but we don't really care about that case, the whole point of adding `MIN_SEPERATION_VALUE` is to neatly handle the case where two segments overlap and generate too many intersections.

TLDR: I think we can just remove the whole `intersection_t_values.iter().fold(...);`

Keavon (approved):
LGTM, just these couple comments.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -52,7 +39,32 @@ impl Subpath {
 		self.get_segment(segment_index).unwrap().tangent(TValue::Parametric(t))
 	}
 
-	/// Returns a normalized unit vector representing the direction of the normal on the subpath based on the parametric `t`-value provided.
```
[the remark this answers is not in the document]
Androxium: Oops, that's my bad, must have accidentally deleted it when rebasing.

PR #1036, opened 2023-02-17
Bezier-rs: Add iframes to rustdoc

hannahli2010:
Add `iframes` containing the UI demos to the bezier-rs rustdocs
 - Add white background
 - Removed header from the `solo` version of the UI

Screenshot below (I was testing by changing the `iframe` to point to localhost)
image

### DIFF
--- a/website/other/bezier-rs-demos/src/style.css
+++ b/website/other/bezier-rs-demos/src/style.css
@@ -2,6 +2,7 @@ html,
 body {
 	font-family: Arial, sans-serif;
 	text-align: center;
+  background-color: white;
 }
 
 body > h1 {
--- a/website/other/bezier-rs-demos/src/style.css
+++ b/website/other/bezier-rs-demos/src/style.css
@@ -2,7 +2,7 @@ html,
 body {
 	font-family: Arial, sans-serif;
 	text-align: center;
-  background-color: white;
+	background-color: white;
 }
 
 body > h1 {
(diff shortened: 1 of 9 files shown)


Androxium (approved):
LGTM. One small nit I have would be removing the function name from the iframe but I don't think it's mandatory.

Keavon:
> LGTM. One small nit I have would be removing the function name from the iframe but I don't think it's mandatory.

The `title` attribute?

RobNadal:
image
It looks like because we don't set a background on our page, the default in the rust docs dark mode is difficult to read. I also agree with @Androxium that we could remove the title from within the iframe. If we do that, we should probably move the tvariant selector to be under the visualizations so it doesn't look like a title itself

Keavon:
The `title` attribute is just for screen readers. It isn't rendered on the page. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#accessibility_concerns

RobNadal:
No, I think @Androxium and I are referring to removing the `div` that says the name of the visualization (e.g. "Length"), since it already says it in the function definition of the rustdocs

Keavon:
Gotcha, that's fine to remove if you wish.

hannahli2010:
I already removed the div with the title in this PR. The screenshot still has the title because the `iframe` displays the the live site which doesn't have the change yet.

I'll add a white background shortly.

Keavon:
I'll be sure to deploy as soon as this is merged so you can get the updates to the hosted page.

website/other/bezier-rs-demos/src/style.css
```diff
@@ -2,6 +2,7 @@ html,
 body {
 	font-family: Arial, sans-serif;
 	text-align: center;
+  background-color: white;
```
Keavon, line 5: This appears to be misaligned.

Keavon (approved):

PR #1037, opened 2023-02-18
Bezier-rs: Add function to smoothly join bezier curves

RobNadal:
- Joins two bezier curves smoothly up to the first derivative

image

### DIFF
--- a/libraries/bezier-rs/src/bezier/solvers.rs
+++ b/libraries/bezier-rs/src/bezier/solvers.rs
@@ -52,15 +52,19 @@ impl Bezier {
 		}
 	}
 
-	/// Returns a normalized unit vector representing the tangent at the point `t` along the curve.
-	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#bezier/tangent/solo" title="Tangent Demo"></iframe>
-	pub fn tangent(&self, t: TValue) -> DVec2 {
-		let t = self.t_value_to_parametric(t);
+	/// Returns the non-normalized vector representing the tangent at the point `t` along the curve.
+	fn non_normalized_tangent(&self, t: f64) -> DVec2 {
 		match self.handles {
 			BezierHandles::Linear => self.end - self.start,
 			_ => self.derivative().unwrap().evaluate(TValue::Parametric(t)),
 		}
-		.normalize()
+	}
+
+	/// Returns a normalized unit vector representing the tangent at the point `t` along the curve.
+	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#bezier/tangent/solo" title="Tangent Demo"></iframe>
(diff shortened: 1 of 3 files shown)


RobNadal:
I prototyped this function back when I first started subpath outline in December, but I don't think I'll actually need it to complete outline. I still the functionality is useful though, so I separated it out and am putting it in its own PR.

libraries/bezier-rs/src/bezier/solvers.rs
```diff
@@ -386,6 +390,15 @@ impl Bezier {
 		.flat_map(|bezier| self.intersections(bezier, None, None))
 		.collect()
 	}
+
+	/// Returns a cubic bezier which joins this and the provided bezier curves.
+	/// The resulting path formed by the Bezier curves is smooth up to the first derivative.
```
hannahli2010, line 395: Not super sure if "smooth" is the best descriptor here but I don't feel too strongly about it.

RobNadal: I'll update the comment to say 'continuous' instead.

libraries/bezier-rs/src/bezier/solvers.rs
```diff
@@ -390,6 +394,15 @@ impl Bezier {
 		.flat_map(|bezier| self.intersections(bezier, None, None))
 		.collect()
 	}
+
+	/// Returns a cubic bezier which joins this and the provided bezier curves.
```
Keavon, line 398:
```suggestion
	/// Returns a cubic bezier which joins this with the provided bezier curve.
```

libraries/bezier-rs/src/bezier/solvers.rs
```diff
@@ -390,6 +394,15 @@ impl Bezier {
...
+	/// The resulting path formed by the Bezier curves is continuous up to the first derivative.
+	/// <iframe frameBorder="0" width="100%" height="325px" src="https://graphite.rs/bezier-rs-demos#bezier/join/solo" title="Join Demo"></iframe>
+	pub fn join(&self, bezier2: &Bezier) -> Bezier {
```
Keavon, line 401: I think `other` would be a more traditional name for this kind of code used in conjunction with `self`.

website/other/bezier-rs-demos/src/features/bezier-features.ts
```diff
@@ -483,6 +483,54 @@ const bezierFeatures = {
...
+					[95, 195],
+					[155, 145],
+				]);
+			}
+			return bezier.join([
+				[140, 150],
+				[85, 110],
+				[65, 180],
+				[30, 140],
+			]);
```
Keavon, line 508: It'd be a bit more idiomatic to use `if-elseif-else` to avoid needing to use the explicit `return` keyword.

Keavon (approved):
Merge after changes. Thanks.

PR #1039, opened 2023-02-19
Bezier-rs: Subpath offset and bezier offset improvements  

RobNadal:
- Improved bezier offset and scale algorithms:
  - Added smoothing between sequential segments
  - Modified scale to convert quadratic segments to cubics, as they are more expressive
  
  - Before and after: 
  - image
  - Removed minor optimization in reduce which causes jumping approximations - offset and outline should now look visually far smoother when moving points / handles

- Improved reduce function to split on inflection points
- Added `to_linear`, `to_quadratic` and `to_cubic` functions
- Added in depth `Bezier::offset` testing
- Modified `Bezier::offset` and `Bezier::outline` to return Subpaths
  - Requires gap filling when the reduce offset encounters irreducible curves (smooth join when situation arises)
- Added subpath offset
- Added subpath intersections function to get intersections between two subpaths
- Decreased magnitude of acceptable distances in visualization of offsets

image


libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -75,6 +77,126 @@ impl Subpath {
...
+		let mut drop_common_point = vec![true; self.len()];
+
+		if distance != 0. {
+			// Clip or joining consecutive Subpaths
+			for i in 0..subpaths.len() - 1 {
+				let j = i + 1;
+				let subpath1 = &subpaths[i];
+				let subpath2 = &subpaths[j];
+
+				let last_segment = subpath1.get_segment(subpath1.len_segments() - 1).unwrap();
```
hannahli2010, line 121: Just want to confirm: We're sure these subpaths won't be empty right?

RobNadal:
Per Keavon's comments in on the Graphite discord, we should expose the conversion from quadratic to cubic Beziers in a user facing function. We should also use favor the naming convention of `this` or `mut_self` rather than `copy`.

libraries/bezier-rs/src/subpath/core.rs
```diff
@@ -31,6 +31,44 @@ impl Subpath {
...
+		if !closed {
+			manipulator_groups.push(ManipulatorGroup {
+				anchor: last.end(),
+				in_handle: last.handle_end(),
+				out_handle: None,
+			});
+			return Subpath::new(manipulator_groups, false);
+		}
+
+		manipulator_groups[0].in_handle = last.handle_end();
```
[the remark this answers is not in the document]
RobNadal: We're ignoring it purposely, per the function comment

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -75,6 +77,126 @@ impl Subpath {
...
+	/// Reduces the segments of the subpath into simple subcurves, then scales each subcurve a set `distance` away.
+	/// The intersections of segments of the subpath are joined using the method specified by the `joint` argument.
+	/// <iframe frameBorder="0" width="100%" height="375px" src="https://graphite.rs/bezier-rs-demos#subpath/offset/solo" title="Offset Demo"></iframe>
+	pub fn offset(&self, distance: f64, joint: Joint) -> Subpath {
+		assert!(self.len_segments() > 1, "Cannot outline an empty Subpath.");
+
+		let mut subpaths = self.iter().map(|bezier| Subpath::from_beziers(bezier.offset(distance), false)).collect::<Vec<Subpath>>();
+		let mut drop_common_point = vec![true; self.len()];
+
+		if distance != 0. {
```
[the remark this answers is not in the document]
RobNadal: Wow, thats a really good point, there's no need to reduce

libraries/bezier-rs/src/subpath/transform.rs
RobNadal: I don't believe its possible for the subpaths to be empty

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -1,11 +1,39 @@
...
+			BezierHandles::Cubic { handle_start, handle_end } => (handle_start + handle_end) / 2.,
+		};
+		Bezier::from_quadratic_dvec2(self.start, handle, self.end)
+	}
+
+	/// Returns a cubic approximation of the given [Bezier].
+	pub fn to_cubic(&self) -> Bezier {
+		let (handle_start, handle_end) = match self.handles {
+			BezierHandles::Linear => (self.start, self.end),
+			BezierHandles::Quadratic { handle } => (self.start + (2. / 3.) * (handle - self.start), self.end + (2. / 3.) * (handle - self.end)),
```
Keavon, line 31: Comment this reference for the math: https://stackoverflow.com/a/63059651/775283

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -1,11 +1,39 @@
...
+			BezierHandles::Cubic { handle_start, handle_end } => (handle_start, handle_end),
```
Keavon, line 32: Can you just `return self` here? It might be more performant.

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -312,78 +339,107 @@ impl Bezier {
...
+		if self.handles != BezierHandles::Linear {
+			scaled.smooth_cubic_paths();
+		}
+
+		scaled
 	}
 
 	/// Version of the `offset` function which scales the offset such that the start of the offset is `start_distance` from the original curve, while the end of
 	/// of the offset is `end_distance` from the original curve. The curve transitions from `start_distance` to `end_distance` gradually, proportional to the
 	/// distance along the equation (`t`-value) of the curve. Similarily to the `offset` function, the returned result is an approximation.
```
Keavon, line 371: Spelling: Similarily -> Similarly

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -695,71 +752,137 @@ mod tests {
...
+					.iter()
+					.map(|t| {
+						let offset_point = offset_segment.evaluate(TValue::Parametric(*t));
+						let closest_point_t = bezier.project(offset_point, ProjectionOptions::default());
+						let closest_point = bezier.evaluate(TValue::Parametric(closest_point_t));
+						let actual_distance = offset_point.distance(closest_point);
+
+						dbg!(actual_distance);
+						dbg!(expected_distance);
+						dbg!(err_threshold);
```
Keavon, line 804: Are these supposed to be committed?

libraries/bezier-rs/src/subpath/core.rs
```diff
@@ -34,6 +34,47 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 		)
 	}
 
+	/// Creates a subpath from a vector of [Bezier]. When two consecutive Bezier do not share an end and start point, this function
```
Keavon, line 37: "vector of [Bezier]" -> "slice of [Bezier]"

"two consecutive Bezier" -> "two consecutive Beziers"

libraries/bezier-rs/src/subpath/manipulators.rs
```diff
@@ -34,6 +34,47 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 		self.manipulator_groups[segment_index % number_of_groups].out_handle = first.handle_start();
 		self.manipulator_groups[(segment_index + 2) % number_of_groups].in_handle = second.handle_end();
 	}
+
+	/// Append a [Bezier] to the end of a subpath from a vector of [Bezier].
+	/// The `append_type` parameter determines how the function behaves when the subpath's last anchor is not equal to the Bezier's start point.
+	/// - `IgnoreStart`: drops the bezier's start point in favour of the subpaths last anchor
+	/// - `SmoothJoin(f64)`: joins the subpaths endpoint with the beziers start with a another Bezier segment that is continuous up to the second derivative
```
Keavon, line 41: "subpaths" -> "subpath's"
"favour" -> "favor"

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -272,6 +274,183 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 			manipulator_group.apply_transform(affine_transform);
 		}
 	}
+
+	// Smooths a Subpath up to the first derivative, using a weighted averaged based on segment length.
+	pub(crate) fn smooth_cubic_paths(&mut self) {
+		for i in 0..self.len() - 2 {
```
Keavon, line 280: In other parts of this PR, the iterators range from 1 rather than 0, and the indexing happens for `i - 1` and `i` rather than `i` and `i + 1` which is nice because it more clearly avoids OB1 mistakes since it's obvious that `i - 1` when `i` starts at 1 won't go below 0. Would it make sense to adopt that pattern here too, with a `for i in 1..self.len() - 1` loop (and changes to the usage of i below)?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -272,6 +274,183 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+			self.manipulator_groups[i + 1].in_handle = Some((average_unit_tangent / 3. * -1.) * end_tangent.length() + end_point);
+
+			let start_point = second_bezier.start();
+			self.manipulator_groups[i + 1].out_handle = Some((average_unit_tangent / 3.) * start_tangent.length() + start_point);
+		}
+	}
+
+	// Helper function to clip overlap of two intersecting open Subpaths. Returns an optional, as intersections may not exist for certain arrangements and distances.
+	// TODO: If a segment curls back on itself tightly enough it could intersect again at the portion that would be trimmed. This could cause the subpaths to be clipped
+	// at the incorrect location. This can be avoided by first trimming the two subpaths at any extrema, effectively ignoring loopbacks.
```
Keavon, line 305: Prefix this line with `// TODO: ` also.

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -272,6 +274,183 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 			manipulator_group.apply_transform(affine_transform);
 		}
 	}
+
+	// Smooths a Subpath up to the first derivative, using a weighted averaged based on segment length.
```
Keavon, line 278: Doc comment with three slashes.

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -272,6 +274,183 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+
+		Subpath::new(manipulator_groups, self.closed)
+	}
+
+	/// Outline returns a single closed subpath (if the original subpath was open) or two closed subpaths (if the original subpath was closed) that forms
+	/// an approximate outline around the subpath at a specified distance from the curve. Outline takes the following parameters:
+	/// - `distance` - The outline's distance from the curve.
+	/// - `joint` - The joint type used to cap the endpoints of open bezier curves, and join successive subpath segments.
+	/// <iframe frameBorder="0" width="100%" height="375px" src="https://graphite.rs/bezier-rs-demos#subpath/outline/solo" title="Outline Demo"></iframe>
+	pub fn outline(&self, distance: f64, joint: Joint) -> (Subpath<ManipulatorGroupId>, Option<Subpath<ManipulatorGroupId>>) {
```
Keavon, line 436: Ah, right, here's a case where we begin needing the `Path` data type (which is composed of multiple `Subpath`s) so a single result can be returned that might have a hole in it (for the case when the input Subpath is closed). But for now, just add a TODO comment to change this function's return type to a Path once we have that data type so it can represent the potential hole.

libraries/bezier-rs/src/utils.rs
```diff
@@ -28,6 +28,13 @@ pub enum SubpathTValue {
 	GlobalEuclideanWithinError { t: f64, error: f64 },
 }
 
+#[derive(Copy, Clone)]
+pub enum Joint {
+	Mitre,
+	Blunt,
+	Rounded,
+}
```
Keavon, line 36: Please use the names from the SVG spec: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-linejoin#usage_context (note the spelling for "Miter")

Keavon (approved):

Keavon (approved):
Awesome!

libraries/bezier-rs/src/subpath/manipulators.rs
RobNadal: Whoops, Canadian-isms showing

libraries/bezier-rs/src/subpath/transform.rs
RobNadal: Sure, I believe its an easy change.

libraries/bezier-rs/src/subpath/transform.rs
RobNadal: I'd prefer to leave `clip_simple_subpaths` private, as it relies on two simplifying assumptions:
- The Subpaths being clipped represent only simple Bezier
- Loopbacks, the edge case I have described in the doc comments, are _very_ rare
These assumptions are both generally true while clipping the subpaths generated in `offset`/`outline`, but I don't know how useful they actually be to hypothetical users.

We could probably make `smooth_cubic_paths` public in the future, but we'd need to think about how to handle the visualization further because our visualizations allow users to play with the handles, but this function also moves handles in the smoothing process. This function also relies on some simplifying assumptions that are only true in the case of offset / outline, so we'd need to do some more work before presenting it to the users, and I don't want to bloat this PR further.

libraries/bezier-rs/src/subpath/transform.rs
RobNadal: Let's chat about the `Path` data type in our next meeting. Perhaps we could add something relatively simple in the short term (in a future PR of course).

libraries/bezier-rs/src/utils.rs
RobNadal: Ah good catch, I made these names up as a placeholder until I verified the real ones. I was pretty close!

libraries/bezier-rs/src/subpath/transform.rs
Keavon: Sounds good, that's a more than reasonable explanation, no need to make them public if it's not intended to be.

PR #1044, opened 2023-02-19
Make image downscaling have a configurable factor

### LINKED ISSUE #1046: Add a downscale control to the Image node for performance

0HyperCube:
Embedded images are no longer downscaled whilst importing from the frontend.

Closes #1046


Keavon:
I think the best option, even though it would take a bunch of RAM and file size when saving the document, is still ingesting the full resolution image into the backend and downscaling (if <100%) in the Rust code based on a downscale percentage slider on the Image node, which would be cached. Otherwise it would compromise the UX from either a usefulness or a sluggishness perspective, while this option just trades off RAM usage (which isn't perfect, but it's better than the alternatives). I filed an issue (#1046) for this. Thanks :)

TrueDoctor (approved):
Merging this now and implementing downscaling in a new pr

PR #1047, opened 2023-02-20
Add 'View > Zoom to Selected' action

### LINKED ISSUE #989: Approachable mini starter tasks for new contributors
All of these are approachable as a first issue for beginners. Those specifically marked as "easy" should be simple enough for less experienced Rust programmers, but they should all be doable with some patience. Before beginning any of these, please ask for clarification about what the issue precisely means and make sure you can reproduce the issue firsthand to understand what's needed. Ask @Keavon for a video clip in Discord if clarification is needed.

This issue is meant to list all the small tasks that didn't warrant their own issue. Please check the [Good First Issue](https://github.com/GraphiteEditor/Graphite/labels/Good%20First%20Issue) tag for a list of other issues that suit this. The majority of those should be reasonably approachable for new Rust programmers, although some of those might assume a bit more Rust experience. Please ask if unsure.

*All gone for now, this issue is closed for the time being*

ajjahn:
Partially resolves #989 - Pan/zoom to the selected layer(s) when hitting the Period or Numpad Period key.

### DIFF
--- a/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
+++ b/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
@@ -223,7 +223,7 @@ impl PropertyHolder for MenuBarMessageHandler {
 				"View".into(),
 				MenuBarEntryChildren(vec![vec![
 					MenuBarEntry {
-						label: "Center View on Selected".into(),
+						label: "Zoom to Selected".into(),
 						shortcut: action_keys!(NavigationMessageDiscriminant::FitViewportToSelection),
 						action: MenuBarEntry::create_action(|_| NavigationMessage::FitViewportToSelection.into()),
 						..MenuBarEntry::default()
--- a/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
+++ b/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
@@ -151,21 +151,13 @@ impl PropertyHolder for MenuBarMessageHandler {
 							..MenuBarEntry::default()
 						},
 					],
-					vec![
-						MenuBarEntry {
-							label: "Delete Selected".into(),
-							icon: Some("Trash".into()),
-							shortcut: action_keys!(DocumentMessageDiscriminant::DeleteSelectedLayers),
-							action: MenuBarEntry::create_action(|_| DocumentMessage::DeleteSelectedLayers.into()),
-							..MenuBarEntry::default()
-						},
-						MenuBarEntry {
-							label: "Center View on Selected".into(),
-							shortcut: action_keys!(NavigationMessageDiscriminant::FitViewportToSelection),
-							action: MenuBarEntry::create_action(|_| NavigationMessage::FitViewportToSelection.into()),
-							..MenuBarEntry::default()
-						},
-					],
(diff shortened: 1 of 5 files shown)


Keavon:
Thanks! Can you also add this to the Layer section of the menu bar? It can be titled "Center View on Selected".

ajjahn:
> Thanks! Can you also add this to the Layer section of the menu bar? It can be titled "Center View on Selected".

Sure, I'll make an attempt. I'm not sure I know exactly where you're referring to though, can you demonstrate via a screenshot/mockup of what you're looking for?

Keavon:
image

It can go in the same grouping just below "Delete Selected".

editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
```diff
@@ -151,13 +151,21 @@ impl PropertyHolder for MenuBarMessageHandler {
...
+							shortcut: action_keys!(DocumentMessageDiscriminant::DeleteSelectedLayers),
+							action: MenuBarEntry::create_action(|_| DocumentMessage::DeleteSelectedLayers.into()),
+							..MenuBarEntry::default()
+						},
+						MenuBarEntry {
+							label: "Center View on Selected".into(),
+							shortcut: action_keys!(NavigationMessageDiscriminant::FitViewportToSelection),
+							action: MenuBarEntry::create_action(|_| NavigationMessage::FitViewportToSelection.into()),
+							..MenuBarEntry::default()
+						},
```
ajjahn, line 167: Done. Is there a method to disable this menu option when there are no items selected?

ajjahn: Also, wondering would this menu option make more sense under the "View" menu?

Keavon: Great question! Unfortunately we don't really have that infrastructure yet, which is why all our menu bar entries don't get grayed out at the appropriate times yet. We just have to make sure we don't do anything unexpected (crash, throw errors, etc.) when it's called at an inappropriate time. The way to properly deal with this is with some kind of event-based model that should automatically update the UI according to changes in context. Otherwise it would require lots of message dispatching from dozens of disparate locations all to update specific parts of the UI which is fragile and super easy to do incorrectly. So we really need that infrastructure to update the UI based on contextual events.

Keavon: > Also, wondering would this menu option make more sense under the "View" menu?

You're right, I wasn't thinking. Let's call it "Zoom to Selected" and put it in the first slot of the View menu. Thanks for asking about that!

editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
```diff
@@ -222,6 +222,12 @@ impl PropertyHolder for MenuBarMessageHandler {
 			MenuBarEntry::new_root(
 				"View".into(),
 				MenuBarEntryChildren(vec![vec![
+					MenuBarEntry {
+						label: "Center View on Selected".into(),
```
Keavon, line 226: "Zoom to Selected"

Keavon (approved):

PR #1049, opened 2023-02-21
Add lock angle modifier to the Pen tool

### LINKED ISSUE #1048: Add "Lock Angle" to Pen tool
- [ ] The current `[Shift] Break Handle` behavior should become <kbd>Alt</kbd> instead of <kbd>Shift</kbd>
- [ ] The current `[Ctrl] Snap 15°` behavior should become <kbd>Shift</kbd> instead of <kbd>Ctrl</kbd>
- [ ] Implement the extend-in-this-direction feature called `[Ctrl] Lock Angle` that exists in the Line tool, but doesn't yet exist in the Pen tool

ajjahn:
- [x] The current [Shift] Break Handle behavior should become Alt instead of Shift
- [x] The current [Ctrl] Snap 15° behavior should become Shift instead of Ctrl
- [x] Implement the extend-in-this-direction feature called [Ctrl] Lock Angle that exists in the Line tool, but doesn't yet exist in the Pen tool

Fixes #1048

### DIFF
--- a/editor/src/messages/tool/tool_messages/pen_tool.rs
+++ b/editor/src/messages/tool/tool_messages/pen_tool.rs
@@ -54,6 +54,7 @@ pub enum PenToolMessage {
 	PointerMove {
 		snap_angle: Key,
 		break_handle: Key,
+		lock_angle: Key,
 	},
 	Undo,
 	UpdateOptions(PenOptionsUpdate),
@@ -148,6 +149,7 @@ struct PenToolData {
 	should_mirror: bool,
 	// Indicates that curve extension is occurring from the first point, rather than (more commonly) the last point
 	from_start: bool,
+	angle: f64,
 }
 
 impl Fsm for PenToolFsmState {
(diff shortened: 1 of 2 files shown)


editor/src/messages/tool/tool_messages/pen_tool.rs
```diff
@@ -399,10 +401,11 @@ impl Fsm for PenToolFsmState {
 					tool_data.should_mirror = true;
 					process().unwrap_or(PenToolFsmState::PlacingAnchor)
 				}
-				(PenToolFsmState::DraggingHandle, PenToolMessage::PointerMove { snap_angle, break_handle }) => {
+				(PenToolFsmState::DraggingHandle, PenToolMessage::PointerMove { snap_angle, break_handle, lock_angle }) => {
 					let mut process = || {
 						// Get subpath
-						let layer_path = tool_data.path.as_ref()?;
+						let data = tool_data.clone();
+						let layer_path = data.path.as_ref()?;
```
ajjahn, line 408: Is there a better way to deal with the exclusive tool_data ref being borrowed immutably here?

TrueDoctor: Couldn't you just move the angle computation above the layer_path declaration? From the looks of it you only need a mutable ref for the computation and can safely demote it to a shared ref afterwards.

ajjahn: I'm not sure that will be possible. The calculation needs `last_anchor.position`, which is worked out something like this: `tool_data -> layer_path -> subpath -> manipulator_groups -> last_anchor`.

0HyperCube: Instead of passing `tool_data` to the `compute_snapped_angle`, just pass `&mut tool_data.angle` (rust allows you to borrow different fields of a struct at the same time).

Keavon (approved):

PR #1053, opened 2023-02-23
Introduce scopes

TrueDoctor:
Fixes the current blend node implementation


Keavon:
TODO as discussed:
- Make them work for all data types, so they can work with vector data after the Image Frame node
- Put them in a parent graph which wraps the user-facing graph
    - Make the breadcrumb trails actually show the nodes of inner/wrapped subgraphs
    - Add a `⋮` checkbox to the breadcrumb trail to show the outer/wrapper subgraphs
    - Faster alternative: temporarily just put the scope nodes in the Input and Output nodes

TrueDoctor:
> TODO as discussed:
> 
> * Make them work for all data types, so they can work with vector data after the Image Frame node
> * Put them in a parent graph which wraps the user-facing graph
>   
>   * Make the breadcrumb trails actually show the nodes of inner/wrapped subgraphs
>   * Add a `⋮` checkbox to the breadcrumb trail to show the outer/wrapper subgraphs
>   * Faster alternative: temporarily just put the scope nodes in the Input and Output nodes

You currently need to add an implementation for other data types but that is pretty simple to do. I decided to choose the faster alternative for now

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -663,15 +729,43 @@ impl DocumentNodeType {
 	}
 }
 
+pub fn wrap_network_in_scope(network: NodeNetwork) -> NodeNetwork {
+	assert_eq!(network.inputs.len(), 1, "Networks wrapped in scope must have exactly one input");
+	let input_type = network.nodes[&network.inputs[0]].inputs.iter().find(|&i| matches!(i, NodeInput::Network(_))).unwrap().clone();
```
0HyperCube, line 734: Perhaps we could do something in the case of multiple network inputs (either adding multiple scopes or adding an error) to avoid strange issues?

0HyperCube (approved):
Great work!

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
TrueDoctor: I think this should basically never happen because a (top level) network can only really have one input (it is basically a function) . I think what we will end up doing is to just not provide the ui to be able to create a network with multiple inputs

TrueDoctor: well, it could be used else where but only in places with a single input. If we were to allow multiple scopes side by side it would be possible to access not yet initialized data. Until we have some sort of borrow checker that is just a limitation of the system

0HyperCube: Sounds good.

TrueDoctor: image

PR #1054, opened 2023-02-23
Bézier-rs: Add ids to manipulator groups

0HyperCube:
Each manipulator group now has a generic id type that implements the `ManipulatorGroupId` trait. This allows the subpaths to integrate with Graphite's UUID system without having to depend on additional libraries or adding extra overhead to other users who may not need the ids.


hannahli2010:
I personally feel like it's a little confusing for both the generic data type and the trait to be named `ManipulatorGroupId`, but I don't feel too strongly about it. Otherwise, it looks good to me!

0HyperCube:
Do you have a preferred name for the generic / trait @hannahli2010?

0HyperCube:
Thanks - I've renamed as suggested @RobNadal.

PR #1057, opened 2023-02-25
Refactor vector data type from document-legacy into node graph

0HyperCube:
Part of #997

The VectorData struct has now been added to graphene_core (along with the relevant structs for style).


Keavon (approved):
Looks good 👍

PR #1058, opened 2023-02-25
Bézier-rs: Add utils to subpath

0HyperCube:
- Constructors for subpaths (taken from the subpath implementation in graphite currently).
- Transforming subpaths
- Features to enable derives for serde and dyn-any (although I had to implement static type manually on some occasions because the macro is broken).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,7 +281,9 @@ checksum = "9e1b586273c5702936fe7b7d6896644d8be71e6314cfe09d3167c95f712589e8"
 name = "bezier-rs"
 version = "0.1.0"
 dependencies = [
+ "dyn-any",
  "glam",
+ "serde",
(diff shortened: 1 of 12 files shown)


Keavon (approved):
Looks nice. I wonder if, since much of this was copied from other Graphite code, we can also delete that code, @0HyperCube?

@hannahli2010 @RobNadal @Androxium perhaps you'd have the time to make the changes I suggested in this code review set to save @0HyperCube's time? Thanks.

0HyperCube:
> Looks nice. I wonder if, since much of this was copied from other Graphite code, we can also delete that code, @0HyperCube?

Not without breaking all of the tools in the editor currently. However once the migration is complete then we can delete the duplicate `Subpath` struct.

PR #1060, opened 2023-02-27
Create opposing handle on dragging with shift

ShouvikGhosh2048:
[From `#code-todo-list` on Discord:](https://discord.com/channels/731730685944922173/881073965047636018/1077737846070317206)
Deleting a handle (turning a cubic into a quadratic bezier) behaves different from dragging a handle into its anchor. In the latter case, holding Shift to mirror the handle-being-dragged's length (implemented in #1009) works. But if the handle is deleted with the Delete key, that doesn't work. We want to fix the actual cause so the two act equivalently, and then we want to fix it so the Shift key works in both situations even after the fix likely breaks it for both situations.


editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -212,8 +216,21 @@ impl ShapeEditor {
...
+	/// 3. If the handle selected has an opposing handle, we need manipulator group's angle mirroring to be true.
+	///
+	/// If (1) and (2) are true and there is no opposing handle, we create one
+	/// (irrespective of angle mirroring) and set angle mirroring to true.
+	///
+	/// We return the opposing handle lengths.
+	///
+	/// toggle_angle_mirroring decides whether the current angle mirroring will be toggled.
+	/// It is required as messages don't immediately change the manipulator groups.
+	pub fn mirror_opposing_handles(&self, document: &Document, toggle_angle_mirroring: bool, responses: &mut VecDeque<Message>) -> OpposingHandleLengths {
```
ShouvikGhosh2048, line 233: As mentioned in the comment above, I pass a `toggle_angle_mirroring` to check whether the angle mirroring will be toggled. This is required as the angle mirroring won't be updated immediately (instead messages to toggle are added to `responses`).

Is this fine? Is there a better way to do this?

0HyperCube: This is fine.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -303,15 +336,36 @@ impl ShapeEditor {
...
 
-					if let Some(offset) = (opposing_handle.position - anchor.position).try_normalize() {
-						let new_opposing_handle_position = anchor.position + offset * (*opposing_handle_length);
-						assert!(new_opposing_handle_position.is_finite(), "Opposing handle not finite!");
+					if let Some(opposing_handle_length) = opposing_handle_length {
+						let opposing_handle = if let Some(opposing_handle) = manipulator_group.opposing_handle(handle) {
+							opposing_handle
+						} else {
+							return;
+						};
```
0HyperCube, line 344:
```rs
let Some(opposing_handle) = manipulator_group.opposing_handle(handle) else {
	return;
};
```
is slightly more clear.

0HyperCube (approved):
Looks good :)

editor/src/messages/tool/common_functionality/shape_editor.rs
ShouvikGhosh2048: Done

Keavon:
Closing for the reason described in [this thread on Discord](https://discord.com/channels/731730685944922173/731738914812854303/1094135778067615835).

PR #1062, opened 2023-03-01
Add transform information next to the image data (#1055)

### LINKED ISSUE #1055: Combine Bitmap and Transform data into an Image data type
Refactor the separate bitmap + transform data types that are run separately into a new Image data type composed of bitmap + transform data. All nodes work on Image data. No need to route the transform data separately. This makes the Blend node able to work on Images even if they're not the same size. It lets us also entirely remove the concept of a transform belonging to the node graph layer. Then the Output node can generically take Vector or Image data, and in the Image case, it runs it through an Image Frame node automatically (thus solving the issue with previewing a bitmap node causing an error for users currently).

AlexandruIca:
Since the image contains a `DAffine2` field now, the `Hash` trait implementation can't just be derived, so it had to be done manually. The `Hash` implementation ignores the `transform` field.

Closes #1055.


node-graph/gcore/src/raster.rs
```diff
@@ -308,24 +325,31 @@ mod image {
...
 		/// Generate Image from some frontend image data (the canvas pixels as u8s in a flat array)
 		pub fn from_image_data(image_data: &[u8], width: u32, height: u32) -> Self {
 			let data = image_data.chunks_exact(4).map(|v| Color::from_rgba8(v[0], v[1], v[2], v[3])).collect();
-			Image { width, height, data }
+			Image {
+				width,
+				height,
+				data,
+				transform: DAffine2::IDENTITY,
+			}
```
TrueDoctor, line 347: try adding the transform as a parameter here and fix the errors that generates

TrueDoctor (approved):
Will do some more testing and regression fixing in #1053

PR #1063, opened 2023-03-02
Bezier-rs: Eliminate simplest intersection errors in Bezier::offset

RobNadal:
- Trim offsets from Bezier curves which intersect, when the original Bezier itself does not intersect
- Before:
image
- After: 
image


Androxium:
Is this an accepted edge case?
Screenshot from 2023-03-03 15-49-20

RobNadal:
Thats super bizarre... what might be happening is in the examples you have sent there is actually multiple intersections. We only ever cut out the first intersection we find. That would explain why the left part of the curve is trimmed out the self-intersection properly, but on the left part of the curve its not

Keavon:
BTW, please make this trimming feature optional since there might be cases where a user wants the original behavior.

Androxium:
I see, that makes sense to me. Do we want the reduction applied to all outline intersections, or just the first? If the work to make it apply to all is too complex, we can do it in a different PR.

PR #1064, opened 2023-03-02
Fix crash when using Imaginate tool

0HyperCube:
Fixes crash when using the imaginate tool, introduced by #1053.


TrueDoctor (approved):
LGTM

PR #1066, opened 2023-03-03
Disable select tool widgets based on selected layer count

tlaplaca:
Subsection of #989 :

- Gray out the Select tool's pivot basis widget when no layers are selected
- Gray out the Select tool's alignment buttons when fewer than two layers are selected
- Gray out the Select tool's boolean operation buttons when fewer than two vector shape layers are selected


editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1987,6 +1991,215 @@ impl DocumentMessageHandler {
 		);
 	}
 
+	pub fn update_tool_options_widgets(&self, responses: &mut VecDeque<Message>) {
```
Keavon, line 1994: I'm noticing this code is all added (rather than being moved from somewhere else). Except this feature doesn't add new widgets. This makes me assume the code here redundantly duplicates the code for building a widget layout, resulting in two separate places that the same things are defined, which isn't maintainable. Unless my interpretation is wrong, can you please do some refactors to avoid the redundant definitions? I wouldn't normally expect a bug fix to add (without also removing) over 200 lines of code. Thanks!

PR #1071, opened 2023-03-07
Bezier-rs: Made wasm imports async in demos

RobNadal:
Bugfix for the issue that caused the following error:
image

wasm-bindgen documentation generally advises that wasm imports must be asynchronous. It seems that we weren't doing that, but were still somehow getting away with it. The wasm package was recently updated in Graphite, at which point the interactive demos broke because, likely because our synchronous imports caught up with us.


Keavon:
Please use `async`/`await` instead of `.then()`.

Keavon (approved):
Nice and clean!

PR #1076, opened 2023-03-10
Move G/R/S to tools folder

0HyperCube:
As discussed in discord. Should also make it easier to integrate this system with the path tool without duplicating too much code.


Keavon (approved):
QA looks good

PR #1079, opened 2023-03-15
Add Range parameter to the Threshold node

isiko:
Adds a Range to the Threshold node. Everything that is inside the range is painted white and everything outside is painted black.


Keavon (approved):
Thank you!

PR #1080, opened 2023-03-15
Add the Mask node

isiko:
Adds a Mask Node

This node behaves weirdly when using the 'Input' node as a Mask, this might possibly be maybe partially "fixed" when merging #1065


Keavon:
Blocked on the crash (relevant diagnostics included in the committed code) which we probably need @TrueDoctor to help with. Unless @isiko you are able to figure out the reason why `mask_pixel.r()` can go negative or even to `NaN` when a node like Levels is used upstream of the mask's Stencil input.

TrueDoctor:
Not blocked on the crash because that is caused by the levels node

Keavon (approved):

PR #1081, opened 2023-03-17
Bezier-rs: Add curvature and rotate functions to subpath

linda-zheng:
Added support for `curvature`, `rotate`, `rotate_about_point` in the Subpath class.

### DIFF
--- a/libraries/bezier-rs/src/subpath/transform.rs
+++ b/libraries/bezier-rs/src/subpath/transform.rs
@@ -333,6 +333,19 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 		Some((clipped_subpath1, clipped_subpath2.unwrap()))
 	}
 
+	/// Returns a subpath that results from rotating the path around the origin by the given angle (in radians).
+	/// <iframe frameBorder="0" width="100%" height="375px" src="https://graphite.rs/bezier-rs-demos#subpath/rotate/solo" title="Rotate Demo"></iframe>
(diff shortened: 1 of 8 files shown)


Androxium:
Getting these errors in the console when loading
Screenshot from 2023-03-17 13-39-50
 the web page

linda-zheng:
Hmm... it seems like these errors are stemming from the bezier curvature function. If you enable the linear bezier curve, we see the same console errors. Perhaps it's best to just not display the curvature for linear curves within the subpath?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -325,6 +326,19 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
 		Some((clipped_subpath1, clipped_subpath2.unwrap()))
 	}
 
+	/// Returns a subpath that results from rotating the path around the origin by the given angle (in radians).
+	/// <iframe frameBorder="0" width="100%" height="375px" src="https://graphite.rs/bezier-rs-demos#subpath/rotate/solo" title="Rotate Demo"></iframe>
+	pub fn rotate(&self, angle: f64) -> Subpath<ManipulatorGroupId> {
```
hannahli2010, line 331: I'm wondering if it'd be cleaner to use the `apply_transform` function that 0HyperCube added to do this transformation instead.

Keavon:
Now that https://github.com/GraphiteEditor/Graphite/pull/1085 is merged, there are some breakages when you run the site.

Keavon (approved):

PR #1082, opened 2023-03-17
Bezier-rs: Add lookup table function for subpath and make it support euclidean parameterization

linda-zheng:
Added `compute_lookup_table` to subpath class.
Added `tvalue_type: Option<TValueType>` to `compute_lookup_table` in both bezier and subpath classes which allows computation based on euclidean distances rather than parametric.

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -375,7 +375,7 @@ impl ShapeState {
 
 		for subpath in &vector_data.subpaths {
 			for (manipulator_index, bezier) in subpath.iter().enumerate() {
-				let t = bezier.project(layer_pos, projection_options);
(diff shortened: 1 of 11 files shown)


Androxium (approved):
LTGM. Something to keep in mind for the future would be figuring out more efficient ways to compute the euclidean lookup table for a curve as the performance drop on the interactive website when manipulating an example in parametric mode vs. euclidean is quite noticeable.

Keavon (dismissed):

Keavon:
Now that #1085 is merged, there are some breakages when you run the site.

Keavon (approved):

Keavon:
Hi @linda-zheng, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

PR #1083, opened 2023-03-18
Bezier-rs: Add joins and caps to offsets and outlines

hannahli2010:
- Add the following caps: Butt, Round, and Square
- Add the following joins: Bevel, Miter, Round

- Handle bad offset cases where the object being offset was essentially a single point
- Fix issue with attempting to normalize a zero vector
- Change `reduced_curves_and_t_values` to return pairs of `t` values

### DIFF
--- a/libraries/bezier-rs/src/bezier/core.rs
+++ b/libraries/bezier-rs/src/bezier/core.rs
@@ -211,6 +211,14 @@ impl Bezier {
 
 		self_points.len() == other_points.len() && self_points.into_iter().zip(other_points.into_iter()).all(|(a, b)| a.abs_diff_eq(b, max_abs_diff))
 	}
+
+	/// Returns true if the start, end and handles of the Bezier are all at the same location
+	pub fn is_single_point(&self) -> bool {
+		let points = self.get_points().collect::<Vec<DVec2>>();
+		let start = self.start();
+
+		points.iter().all(|point| point.abs_diff_eq(start, MAX_ABSOLUTE_DIFFERENCE))
+	}
 }
 
 #[cfg(test)]
--- a/libraries/bezier-rs/src/bezier/core.rs
+++ b/libraries/bezier-rs/src/bezier/core.rs
@@ -214,10 +214,9 @@ impl Bezier {
 
 	/// Returns true if the start, end and handles of the Bezier are all at the same location
 	pub fn is_point(&self) -> bool {
-		let points = self.get_points().collect::<Vec<DVec2>>();
 		let start = self.start();
 
-		points.iter().all(|point| point.abs_diff_eq(start, MAX_ABSOLUTE_DIFFERENCE))
+		self.get_points().all(|point| point.abs_diff_eq(start, MAX_ABSOLUTE_DIFFERENCE))
 	}
 }
 
--- a/libraries/bezier-rs/src/bezier/core.rs
+++ b/libraries/bezier-rs/src/bezier/core.rs
@@ -213,7 +213,7 @@ impl Bezier {
 	}
 
 	/// Returns true if the start, end and handles of the Bezier are all at the same location
-	pub fn is_single_point(&self) -> bool {
+	pub fn is_point(&self) -> bool {
 		let points = self.get_points().collect::<Vec<DVec2>>();
 		let start = self.start();
 
--- a/libraries/bezier-rs/src/bezier/transform.rs
+++ b/libraries/bezier-rs/src/bezier/transform.rs
@@ -158,7 +158,7 @@ impl Bezier {
 		// Verify the angle formed by the endpoint normals is sufficiently small, ensuring the on-curve point for `t = 0.5` occurs roughly in the center of the polygon.
 		let normal_0 = self.normal(TValue::Parametric(0.));
 		let normal_1 = self.normal(TValue::Parametric(1.));
-		let endpoint_normal_angle = (normal_0.x * normal_1.x + normal_0.y * normal_1.y).acos();
+		let endpoint_normal_angle = (normal_0.x * normal_1.x + normal_0.y * normal_1.y).min(1.).acos();
 		endpoint_normal_angle < SCALABLE_CURVE_MAX_ENDPOINT_NORMAL_ANGLE
 	}
 
@@ -176,10 +176,10 @@ impl Bezier {
 	/// The function takes the following parameter:
 	/// - `step_size` - Dictates the granularity at which the function searches for reducible subcurves. The default value is `0.01`.
 	///   A small granularity may increase the chance the function does not introduce gaps, but will increase computation time.
-	pub(crate) fn reduced_curves_and_t_values(&self, step_size: Option<f64>) -> (Vec<Bezier>, Vec<f64>) {
+	pub(crate) fn reduced_curves_and_t_values(&self, step_size: Option<f64>) -> (Vec<Bezier>, Vec<[f64; 2]>) {
 		// A linear segment is scalable, so return itself
 		if let BezierHandles::Linear = self.handles {
-			return (vec![*self], vec![0., 1.]);
+			return (vec![*self], vec![[0., 1.]]);
 		}
 
 		let step_size = step_size.unwrap_or(DEFAULT_REDUCE_STEP_SIZE);
@@ -192,7 +192,7 @@ impl Bezier {
 
 		// Split each subcurve such that each resulting segment is scalable.
 		let mut result_beziers: Vec<Bezier> = Vec::new();
-		let mut result_t_values: Vec<f64> = vec![extrema[0]];
+		let mut result_t_values: Vec<[f64; 2]> = vec![];
 
 		extrema.windows(2).for_each(|t_pair| {
 			let t_subcurve_start = t_pair[0];
@@ -201,7 +201,7 @@ impl Bezier {
 			// Perform no processing on the subcurve if it's already scalable.
 			if subcurve.is_scalable() {
 				result_beziers.push(subcurve);
-				result_t_values.push(t_subcurve_end);
+				result_t_values.push([t_subcurve_start, t_subcurve_end]);
 				return;
 			}
 
@@ -209,6 +209,7 @@ impl Bezier {
 			let mut segment: Bezier;
 			let mut t1 = 0.;
 			let mut t2 = step_size;
+			let mut is_prev_valid = false;
 			while t2 <= 1. + step_size {
 				segment = subcurve.trim(TValue::Parametric(t1), TValue::Parametric(f64::min(t2, 1.)));
 				if !segment.is_scalable() {
@@ -216,14 +217,21 @@ impl Bezier {
 
 					// If the previous step does not exist, the start of the subcurve is irreducible.
 					// Otherwise, add the valid segment from the previous step to the result.
-					if f64::abs(t1 - t2) >= step_size {
+					if is_prev_valid {
 						segment = subcurve.trim(TValue::Parametric(t1), TValue::Parametric(t2));
-						result_beziers.push(segment);
-						result_t_values.push(t_subcurve_start + t2 * (t_subcurve_end - t_subcurve_start));
+						if segment.is_scalable() {
+							result_beziers.push(segment);
+							result_t_values.push([t_subcurve_start + t1 * (t_subcurve_end - t_subcurve_start), t_subcurve_start + t2 * (t_subcurve_end - t_subcurve_start)]);
+						} else {
+							t2 = t1 + step_size;
+						}
 					} else {
-						return;
+						t2 = t1 + step_size;
 					}
 					t1 = t2;
+					is_prev_valid = false;
+				} else {
+					is_prev_valid = true;
 				}
 				t2 += step_size;
 			}
@@ -232,7 +240,7 @@ impl Bezier {
 				segment = subcurve.trim(TValue::Parametric(t1), TValue::Parametric(1.));
 				if segment.is_scalable() {
 					result_beziers.push(segment);
-					result_t_values.push(t_subcurve_end);
+					result_t_values.push([t_subcurve_start + t1 * (t_subcurve_end - t_subcurve_start), t_subcurve_end]);
 				}
 			}
 		});
@@ -589,7 +597,7 @@ impl Bezier {
 mod tests {
 	use super::*;
 	use crate::compare::{compare_arcs, compare_points, compare_vec_of_points};
-	use crate::utils::{Joint, TValue};
+	use crate::utils::{Cap, TValue};
 	use crate::EmptyId;
 
 	#[test]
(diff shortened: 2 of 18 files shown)


Androxium:
It looks really good! A really small nit I found was with the inner outline generated, at certain distances, it doesn't look like they connect at a point, looks like they overlap just barely. This screenshot was taken at 250% zoom so the misalignment is exaggerated. I think the current level of precision is fine, I just wanted to point this out.

Screenshot from 2023-03-19 17-37-58

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -123,6 +124,118 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+		let center_to_arc_point = arc_point - center;
+
+		(
+			left - center_to_left.perp() * HANDLE_OFFSET_FACTOR,
+			ManipulatorGroup::new(
+				arc_point,
+				Some(arc_point + center_to_arc_point.perp() * HANDLE_OFFSET_FACTOR),
+				Some(arc_point - center_to_arc_point.perp() * HANDLE_OFFSET_FACTOR),
+			),
+			right + center_to_right.perp() * HANDLE_OFFSET_FACTOR,
```
RobNadal, line 223: Could we eliminate some of the duplication between  `round_line_join(...)` and `round_cap(...)` with a helper functions that builds `left`, `right` and  the manipulator group using the `arc_point`, `center_to_arc_point` and `handle_offset_factor` arguments?

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -123,6 +124,118 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+		)
+	}
+
+	/// Returns the two manipulator groups that create a sqaure cap between the end of `self` and the beginning of `other`.
+	pub(crate) fn square_cap(&self, other: &Subpath<ManipulatorGroupId>) -> [ManipulatorGroup<ManipulatorGroupId>; 2] {
```
RobNadal, line 228: Unless I'm mistaken, I believe the SVG spec allows the distance the square cap extends beyond the outline to be configurable. Could we accomplish this by modifying the `square_cap` variant to wrap an `f64`?  Might be a good idea to add if its relatively easy (no need to include it in the UI demos)

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -505,4 +618,153 @@ mod tests {
 	}
 
 	// TODO: add more intersection tests
+
+	#[test]
+	fn round_join_counter_clockwise_rotation() {
```
RobNadal, line 623: Could you add a comment explaining the scenario this test case (and the one below) is covering? I noticed that the Subpath definitions are different and the offset distance is positive vs negative, but I don't really understand why and how that relates to being a clockwise / counter clockwise rotation

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -505,4 +618,153 @@ mod tests {
...
+
+		let offset = subpath.reverse().offset(10., utils::Join::Round);
+		let mut str = String::new();
+		offset.to_svg(&mut str, "stroke=\"black\" stroke-width=\"2\" fill=\"none\"".to_string(), String::new(), String::new(), String::new());
+		println!("{}", str);
+	}
+
+	#[test]
+	fn round_join_test_2() {
+		// TODO: Case where almost linear join gets wonky??
```
RobNadal, line 739: Did you mean to leave this comment in?

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -505,4 +618,153 @@ mod tests {
...
+		);
+
+		let offset = subpath.reverse().offset(10., utils::Join::Round);
+		let mut str = String::new();
+		offset.to_svg(&mut str, "stroke=\"black\" stroke-width=\"2\" fill=\"none\"".to_string(), String::new(), String::new(), String::new());
+		println!("{}", str);
+	}
+
+	#[test]
+	fn round_join_test_2() {
```
RobNadal, line 738: Could you give this test a more detailed name?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -326,17 +329,22 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
-	pub fn offset(&self, distance: f64, joint: Joint) -> Subpath<ManipulatorGroupId> {
+	/// The intersections of segments of the subpath are joined using the method specified by the `join` argument.
+	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#subpath/offset/solo" title="Offset Demo"></iframe>
+	pub fn offset(&self, distance: f64, join: Join) -> Subpath<ManipulatorGroupId> {
 		assert!(self.len_segments() > 1, "Cannot offset an empty Subpath.");
 
 		// An offset at a distance 0 from the curve is simply the same curve
-		if distance == 0. {
+		// An offset of a single point is not defined
+		if distance == 0. || self.len_segments() == 1 {
```
RobNadal, line 339: Did you mean `len()` here? An offset of a Subpath with a `len_segments == 1` is basically just `Bezier::offset`, so isn't it still valid?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -359,25 +367,42 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+						subpaths[i].manipulator_groups[last_index].out_handle = Some(out_handle);
+						subpaths[i].manipulator_groups.push(round_point.clone());
+						subpaths[j].manipulator_groups[0].in_handle = Some(in_handle);
 						drop_common_point[j] = false;
 					}
-					_ => unimplemented!(),
 				}
+			} else {
+				// Otherwise, default to the bevel join
+				drop_common_point[j] = false;
```
RobNadal, line 405: We seem to be setting `drop_common_point[j] = false` in every arm of the if-statement and match-block. Do we even need this boolean array anymore? The original goal of this boolean array was to drop duplicate points in cases where the outline lined up perfectly (ie. no joining / clipping necessary). Could you check to make sure we're still doing that?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -509,6 +576,43 @@ mod tests {
...
+					out_handle: None,
+					in_handle: None,
+					id: EmptyId,
+				},
+			],
+			false,
+		);
+
+		let outline = subpath.outline(10., crate::Join::Round, crate::Cap::Round).0;
+		assert_eq!(outline.len(), 25);
```
RobNadal, line 612: This feels like it could end up being somewhat flaky. Could you instead assert that none of the consecutive manipulator groups fall on top of one another?

libraries/bezier-rs/src/utils.rs
```diff
@@ -29,10 +29,17 @@ pub enum SubpathTValue {
 }
 
 #[derive(Copy, Clone)]
```
RobNadal, line 31: Could you insert a short comment about this enum's pupose (connects discontinuities in outline / offsets) and then insert a link to the SVG reference? (and do the same for `Cap`)

RobNadal (commented):
Thanks for all the effort you put into this feature @hannahli2010, looks amazing!

libraries/bezier-rs/src/subpath/transform.rs
hannahli2010: The condition on line 360 should still be catching that right?

libraries/bezier-rs/src/subpath/solvers.rs
hannahli2010: I don't think that is part of the SVG spec based on this: https://www.w3.org/TR/SVG2/painting.html#LineCaps.
However, I realized that I did not handle the single point case correctly so I fixed and added tests for that!

RobNadal: Looks good!

libraries/bezier-rs/src/subpath/transform.rs
RobNadal: Can you extract the `drop_common_point[j] = false;` to the top of the if-statement, since you're setting it true in every arm of the match statement anyway?

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -404,44 +422,48 @@ impl Bezier {
 	/// The 'caps', the linear segments at opposite ends of the outline, intersect the original curve at the midpoint of the cap.
 	/// Outline takes the following parameter:
 	/// - `distance` - The outline's distance from the curve.
-	/// <iframe frameBorder="0" width="100%" height="375px" src="https://graphite.rs/bezier-rs-demos#bezier/outline/solo" title="Outline Demo"></iframe>
-	pub fn outline<ManipulatorGroupId: crate::Identifier>(&self, distance: f64) -> Subpath<ManipulatorGroupId> {
-		let first_segment = self.offset(distance);
-		let third_segment = self.reverse().offset(distance);
+	/// <iframe frameBorder="0" width="100%" height="400px" src="https://graphite.rs/bezier-rs-demos#bezier/outline/solo" title="Outline Demo"></iframe>
+	pub fn outline<ManipulatorGroupId: crate::Identifier>(&self, distance: f64, cap: Cap) -> Subpath<ManipulatorGroupId> {
+		let (first_segment, third_segment) = if self.is_single_point() {
```
RobNadal, line 427: In `subpath::transform` I think we called these variables `pos_offset, neg_offset`, which I think I like better. Could we use those names here to be consistent?

libraries/bezier-rs/src/bezier/core.rs
```diff
@@ -201,6 +201,14 @@ impl Bezier {
 
 		self_points.len() == other_points.len() && self_points.into_iter().zip(other_points.into_iter()).all(|(a, b)| a.abs_diff_eq(b, max_abs_diff))
 	}
+
+	/// Returns true if the start, end and handles of the Bezier are all at the same location
+	pub fn is_single_point(&self) -> bool {
```
RobNadal, line 206: Tiny nit: if 'point' is singular anyway, saying 'single point' feels redundant. Could we change this function name to `is_point`?

libraries/bezier-rs/src/subpath/transform.rs
```diff
@@ -359,25 +372,42 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+						subpaths[i].manipulator_groups[last_index].out_handle = Some(out_handle);
+						subpaths[i].manipulator_groups.push(round_point.clone());
+						subpaths[j].manipulator_groups[0].in_handle = Some(in_handle);
 						drop_common_point[j] = false;
 					}
-					_ => unimplemented!(),
 				}
+			} else {
+				// Otherwise, default to the bevel join
+				// drop_common_point[j] = false;
```
RobNadal, line 410: Dead branch

libraries/bezier-rs/src/bezier/transform.rs
```diff
@@ -201,29 +201,37 @@ impl Bezier {
...
+						if segment.is_scalable() {
+							result_beziers.push(segment);
+							result_t_values.push([t_subcurve_start + t1 * (t_subcurve_end - t_subcurve_start), t_subcurve_start + t2 * (t_subcurve_end - t_subcurve_start)]);
+						} else {
+							t2 = t1 + step_size;
+						}
 					} else {
-						return;
+						t2 = t1 + step_size;
 					}
```
Keavon, line 230: You should probably be able to combine these else cases.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -123,6 +124,92 @@ impl<ManipulatorGroupId: crate::Identifier> Subpath<ManipulatorGroupId> {
...
+	pub(crate) fn miter_line_join(&self, other: &Subpath<ManipulatorGroupId>) -> Option<ManipulatorGroup<ManipulatorGroupId>> {
+		let in_segment = self.get_segment(self.len_segments() - 1).unwrap();
+		let out_segment = other.get_segment(0).unwrap();
+		let in_tangent = in_segment.tangent(TValue::Parametric(1.));
+		let out_tangent = out_segment.tangent(TValue::Parametric(0.));
+
+		let normalized_in_tangent = in_tangent.normalize();
+		let normalized_out_tangent = out_tangent.normalize();
+
+		// The tangents must not be parallel for the miter join
```
Keavon, line 138: Can you also implement https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stroke-miterlimit please? We need this in Graphite.

Keavon (approved):
Please resolve the comments and merge conflicts, then you're good to merge. Nice work!

libraries/bezier-rs/src/bezier/transform.rs
hannahli2010: To combine them, I'd want to refactor as:
```
if is_prev_valid && segment.is_scalable() { .. }
```
However, that would involve computing `subcurve.trim(...)` every time despite the fact that it isn't always necessary. Is this alright to do?

website/other/bezier-rs-demos/src/features/subpath-features.ts
```diff
@@ -129,6 +130,8 @@ const subpathFeatures = {
 				step: 1,
 				default: 10,
 			},
+			joinOptions,
+			capOptions,
```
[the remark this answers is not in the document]
hannahli2010: image

libraries/bezier-rs/src/bezier/core.rs
```diff
@@ -201,6 +201,14 @@ impl Bezier {
 
 		self_points.len() == other_points.len() && self_points.into_iter().zip(other_points.into_iter()).all(|(a, b)| a.abs_diff_eq(b, max_abs_diff))
 	}
+
+	/// Returns true if the start, end and handles of the Bezier are all at the same location
+	pub fn is_point(&self) -> bool {
+		let points = self.get_points().collect::<Vec<DVec2>>();
```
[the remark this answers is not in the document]
hannahli2010: Very good catch! I will change that.

libraries/bezier-rs/src/bezier/transform.rs
Keavon: I noticed it might require rearrangements that would hurt performance, but I thought I'd ask in case you could find another less obvious way that didn't occur to me. If you think this is the only way to arrange that logic without causing extra computational cost, it's fine to leave it as is.

libraries/bezier-rs/src/subpath/solvers.rs
hannahli2010: I'm planning to implement this in a different PR so that we can merge this one first.

I'm thinking of adding this feature via an additional optional param on `offset` and `outline` - does that sound alright to you @Keavon ?

PR #1106, opened 2023-04-09
Add cache clearing to stop the memory leak

### LINKED ISSUE #1103: Sending rasterized images from JS to Rust runs out of wasm memory
When dragging a node graph frame around which now rasterizes every frame thanks to #1100, the JS is sending lots and lots of binary data to the Rust code. This also happens when repeatedly using Imaginate to generate many images (including the in-progress generation preview images) after enough generations (like 50 or more). It ends up using more memory than the wasm allocator can handle, since apparently we're not freeing the memory or something. It results in a freeze of the editor and spamming the console.

We run into this issue with wasm-bindgen where we use up the 2GB of memory and JS thinks the pointer is negative: <https://github.com/rustwasm/wasm-bindgen/issues/2388>. That was fixed six weeks ago in <https://github.com/rustwasm/wasm-bindgen/pull/3310> but doesn't seem to have made it into a release yet. It doubles the available memory to 4GB which is the maximum until [wasm memory64 becomes available](https://webassembly.org/roadmap/#:~:text=%5Bb%5D-,Memory64,-%5Ba%5D). Once we can use the full 4GB, we should also add proper error handling to tell the user what has happened if memory has been all used up so it doesn't result in a mysterious ceasing to work.

Increasing that allocation limit would be important, but more importantly, we need to fix the cause of the problem which is that memory seems to not be freed, or at least not be available again for use by the allocator.

0HyperCube:
Clears the node graph cache between runs, leading to reduced memory usage.

Closes #1103

### DIFF
--- a/node-graph/gstd/src/memo.rs
+++ b/node-graph/gstd/src/memo.rs
@@ -2,6 +2,8 @@ use graphene_core::Node;
 
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
+use std::pin::Pin;
+use std::sync::atomic::AtomicBool;
 use xxhash_rust::xxh3::Xxh3;
 
 /// Caches the output of a given Node and acts as a proxy
@@ -9,7 +11,7 @@ use xxhash_rust::xxh3::Xxh3;
 pub struct CacheNode<T, CachedNode> {
 	// We have to use an append only data structure to make sure the references
 	// to the cache entries are always valid
-	cache: boxcar::Vec<(u64, T)>,
+	cache: boxcar::Vec<(u64, T, AtomicBool)>,
 	node: CachedNode,
 }
 impl<'i, T: 'i, I: 'i + Hash, CachedNode: 'i> Node<'i, I> for CacheNode<T, CachedNode>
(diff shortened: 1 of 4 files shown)


node-graph/gstd/src/memo.rs
```diff
@@ -72,8 +82,16 @@ impl<'i, T: 'i + Hash> Node<'i, Option<T>> for LetNode<T> {
...
+			self.cache = boxcar::vec![last];
+		}
+	}
 }
 
+impl<T> std::marker::Unpin for LetNode<T> {}
+
 impl<T> LetNode<T> {
 	pub fn new() -> LetNode<T> {
 		LetNode { cache: boxcar::Vec::new() }
```
TrueDoctor, line 97: This is actually the most relevant part, we can basically replace the box car with a single unsafe cell because this should during the runtime of the graph only ever have one value which can be thrown away at the end

TrueDoctor: Nevermind didn't see this was already implemented for the cache node

node-graph/interpreted-executor/src/executor.rs
```diff
@@ -107,36 +107,42 @@ impl BorrowTree {
...
-		let output = node.node.eval(Box::new(input));
+		let reader = node.read().unwrap();
+		let output = reader.node.eval(Box::new(input));
 		dyn_any::downcast::<O>(output).ok().map(|o| *o)
 	}
-	pub fn eval_any<'i, 's: 'i>(&'s self, id: NodeId, input: Any<'i>) -> Option<Any<'i>> {
+	pub fn eval_any<'i>(&'i self, id: NodeId, input: Any<'i>) -> Option<Any<'i>> {
 		let node = self.nodes.get(&id)?;
-		Some(node.node.eval(input))
+		Some(unsafe { (*((&*node.read().unwrap()) as *const NodeContainer)).node.eval(input) })
```
TrueDoctor, line 145: Oof I dislike the evaluation being an unsafe operation but I guess its fine because it only is a lifetime extension

TrueDoctor (approved):
We should ideally let miri run on a test that evaluates the nodegraph multiple times to check if this contains any subtle UB but this looks fine for now

node-graph/interpreted-executor/src/executor.rs
0HyperCube: That is a big ugly - I wasn't quite sure how to resolve this though.

PR #1153, opened 2023-04-23
Channel seperator node

isiko:
Adds a node for separating the alpha and RGB channels of an image. This PR also adds some Blend modes to recombine the separate channels, those are intended as a temporary solution and will be combined into a new node soon.


isiko:
(Don't Merge yet, there are some weird WASM errors that don't show up with cargo check)

Keavon:
Was the plan to also add channel combine nodes in this PR? (A separate PR would be fine too if you prefer that.) And are there any further improvements planned in this PR?

isiko:
I just rebased the branch, if the checks all pass this is ready for review, I'll add the 'combine' node in a new PR.
Also, I think the appearance of the node could maybe be improved, as it has 3 equally important outputs and no 'main' output, which makes it not that intuitive to use. It would be nice if you could have a look at that, as I'm not that involved in the front-end part of things.

Keavon:
Thanks for pointing out the fact that its input/output structure is rather ugly, I forgot to actually test it and comment on that! We want to have zero primary outputs and three (R/G/B) return value (i.e. non-primary) outputs.

We also want to name this "Split Channels". And what do you think about adding a fourth channel for alpha? The channel outputs themselves should probably be shortened to just "Red", "Green", "Blue", (and "Alpha"?).

Keavon:
As discussed on Discord, please also remove the three checkbox parameters and make the data output be the Luma type instead of the Color type, and add a "Luma to Color" node which takes the one channel and splats it to three channels (resulting visually in a gray color).

And since there's a single issue for this (#1169), would you actually mind doing the "Combine Channels" node in this PR too? I realize that'll delay its merge but it would help with organization. If you still prefer to merge this first, just say so and we can do that anyways :)

Keavon:
To summarize the desired functionality for the polished version of this PR feature set:
- Split Channels node. One primary input (Color), zero primary outputs. Zero parameter inputs, three (or four?) returned outputs named "Red", "Green", "Blue" of type Luma. Consider also including "Alpha" as a fourth.
- Combine Channels node. Zero primary inputs, one primary output (Color). Three (or four?) parameter inputs taking Luma parameter inputs named "Red", "Green", "Blue" (and "Alpha"?), zero returned outputs.
- Luma to Color node. One primary input (Luma), one primary output (Color). Splats the single channel to the three channels and gives an alpha value of 1.

isiko:
> We also want to name this "Split Channels". And what do you think about adding a fourth channel for alpha? The channel outputs themselves should probably be shortened to just "Red", "Green", "Blue", (and "Alpha"?).

Should be all done :+1:

node-graph/gstd/src/raster.rs
```diff
@@ -174,6 +174,65 @@ fn compute_transformed_bounding_box(transform: DAffine2) -> Bbox {
...
+
+	// Transforms a point from the background image to the forground image
+	let bg_to_fg = image.transform() * DAffine2::from_scale(1. / image_size);
+	let area = bg_to_fg.transform_point2(DVec2::new(1., 1.)) - bg_to_fg.transform_point2(DVec2::ZERO);
+
+	for y in 0..image.height() {
+		for x in 0..image.width() {
+			let image_point = DVec2::new(x as f64, y as f64);
+			let mut mask_point = bg_to_fg.transform_point2(image_point);
+			let local_mask_point = stencil.transform().inverse().transform_point2(mask_point);
```
nat-rix, line 218: Calculating the inverse of an affine transformation is a bit costly, so you probably want to do that before the loop.

isiko: Fixed :+1:

Keavon:
- [ ] *Combine Channels* shouldn't have a primary input
- [ ] *Split Channels* shouldn't have a primary output
- [ ] I think it'd be best to also add Alpha to both
- [ ] With the same reasoning, *Extract Alpha* can be a choice in the *Extract Channel* node instead of being a separate node

I also ran into a crash, twice, involving the usage of the *Extract Alpha* node but I can't seem to be able to find a reproducible case. Here's its stack trace:
image

node-graph/gstd/src/raster.rs
```diff
@@ -174,6 +174,66 @@ fn compute_transformed_bounding_box(transform: DAffine2) -> Bbox {
...
+	let image_size = DVec2::new(image.width() as f64, image.height() as f64);
+	let mask_size = stencil.transform().decompose_scale();
+
+	if mask_size == DVec2::ZERO {
+		return image;
+	}
+
+	// Transforms a point from the background image to the forground image
+	let bg_to_fg = image.transform() * DAffine2::from_scale(1. / image_size);
+	let area = bg_to_fg.transform_point2(DVec2::new(1., 1.)) - bg_to_fg.transform_point2(DVec2::ZERO);
```
nat-rix, line 212: In my understanding of maths this whole line is just the same as:
```rust
let area = bg_to_fg.transform_vector2(DVec2::ONE);
```
(`.transform_vector2` is an alias of `.matrix2 *` and ignores the translation)

TrueDoctor:
> With the same reasoning, Extract Alpha can be a choice in the Extract Channel node instead of being a separate node
 
 But not every colour has alpha so it makes sense to model the proto nodes after the color traits

isiko:
> * [ ]  _Combine Channels_ shouldn't have a primary input
> * [ ]  _Split Channels_ shouldn't have a primary output

I think I talked to @TrueDoctor about this when I implemented the split node, and if I remember correctly he said that it's not really possible to do this from the backend (the first input/output is always interpreted as main I/O by the frontend). I could maybe try to figure out where in the code this would need to be added, but I would assume that would be beyond the scope of this PR (and also probably my knowledge of the codebase).

> * [ ]  I think it'd be best to also add Alpha to both

That's actually already implemented: The 'main' I/O represents the Alpha Channel

> * [ ]  With the same reasoning, _Extract Alpha_ can be a choice in the _Extract Channel_ node instead of being a separate node

Initially, I had implemented it that way, but @TrueDoctor suggested splitting it into a separate node. While I can combine them again, it would make the code less elegant. Currently, I use the `RedGreenBlue` Enum to determine the channel for extraction/insertion. If I add the alpha channel, I would need to create a new Enum with the same value + Alpha.

> I also ran into a crash, twice, involving the usage of the _Extract Alpha_ node but I can't seem to be able to find a reproducible case. Here's its stack trace: image

I'll look into that :+1:

node-graph/gstd/src/raster.rs
isiko: Done in next Push :+1:

TrueDoctor:
> I think I talked to @TrueDoctor about this when I implemented the split node, and if I remember correctly he said that it's not really possible to do this from the backend (the first input/output is always interpreted as main I/O by the frontend). I could maybe try to figure out where in the code this would need to be added, but I would assume that would be beyond the scope of this PR (and also probably my knowledge of the codebase).

Not having a primary input is sort of possible by just setting the primary input to the empty tuple in the document node definition but I don't think that is possible for primary outputs atm. I guess the only difference is that the primary output is not named and the others are but from a code perspective they are identical

Keavon:
> That's actually already implemented: The 'main' I/O represents the Alpha Channel

I don't understand what you mean by that, could you please rephrase or elaborate? To clarify my intent, I am saying that the "Alpha" channel should be output as a fourth parameter and input as a fourth parameter so the user can shuffle it around like the other channels.

nat-rix:
test
I am a bit confused what the channel split and combine nodes do here.

1. Why is the result of only connecting the red channel from the input image with the red channel of the output image a grayscale? Shouldn't it be a red image?
2. When I do the same with blue or/and green instead, the image goes blank
3. The original image (left) and the image with all channels splitted and merged again (right) somehow aren't exactly the same. If you take a close look at the zoomed in image section, you'll notice some artifacts like the red "dots" at the border of the petal. Could it be that some color information gets lost in this process?

Keavon:
> Shure; The Split should have 4 Outputs. 3 Labeled Red, Green, and Blue and one unlabeled one (the Primary Output). For Images without an Alpha Channel, this outputs a white Image. The same happens with the Combine node, the unlabeled (primary) Input uses a Mask Node behind the scenes to insert an Alpha Channel into the Image. You can see this by inserting some data into the Alpha Channel of an Image using the Mask Node and then feeding the Result into a Split Node.

I see, thanks for explaining. Unfortunately that's rather unintuitive to make the alpha channel use the primary input/output though. I'd rather we lablel it "Alpha" and place it after the RGB channels.

isiko:
> I'd rather we lablel it "Alpha" and place it after the RGB channels.

Shure, but then the `Red` Channel would become the primary I/O, indicating a difference between the `Red` Channel and the `Green`/`Blue` Channel. If that is fine I can implement that change.

Keavon:
I was suggesting that the primary input/output be removed entirely from both nodes.

isiko:
> I was suggesting that the primary input/output be removed entirely from both nodes.

Yeah, but I thought that wasn't possible at the Moment? At least that's what I got from [this](https://github.com/GraphiteEditor/Graphite/pull/1153#issuecomment-1547711302) message.

Keavon:
From that, it sounds like a lack of primary input is possible but not a lack of primary output. I'd suggest implementing that and then we file a followup issue to support optional primary outputs and a removal of the node's primary output (which should be set to output nothing).

isiko:
The IO stuff should be implemented now, I have a solution for the Issues nat-rix brought up and am currently implementing that (Should be done soonTM, the biggest hurdle currently is figuring out how the `node_registry` works)

nat-rix:
This is great!
Btw, when connecting the green channel of split channels with the green channel of combine channels, the editor crashes with the assertion:
image

isiko:
> When connecting the green channel of split channels with the green channel of combine channels, the editor crashes with the assertion

Thanks, I know what's causing that, that's also going to be fixed by the stuff I'm doing in the node_registry

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -371,6 +371,35 @@ fn static_nodes() -> Vec<DocumentNodeType> {
 			outputs: vec![DocumentOutputType::new("Image", FrontendGraphDataType::Raster)],
 			properties: node_properties::mask_properties,
 		},
+		DocumentNodeType {
+			name: "Insert Channel",
+			category: "Image Adjustments",
+			identifier: NodeImplementation::proto("graphene_std::raster::InsertChannelNode<_, _, _, _>"),
+			inputs: vec![
+				DocumentInputType::value("Image", TaggedValue::ImageFrame(ImageFrame::empty()), true),
+				DocumentInputType::value("Stencil", TaggedValue::ImageFrame(ImageFrame::empty()), true),
```
Keavon, line 380: Rename Stencil to Replacement

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -371,6 +371,35 @@ fn static_nodes() -> Vec<DocumentNodeType> {
...
+				DocumentInputType::value("ColorChannel", TaggedValue::RedGreenBlue(RedGreenBlue::Red), false),
```
Keavon, line 381: Rename ColorChannel (shown here, and "Target Channel" shown in the Properties panel) to Replace.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -483,6 +512,86 @@ fn static_nodes() -> Vec<DocumentNodeType> {
 			outputs: vec![DocumentOutputType::new("Image", FrontendGraphDataType::Raster)],
 			properties: node_properties::luminance_properties,
 		},
+		DocumentNodeType {
+			name: "Extract Channel",
+			category: "Image Adjustments",
+			identifier: NodeImplementation::proto("graphene_core::raster::ExtractChannelNode<_>"),
+			inputs: vec![
+				DocumentInputType::value("Image", TaggedValue::ImageFrame(ImageFrame::empty()), true),
+				DocumentInputType::value("Channel", TaggedValue::RedGreenBlue(RedGreenBlue::Red), false),
```
Keavon, line 521: Rename Channel/"Target Channel" (inconsistency) to **From**.

PR #1159, opened 2023-04-24
Fix too many locals in wasm binary

nat-rix:
Split function into smaller non-inlinable ones.
This fixes the issue that rust creates too many wasm locals in non-optimized builds. That lead to a [compile error](https://github.com/rustwasm/wasm-pack/issues/981) in such builds.

### DIFF
--- a/node-graph/interpreted-executor/src/node_registry.rs
+++ b/node-graph/interpreted-executor/src/node_registry.rs
@@ -68,48 +68,59 @@ macro_rules! register_node {
 	};
 }
 macro_rules! raster_node {
-	($path:ty, params: [$($type:ty),*]) => {
-		vec![
-		(
-			NodeIdentifier::new(stringify!($path)),
-			|args| {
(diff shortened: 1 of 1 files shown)


node-graph/interpreted-executor/src/node_registry.rs
```diff
@@ -68,48 +68,56 @@ macro_rules! register_node {
...
-				let params = vec![$(value_fn!($type)),*];
-				NodeIOTypes::new(concrete!(ImageFrame<Color>), concrete!(ImageFrame<Color>), params)
-			},
-		)
-		]
-	}
+	($path:ty, params: [$($type:ty),*]) => {{
+		// this function could also be inlined but serves as a workaround for
+		// [wasm-pack#981](https://github.com/rustwasm/wasm-pack/issues/981).
+		// The non-inlining function leads to less locals in the resulting wasm
```
TrueDoctor, line 74: ~less~ -> fewer

TrueDoctor (approved):

PR #1160, opened 2023-04-25
Upgrade npm dependencies and add "@parcel/config-default"

iacore:
"@parcel/config-default" is needed to run on other package managers like pnpm even though it seems to have worked okay on regular npm.

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -17,9 +17,9 @@
 				"@parcel/transformer-inline-string": "^2.8.3",
 				"@parcel/transformer-webmanifest": "^2.8.3",
 				"@types/license-checker-webpack-plugin": "^0.2.1",
-				"@types/node": "^18.16.1",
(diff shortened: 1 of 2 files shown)


Keavon:
@TrueDoctor can you help figure out what's wrong with the new GitHub Actions script/config which is causing CI to fail on this?

TrueDoctor:
I think this might be due to the pr not originating from our repo, hence we don't have access to post comments to that repos pr

TrueDoctor:
@iacore  you need to include your `package-lock.json` I tried to push mine but that seems to fail on the ci. You can try to push your version to see if that works

iacore:
This is npm being strict.

https://github.com/HellButcher/parcel-transformer-svelte3-plus/pull/4

Keavon:
Looks like that's updated and published!

Keavon (approved):
Thanks

PR #1165, opened 2023-04-26
Add monitor node for graph introspection

TrueDoctor:
Adds a `serialize` function on Node which can be implemented by nodes to
allow introspecting their content. This pr also adds a Monitor Node that
always caches the last value it was evaluated with.

Test Plan: -

### DIFF
--- a/node-graph/gstd/src/memo.rs
+++ b/node-graph/gstd/src/memo.rs
@@ -1,9 +1,11 @@
 use graphene_core::Node;
+use serde::{Deserialize, Serialize};
 
 use std::hash::{Hash, Hasher};
 use std::marker::PhantomData;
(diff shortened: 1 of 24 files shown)


node-graph/gstd/src/memo.rs
```diff
@@ -49,6 +51,39 @@ impl<T, CachedNode> CacheNode<T, CachedNode> {
 	}
 }
 
+/// Caches the output of a given Node and acts as a proxy
+#[derive(Default)]
+pub struct MonitorNode<T, CachedNode> {
+	// We have to use an append only data structure to make sure the references
```
0HyperCube, line 57: Outdated comment.

0HyperCube (approved):

PR #1182, opened 2023-04-27
Fix #1174: Spline does not have mirror handles enabled

### LINKED ISSUE #1174: Spline does not have mirror handles enabled
A shape produced by the spline tool does not have mirror handles enabled, meaning that when you switch to the path tool and drag a handle, the opposite one does not automatically move. The expected behaviour is that it should remain mirrored, similar to when a similar action is taken with an ellipse. To resolve this, the mirror handle setting should be enabled when creating a spline in `spline_tool.rs`, you can use `ellipse_tool.rs` as reference for setting the mirror.

Chase-Percy:
Fixes #1174

---

Sets the spline mirror angle handles to true by default in the add_spline function.


editor/src/messages/tool/tool_messages/spline_tool.rs
```diff
@@ -264,8 +265,20 @@ fn add_spline(tool_data: &SplineToolData, global_tool_data: &DocumentToolData, s
...
+	// Set the manipulator groups to have their handle angles mirrored by default
+	for manipulator_group in manipulator_groups {
+		responses.add(GraphOperationMessage::Vector {
+			layer: layer_path.clone(),
+			modification: VectorDataModification::SetManipulatorHandleMirroring {
+				id: manipulator_group.id,
+				mirror_angle: true,
+			},
+		});
+	}
```
Keavon, line 280: @0HyperCube this is now duplicate code in two places (also ellipse_tool.rs:135), is there a good place this could be extracted to to avoid the duplication? @Chase-Percy you're also welcome to try and find somewhere before he responds with a suggestion.

Keavon (approved):
I haven't gotten to test it yet but this code looks nice and clean!

editor/src/messages/tool/tool_messages/spline_tool.rs
0HyperCube: Perhaps `editor/src/messages/tool/common_functionality/graph_modification_utils.rs`?

0HyperCube (approved):
Looks good.

editor/src/messages/tool/tool_messages/spline_tool.rs
Chase-Percy: Sure, I can refactor it into there if you would like to go ahead with that.

Chase-Percy: I wasn't 100% sure how to name the function without it being massive. Feel free to suggest something else. [243f37b](https://github.com/GraphiteEditor/Graphite/pull/1182/commits/243f37b8fc2497729d1eddaa71371067b60bb6b5)

0HyperCube: Perhaps `mirror_all_handles`?

PR #1184, opened 2023-04-27
Implement the Brush without relying on a stamp texture

TrueDoctor:
Test Plan: Test the BrushNode in the editor


TrueDoctor:
@Keavon just measured a performance improvement of 400ms -> 187ms with this pr, looking promising

TrueDoctor:
The debug statement should remain, it can be used to track down unnecessary clones

Keavon:
We can add it back as a `trace!()` or `warn!()` then, but we don't commit `debug!()` anywhere in our code besides that one.

TrueDoctor:
> We can add it back as a `trace!()` or `warn!()` then, but we don't commit `debug!()` anywhere in our code besides that one.

The Debug node is as its name implies basically a debug statement itself and usages of it should never be committed, but the node itself is fine

node-graph/gstd/src/brush.rs
```diff
@@ -140,7 +133,7 @@ fn erase(input: (Color, Color), flow: f64) -> Color {
 	Color::from_unassociated_alpha(input.r(), input.g(), input.b(), alpha)
 }
 
-#[node_fn(BrushTextureNode)]
+#[node_fn(BrushStampGeneratorNode)]
 fn brush_texture(diameter: f64, color: Color, hardness: f64, flow: f64) -> BrushStampGenerator<Color> {
```
Keavon, line 137: Rename to `fn brush_stamp_generator_node(...`

Keavon (approved):
Great perf improvement!

PR #1189, opened 2023-04-29
Hide edge aliasing of hard brushes with a blur

TrueDoctor:
Implement analytical AA around the edges of the brush.

Test Plan: Draw a stamp with varying hardness and look at the results


TrueDoctor:
image

Keavon (approved):

PR #1190, opened 2023-04-29
Implement Infrastructure to reuse previous frames for brush drawing

TrueDoctor:
Implement Infrastructuro to reuse the previous evaluation of the
node graph to blend the new stroke with instead of drawing the
entire trace from scratch.
This does not transition to a blending based approach because that still
caused regressions but allows the brush node to work with input data
natively.

Test Plan:
- Use the brush tool in the editor and check for regressions
- Evaluate the performance


node-graph/gstd/src/memo.rs
```diff
@@ -53,19 +54,17 @@ impl<T, CachedNode> CacheNode<T, CachedNode> {
...
+#[cfg(feature = "serde")]
+impl<'i, T: 'i + Serialize + Clone> Node<'i, T> for MonitorNode<T> {
 	type Output = T;
-	fn eval(&'i self, input: I) -> Self::Output {
-		let output = self.node.eval(input);
-		*self.output.lock().unwrap() = Some(output.clone());
-		output
+	fn eval(&'i self, input: T) -> Self::Output {
+		*self.output.lock().unwrap() = Some(input.clone());
+		log::debug!("Initializing monitor node with");
```
Keavon, line 66: Remove

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -106,6 +106,18 @@ fn static_nodes() -> Vec<DocumentNodeType> {
...
+			name: "Monitor",
+			category: "General",
+			identifier: NodeImplementation::proto("graphene_core::ops::IdNode"),
+			inputs: vec![DocumentInputType {
+				name: "In",
+				data_type: FrontendGraphDataType::General,
+				default: NodeInput::value(TaggedValue::None, true),
+			}],
+			outputs: vec![DocumentOutputType::new("Out", FrontendGraphDataType::General)],
+			properties: |_document_node, _node_id, _context| node_properties::string_properties("The Monitor node stores the value of it's last evaluation"),
```
Keavon, line 119: `it's` -> `its`

Keavon (changes requested):
Perf problems as discussed in Discord. Namely:
- [x] Bounding box includes document origin
- [x] Releasing the click makes the stroke get drawn twice, appearing more bold
- [ ] Sometimes strokes disappear after being drawn
- [ ] Slowness shouldn't grow with the length of a (continuous) brush stroke, since it should reuse the previous frame as this PR title claims
- [ ] Slowness shouldn't grow with the size of the drawing area, since it should reuse the previous frame (exception: growing the bounding box requires a new allocation and blitting operation, that's reasonable to slow down in that case)
- [ ] We don't want the user to have to ever see the Monitor node

PR #1196, opened 2023-05-03
Fix "Failed to execute" dialog showing up at times

TrueDoctor:
Only remove orphaned nodes when they are still orphans in the current
execution iteration.

Test Plan:
Try to cause the "Failed to execute" bug and see if it
persists


Keavon (approved):
🎉

PR #1208, opened 2023-05-11
Fix node copy paste issues

Chase-Percy:
- Allows input frame nodes to be copied.
- Fixes pasting of the output node. It was looping forever since the while loop was dependent on an iter.all() method, but all() always returns true on an empty iter.
- Fixes the duplication (ctrl-D) of nodes.


Keavon:
It's looking like the Input Frame node, if you duplicate it, shows an error dialog if you try to hook it up while hooking it back up to the original works fine. Is that something you could look into please, if you can figure it out based on what True Doctor mentioned in Discord? Thanks!

Keavon (approved):

PR #1209, opened 2023-05-11
Decouple node graph execution

TrueDoctor:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1773,6 +1773,7 @@ dependencies = [
  "wasm-bindgen",
  "wasm-bindgen-futures",
  "wasm-bindgen-test",
+ "web-sys",
 ]
 
 [[package]]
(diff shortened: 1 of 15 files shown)


Keavon:
Regressions:

- [ ] Pen tool: this is happening, but I can't seem to figure out when it does or doesn't occur ([video](https://files.keavon.com/-/BlackDazzlingAmericanmarten/capture_9_.mp4))
- [ ] Pen tool: hitting <kbd>Esc</kbd> doesn't clear the curve segment extension that was in-progress before being cleared
- [ ] Path tool: double clicking a point to toggle smooth/sharp doesn't update its overlays
- [ ] Path tool: holding <kbd>Alt</kbd> while dragging an anchor or handle causes it to move at a different rate and get offset from the cursor ([video](https://files.keavon.com/-/CostlyCompleteXenoposeidon/capture_10_.mp4))

Keavon (approved):

PR #1214, opened 2023-05-12
Curves image adjustment node

nat-rix:


Keavon:
Looks like this is taking shape and starting to work now! I'm noticing that it breaks (and throws a console error) after a little while of dragging the points around, at which point the laggy UI stops being laggy (since it's not doing anything anymore from then on).

nat-rix:
> I'm noticing that it breaks (and throws a console error) after a little while of dragging the points around

Could you please provide an error log? I can't reproduce that.

Keavon:
> Could you please provide an error log? I can't reproduce that.

It should definitely just happen by using it for a bit, maybe try and be aggressive and drag it around a lot every frame. I'm using Chrome in case that makes any difference. I'm adding this downstream of the Downscale node for a 4000x3000 image. The only error it prints is this:

image

Note, that is not referring to hitting a Rust `unreachable!();` macro, but rather [this wasm instruction](https://developer.mozilla.org/en-US/docs/WebAssembly/Reference/Control_flow/unreachable).

nat-rix:
Huh, interesting. This looks like an out of memory error. My first intuition is that this happens, because too many non-completed events stack up in the event queue. But I'll have
a look into this a bit later

nat-rix:
The bug is also happening on master when using e.g. the blend node. I think this is probably "normal", when storing such a huge image uncompressed and then making clones of it. But to be sure, I'll investigate further, which kind of `TaggedValue` is cloned there for which reason to produce an oom.

Keavon:
Okay, good to know it's not an issue with the curves node, just a symptom of making frequent adjustments. Weird that the Downscale node isn't mitigating. I'll use smaller images to test from now on. Thanks for creating an issue for that.

TrueDoctor:
I've updated the branch to use the new master, I think it makes sense to merge this pretty soon (even if it is not completely done) to prevent further code rot

TrueDoctor:
This does acutally work surprisingly well, I can fix the tests and I think we might be good to merge then

Keavon:
I built this branch and don't see the Curves node in the layer add menu, do you know what's up with that?

TrueDoctor:
@nat-rix do you know what would most likely be the culprit for the failing tests?

TrueDoctor:
> I built this branch and don't see the Curves node in the layer add menu, do you know what's up with that?

The build link works, can you retry that?

TrueDoctor:
@Keavon It is unfortunate you had to revert the brightness contrast changes. It would be very good if we could reuse that code instead of duplicating the entire file + we lose the improvements to the code structure

Keavon:
Yes, but that'll have to be a future contributor's task since the more important priority right now is just getting this merged in its current (plus slightly fixed up) state. You reminded me that I forgot to commit the TODO comment about replacing that someday.

Keavon (approved):

Keavon:
Hi @nat-rix, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

Keavon:
Hi @nat-rix, we're grateful of your past code contribution to Graphite, but unless we hear from you by the **end of this weekend** with agreement about relicensing your code to include the MIT license, we will begin steps to remove your code contributions **so that you will no longer be a Graphite code contributor**. We'd much rather avoid this outcome, so please respond now.

Issue #1171, opened 2023-04-26
Replace legacy artboards with artboard nodes


Keavon:
This is an example in our mockup of how artboards will work, with an Artboard node feeding into a layer, and the layer's vertical output being given to the graph's output:

![image](https://user-images.githubusercontent.com/4388688/234689702-e698978b-6da0-4902-9c21-a1f7dbca87bb.png)

```[tasklist]
- [x] #1205 - Add an Artboard node which takes `GraphicElement` input and produces `Artboard` or `ArtboardGroup` output data.
- [x] #1224 - These node-based artboards will need to live in a new node graph attached directly to the document itself, rather than being part of any one layer. This special node graph will become the singular node graph representing a whole document once the transition away from legacy layers is complete.
- [x] #1221 - The output from that special node graph is of type `ArtboardGroup` and it should enter the SVG output renderer and become an appropriately colored background rectangle as well as a definition for a [clipping mask](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Clipping_and_masking). (A checkered transparency background (defined in CSS) should also be included in the viewport, but in a way that causes it to be excluded from things that rasterize the artboard.)
- [x] #1222 - The Artboard tool will also need to be updated to work with node-based artboards, plus any editor features like the New Document dialog and *Document* > *Clear Artboards*.
- [x] #1223 - Legacy artboard code can be removed.
```

Keavon:
Closed by 595f5462496dfc5fc702b0853d336898dde65c3e

PR #1217, opened 2023-05-15
Document graph

0HyperCube:
Part of #1171

- [x] Each document has a node graph
- [x] Graph type is displayed
- [x] Properties are usable for all node graphs


node-graph/gcore/src/graphic_element/renderer/quad.rs
```diff
@@ -46,6 +46,10 @@ impl core::ops::Mul<Quad> for DAffine2 {
 	type Output = Quad;
 
 	fn mul(self, rhs: Quad) -> Self::Output {
-		Quad(rhs.0.map(|point| self.transform_point2(point)))
+		let mut output = Quad::default();
+		for (i, point) in rhs.0.iter().enumerate() {
+			output.0[i] = self.transform_point2(*point);
+		}
+		output
```
Keavon, line 53: @0HyperCube my merge conflict resolution had to pick between two options for how this function should end up. I basically guessed which one it should be, so please confirm this is as intended. Thanks.

Keavon (approved):
Code looks good!

PR #1220, opened 2023-05-17
Replace Fill node's "Use Gradient"/"Use Solid" buttons with toggle

### LINKED ISSUE #1219: Replace Fill node's "Use Gradient"/"Use Solid Color" button with a radio button
image

If you draw a shape, its Fill node has that, with a "Use Gradient" (and if you click it, "Use Solid Color") button at the bottom.

image

I'd like to replace those two buttons with a radio button toggle, just like you see in that second screenshot with "Linear" and "Radial".

Instead of being at the bottom, it should be at the top. It should say "Solid" and "Gradient" as its two choices. I suggest copying the Gradient Type (Linear/Radial) row and modifying it to do what the "Use Gradient"/"Use Solid Color" buttons currently do. A very good way to approach any new task if to locate the code by searching the whole project for a unique-looking string, in this case "Use Solid Color" would be a good thing to search for. For the enum that backs the radio button, you can use this one called `FillType`: <https://github.com/GraphiteEditor/Graphite/blob/duplicating_layers_layer_tree/node-graph/gcore/src/vector/style.rs#L210-L217> (as a tip to explain my thought process, I searched the project for `::Gradient` and forced it to match only the full word and be case-sensitive, and I found that enum in use in a few places)

rotteegher39:
Closes #1219

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
@@ -317,21 +317,18 @@ fn fill_type_widget(document_node: &DocumentNode, node_id: u64, index: usize, na
 		exposed: false,
 	} = &document_node.inputs[index]
 	{
-		let entries = [("Solid", FillType::Solid), ("Gradient", FillType::Gradient)]
-			.into_iter()
-			.map(|(name, val)| RadioEntryData::new(name)
-				.tooltip("Switch type of filling.\nChanging from gradient to solid keeps the 0% stop color")
-				.on_update(update_value(move |_| TaggedValue::FillType(val), node_id, index)))
-			.collect();
+		let entries = vec![
+			RadioEntryData::new("Solid").on_update(update_value(move |_| TaggedValue::FillType(FillType::Solid), node_id, index)),
+			RadioEntryData::new("Gradient").on_update(update_value(move |_| TaggedValue::FillType(FillType::Gradient), node_id, index)),
+		];
 
-		add_blank_assist(&mut widgets);
 		widgets.extend_from_slice(&[
 			WidgetHolder::unrelated_separator(),
 			RadioInput::new(entries)
 				.selected_index(match fill_type {
-					FillType::None => 0 as u32,
-					FillType::Solid => 0 as u32,
-					FillType::Gradient => 1 as u32,
+					FillType::None => 0,
+					FillType::Solid => 0,
+					FillType::Gradient => 1,
 				})
 				.widget_holder(),
 		]);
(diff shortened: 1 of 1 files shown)


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -310,6 +310,35 @@ fn line_join_widget(document_node: &DocumentNode, node_id: u64, index: usize, na
...
+	} = &document_node.inputs[index]
+	{
+		let entries = [("Solid", FillType::Solid), ("Gradient", FillType::Gradient)]
+			.into_iter()
+			.map(|(name, val)| RadioEntryData::new(name)
+				.tooltip("Switch type of filling.\nChanging from gradient to solid keeps the 0% stop color")
+				.on_update(update_value(move |_| TaggedValue::FillType(val), node_id, index)))
+			.collect();
+
+		add_blank_assist(&mut widgets);
```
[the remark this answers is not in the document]
rotteegher39: `add_blank_assist()` is actually unnecessary if you pass the `blank_assist: bool` as true where this function is called in `fill_properties()`
It was called with 'false' bool input for some reason. I didn't know that that `blank_assist` input parameter meant and just put false for some reason while was testing.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -310,19 +310,48 @@ fn line_join_widget(document_node: &DocumentNode, node_id: u64, index: usize, na
...
+			RadioEntryData::new("Solid").on_update(update_value(move |_| TaggedValue::FillType(FillType::Solid), node_id, index)),
+			RadioEntryData::new("Gradient").on_update(update_value(move |_| TaggedValue::FillType(FillType::Gradient), node_id, index)),
+		];
+
+		widgets.extend_from_slice(&[
+			WidgetHolder::unrelated_separator(),
+			RadioInput::new(entries)
+				.selected_index(match fill_type {
+					FillType::None => 0,
+					FillType::Solid => 0,
```
Keavon, line 330:
```
FillType::None | FillType::Solid => 0,
```
would make the intent clearer.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -310,19 +310,48 @@ fn line_join_widget(document_node: &DocumentNode, node_id: u64, index: usize, na
 	LayoutGroup::Row { widgets }
 }
 
+fn fill_type_widget(document_node: &DocumentNode, node_id: u64, index: usize, name: &str, blank_assist: bool) -> LayoutGroup {
```
Keavon, line 313: Let's remove the `name` and `blank_assist` parameters from this function signature, and just hard-code their values in the line below in place of their variables. So in other words, move the hard-coded values from the caller of this function to the `let mut widgets = start_widgets(...);` line.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -310,19 +310,48 @@ fn line_join_widget(document_node: &DocumentNode, node_id: u64, index: usize, na
...
+					FillType::Solid => 0,
+					FillType::Gradient => 1,
+				})
+				.widget_holder(),
+		]);
+	}
+	LayoutGroup::Row { widgets }
+}
+
 fn gradient_type_widget(document_node: &DocumentNode, node_id: u64, index: usize, name: &str, blank_assist: bool) -> LayoutGroup {
```
Keavon, line 339: Same as above about inlining the `name` and `blank_assist` variables.

Issue #912, opened 2022-12-25
New nodes


Keavon:
This is a living, growing list.

### Noise

- [ ] Pixel Noise
- [ ] Perlin and/or Simplex Noise
- [ ] Voronoi Noise

### Filters

- [ ] #911

giannissc:
A good resource for nodes (vulkan compute shaders) useful in photo editing can be found in [vkdt](https://github.com/hanatos/vkdt/tree/master/src/pipe/modules). Also Aurelien's [blog](https://eng.aurelienpierre.com/2021/02/image-processing-does-not-kill-people-and-its-a-shame/#do-the-right-thing-the-right-way-at-the-right-time) has quite a few gems (if you don't mind his language) @Keavon

Keavon:
Neat, thanks for the resources!

defini-tiv:
Is there a node that combines scalars to vectors?
Maybe im missing something but seems pretty essential for me. Of course Chnging the vector2 value node to accept f64 as input would suffice.

PR #1226, opened 2023-05-20
New nodes: Resample Points, Spline from Points

Chase-Percy:
Part of #912


Keavon:
Hi, can you please remind (or update) me of the status of this issue? Thanks!

Chase-Percy:
> Hi, can you please remind (or update) me of the status of this PR? Thanks!

Hey sorry just saw this. The last update I had is in the discord task help section. It's mostly done except for an edge case I posted about.

Keavon:
"Density" should probably be renamed "Spacing" since it's the inverse of density.

Keavon:
The fact that this works on text now makes this my favorite node!

image

But unfortunately it seems to be broken for shapes created with these tools:

- Line
- Rectangle
- Ellipse
- Polygon

The others work, even with closed shapes. @0HyperCube do you have any ideas why that might be?

0HyperCube:
@Keavon Works fine. Note that the spacing is specified in vector data space so the rectangle / line / ellipse / polygon which are unit size require a spacing of <1.

Keavon:
@0HyperCube Oh, I see. Would that be difficult to solve by making it use the current transform?

Keavon (approved):

PR #1228, opened 2023-05-20
Optimize colorspace conversion

orlp:
This is a work-in-progress to help optimize the brush tool, but it is generally applicable.


orlp:
As a simple test I made a new 1920x1080 image and dragged the brush tool from the topleft corner to the bottomright. I looked at the profile of the last event in each.

Profile before, on `master` (with wasm-opt but with symbols enabled):

image

Profile after (which in general is quite a bit faster thanks to `lto = "thin"`):

image

orlp:
The improvements aren't as dramatic on https://dev.graphite.rs/ vs https://color-conv-opt.graphite.pages.dev/ due to wasm-opt without debug symbols being able to do inlining similar to `lto = "thin"`, but I still find this to be a 50% speedup overall roughly.

node-graph/gcore/src/raster/discrete_srgb.rs
```diff
@@ -0,0 +1,185 @@
...
+    0x66f, 0x66f063b, 0xcaa0515, 0x11c00773, 0x193305dc, 0x1f1004f3, 0x24030481, 0x28850773,
+    0x2ff9065e, 0x365805a1, 0x3bfa0547, 0x414108f7, 0x4a3907d8, 0x52110709, 0x591b06aa, 0x5fc50b70,
+    0x6b350a18, 0x754e091c, 0x7e6b08aa, 0x87160ef1, 0x96070d3e, 0xa3460bfc, 0xaf430b6c, 0xbaaf13bd,
+    0xce6d1187, 0xdff40fe3, 0xefd70f28
+];
+
+#[inline]
+pub fn float_to_srgb_u8(mut f: f32) -> u8 {
+	// Clamp f to [0, 1], with a negated condition to handle NaNs as 0.
+	if !(f >= 0.0) {
```
Keavon, line 58: Any reason not to use `f < 0.` here?

orlp: Yes, that's what the comment is about. `NaN >= 0.0` is `false`, which means the branch is taken and we thus convert `NaN` to `0`. If you write `f < 0.0` then `NaN` is left unchanged messing up the rest of the algorithm.

Keavon: Ah, I overlooked the comment 👍

TrueDoctor (approved):
Looks good to me

PR #1233, opened 2023-05-24
Add some Improvements to Mask Node

isiko:
These Improvements were proposed by nat-rix for #1153 but also apply to the Mask Node


TrueDoctor (approved):
Looks good

PR #1242, opened 2023-05-25
Make Brush tool use per-stroke options and improve its performance

### LINKED ISSUE #1237: Drawing on rotated / scaled brush layers is incorrect
The mouse trace coordinates for rotated / scaled brush layers are not computed correctly.

There is also the issue of brush radius, should this scale up/down with the brush layer or remain constant? E.g. if I draw with a line with a 40 pixel brush, scale the brush layer x2, and then do another stroke with the same 40px brush, should the two lines have the same size (both 80px), or should the second line still use a 40px brush size?

### LINKED ISSUE #1252: Brush tool should allow per-stroke brush options
This means the Properties panel for a Brush node will no longer let you nondestructively edit the brush settings like radius and hardness and color. Someday we can, with the full node system, regain that sort of functionality.

orlp:
This pull request introduces per-stroke parameters, meaning each stroke of the brush can have an independent size, color, etc.

It also introduces the new spacing parameter that allows you to put blits along the brush path closer or further apart.

It also contains some optimizations that makes drawing overall quite a bit faster.

Closes #1252. Closes #1237.

### DIFF
--- a/editor/src/messages/tool/tool_messages/brush_tool.rs
+++ b/editor/src/messages/tool/tool_messages/brush_tool.rs
@@ -40,7 +40,7 @@ impl Default for BrushOptions {
 			diameter: 40.,
 			hardness: 50.,
 			flow: 100.,
-			spacing: 10.,
+			spacing: 50.,
 			color: ToolColorOptions::default(),
 		}
 	}
--- a/editor/src/messages/tool/tool_messages/brush_tool.rs
+++ b/editor/src/messages/tool/tool_messages/brush_tool.rs
@@ -269,8 +269,18 @@ impl Fsm for BrushToolFsmState {
 		use BrushToolFsmState::*;
 		use BrushToolMessage::*;
 
-		let transform = document.document_legacy.root.transform;
+		let viewport_to_docspace = document.document_legacy.root.transform.inverse();
+		// let mut docspace_to_layerspace = DAffine2::default();
+		// if let Some(layer_path) = &tool_data.path {
+		// 	if let Some(transform) = get_layer_transform(document, layer_path) {
+		// 		docspace_to_layerspace = transform.inverse();
+		// 	} else {
+		// 		log::error!("could not get layer transform");
+		// 	}
+		// }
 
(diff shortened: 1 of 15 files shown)


orlp:
I'm done with work for today, but I do notice a strange 'instability' issue while drawing - the blits for some reason move around a bit.

0HyperCube:
@orlp Great work on this so far, it works really nicely. 

I've added a commit to your branch for inserting a brush stroke node into a layer with an image in (just drag a png into the viewport and then draw with the brush tool on top of it). As part of that I've also committed a fix for the transform issue you were having - hopefully that is helpful.

Keavon:
Regression to fix before merging:

0HyperCube:
I don't have time to fix this, but I suspect it is something to do with the `ExtendImageNode` as this only starts happening when the brush strokes are not entirely contained within the background image.

editor/src/messages/tool/tool_messages/brush_tool.rs
```diff
@@ -255,78 +276,68 @@ impl Fsm for BrushToolFsmState {
...
+					if new_layer {
 						responses.add(DocumentMessage::DeselectAllLayers);
-						tool_data.path = Some(document.get_path_for_new_layer());
+						tool_data.layer_path = document.get_path_for_new_layer();
 					}
-
-					let pos = transform.inverse().transform_point2(input.mouse.position);
-
-					tool_data.points.push(vec![pos]);
+					let layer_position = tool_data.transform.inverse().transform_point2(document_position);
```
Keavon, line 292: @0HyperCube do you have any thoughts about this?

https://files.keavon.com/-/VibrantQuerulousAustralianfreshwatercrocodile/capture_8_.mp4

Keavon:
@orlp we'll merge this now and have you fix the bounds expansion bug (video above) in a followup PR when you get the chance. Thanks.

Keavon (approved):

TrueDoctor (approved):

PR #1243, opened 2023-05-26
Add the first basic version of the GPU blend node

AlexandruIca:
This PR add a blend node that uses the GPU. Expected behavior should be identical to the existing blend node.


TrueDoctor (approved):
Would like some changes but I can do that myself after merge to keep this as close to master as possible

PR #1248, opened 2023-05-26
Fix some node connection UI issues

0HyperCube:
Sometimes the disconnecting nodes was broken, when some inputs were not exposed.


Keavon (approved):
❤️

PR #1250, opened 2023-05-27
Fix Imaginate by porting its JS roundtrip code to graph-based async execution in Rust

### LINKED ISSUE #1257: Migrate Imaginate code out of JS, into the node graph itself

nat-rix:
Closes #1257


node-graph/gcore/src/application_io.rs
```diff
@@ -112,10 +112,25 @@ impl<T: ApplicationIo> ApplicationIo for &T {
...
+}
+
+pub trait NodeGraphUpdateSender {
+	fn send(&self, message: NodeGraphUpdateMessage);
+}
+
+pub trait GetImaginatePreferences {
+	fn get_host_name(&self) -> &str;
+}
+
```
TrueDoctor, line 127: would it make more sense to put this into the application io trait?

node-graph/gstd/src/http.rs
```diff
@@ -25,6 +25,7 @@ pub struct EvalSyncNode {}
 fn eval_sync<F: Future + 'input>(future: F) -> F::Output {
 	let future = futures::future::maybe_done(future);
 	futures::pin_mut!(future);
+	info!("eval_sync: {}", std::any::type_name::<F::Output>());
 	match future.as_mut().take_output() {
 		Some(value) => value,
 		_ => panic!("Node construction future returned pending"),
```
TrueDoctor, line 31: This function can probably be removed completely right?

node-graph/interpreted-executor/src/node_registry.rs
```diff
@@ -444,18 +445,59 @@ fn node_registry() -> HashMap<NodeIdentifier, HashMap<NodeIOTypes, NodeConstruct
...
+						fn_type!(bool),
+						fn_type!(f64),
+						fn_type!(ImaginateMaskStartingFill),
+						fn_type!(bool),
+						fn_type!(bool),
+						fn_type!(ImaginateCache),
+					],
+				),
+			),
+		],
```
TrueDoctor, line 500: could you just use the `async_node` macro here?

TrueDoctor (approved):
I like the overall code structure and design!
I'll leave it up to you if you want to merge this now and  make a second pr to address the comments or if you want to do that before the mereg

Keavon:
Looking great! Some QA notes:
- [ ] After the image generation reaches 100%, it says "Uploading Input Image" briefly before displaying the image
- [ ]The "Progress" text should say "Ready" instead of "Nothing to do" like in the version on master
- [ ]It seems the in-progress image previews are gone, can those be restored?
- [ ]The seed number input seems to have become a slider which isn't desired
- [ ]Creating a new Imaginate layer with the Imaginate tool doesn't start blank, it instead fills in the image from the previous generation, and the Clear button doesn't even clear it
- [ ]The "Resolution" parameter (based on the size rounded to 64) doesn't update while resizing (with the Imaginate or Select tool), only after completing the resize
- [ ]After setting the Imaginate server URL in the preferences, reloading the browser page defaults back to `localhost:7860`. You have to go into Preferences, change the URL to something else, and change it back to the desired value to make it read the saved value
- [ ]I get this failure when trying to to img2img:
image

Keavon:
Since the current master branch is fully broken, we might want to merge this now or soon and do followup PRs for additional fixes. I'll let you decide how you want to handle that.

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -218,70 +216,34 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
...
+				#[cfg(not(test))]
+				if let Some(fut) = self.persistent_data.imaginate.initiate_server_check() {
+					wasm_bindgen_futures::spawn_local(async move {
+						let () = fut.await;
+						use wasm_bindgen::prelude::*;
+
+						#[wasm_bindgen(module = "/../frontend/src/wasm-communication/editor.ts")]
+						extern "C" {
+							#[wasm_bindgen(js_name = injectImaginatePollServerStatus)]
+							fn inject();
```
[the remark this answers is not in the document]
nat-rix: I didn't find a better way to fire an event after the future resolves. The wasm event handler must be woken up by javascript to actually run.

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -36,6 +33,7 @@ pub struct PortfolioMessageHandler {
 impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &PreferencesMessageHandler)> for PortfolioMessageHandler {
 	#[remain::check]
 	fn process_message(&mut self, message: PortfolioMessage, responses: &mut VecDeque<Message>, (ipp, preferences): (&InputPreprocessorMessageHandler, &PreferencesMessageHandler)) {
+		self.persistent_data.imaginate.poll_server_check();
```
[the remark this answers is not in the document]
nat-rix: Actually that line of code isn't needed at all and i've just forgot to delete it. Thanks for pointing out though.

nat-rix:
@Keavon 

> After the image generation reaches 100%, it says "Uploading Input Image" briefly before displaying the image

I changed the text to "Uploading Image" if this was the problem. Uploading means here the server uploads stuff to us. I guess "Downloading Image" would be a better name here...
when thinking about it, "Uploading Input Image" was just an extremely stupid name for this xd.
- [ ] TODO: to be resolved [1]

> The "Progress" text should say "Ready" instead of "Nothing to do" like in the version on master

Resolved.

> It seems the in-progress image previews are gone, can those be restored?

What do you mean by in-progress image previews? Should I display the images that the api sends as a response to `progress` requests while generating? If so that should be possible, but also isn't completely trivial to implement. The latest-stable doesn't didn't have this feature, so I wasn't aware of that.
- [ ] TODO: to be resolved [1]
 
> The seed number input seems to have become a slider which isn't desired

Resolved.

> Creating a new Imaginate layer with the Imaginate tool doesn't start blank, it instead fills in the image from the previous generation, and the Clear button doesn't even clear it

Oh oof. I think I know where this comes from. I'll fix that later.
- [ ] TODO: to be resolved [1]

> The "Resolution" parameter (based on the size rounded to 64) doesn't update while resizing (with the Imaginate or Select tool), only after completing the resize

Also going fixing in a later commit.
- [ ] TODO: to be resolved [1]

> After setting the Imaginate server URL in the preferences, reloading the browser page defaults back to `localhost:7860`. You have to go into Preferences, change the URL to something else, and change it back to the desired value to make it read the saved value

Also going fixing in a later commit.
- [ ] TODO: to be resolved [1]

> I get this failure when trying to to img2img:

Can't reproduce. Looks like a simple network error. Are you sure you haven't configured the server to the wrong host?

> Since the current master branch is fully broken, we might want to merge this now or soon and do followup PRs for additional fixes. I'll let you decide how you want to handle that.

I'm absolutely ok with this being merged now. Less rebasing for me in the future.

- [ ] TODO: implement masking [1]

[1] these serve as markers to show what has to be done (in followup prs). And to have visual rewards when fixing bugs. Nothing feeling more rewarding than clicking on a check box.

node-graph/gcore/src/application_io.rs
nat-rix: I don't know. There are no docs explaining what ApplicationIo and EditorApi actually mean. Would you explain it to me?

node-graph/interpreted-executor/src/node_registry.rs
nat-rix: I could, after I figure out how to do the `ImaginateCache`-thing properly. Currently it seems that different nodes share the same cache what obviously isn't desired

editor/src/messages/portfolio/portfolio_message_handler.rs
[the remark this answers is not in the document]
TrueDoctor: Okay, makes sense, In the future we might want to abstract that as its own feature then, basically adding the functionality of spawning a task that when done dispatches a message.

node-graph/gcore/src/application_io.rs
TrueDoctor: the idea is to provide a platform independent interface for interacting with the outside world + Provide information about the editor state. we currently store the editor state stuff in EditorApi and facilitate the io interaction through the ApplicationIo trait but we might restructure this anyway in the future so it probably does not really matter for now

node-graph/interpreted-executor/src/node_registry.rs
TrueDoctor: nodes are considered equal if they receive the same inputs and have the same implementation. 

> We generate a stable node id for each node which you think of sort of like like a Gödel number for a turing machine. It is a unique identifier which describes the mapping from input -> output
the following always has to hold true
let x, y ∈ StableNodeIds
>
> x = y ⇒ ∀i ∈ Input: x.eval(i) == y.eval(i)

So it feels like it would make more sense to store the cache in the node itself anyway instead of feeding it in as an input

Issue #1247, opened 2023-05-26
Use a texture sampler instead of looping over pixels in the Blend node (and others)


Keavon:
The Blend node loops through all the pixels, which is a limitation causing ugly texture rendering and especially bad handling or rotation (and scaling) because it performs something like nearest neighbor sampling. The result is a very ugly image output and artifacts like #1132 and pixels "jumping" with an offset causing a blended image to appear one pixel to the side of its original.

Even on GPU, this for loop is converted into a parallelized execution of the loop's body, meaning the GPU is still just performing the nearest neighbor lookup in parallel. Instead, we want to utilize the texture sampler capabilities of the GPU, which would perform accurate pixel lookups with built-in texture interpolation (bilinear or bicubic instead of nearest neighbor). An added challenge is that `rust-gpu` currently does not provide any way to run an equivalent texture sampler concept on the CPU, since its crate provides the `Image` and `Sampler` API only when compiling to the GPU architecture (see https://github.com/EmbarkStudios/rust-gpu/issues/1065). Even if that blocks the CPU implementation, this issue should at least be able to be solved on GPU.

This probably affects more than just the Blend node— basically anything that needs to take an image input and sample from it with some offset/scale/rotation.

PR #1254, opened 2023-05-28
Make brush stamp blitting numerically stable

orlp:
Prevents the brush trail from 'jumping around' due to numerical instability.

Doesn't solve https://github.com/GraphiteEditor/Graphite/issues/1247, but gets rid of the most egregious effect users could instantly notice.


Keavon:
There's definitely a rather weird bug here...

https://github.com/GraphiteEditor/Graphite/assets/4388688/fa54d810-7658-4850-88c5-e510ee4b340b

orlp:
This is superseded by my later work on the brush-blend-mode branch.

PR #1260, opened 2023-05-30
Hide snapping overlays when snapping to a point

ShouvikGhosh2048:


ShouvikGhosh2048:
Currently this PR hides snapping overlays when we snap to a point.
i) Should we show the overlay for the point currently being snapped to?
ii) The initial requirement was to hide snapping overlays when handles are dragged into their anchor while using the path tool. The PR however hides snapping overlays irrespective of the tool / whether a handle or anchor is being dragged. Is this fine?
@Keavon

Keavon:
Thanks, I'll try to get a chance to test this branch and, with the context it provides, answer those questions ASAP. Hopefully in the next few days if I can manage that.

Keavon:
The behavior seems to work as desired. I mostly comprehended your question number (ii) but not fully, however I don't notice any issues.

Keavon (approved):

Keavon (approved):
Sorry this took so long! Bad timing since I went on vacation and was urgently packing when this came in.

PR #1261, opened 2023-05-30
Brush blend modes and erase/restore

orlp:
This pull requests adds blend modes for strokes, as well as a new erase/restore mode for the brush layer.


Keavon:
Crash: after creating at least one Brush layer (you don't even have to add the Draw Canvas node, just draw a simple stroke), closing the document tab crashes the editor. This doesn't happen on master.

Keavon (approved):

PR #1267, opened 2023-05-31
New node: Pixel Noise

isiko:
Adds Pixel Noise, Perlin Noise, some bad Simplex Noise, and Voronoi Noise

For the last 3, I'm using [noise-rs](https://github.com/Razaekel/noise-rs), a library that seems to have a few bugs but still is the best I could find. I think through forking it I have at least fixed the major bugs (there was one in the Perlin noise generator that would crash the editor when getting high inputs), but the simplex noise still has artifacts so I'll probably remove that node. If it's important that we have simplex noise, I could of course look to see if I find the error though.

Pixel noise on the other hand does not use any libraries (except for a randomizer) and as it was pointed out that it would be nice to have other similar types of noise (like for example pink noise), everything is already set up for new noise types to be implemented easily. At the moment it can generate white noise.


Keavon:
Some quick feedback:
- All: Please put 'Seed' after 'Width' and 'Height'
- All: Rename 'Frequency' to 'Scale'
- All: Rename 'Allow Tiling' to 'Tile', and add it for those it's missing from
- Voronoi: Add a choice between 'F1' and 'F2'
- Simplex: Disable this node since it's totally broken
- Perlin: It seems to like to clip the black regions... is the math right?

TrueDoctor:
@isiko can you work on reducing this to the noise nodes which currently work so we can already merge some of the nodes?

Keavon:
The plan, as discussed when the three of us last met in-person, is to attempt to port https://github.com/Auburn/FastNoiseLite using the desktop app test tool provided by that project as a reference, ensuring correctness of the port against the reference. I'm okay with abandoning this PR instead of trying to rebase it, then starting with the Rust port in a separate context (not necessarily requiring the Graphite codebase while porting). The port doesn't have to be complete, but the nice thing is that that repo provides the same implementation in many different languages and they are all respectively written in a single file (~2.5k lines but heavily commented). It might even be worth considering an upstream contribution if they'd be interested in adopting a Rust version if it can be faithfully ported.

isiko:
I have some exams coming up, so I won't have much time to work on the Perlin/Voronoi/... node in the near future, but the pixel noise node should be good to go, so I think it would be best to remove the stuff we need the lib for, merge the pixel noise node, and when I'm done with the exams I can work on the library in a new repo.

Keavon:
That sounds perfect. Also today I began manually transpiling/porting the C# code to Rust. I'm about 1/3 through the syntax changes. The logic should remain all identical.

0HyperCube (approved):
The code looks good and the node works well.

Keavon:
Node works well! Glad to have this 😎

PR #1268, opened 2023-06-01
Add blit caching and blend modes to Brush tool

orlp:
This pull request is a follow-up of https://github.com/GraphiteEditor/Graphite/pull/1261, improving the performance by caching blits and blends.

Now, no matter how many times you brush, or how long one stroke is, the performance stays relatively constant.


Keavon:
Crash: after creating at least one Brush layer (you don't even have to add the Draw Canvas node, just draw a simple stroke), closing the document tab crashes the editor. This doesn't happen on master.

TrueDoctor (approved):

PR #1276, opened 2023-06-02
Add the Extract Opaque node

isiko:
Title says it all :D; the new node takes an image and makes it opaque


isiko:
It seems like this was caused by some floating point errors in the luminance calculation for colors. This would result in values that were just slightly out of the allowed Range (i.e. `0.0 - 1.0`). I fixed it by just clamping the output of that function to that range.

isiko:
(After moving the commit back to the right branch, this should be fixed now :upside_down_face: )

isiko:
Wait, never Mind, this seems to be an issue with the levels node

Keavon:
The Extract Opaque node still crashes the editor on [this image](https://upload.wikimedia.org/wikipedia/commons/9/9d/Alpha_transparency_image.png) (note that you have to download the image to your desktop before dragging it in, you can't copy the image from the browser).

Keavon (approved):

PR #1282, opened 2023-06-03
Minimum viewport scale cover when zooming

### LINKED ISSUE #1281: Prevent the document from getting zoomed out of view
The user can zoom the artwork to a ludicrously small size, way smaller than a viewport pixel. But we want to support those scales if the user is making a really large piece of artwork (say, a map of a country that's to-scale). We should cap the zooming out so it never goes to less than 1% of the size of the [union of artboards + artwork] bounding box.

~~Likewise, the user can scroll/pan and move the artwork very far out of the viewport. To prevent users from losing it, we should ensure there is always at least one pixel of artwork visible in the viewport by clamping any attempts to pan or scroll beyond the [union of artboards + artwork] bounding box. Panning can also occur by Ctrl+scroll zooming when the mouse is pointing in a corner of the viewport.~~

0HyperCube:
Closes #1281

The artwork must cover 1% of the viewport (on either axis) at all times when zooming.


editor/src/consts.rs
```diff
@@ -5,6 +5,7 @@ pub const VIEWPORT_ZOOM_WHEEL_RATE: f64 = (1. / 600.) * 3.;
 pub const VIEWPORT_ZOOM_MOUSE_RATE: f64 = 1. / 400.;
 pub const VIEWPORT_ZOOM_SCALE_MIN: f64 = 0.000_000_1;
 pub const VIEWPORT_ZOOM_SCALE_MAX: f64 = 10_000.;
+pub const VIEWPORT_MINIMUM_FRACTION_COVER: f64 = 0.01;
```
Keavon, line 8: `VIEWPORT_ZOOM_MINIMUM_FRACTION_COVER`

Keavon (approved):

PR #1285, opened 2023-06-04
Add line extension snapping mode

ChrisMend19:
For hotkey-based line extension, hitting the key "V" while using the path tool enables snapping to whichever of the one or two imaginary infinite lines the being-dragged point's current position is closer to.


Keavon:
I'm concerned that this adds 830 lines of code, that's way too much for a feature like this I would think. @TrueDoctor @0HyperCube could I please get your thoughts on this and how to proceed?

Keavon:
Moved to #1355

PR #1288, opened 2023-06-05
Remove warnings from build and satisfied clippy

gautamprikshit1:
Removed warnings from cargo build process


node-graph/gcore/src/value.rs
```diff
@@ -115,6 +115,7 @@ impl<'i, T: Clone + 'i> Node<'i, ()> for DebugClonedNode<T> {
 }
 
 impl<T: Clone> DebugClonedNode<T> {
+	#[allow(clippy::new_ret_no_self)]
 	pub const fn new(value: T) -> ClonedNode<T> {
```
0HyperCube, line 119: This function should return a `DebugClonedNode`, this is probably just a copy/paste error.

PR #1289, opened 2023-06-05
Fix Levels Node

isiko:
The Levels Node was previously outputing channel values out side the range of 0..1, this should be fixed now.
This was brought up in #1276


0HyperCube (approved):
Looks good.

PR #1304, opened 2023-06-08
Fix zoom with scroll preference reset to default correctly

Kirirou:
Closes [graphite discord link to issue](https://discord.com/channels/731730685944922173/881073965047636018/1111163255843475466)


Keavon:
@TrueDoctor or @0HyperCube could you please pull and build this and confirm that it fixes the issue? Thanks!

rotteegher39:
I verified that it works as intended several times.

When Zoom With Scroll preference is turned on, Reset To Defaults button resets the scroll mapping back to default as inteded.

0HyperCube (approved):
This works very well, nice catch.

PR #1312, opened 2023-06-12
Improve 'add node' menu

### LINKED ISSUE #1280: Better node insertion dropdown
Currently when a new node is inserted into the node graph through the right click menu, the list is difficult to work with.
image
A scroll bar and a collapsed dropdown for each category would help with this.

FadedBronze:
Improved node adding list with dropdowns and scrolling.

This is aiming to fix: 
Better node insertion dropdown

Closes #1280

### DIFF
--- a/frontend/src/components/panels/NodeGraph.svelte
+++ b/frontend/src/components/panels/NodeGraph.svelte
@@ -18,6 +18,8 @@
 	const WHEEL_RATE = (1 / 600) * 3;
 	const GRID_COLLAPSE_SPACING = 10;
 	const GRID_SIZE = 24;
+	const ADD_NODE_MENU_WIDTH = 180;
+	const ADD_NODE_MENU_HEIGHT = 200;
 
 	const editor = getContext<Editor>("editor");
 	const nodeGraph = getContext<NodeGraphState>("nodeGraph");
@@ -55,8 +57,8 @@
 		if (!bounds) return;
 		const { width, height } = bounds;
 
-		appearRightOfMouse = nodeListX > width - 90;
-		appearAboveMouse = nodeListY > height - 100;
+		appearRightOfMouse = nodeListX > width - ADD_NODE_MENU_WIDTH / 2;
+		appearAboveMouse = nodeListY > height - ADD_NODE_MENU_HEIGHT / 2;
 	})();
 
 	$: linkPathInProgress = createLinkPathInProgress(linkInProgressFromConnector, linkInProgressToConnector);
@@ -75,7 +77,7 @@
 
 	type NodeCategoryDetails = {
 		nodes: FrontendNodeType[];
-		shouldBeOpen: boolean;
+		open: boolean;
 	};
 
 	function buildNodeCategories(nodeTypes: FrontendNodeType[], searchTerm: string): [string, NodeCategoryDetails][] {
@@ -89,17 +91,17 @@
 			}
 
 			const category = categories.get(node.category);
-			let shouldBeOpen = nameIncludesSearchTerm;
+			let open = nameIncludesSearchTerm;
 			if (searchTerm.length === 0) {
-				shouldBeOpen = false;
+				open = false;
 			}
 
 			if (category) {
-				category.shouldBeOpen = shouldBeOpen;
+				category.open = open;
 				category.nodes.push(node);
 			} else
 				categories.set(node.category, {
-					shouldBeOpen,
+					open: open,
 					nodes: [node],
 				});
 		});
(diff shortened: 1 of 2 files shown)


FadedBronze:
Switched the default arrows for the dropdown arrows. Also, made it so the 'add node' menu stays in the node graph.

image
image
image
image

0HyperCube:
This works really well @FadedBronze. I have some other thoughts for UX improvements, but they were not in the original requirements so we can merge without them if you prefer:
- A fixed width for the node panel. Currently when you expand some sections like the GPU section, the width of the node list menu changes. I think it would be better if the text just wrapped if it was too long.
- If the user opens the node panel at the very bottom of the window, the list should appear above the mouse. This ensures that the list is fully visible after opening it.
- Automatically expand the dropdown when the user types in the search box, allowing them to immediately see the node they want.

FadedBronze:
@0HyperCube Those are some good ideas. I will add them when I get the time. Around how large should the width of the nodegraph to be?

0HyperCube:
Thank you for the contribution @FadedBronze. For the width of the menu, I would say probably around 180px (although that is completely arbitrary).

frontend/src/components/panels/NodeGraph.svelte
```diff
@@ -46,6 +46,19 @@
...
+	let appearAboveMouse = false;
+	let appearRightOfMouse = false;
+
+	$: (() => {
+		const bounds = graph?.div()?.getBoundingClientRect();
+		if (!bounds) return;
+		const { width, height } = bounds;
+
+		appearRightOfMouse = nodeListX > width - 90;
+		appearAboveMouse = nodeListY > height - 100;
```
0HyperCube, line 59: Perhaps we should declare the constants `90` and `100` at the top of the `<script>` tag for easy access?

frontend/src/components/panels/NodeGraph.svelte
```diff
@@ -60,16 +73,35 @@
 		return sparse;
 	}
 
-	function buildNodeCategories(nodeTypes: FrontendNodeType[], searchTerm: string): [string, FrontendNodeType[]][] {
-		const categories = new Map();
+	type NodeCategoryDetails = {
+		nodes: FrontendNodeType[];
+		shouldBeOpen: boolean;
```
0HyperCube, line 78: Would simply calling this `open` be clearer?

frontend/src/components/panels/NodeGraph.svelte
```diff
@@ -497,18 +529,27 @@
...
-						{#each nodeCategory[1] as nodeType}
-							<TextButton label={nodeType.name} action={() => createNode(nodeType.name)} />
-						{/each}
-					</LayoutCol>
-				{:else}
-					<TextLabel>No search results</TextLabel>
-				{/each}
+				<div style="height: 200px; overflow-y: scroll;" on:wheel|stopPropagation>
+					{#each nodeCategories as nodeCategory}
+						<details style="display: flex; flex-direction: column;" open={nodeCategory[1].shouldBeOpen}>
```
0HyperCube, line 540: We generally prefer adding styles to the css rather than inline.

frontend/src/wasm-communication/messages.ts
```diff
@@ -89,6 +89,7 @@ export class NodeGraphOutput {
 }
 
 export class FrontendNode {
+	[x: string]: any;
```
0HyperCube, line 92: I'm not quite sure what this does?

frontend/src/wasm-communication/messages.ts
```diff
@@ -107,6 +108,7 @@ export class FrontendNode {
 	readonly disabled!: boolean;
 
 	readonly thumbnailSvg!: string | undefined;
+	category: any;
```
0HyperCube, line 111: For consistency with the other entries here, this should probably be `readonly category!: string;`

0HyperCube (changes requested):
Works well, thanks for these changes. I've just got a few minor code style changes before we merge.

frontend/src/wasm-communication/messages.ts
FadedBronze: It seems I accidently included that file It doesn't do anything, Sorry.

frontend/src/wasm-communication/messages.ts
FadedBronze: That doesn't do anything.

frontend/src/components/panels/NodeGraph.svelte
FadedBronze: Ok, good to know. I was thinking since was only two it doesn't make sense to make a class, but this does avoid confusion of CSS in two places.

frontend/src/components/panels/NodeGraph.svelte
FadedBronze: yes.

frontend/src/components/panels/NodeGraph.svelte
FadedBronze: yes, that does make it clearer.

0HyperCube (approved):
Thank you for finishing up these changes, it really improves the node graph user experience. I hope you enjoyed helping out with this!

PR #1313, opened 2023-06-16
conditionally enable Select Tool's buttons

dhruv-ahuja:
This PR aims to fix two minor issues mentioned in https://github.com/GraphiteEditor/Graphite/issues/989:
- Gray out the Select tool's alignment buttons when fewer than two layers are selected
- ~~Gray out the Select tool's boolean operation buttons when fewer than two vector shape layers are selected~~

Currently done with fixes for the first issue.

Since the Boolean operator feature is under works, I will pick up the other small issue of disabling pivot button when no layers are selected.

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -126,6 +126,7 @@ impl PropertyHolder for SelectTool {
 			})
 			.collect();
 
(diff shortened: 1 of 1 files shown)


dhruv-ahuja:
Added logic to only enable the Pivot button if some layer is selected, as per discussion :)

0HyperCube (approved):
Looks good, thanks for this contribution

PR #1314, opened 2023-06-19
Minor Tooling and UI Improvements

dhruv-ahuja:
This PR will aim to fix and make minor improvements for the tools. 
Working towards the following:
- [x] menu buttons and several individual elements should be disabled when no file is opened
- [x] stopping the fill and gradient tools from working on a bitmap image
The same issues are also mentioned [here](https://github.com/GraphiteEditor/Graphite/issues/1315).

### DIFF
--- a/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
+++ b/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
@@ -14,6 +14,7 @@ impl MessageHandler<MenuBarMessage, bool> for MenuBarMessageHandler {
 	#[remain::check]
 	fn process_message(&mut self, message: MenuBarMessage, responses: &mut VecDeque<Message>, _data: bool) {
 		use MenuBarMessage::*;
+		self.has_active_document = _data;
 
 		#[remain::sorted]
 		match message {
@@ -198,7 +199,6 @@ impl PropertyHolder for MenuBarMessageHandler {
 				action: MenuBarEntry::create_action(|_| FrontendMessage::TriggerVisitLink { url: "https://graphite.rs".into() }.into()),
 				..Default::default()
 			},
-			// TODO: ensure that menubar is re-rendered when a document opens
 			MenuBarEntry::new_root(
 				"File".into(),
 				true,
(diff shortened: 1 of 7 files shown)


editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
```diff
@@ -6,12 +6,15 @@ use crate::messages::portfolio::document::utility_types::clipboards::Clipboard;
...
-pub struct MenuBarMessageHandler {}
+pub struct MenuBarMessageHandler {
+	has_inactive_document: bool,
+}
 
-impl MessageHandler<MenuBarMessage, ()> for MenuBarMessageHandler {
+impl MessageHandler<MenuBarMessage, bool> for MenuBarMessageHandler {
 	#[remain::check]
-	fn process_message(&mut self, message: MenuBarMessage, responses: &mut VecDeque<Message>, _data: ()) {
+	fn process_message(&mut self, message: MenuBarMessage, responses: &mut VecDeque<Message>, _data: bool) {
```
0HyperCube, line 15: Can we rename `_data` to something more informative?

editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
```diff
@@ -6,12 +6,15 @@ use crate::messages::portfolio::document::utility_types::clipboards::Clipboard;
 use crate::messages::prelude::*;
 
 #[derive(Debug, Clone, Default)]
-pub struct MenuBarMessageHandler {}
+pub struct MenuBarMessageHandler {
+	has_inactive_document: bool,
```
0HyperCube, line 10: Perhaps it would be clearer to rename to `no_active_document`?

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -91,6 +93,7 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
 					responses.add(DocumentMessage::ClearLayerTree);
 					let hint_data = HintData(vec![HintGroup(vec![])]);
 					responses.add(FrontendMessage::UpdateInputHints { hint_data });
+					self.menu_bar_message_handler.process_message(MenuBarMessage::SendLayout, responses, false);
```
0HyperCube, line 96: Note that, to avoid unnecessary UI updates, this only really needs to be called:
- PortfolioMessage::DeleteDocument after setting self.active_document_id = None;
- PortfolioMessage::DestroyAllDocuments after setting self.active_document_id = None;
- PortfolioMessage::SetActiveDocument after setting self.active_document_id = Some(...)
These are the only times that the active document changes.

You can also use `responses.add(MenuBarMessage::SendLayout);` which is more concise and less error prone.

0HyperCube (changes requested):
Great work on this! I've just got a few minor points for cleanup before merging.

0HyperCube (approved):
Looks good - thanks again for this useful UX improvement!

PR #1321, opened 2023-07-04
Add Snapping Options to the Snap Dropdown Menu

### LINKED ISSUE #1319: Snapping options dropdown
Populate the dropdown menu for the snapping system shown below. Some initial features would be a checkbox for snap to node and a checkbox for snap to bounding box, however this will be added to as more snapping features are added.
The snapping dropdown
The inkscape snapping options box for inspiration (although we don't need icons for the initial version):
Inkskape snapping options
In order to add this, we need to change the `PopoverButton` widget to have `options_widgets: WidgetLayout`. We also need to modify `PopoverButton` in `frontend/src/wasm-communication/messages.ts` and change the svelte component `frontend/src/components/widgets/buttons/PopoverButton.svelte` to display the options instead of the placeholder text.

dhruv-ahuja:
Closes #1319


frontend/src/components/widgets/inputs/CheckboxInput.svelte
```diff
@@ -47,7 +47,7 @@
 <style lang="scss" global>
 	.checkbox-input {
 		flex: 0 0 auto;
-		align-items: center;
+		align-items: right;
```
0HyperCube, line 50: This change appears to move the checkbox in a strange way. Is it needed?

frontend/src/components/widgets/inputs/CheckboxInput.svelte
```diff
@@ -75,6 +75,10 @@
 				}
 			}
 
+			.label-text {
+				margin-left: 8px;
+			}
```
0HyperCube, line 80: This style is unused I think?

0HyperCube (approved):
Works very well! Thanks for your work on this improvement (and sorry again that I forgot to review).

0HyperCube:
@Keavon Clicking the label doesn't toggle the checkbox which I would expect. This is because the `for` attribute is not set for the label as it is a separate widget.

Keavon:
> @Keavon Clicking the label doesn't toggle the checkbox which I would expect. This is because the `for` attribute is not set for the label as it is a separate widget.

Yeah, that's on my (mental) todo list to make labels support an association with a checkbox for that.

frontend/src/components/widgets/inputs/CheckboxInput.svelte
dhruv-ahuja: Reverted, had forgotten to undo this.

frontend/src/components/widgets/inputs/CheckboxInput.svelte
dhruv-ahuja: Yes, removed now. Thank you.

PR #1328, opened 2023-07-16
Artboard nodes

### LINKED ISSUE #1221: Render node-based artboards
Part of #1171

The output from that special node graph is of type `ArtboardGroup` and it should enter the SVG output renderer and become an appropriately colored background rectangle as well as a definition for a [clipping mask](https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/Clipping_and_masking).
    
(A checkered transparency background (defined in CSS) should also be included in the viewport, but in a way that causes it to be excluded from things that rasterize the artboard.)

0HyperCube:
Closes #1221
Cool Artboards


Keavon (approved):
There are still a few bugs (and I even managed a crash) arising from the quantum entanglement of the old and new artboards, but I fixed a few of the most severe issues and this should be fine for now until you can remove the quantum entanglement with the old code. This is a fantastic start with a clear view to the finish line now, thank you for your efforts!

PR #1331, opened 2023-07-23
Fix nudge resizing having the wrong pivot 

### LINKED ISSUE #1317: Make arrow key nudge-resizing operate in document space
The nudge system correctly moves by 1px (or 10px when <kbd>Shift</kbd> is held down) relative to the document (so the nudge happens regardless of viewport zoom). However, nudge-resizing is currently viewport-relative which isn't desired, and this means the zoom level ends up factoring itself into your new scale value (which can be especially undesirable if the view is at a noninteger ratio, causing the selected layers to render with antialiasing because they will have lost grid alignment).

- Nudging is using the arrow keys to move the selected layers around within the canvas.
- Nudge resizing is using the arrow keys, while also holding <kbd>Alt</kbd>, to resize the bottom and right edges of the selected layers (or top and left if <kbd>Ctrl</kbd> is also pressed)

---

The following might no longer be relevant after the unified node graph refactor when the nudge system was rewritten to use the document graph (in #1501). But last time we tried switching from viewport space nudging to document space nudging, the following problem occurred. It's left here for context if necessary.

> Currently, if you select a layer and change its size with <kbd>alt</kbd> + arrow key then the layer gets a seemingly arbitrary translation also. This is related to `DocumentMessage::NudgeSelectedLayers`. A reasonable approach would be to not change the translation of the layer, so that the pivot does not move.
> 
> Note that we don't want to use the actual pivot point for the layer (image, controlled by dragging it or this widget: image). We want to use the top left of the bounding box since that's more useful for positioning the layer which is what this nudge-resize feature is used for.
> 
> This is a regression introduced in #1114, which correctly switched from viewport-space to document-space for nudging but introduced this functionality regression for nudge resizing. A correct fix will need to change how the transforms calculate the final size so it doesn't cause an unintended translation.

dhruv-ahuja:
Closes #1317

https://github.com/GraphiteEditor/Graphite/assets/83733638/7caeb357-9ea8-4fd2-ba66-8412016fb92e


dhruv-ahuja:
Closing this one as per discussion.

PR #1332, opened 2023-07-26
Show disabled menu bar entries with no active document

dhruv-ahuja:
Making the change as advised. This change will allow the user to see all possible submenus without being able to use them if there is no active document.
This somewhat reverts the behaviour of one previous PR.


Keavon:
Looks great, thank you!

Keavon (approved):

PR #1336, opened 2023-07-28
Update GitHub Actions CI action versions

waywardmonkeys:
These are generating warnings in the GitHub Actions UI as they are being forced to use a more current node.


Keavon (approved):
Thanks!

Keavon:
Thank you for this! We have a number of issues with our CI pipeline (it has basically been half-broken for the past couple months). Would you be interested in working with us to get that all resolved? It would be most helpful to the project, if you're willing.

PR #1342, opened 2023-07-30
Fix artboard thumbnails

0HyperCube:
The layer nodes now have thumbnails again. These were broken because the string identifer pointed to "graphene_std" instead of "graphene_core", the document id paths are different, and because the nodes are now not part of a layer.

### DIFF
--- a/editor/src/node_graph_executor.rs
+++ b/editor/src/node_graph_executor.rs
@@ -29,6 +29,22 @@ use std::rc::Rc;
 use std::sync::mpsc::{Receiver, Sender};
 use std::sync::Arc;
 
+/// Identifies a node graph, either the document graph or a node graph associated with a legacy layer.
+#[derive(Clone, Copy, Hash, PartialEq, Eq, Debug)]
+pub enum GraphIdentifier {
+	DocumentGraph,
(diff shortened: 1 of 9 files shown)


editor/src/node_graph_executor.rs
```diff
@@ -271,7 +263,7 @@ pub struct NodeGraphExecutor {
 	receiver: Receiver<NodeGraphUpdate>,
 	// TODO: This is a memory leak since layers are never removed
 	pub(crate) last_output_type: HashMap<Vec<LayerId>, Option<Type>>,
-	pub(crate) thumbnails: HashMap<LayerId, HashMap<NodeId, SvgSegmentList>>,
+	pub(crate) thumbnails: HashMap<Option<LayerId>, HashMap<NodeId, SvgSegmentList>>,
```
TrueDoctor, line 266: I don't really like this being an option, feels like a hack. Or I just don't understand the reasoning, could you add a comment in that case?

TrueDoctor (approved):
LGTM

I don't love the repurposing of the hash map for legacy layers and the new document graph but I guess that is fine for now, we just need to remember to change it back afterwards

PR #1346, opened 2023-07-31
Switch to the widget builder pattern on all remaining layouts

0HyperCube:
Use the widget builder pattern everywhere. I don't think I broke anything but you may wish to test again.


TrueDoctor (approved):
The code looks good, did not check in detail but looks good

Keavon (approved):
So much more ergonomic, thank you. (And those lines saved, wow!)

Cargo.toml
```diff
@@ -28,7 +28,7 @@ resolver = "2"
 exclude = ["node-graph/gpu-compiler"]
 
 [workspace.dependencies]
-specta = { git = "https://github.com/0HyperCube/specta.git", rev = "d28cd72ea4ea2f6eaa85e0622d3a98340dbb43d3", features = [
+specta = { git = "https://github.com/0HyperCube/specta.git", rev = "c47a22b4c0863d27bc47529f300de3969480c66d", features = [
```
Keavon, line 31: Just checking, does the reason you had to maintain your own fork still persist or would we be able to switch back to their official repo now? @0HyperCube

0HyperCube: Two reasons @Keavon 
- They specify glam=0.22 whereas we use glam=0.24 so the encoding doesn't work.
- Their current release doesn't allow doc comments and produces a compile error.
```
error[E0659]: `doc` is ambiguous
  --> editor/src/messages/layout/utility_types/misc.rs:27:2
   |
27 |     /// The color swatch for the working colors and a flip and reset button found at the bottom of the tool shelf.
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ambiguous name
   |
   = note: ambiguous because of a name conflict with a builtin attribute
   = note: `doc` could refer to a built-in attribute
note: `doc` could also refer to the derive helper attribute defined here
  --> editor/src/messages/layout/utility_types/misc.rs:4:83
   |
4  | #[derive(PartialEq, Clone, Debug, Hash, Eq, Copy, Serialize, Deserialize, specta::Type)]
   |                                                                                   ^^^^
```

Keavon: Gotcha, thanks. Would you mind including that in a comment?

PR #1355, opened 2023-08-03
Add line extension snapping mode

Keavon:
For hotkey-based line extension, hitting the key "V" while using the path tool enables snapping to whichever of the one or two imaginary infinite lines the being-dragged point's current position is closer to.

[Reopens #1285]


Keavon:
@0HyperCube I ended up rebasing it, but had to create a new PR for this. In response to my question for you in that earlier PR (#1285)...

> I'm concerned that this adds 830 lines of code, that's way too much for a feature like this I would think. @TrueDoctor @0HyperCube could I please get your thoughts on this and how to proceed?

...we can now see the diff again properly. Do you think any of this code is salvageable, or is it just way too badly structured and we should just put this issue back up for grabs so someone else can implement it? (Please prioritize your current focus over spending time answering this for the next week or two, though, I'll remind you about this then.)

0HyperCube (commented):
This PR seems to be already merged with grid?

Just looking at the path tool modifications, I would agree that accessing the next and previous anchors and storing them, then projecting the mouse position onto these rays should not be >50 lines.

Keavon:
> This PR seems to be already merged with grid?

I believe grid and this branch both have a common commit history that diverge at some point. But yeah, 830 lines to do something that should be done in 50 is not worth merging, so I agree with closing this.

PR #1359, opened 2023-08-04
Add config for nix-shell

isiko:
This adds a development environment for Nix users, which automatically installs all libraries needed for building and fixes some nix stuff.


shell.nix
```diff
@@ -0,0 +1,50 @@
+let
```
Keavon, line 1: Please add a comment at the top of the file which provides a detailed explanation of what the purpose of this file is. Assume no knowledge of what nix-shell or nix is.

isiko: I added a comment, I hope that's descriptive enough :+1:

Keavon: Thanks. But since I don't know Nix is, this doesn't really explain anything to me. However it looks like it's much more useful now for people who do directly intend to use it. Adding a description of why this file exists, assuming no knowledge of Nix, would be important. Thank you!

isiko: I asked TrueDoctor for some Input and added a bit to the description :+1:

PR #1368, opened 2023-08-06
Decrease graph compilation time significantly

TrueDoctor:

### DIFF
--- a/node-graph/graph-craft/src/proto.rs
+++ b/node-graph/graph-craft/src/proto.rs
@@ -462,26 +462,31 @@ impl ProtoNetwork {
 
 	fn reorder_ids(&mut self) {
 		let order = self.topological_sort();
-		// Map of node ids to indexes (which become the node ids as they are inserted into the borrow stack)
-		let lookup: HashMap<_, _> = order.iter().enumerate().map(|(pos, id)| (*id, pos as NodeId)).collect();
-		self.nodes = order
-			.iter()
-			.enumerate()
-			.map(|(pos, id)| {
-				let node = self.nodes.swap_remove(self.nodes.iter().position(|(test_id, _)| test_id == id).unwrap()).1;
-				(pos as NodeId, node)
-			})
-			.collect();
-		self.replace_node_references(&lookup, false);
-		assert_eq!(order.len(), self.nodes.len());
-	}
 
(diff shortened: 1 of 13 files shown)


node-graph/graph-craft/src/proto.rs
```diff
@@ -340,45 +333,59 @@ impl ProtoNetwork {
...
-				let pre_node_input = inputs.get(input_node as usize).expect("input node should exist");
-				let pre_path = paths.get(input_node as usize).expect("input node should exist");
-				Some((input_node, *id, pre_node_input.clone(), pre_path.clone()))
-			} else {
-				resolved.insert(*id);
-				None
+		let max_id = self.nodes.len() as NodeId - 1;
+
+		// Collect outward edges once
+		let outwards_edges = self.collect_outwards_edges();
```
Keavon, line 342: I have to request we use "upstream" or "downstream" since "outwards" conveys no meaning to me and I have no clue if that means upstream or downstream. As a reminder, upstream means easier in the execution, more to the left of the graph. Water flowing through a stream, or data through the graph, progresses downstream and rightward to the output over time.

Keavon (approved):

node-graph/graph-craft/src/proto.rs
TrueDoctor: Outwards means the arrow leaves the node and inward means the opposite I've always found that to be quite intuitive, especially considering that we at this point do not yet necessarily have a DAG and without a dag the notion of upstream and downstream doesn't really make sense right?
But this is not a hill I would like to die on

Keavon: To me "outward" implies "radiating outward in all directions" and conveys no meaning to me. Now that you have explained it, I suppose I see your logic for the choice of word but it's not something I would be able to understand without learning and remembering the alternative word choice. Is there a reason that the standard upstream/downstream terminology confuses you?

TrueDoctor: I always mix up upstream and downstream. I think another reason is that there is not direct equivalent term in german so there is not really something to map that against. The notion of not having a direction is actually accurate as we at this stage don't necessarily have an established flow yet, that basically happens after the topological sort. But yeah idk, hypercube chose the terminology initially

PR #1370, opened 2023-08-07
Correct Snapping Offsets After Canvas Transformation When Using Snap-Reliant Tools

dhruv-ahuja:
This PR aims to fix the current state of tools using snapping, where they currently use old snapping offsets even when the canvas has transformed through panning.


dhruv-ahuja:
I've reverted the LineTool changes as discussed, since its behaviour seems to be attached to changes required for the nudge resize change for the SelectTool.

TrueDoctor (approved):

PR #1373, opened 2023-08-07
Add Circular Repeat Node

isiko:
Repeats a given set of Vector Data in a circular pattern a specified amount of times.


TrueDoctor (approved):
Should be good after those nitpicks have been adressed

isiko:
Should be all fixed now

PR #1376, opened 2023-08-08
Add bounding box node

isiko:
Generates a rectangle based on the bounding box of the input vector data


isiko:
There's a weird Issue with moving the output data after the bouning box is added, but I don't really see how that could be caused by my code and seems more like some kind of caching issue
[Screencast from 2023-08-08 17-27-49.webm](https://github.com/GraphiteEditor/Graphite/assets/44641688/ad36e159-fdc6-42a2-a0cf-f04b1ac621c0)

isiko:
Should be fixed now

TrueDoctor (approved):

PR #1381, opened 2023-08-10
Fix undo not immediately removing copy/pasted layer

xkef:
This should fix the following item in the approachable starters (#989):

> Undo does not work when copy/pasting layers (or groups), it should remove the layer that was pasted. [(As posted in Discord)](https://discord.com/channels/731730685944922173/881073965047636018/1080691819593076847)

The problem was the ordering of the responses vector. I don't quite understand what the rational was to use `add_front` for `DocumentOperation::InsertLayer`.

### DIFF
--- a/editor/src/messages/portfolio/portfolio_message_handler.rs
+++ b/editor/src/messages/portfolio/portfolio_message_handler.rs
@@ -393,16 +393,16 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
 						for entry in data.iter().rev() {
 							let destination_path = [shallowest_common_folder.to_vec(), vec![generate_uuid()]].concat();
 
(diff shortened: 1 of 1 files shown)


TrueDoctor:
Can you run git blame to figure out when this was changed to add_front?

xkef:
> Can you run git blame to figure out when this was changed to add_front?

I think it always was `add_front` (before https://github.com/GraphiteEditor/Graphite/pull/1186/files#diff-e57d2462cb27f93fc524ef7a09bec9c23293486221d207fca2563eb2173d5922R410-R438 it was `push_front`). Originally, it was added like this: https://github.com/GraphiteEditor/Graphite/pull/564/files#diff-bc51ada277cc51d596384df21f5d513c0b78fa8b78e65ca4b673ddee122abf08R369-R398.

PR #1383, opened 2023-08-11
Add more Math Nodes

isiko:
Added a few more mathematical operations as nodes. With this PR, all operations mentioned in #912 should be implemented + some other operations like rounding and the natural logarithm.

# Operations Implemented by this PR:
* Floor
* Ceil
* Round
* Absolute Value
* Logarithm
* Natural Logarithm
* sin
* cos
* tan


node-graph/gcore/src/ops.rs
```diff
@@ -83,6 +83,80 @@ where
...
+	input.ln()
+}
+
+// Sine
+pub struct SineParameterNode;
+
+#[node_macro::node_fn(SineParameterNode)]
+fn ln(input: f32) -> f32 {
+	input.sin()
+}
```
TrueDoctor, line 142: The I think none of these nodes should have Parameter in their name as they don't actually take a second argument. (With AddParameter, the parameter is added to the primary input, this is not the case here)

isiko: Sorry for that, I just copy-pasted without thinking. Should be fixed now

PR #1386, opened 2023-08-13
Add 'select all points' method to ShapeState struct

### LINKED ISSUE #1318: Select all points in path tool
The user should be able to press <kbd>Ctrl</kbd><kbd>A</kbd> to select all the points. This should happen only in the path tool, and should not change the currently selected layers.

In order to achieve this, a new function should be added to the `ShapeState` struct that allows selection of all points in all layers. This function should iterate through `&mut self.selected_shape_state`, accessing the current vector data for each layer. For each group in `vector_data.manipulator_groups()` and for each selected type in `[SelectedType::Anchor, SelectedType::InHandle, SelectedType::OutHandle]`, we need to run `state.select_point(ManipulatorPointId::new(group.id, selected_type));`.

gautamprikshit1:
Closes #1318


gautamprikshit1:
@Keavon @0HyperCube

0HyperCube (approved):

editor/src/messages/input_mapper/default_mapping.rs
```diff
@@ -177,6 +177,7 @@ pub fn default_mapping() -> Mapping {
 		entry!(KeyDown(Lmb); action_dispatch=PathToolMessage::DragStart { add_to_selection: Shift }),
 		entry!(PointerMove; refresh_keys=[Alt, Shift], action_dispatch=PathToolMessage::PointerMove { alt_mirror_angle: Alt, shift_mirror_distance: Shift }),
 		entry!(KeyDown(Delete); action_dispatch=PathToolMessage::Delete),
+		entry!(KeyDown(KeyA); modifiers=[Control], action_dispatch=PathToolMessage::SelectAllPoints),
```
Keavon, line 180: You need to also declare that this action is active in the relevant context (likely in the Path tool). See the message in Discord where I gave you that hint (involving the `actions!(...)` macro) for more info. Once implemented, please build your code and open the editor to test that this PR's feature works and then set it to Ready for Review (instead of Draft) status and ping me again for another QA test. Lookin' forward to it :)

Keavon, line 180: You need to also declare that this action is active in the relevant context (likely in the Path tool). See the message in Discord where I gave you that hint (involving the `actions!(...)` macro) for more info. Once implemented, please build your code and open the editor to test that this PR's feature works and then set it to Ready for Review (instead of Draft) status and ping me again for another QA test. Lookin' forward to it :)

Keavon (changes requested):

Keavon:
!build

Keavon:
This looks pretty good from my QA testing. However upon playing with it, realized it would probably be more useful to people if it selects all the anchors but not the handles. And if existing handles are already selected, I think(?) we want to just overwrite that by deselecting all the handles and selecting all the anchors.

Keavon:
!build

Keavon:
There is one other bug I'm noticing which you didn't cause, but it's a bug that interacts with the feature you're adding. I was wondering if you could take on the extra challenge to try and fix it too. Currently, if you have multiple points selected, clicking on one of those selected points (which deselects all but then re-selects that one clicked point) doesn't update the overlays so it visually appears like that click didn't do anything (however, if you hit G to move the selected points around, you'll see how it's indeed the one point that's selected).

So the fix is to find where to update the overlays for this specific code path.

gautamprikshit1:
!build

Keavon:
!build

Keavon (approved):
Awesome 🎉

Issue #1125, opened 2023-04-15
Make the Spline tool non-destructive


Keavon:
Instead of adding points to a temporary list stored in the Spline tool, which gets destructively committed when the user is done and turned into a Bezier curve, move that spline -> Bezier logic into a node and make the Spline tool draw a polyline which gets fed into the spline interpolator node.

0HyperCube:
Part of #134

Keavon:
Closing as part of #134.

PR #1389, opened 2023-08-14
New nodes: shape/curve primitives

Ezbaze:
https://github.com/GraphiteEditor/Graphite/issues/912

New Nodes:
- Line
- Spline
- Rectangle
- Circle
- Elipse
- Polygon
- Star


Keavon:
Shouldn't 'Spline' be called 'Arc'?

Ezbaze:
> Shouldn't 'Spline' be called 'Arc'? Or alternatively, take N inputs if that'd be possible.

I think calling it a 'Arc' would be a bit misleading as you wouldn't really be able to use it in that mannar.

As for the "N inputs", I was thinking of making that a seperate mode inside the node itself or making it into a seperate node entirely?

Technically the 'Spline' node resembles the 'Quadratic Bezier' node in Blender

Keavon:
I can't imagine that this Spline node is all that useful all on its own, but the n-input Spline node likely would be useful (and would act as a superset, so your three-point use case is covered using it). I suggest changing it to support the n-input case.

Ezbaze:
Yes, it would also allow the ( https://github.com/GraphiteEditor/Graphite/issues/1125 ) to be completed.

Additionally I think it would be good to modify how the List Inputs currently function (Eg. Stroke Node -> Dashed Lengths ) to function similarly to how the Gradient Input functions, and ideally it would also be good to have an input that takes more than one input (Eg. Join Geometry Node) to make it easier to join Lists

- Gradient Input
image

- Join Geometry Node
image

TrueDoctor:
> Yes, it would also allow the ( #1125 ) to be completed.
> 
> Additionally I think it would be good to modify how the List Inputs currently function (Eg. Stroke Node -> Dashed Lengths ) to function similarly to how the Gradient Input functions, and ideally it would also be good to have an input that takes more than one input (Eg. Join Geometry Node) to make it easier to join Lists (And it would be useful for other types of nodes like the Vector Ops)
> 
> * Gradient Input
>   image
> * Join Geometry Node
>   image

there is a graphene-idiomatic way to construct lists by basically using vertical compositing to chain together multiple push nodes which you could then feed into the list input. You could then of course hide the layer stack to reduce visual clutter and in general design abstractions which hide this from the user, but that is what has to happen in the backend.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -217,6 +218,49 @@ fn vec_f32_input(document_node: &DocumentNode, node_id: NodeId, index: usize, na
...
+				if parts.len() == 2 {
+					let x = parts[0].trim().trim_end_matches(')').trim_start_matches('(').trim().parse::<f64>().ok()?;
+					let y = parts[1].trim().trim_end_matches(')').trim_start_matches('(').trim().parse::<f64>().ok()?;
+					Some(DVec2 { x: x, y: y })
+				} else {
+					None
+				}
+			})
+			.collect::<Option<Vec<_>>>()
+			.map(TaggedValue::VecDVec2)
```
0HyperCube, line 245: If we want to be more forgiving with the syntax, allowing the style of `<path d=` in SVG, we could do something like:
```rs
string
	.split(|c: char| !c.is_alphanumeric() && !matches!(c, '.' | '+' | '-'))
	.filter(|x| !x.is_empty())
	.map(|x| x.parse::<f64>().ok())
	.collect::<Option<Vec<_>>>()
	.map(|numbers| numbers.chunks_exact(2).map(|vals| DVec2::new(vals[0], vals[1])).collect())
	.map(TaggedValue::VecDVec2);
```

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -217,6 +218,49 @@ fn vec_f32_input(document_node: &DocumentNode, node_id: NodeId, index: usize, na
...
+
+	if let NodeInput::Value {
+		tagged_value: TaggedValue::VecDVec2(x),
+		exposed: false,
+	} = &document_node.inputs[index]
+	{
+		widgets.extend_from_slice(&[
+			Separator::new(SeparatorType::Unrelated).widget_holder(),
+			text_props
+				.value(x.iter().map(|v| v.to_string()).collect::<Vec<_>>().join(", ").replace("]", ")").replace("[", "("))
```
0HyperCube, line 256: Could be shortened to:
```rs
.value(x.iter().map(|v| format!("({}, {})", v.x, v.y)).collect::<Vec<_>>().join(", ")
```

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -1139,6 +1183,58 @@ pub fn modulo_properties(document_node: &DocumentNode, node_id: NodeId, _context
 	vec![operand("Modulo", 1)]
 }
 
+pub fn circle_properties(document_node: &DocumentNode, node_id: NodeId, _context: &mut NodePropertiesContext) -> Vec<LayoutGroup> {
+	let operand = |name: &str, index| {
+		let widgets = number_widget(document_node, node_id, index, name, NumberInput::default(), true);
+
+		LayoutGroup::Row { widgets }
+	};
+	vec![operand("Radius", 1)]
```
0HyperCube, line 1192: There is no real need to have an operand closure only to call it once. Could be shortened to:
```rs
let radius = number_widget(document_node, node_id, 1, "radius", NumberInput::default(), true);
vec![LayoutGroup::Row { widgets: radius }]
```

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -6,19 +6,96 @@ use bezier_rs::Subpath;
 
 use glam::DVec2;
 
-pub struct UnitCircleGenerator;
+#[derive(Debug, Clone, Copy)]
+pub struct UnitCircleGenerator<Radius> {
```
0HyperCube, line 10: Where you have mentioned unit in these node names is outdated (as the output can be of any size).

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -6,19 +6,96 @@ use bezier_rs::Subpath;
...
+	radius_x: RadiusX,
+	radius_y: RadiusY,
+}
+
+#[node_macro::node_fn(UnitElipseGenerator)]
+fn unit_elipse(_input: (), radius_x: f32, radius_y: f32) -> VectorData {
+	let radius_x = radius_x.into();
+	let radius_y: f64 = radius_y.into();
+	let corner1 = DVec2::new(radius_x * -1., radius_y * -1.);
+	let corner2: DVec2 = DVec2::new(radius_x, radius_y);
```
0HyperCube, line 31: Could be simplified to:
```rs
let radius = DVec2::new(radius_x as f64, radius_y as f64);
let corner1 = -radius;
let corner2 = radius;
```

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -6,19 +6,96 @@ use bezier_rs::Subpath;
...
+	size_x: SizeX,
+	size_y: SizeY,
+}
+
+#[node_macro::node_fn(UnitRectangleGenerator)]
+fn unit_square(_input: (), size_x: f32, size_y: f32) -> VectorData {
+	let size_x = (size_x / 2.).into();
+	let size_y = (size_y / 2.).into();
+	let corner1 = DVec2::new(size_x * -1., size_y * -1.);
+	let corner2 = DVec2::new(size_x, size_y);
```
0HyperCube, line 46: The size here is actual half of the side - either rename it or change it to be the actual size:
```rs
let size = DVec2::new(size_x as f64, size_y as f64);
let corner1 = -size / 2.;
let corner2 = size / 2.;
```

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -6,19 +6,96 @@ use bezier_rs::Subpath;
...
+	points: Points,
+	radius: Radius,
+	inner_radius: InnerRadius,
+}
+
+#[node_macro::node_fn(UnitStarGenerator)]
+fn unit_star(_input: (), points: u32, radius: f32, inner_radius: f32) -> VectorData {
+	let points = points.into();
+	let radius = (radius * 2.).into();
+	let inner_radius = (inner_radius * 2.).into();
```
0HyperCube, line 75: These actually refer to the diameter and would ideally be named as such.

node-graph/interpreted-executor/src/node_registry.rs
```diff
@@ -630,7 +630,13 @@ fn node_registry() -> HashMap<NodeIdentifier, HashMap<NodeIOTypes, NodeConstruct
...
 		register_node!(graphene_core::vector::BoundingBoxNode, input: VectorData, params: []),
 		register_node!(graphene_core::vector::CircularRepeatNode<_, _, _>, input: VectorData, params: [f32, f32, u32]),
-		register_node!(graphene_core::vector::generator_nodes::UnitCircleGenerator, input: (), params: []),
+		register_node!(graphene_core::vector::generator_nodes::UnitCircleGenerator<_>, input: (), params: [f32]),
+		register_node!(graphene_core::vector::generator_nodes::UnitElipseGenerator<_, _>, input: (), params: [f32, f32]),
+		register_node!(graphene_core::vector::generator_nodes::UnitRectangleGenerator<_, _>, input: (), params: [f32, f32]),
+		register_node!(graphene_core::vector::generator_nodes::UnitPolygonGenerator<_, _>, input: (), params: [u32, f32]),
+		register_node!(graphene_core::vector::generator_nodes::UnitStarGenerator<_, _, _>, input: (), params: [u32, f32, f32]),
+		register_node!(graphene_core::vector::generator_nodes::UnitLineGenerator<_, _>, input: (), params: [DVec2, DVec2]),
+		register_node!(graphene_core::vector::generator_nodes::UnitSplineGenerator<_>, input: (), params: [Vec<DVec2>]),
```
0HyperCube, line 639: The node names here are outdated as they include `Unit` when they actually output arbitrary sizes.

0HyperCube (approved):
Works well - I'm excited to see these integrated into the tools.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -217,6 +218,37 @@ fn vec_f32_input(document_node: &DocumentNode, node_id: NodeId, index: usize, na
 	}
 	widgets
 }
+
+fn vec_dvec2_input(document_node: &DocumentNode, node_id: NodeId, index: usize, name: &str, text_props: TextInput, blank_assist: bool) -> Vec<WidgetHolder> {
+	let mut widgets = start_widgets(document_node, node_id, index, name, FrontendGraphDataType::Color, blank_assist);
+
+	let from_string = |string: &str| {
+		let string = &Regex::new(r",+").unwrap().replace_all(string, ",").to_string();
```
0HyperCube, line 226: This line is now unnecessary (you can also remove the regex dependency).

Keavon (approved):

PR #1399, opened 2023-08-24
New nodes: Logical boolean operations (OR, AND, XOR, NOT)

isiko:
Adds these boolean operator nodes:
- Not
- And
- OR
- XOR


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -1913,6 +1913,51 @@ fn static_nodes() -> Vec<DocumentNodeType> {
...
+			inputs: vec![
+				DocumentInputType::value("First", TaggedValue::Bool(false), true),
+				DocumentInputType::value("Second", TaggedValue::Bool(false), true),
+			],
+			outputs: vec![DocumentOutputType::new("Output", FrontendGraphDataType::Boolean)],
+			properties: node_properties::logic_operator_properties,
+			..Default::default()
+		},
+		DocumentNodeType {
+			name: "AND",
```
0HyperCube, line 1929: Thoughts on capitalisation @Keavon? They are often capitalised in logic but look slightly out of place with the other title case nodes.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -1913,6 +1913,51 @@ fn static_nodes() -> Vec<DocumentNodeType> {
...
+			category: "Logic",
+			identifier: NodeImplementation::proto("graphene_core::logic::LogicAndNode<_>"),
+			inputs: vec![
+				DocumentInputType::value("First", TaggedValue::Bool(false), true),
+				DocumentInputType::value("Second", TaggedValue::Bool(false), true),
```
0HyperCube, line 1934: We usually set the default exposed values of the second inputs to false (so the last boolean in the `DocumentInputType::value`). This means that they do not by default show up as dots in the node graph leading to reduced clutter. We probably should document this somewhere.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
```diff
@@ -1913,6 +1913,51 @@ fn static_nodes() -> Vec<DocumentNodeType> {
...
+			properties: node_properties::logic_operator_properties,
+			..Default::default()
+		},
+		DocumentNodeType {
+			name: "XOR",
+			category: "Logic",
+			identifier: NodeImplementation::proto("graphene_core::logic::LogicXorNode<_>"),
+			inputs: vec![
+				DocumentInputType::value("First", TaggedValue::Bool(false), true),
+				DocumentInputType::value("Second", TaggedValue::Bool(false), true),
```
0HyperCube, line 1946: Thoughts about naming "Second" here @Keavon? "Operator" is used in the properties for this and I'm wondering if we should try and be consistent with the naming?

node-graph/gcore/src/logic.rs
```diff
@@ -7,3 +7,38 @@ fn log_to_console<T: core::fmt::Debug>(value: T) -> T {
...
+fn logic_and(first: bool, second: bool) -> bool {
+	first && second
+}
+
+pub struct LogicXorNode<Second> {
+	second: Second,
+}
+
+#[node_macro::node_fn(LogicXorNode)]
+fn logic_and(first: bool, second: bool) -> bool {
```
0HyperCube, line 34: The function name should be `logic_xor` (it doesn't matter as it is not used anywhere - just to reduce confusion).

node-graph/gcore/src/logic.rs
```diff
@@ -7,3 +7,38 @@ fn log_to_console<T: core::fmt::Debug>(value: T) -> T {
...
+	//(first && !second) || (!first && second)
```
0HyperCube, line 35: Did you mean to leave this comment in? If so, could you please add a space after the `//`?

0HyperCube (commented):
Looks good - just some minor queries.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
Keavon: I think maybe "Operand A" and "Operand B" might be the best options? I don't love it but can't think of anything better.

0HyperCube: Using "Operand B" for the properties panel as well? Or keep it is just "Operator"?

Keavon: I think we'd want to be consistent so it should be called "Operand B" in the Properties panel also. "Operator" would be incorrect (that refers to the AND/OR/XOR/etc., not the left and right hand values which are the operands).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
Keavon: I think Title Case is best for all of these except XOR since the other three are English words.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
isiko: I also used that naming for the `Min`, `Max` and `Equality`Nodes, should I update those as well?
(I think I originally got the names from the parameters of the functions that implement the nodes)

node-graph/gcore/src/logic.rs
isiko: Sorry I left that in, it will be fixed with the next push

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
isiko: Sure, but wouldn't you want to get both inputs to these nodes from other nodes? Most of these nodes either do nothing or always output the same value if you manually set the second value, so I thought that the user would expose the second input manually anyway.

0HyperCube: Yes - that makes sense.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
0HyperCube: Yes - I think it is probably best to try and keep the things consistent.

isiko:
Should be all fixed now :)

0HyperCube:
The "Not" node has, in the properties "Operand B" where there should be nothing.
Toggling "Operand B" using the expose dot for the other nodes actually toggles the first input.

Keavon (approved):

Keavon:
Hi @isiko, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

PR #1400, opened 2023-08-25
Remove console spam

0HyperCube:
Reduce the amount of unnecessary things logged by:
- Not logging thumbnails for layer nodes.
- Fixing up the passive event listeners (to prevent chrome warning) - note that I had to use a div instead of the LayoutRow div wrapper component (I'm not sure what advantage that provides other than automatically assigning a class and causing confusing css when setting additional styles).
- Fix css being set as `minWidth` instead of `min-width` and so not being recognised (although it doesn't seem to change the layout - it may reduce confusion in future).
- Use `null` instead of `undefined` for no min-width to remove firefox warning (thanks again js for two nil types).
- Comment out bad selectors (where `:hover` was used after a pseudo-element)

### DIFF
--- a/frontend/src/components/Editor.svelte
+++ b/frontend/src/components/Editor.svelte
@@ -207,9 +207,9 @@
 				border: 2px solid transparent;
 				border-radius: 10px;
 
-				&:hover {
-					box-shadow: inset 0 0 0 1px var(--color-6-lowergray);
-				}
+				// &:hover {
+				// 	box-shadow: inset 0 0 0 1px var(--color-6-lowergray);
+				// }
 			}
 
 			&::-webkit-scrollbar-thumb {
@@ -219,9 +219,9 @@
 				border-radius: 10px;
 				margin: 2px;
 
-				&:hover {
-					background-color: var(--color-6-lowergray);
-				}
+				// &:hover {
+				// 	background-color: var(--color-6-lowergray);
+				// }
 			}
 
 			&::-webkit-scrollbar-corner {
--- a/frontend/src/components/Editor.svelte
+++ b/frontend/src/components/Editor.svelte
@@ -206,10 +206,10 @@
 				box-shadow: inset 0 0 0 1px var(--color-5-dullgray);
 				border: 2px solid transparent;
 				border-radius: 10px;
+			}
 
-				// &:hover {
-				// 	box-shadow: inset 0 0 0 1px var(--color-6-lowergray);
-				// }
+			&:hover::-webkit-scrollbar-track {
+				box-shadow: inset 0 0 0 1px var(--color-6-lowergray);
 			}
 
 			&::-webkit-scrollbar-thumb {
@@ -218,10 +218,10 @@
 				border: 2px solid transparent;
 				border-radius: 10px;
 				margin: 2px;
+			}
 
-				// &:hover {
-				// 	background-color: var(--color-6-lowergray);
-				// }
+			&:hover::-webkit-scrollbar-thumb {
+				background-color: var(--color-6-lowergray);
 			}
 
 			&::-webkit-scrollbar-corner {
--- a/frontend/src/components/layout/LayoutCol.svelte
+++ b/frontend/src/components/layout/LayoutCol.svelte
@@ -23,6 +23,7 @@
 	}
 </script>
 
+<!-- Excluded events because these require `|passive` or `|nonpassive` modifiers. Use a <div> for these instead: `on:wheel`, `on:touchmove`, `on:touchstart` -->
 <div
 	class={`layout-col ${className} ${extraClasses}`.trim()}
 	class:scrollable-x={scrollableX}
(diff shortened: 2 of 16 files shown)


frontend/src/components/Editor.svelte
```diff
@@ -219,9 +219,9 @@
 				border-radius: 10px;
 				margin: 2px;
 
-				&:hover {
-					background-color: var(--color-6-lowergray);
-				}
+				// &:hover {
+				// 	background-color: var(--color-6-lowergray);
+				// }
```
Keavon, line 224: Same.

frontend/src/components/layout/LayoutCol.svelte
```diff
@@ -53,7 +53,6 @@
 	on:mouseout
 	on:mouseup
 	on:select
-	on:wheel
```
Keavon, line 56: Why did these three need to be removed?

0HyperCube: If you want to use them then you have to mark them as either `|passive` or `|nonpassive` in order to avoid a warning in chromium. This is impossible to do whilst they are in this ugly wrapper component.

frontend/src/components/views/Graph.svelte
```diff
@@ -567,20 +566,18 @@
 	});
 </script>
 
-<LayoutRow
-	class="graph"
+<div
+	class=".layout-row graph"
```
Keavon, line 570: Shouldn't be a `.` here. Also I'd prefer to not have a DIV mimic a component (`LayoutRow`) which it isn't, so I'd prefer just adding the same CSS properties that `.layout-row` gets to the CSS for `.graph` and that way it can avoid pretending to be a `layout-row`.

frontend/src/components/widgets/buttons/TextButton.svelte
```diff
@@ -22,7 +22,7 @@
 	class:emphasized
 	class:disabled
 	class:sharp-right-corners={sharpRightCorners}
-	style:min-width={minWidth > 0 ? `${minWidth}px` : undefined}
```
Keavon, line 25: That's too bad there's a warning from `undefined` :( In general I try to forcibly use `undefined` everywhere to standardize on it instead of `null`, even so some JS APIs which return `null` get often converted to `undefined` with the `||` operator where that's safe. But if this produces warnings, I guess we gotta stick with `null`.

frontend/src/components/widgets/inputs/CurveInput.svelte
```diff
@@ -234,7 +234,6 @@
 			fill: var(--color-1-nearblack);
 			stroke: var(--color-e-nearwhite);
 			stroke-width: 0.01;
-			r: 0.02;
```
Keavon, line 237: Ah, it looks like this isn't a valid CSS property pre-SVG2. But it will be when/if browsers ever support SVG2. But I think it's still working as intended (at least in Chrome) with this property applied? I don't think removing it is okay without adding it back in the SVG circle elements as a proper attribute. Can you please add it to the relevant `<circle>`s?

frontend/src/components/layout/LayoutCol.svelte
Keavon: I see, I wasn't aware of `|passive| and `|nonpassive` in Svelte. For context, the goal was to port `LayoutCol` and `LayoutRow` from Vue with as few syntactic changes as possible, even though Svelte doesn't handle this concept as nicely as Vue did. Maybe someday I'll go and find a better solution for this. But that'll take a while and I need to experiment with approaches until I find one I like best. So this wrapper component concept is definitely not the cleanest but it's acceptable *for the moment* given the constraints.

For this, can we just do two lines for `on:wheel|passive` and `on:wheel|nonpassive` or does it have to be one or the other? Can we pick a default that will almost always be what we want?

0HyperCube: You have to pick one (as otherwise the event would be forwarded twice). You cannot override your choice. You could choose a default - if you choose passive then no handlers can use preventDefault; if you choose nonpassive then you would never be able to handle scroll passively (which has better performance).

Keavon: Thanks for explaining. I'll have to think about that and dig around in the code tomorrow once I'm not so tired to decide on a default. Or if you can pick one that seems sensible to you (like if you know we have no usages of preventDefault anywhere in the code) then go for that if you'd like. But if in doubt, I can take a closer look tomorrow. Thank you!

TrueDoctor:
The rust code looks good

Keavon (approved):

frontend/src/components/widgets/buttons/TextButton.svelte
Keavon: I did some research and found that an empty string works instead of `null`, which helps us avoid `null` throughout our JS codebase in favor of `undefined`.

PR #1404, opened 2023-08-29
Add Path tool options for editing X/Y point coordinates

### LINKED ISSUE #1229: Add point position inputs to Path tool options bar
In place of these red boxes, I'd like to have two number inputs for X and Y values that display the position of the currently selected point.
image

Other tools have widgets there, such as the Pen tool:
image

That bar where the widgets go is called the options bar.

If no points are selected, or if multiple points are selected, the number inputs should be disabled.

mobile-bungalow:
Closes #1229

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -152,20 +152,22 @@ impl ShapeState {
 		self.selected_shape_state.values().flat_map(|state| &state.selected_points)
 	}
 
-	/// Moves an anchor to a new position, new_position must be in art board space.
+	/// Moves a control point to a new position, new_position must be in art board space.
 	/// Returns Some(()) if successful and None otherwise.
-	pub fn reposition_anchor_point(&self, point: &ManipulatorPointId, responses: &mut VecDeque<Message>, document: &Document, new_position: DVec2, layer_path: &[u64]) -> Option<()> {
-		if point.manipulator_type.is_handle() {
-			warn!("cannot call `reposition anchor` on a handle");
-			return None;
-		}
-
+	pub fn reposition_control_point(&self, point: &ManipulatorPointId, responses: &mut VecDeque<Message>, document: &Document, new_position: DVec2, layer_path: &[u64]) -> Option<()> {
 		let layer = document.layer(layer_path).ok()?;
 		let vector_data = layer.as_vector_data()?;
 		let transform = layer.transform.inverse();
 		let position = transform.transform_point2(new_position - layer.pivot);
 		let group = vector_data.manipulator_from_id(point.group)?;
 		let delta = position - point.manipulator_type.get_position(group)?;
+
+		if point.manipulator_type.is_handle() {
+			responses.add(GraphOperationMessage::Vector {
+				layer: layer_path.to_vec(),
+				modification: VectorDataModification::SetManipulatorHandleMirroring { id: group.id, mirror_angle: false },
+			});
+		}
 
 		let mut move_point = |point: ManipulatorPointId| {
 			let Some(position) = point.manipulator_type.get_position(group) else {
@@ -178,8 +180,10 @@ impl ShapeState {
 		};
 
 		move_point(*point);
-		move_point(ManipulatorPointId::new(point.group, SelectedType::InHandle));
-		move_point(ManipulatorPointId::new(point.group, SelectedType::OutHandle));
+		if !point.manipulator_type.is_handle() {
+			move_point(ManipulatorPointId::new(point.group, SelectedType::InHandle));
+			move_point(ManipulatorPointId::new(point.group, SelectedType::OutHandle));
+		}
 
 		Some(())
 	}
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -27,7 +27,6 @@ impl SelectedLayerState {
 	pub fn clear_points(&mut self) {
 		self.selected_points.clear();
 	}
-
 	pub fn selected_points_count(&self) -> usize {
 		self.selected_points.len()
 	}
@@ -152,8 +151,8 @@ impl ShapeState {
 		self.selected_shape_state.values().flat_map(|state| &state.selected_points)
 	}
 
-	/// Moves a control point to a new position, new_position must be in art board space.
-	/// Returns Some(()) if successful and None otherwise.
+	/// Moves a control point to a `new_position` in document space.
+	/// Returns `Some(())` if successful and `None` otherwise.
 	pub fn reposition_control_point(&self, point: &ManipulatorPointId, responses: &mut VecDeque<Message>, document: &Document, new_position: DVec2, layer_path: &[u64]) -> Option<()> {
 		let layer = document.layer(layer_path).ok()?;
 		let vector_data = layer.as_vector_data()?;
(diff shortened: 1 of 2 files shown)


editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -72,13 +107,49 @@ impl ToolMetadata for PathTool {
...
+			Layout::WidgetLayout(WidgetLayout::default())
+		}
 	}
 }
 
 impl<'a> MessageHandler<ToolMessage, &mut ToolActionHandlerData<'a>> for PathTool {
 	fn process_message(&mut self, message: ToolMessage, responses: &mut VecDeque<Message>, tool_data: &mut ToolActionHandlerData<'a>) {
 		self.fsm_state.process_event(message, &mut self.tool_data, tool_data, &(), responses, true);
+
+		let new_point = get_single_selected_point(&tool_data.document.document_legacy, tool_data.shape_editor);
```
mobile-bungalow, line 148: Is it idiomatic in this code base to run code on every message? most messages can change the selected point so it seemed appropriate here.

Keavon: Good idea to ask! While not terrible, I think it might be slightly more idiomatic to create another message called something like "selected points count changed" and make sure that any message handler which modifies the count calls that. Alternatively, that could be an event, which is a special type of message that allows subscriptions by users who need that information (see `editor/src/messages/broadcast/broadcast_event.rs`).

mobile-bungalow: I'm going to go with the first option. To do the second properly I feel like we would have to make every tool broadcast it's updates to the selected point set, or diff the set repeatedly, which seems out of scope for this PR. Let me add a new message and move the point update code to the relevant match case.

Keavon: Perfect, and yeah, I was mostly mentioning the event concept for your future awareness. I agree using a regular message is likely more appropriate here.

mobile-bungalow: thanks!

mobile-bungalow: I added the message, the `send_layout` call still has to happen outside the main match body because the `LayoutHolder` implementation is on `PathTool`.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -148,6 +152,36 @@ impl ShapeState {
 		self.selected_shape_state.values().flat_map(|state| &state.selected_points)
 	}
 
+	/// Moves an anchor to a new position, new_position must be in art board space.
+	/// Returns Some(()) if successful and None otherwise.
+	pub fn reposition_anchor_point(&self, point: &ManipulatorPointId, responses: &mut VecDeque<Message>, document: &Document, new_position: DVec2, layer_path: &[u64]) -> Option<()> {
+		if point.manipulator_type.is_handle() {
+			trace!("cannot call `reposition anchor` on a handle");
```
0HyperCube, line 159: This is probably better as a warning `warn!()`.

0HyperCube (approved):
Looks good.

0HyperCube:
You might also need to run `cargo fmt`.

editor/src/messages/tool/common_functionality/shape_editor.rs
mobile-bungalow: agreed and fixed.

mobile-bungalow:
For some reason running rustfmt does not correct `let ... else` blocks for me. maybe my toolchain version is out of sync with the actions.

Keavon:
Have you tried `rustup update`? They just added rustfmt support for `let ... else`. It's also possible the outdatedness is on the CI end but I'd try updating yours first.

mobile-bungalow:
yeah that worked. Really did not expect that because I installed rust on this machine two days ago.

Keavon:
Rust 1.72 came out [on the 24th](https://blog.rust-lang.org/2023/08/24/Rust-1.72.0.html) so I'm not sure why it would have installed the older version two days ago. But at least it's fixed now!

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -72,13 +112,47 @@ impl ToolMetadata for PathTool {
 
 impl LayoutHolder for PathTool {
 	fn layout(&self) -> Layout {
-		Layout::WidgetLayout(WidgetLayout::default())
+		if let Some(SingleSelectedPoint { coordinates: DVec2 { x, y }, .. }) = self.tool_data.single_selected_point {
+			let x_loc = NumberInput::new(Some(x))
+				.unit(" px")
+				.label("x")
```
Keavon, line 118: Uppercase please (and `"y"`) also. Also, if you can set `.min_width(120)` that would help avoid the numbers from getting cut off.

Keavon:
This works splendidly for anchors. However it doesn't seem to work for handles, which would be helpful to have. Can you make it work for those too?

mobile-bungalow:
Yes, I can. I wasn't sure if those were in the spec. I'll finish that up tonight.

Keavon:
Perfect. I'm loving your efficiency!

mobile-bungalow:
Set up for handles as well. Notes for the future

1.) manually entering handle coordinates turns mirroring off for handle pairs, as is consistent with illustrator and figma.
2.) Currently there is no way to turn mirroring back on, such as by dragging the handles to be colinear.  However mirroring will be turned on when the user nudges. We should likely make the path editing experience similar to other vector editors and have easier UX on handles in the path tool.

Keavon:
!build

editor/src/messages/tool/tool_messages/path_tool.rs
mobile-bungalow: just saw this, fixing and pushing. sorry for the rush.

Keavon:
Thanks. Do you happen to be proficient enough with Git to edit your branch to remove all the merge conflicts and rebase everything into a linear history on top of master? Or even better, onto `master~1` (the commit before master), since I'd like to merge yours before my WIP commit that's living in master right now (I'll temporarily remove it, merge your PR, then add it back so I can keep editing it). If you don't know how to do all the necessary Git trickery, I can probably do it for you.

mobile-bungalow:
I can resolve the merge conflicts and rebase to master~1, but it seems like there is a lot of churn going on right now through force pushes and i would prefer not to step on your toes if you are trying to get something sensitive with the CI tooling working. Could you do the rebase?

Keavon:
!build

Keavon:
Ok, I removed all the merge commits to make it one nice linear commit chain. Then I rebased it on master (which is done now, after several hours of churn while I had to debug CI runner stuff—unfortunately—directly on the master branch). In the future it'd be cleanest and least error-prone to opt for rebasing over merging master into your PR branch when possible (sometimes that's difficult because of many rebase conflicts, so a merge is easier when you only need to deal with those conflicts once). But what you did is perfect acceptable, it just got a bit icky due to my repeated force pushes to master which is abnormal but was necessary today.

Keavon:
!build

mobile-bungalow:
Totally understand the need for a linear git history. I will rebase in the future, I'm so used to merging.

Keavon:
It's not strictly necessary since every PR gets squashed in the end, but it helps avoid some hairiness while the PR is being worked on. For future reference, there's also this handy button in the UI (just slightly hidden behind a dropdown).

https://github.com/GraphiteEditor/Graphite/assets/4388688/693621ef-3081-4a27-8416-e8b37a79e8dd

PR #1405, opened 2023-08-29
Reconnect links when deleting a node

### LINKED ISSUE #1410: When deleting a node, the upstream/downstream links should be reconnected

dhruv-ahuja:
This PR aims to reconnect the deleted node's list of inputs to point to the deleted node's primary input, if the inputs point to the deleted node's primary outputs.

Closes #1410


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -381,8 +381,23 @@ impl NodeGraphMessageHandler {
...
+					first_input_node = Some(node.inputs[0].clone());
+					warn!("first node: {:?}", first_input_node);
+
+					// fetch output node only if there's an input node for comparison too
+					if network.outputs.len() > 0 {
+						first_output_node = Some(&network.outputs[0]);
+						warn!("first output node: {:?}", first_output_node);
+					}
+				}
+
```
0HyperCube, line 400: There is no guarantee that the deleted node will appear in the hashmap before the other nodes. Hashmaps by definition do not guarantee ordering.

Instead you must find the deleted_node before this loop, and set the first_input_node variable based on this.

You must also ensure that the input actually refers to a node i.e. `if let NodeInput::Node { node_id, .. } = input`.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -381,8 +381,23 @@ impl NodeGraphMessageHandler {
 			warn!("Deleting the output node!");
 			return false;
 		}
+
+		let mut first_input_node: Option<NodeInput> = None;
+		let mut first_output_node: Option<&NodeOutput> = None;
```
0HyperCube, line 386: We should not have a `first_output_node` variable. Please remove it.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -397,12 +412,26 @@ impl NodeGraphMessageHandler {
...
+					let mut refers_to_output_node = false;
+
+					if let Some(input_node) = &first_input_node {
+						if let Some(output_node) = first_output_node {
+							if output_node.node_id == *node_id {
+								refers_to_output_node = true;
+								*input = input_node.clone()
+							}
+						}
+					}
```
0HyperCube, line 426: Let's say you have the following network and delete the greyscale node.
image
We find that the invert node uses the (now deleted) greyscale node as input. We need to modify the input node so that instead refers to something else.

Once we reach this code block, we know that we are dealing with a node with an input that used the deleted node (i.e. the invert node's first input).

We must also check if we are using the primary output from the now deleted greyscale node. You can do this by modifying the above code:
```rs
let NodeInput::Node { node_id, .. } = input else {
	continue;
};
```
to be
```rs
let NodeInput::Node { node_id, output_index, .. } = input else {
	continue;
};
```
Then you can check `if output_index == 0`.

Previously we just reset it to default. Now we need to reset it to the `first_input_node`, if that exists. If it doesn't then we should use the old behaviour `*input = NodeInput::value(tagged_value.clone(), true);`.

0HyperCube (changes requested):

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -381,12 +381,30 @@ impl NodeGraphMessageHandler {
 			warn!("Deleting the output node!");
 			return false;
 		}
+
+		let mut first_input_node: Option<NodeInput> = None;
+
+		if !disconnect_nodes {
+			// check whether the deleting node's first (primary) input is a node
+			for (node_id, node) in network.nodes.iter() {
+				if *node_id == deleting_node_id {
```
0HyperCube, line 390: These two lines could be shortened to `if let Some(node) = network.nodes.get(&deleting_node_id)`

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -381,12 +381,30 @@ impl NodeGraphMessageHandler {
...
+					if node.inputs.len() > 0 {
+						if let NodeInput::Node { .. } = node.inputs[0] {
```
0HyperCube, line 392: These two lines could be shortened to `if matches!(node.inputs.first(), Some(NodeInput::Node { .. }))`

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -381,12 +381,30 @@ impl NodeGraphMessageHandler {
...
+		}
+
 		for (node_id, node) in network.nodes.iter_mut() {
 			if *node_id == deleting_node_id {
 				continue;
 			}
 			for (input_index, input) in node.inputs.iter_mut().enumerate() {
-				let NodeInput::Node { node_id, .. } = input else {
+				// skip iteration if not a node, otherwise extract its values for use
+				// node_id here shadows the node_id in the outer for-loop
```
0HyperCube, line 406: The shadowing here is a bit confusing. To fix this, you could use `let NodeInput::Node { node_id: upstream_node_id, output_index, .. }`. Then you probably don't need this comment (as it is pretty clear that we are skipping if it is not a node input).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -397,20 +415,34 @@ impl NodeGraphMessageHandler {
...
+					}
+
+					if !refers_to_output_node {
+						*input = NodeInput::value(tagged_value.clone(), true);
+					}
 				}
 			}
 			if let DocumentNodeImplementation::Network(network) = &mut node.implementation {
-				Self::remove_references_from_network(network, deleting_node_id);
+				return Self::remove_references_from_network(network, deleting_node_id, disconnect_nodes);
```
0HyperCube, line 437: I've realised that we don't need any recursion here, and so the if let block can actually be deleted.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -436,9 +468,14 @@ impl NodeGraphMessageHandler {
 	}
 }
 
-impl MessageHandler<NodeGraphMessage, (&mut Document, &NodeGraphExecutor, u64, &str)> for NodeGraphMessageHandler {
+impl MessageHandler<NodeGraphMessage, (&mut Document, &NodeGraphExecutor, u64, &str, &InputPreprocessorMessageHandler)> for NodeGraphMessageHandler {
```
0HyperCube, line 471: You might want to consider making this massive tuple into a struct, like:
```rs
pub struct NodeGraphHandlerData<'a> {
    pub document: &mut Document,
    pub node_graph_executor: &NodeGraphExecutor,
    pub document_id: u64,
    pub document_name: &str,
    pub input: &InputPreprocessorMessageHandler
}
```
Passing this around instead of the massive tuple would probably be cleaner.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -512,19 +549,27 @@ impl MessageHandler<NodeGraphMessage, (&mut Document, &NodeGraphExecutor, u64, &
...
 			}
-			NodeGraphMessage::DeleteSelectedNodes => {
+			NodeGraphMessage::DeleteSelectedNodes { disconnect_input_nodes } => {
 				responses.add(DocumentMessage::StartTransaction);
 
+				let disconnect_nodes: bool;
+				if let Some(key) = disconnect_input_nodes {
+					disconnect_nodes = input.keyboard.get(key as usize)
+				} else {
+					disconnect_nodes = false
```
0HyperCube, line 568: This should probably be true?

0HyperCube (commented):
Works very well.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
dhruv-ahuja: Ah actually we only process the nodes' inputs  if this condition is false. Otherwise in case of a true value, that part is totally skipped during code execution.

0HyperCube: I didn't think we would reconnect nodes on Ctrl X - although perhaps we should?

dhruv-ahuja: @Keavon should we reconnect nodes when performing a cut operation?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
dhruv-ahuja: image
Removing the if-let block is causing an `out of bounds access` memory error

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: Keavon says it is fine.

0HyperCube (approved):

Keavon (changes requested):
~~It's currently impossible to delete a node at all.~~ This was a regression in master that I caused without realizing.

Keavon (approved):

PR #1413, opened 2023-09-02
Fix deleting the Path node in "Just a Potted Cactus" demo art

0HyperCube:
Fixes deleting shape nodes from the cactus demo artwork which will soon have to be replaced with a complete document network.

Follow-on to #1411


Keavon (approved):
Thank you!

PR #1415, opened 2023-09-03
Allow toggling smooth/sharp angle from the path tool options

### LINKED ISSUE #1249: Add a sharp/smooth anchor toggle to the Path tool
Similar to #1229, there should be a radio input to toggle between Sharp and Smooth added to the Path tool's options bar.

- Sharp: an anchor's two handles can be any angle
- Smooth: an anchor's two handles are locked at a 180° angle offset

Users can currently toggle between the two while dragging a handle by tapping the <kbd>Alt</kbd> key. This toggle widget in the Properties panel should help visualize the current state of things if an anchor is selected, and/or one or both of its two handles.

If one handle is selected, clicking the toggle to change to Smooth should cause the opposite handle to rotate (but not change length) to be 180° from the selected one. If an anchor and/or both handles are selected (meaning there isn't one specific handle), then clicking switching to Smooth should rotate the angles of both handles to their average rotation (also without affecting their lengths). When switching from Smooth the Sharp, don't change their angles or lengths.

mobile-bungalow:
Closes #1249

Notes: It seems like there is something wrong with alt-drag smooth/sharp swapping and undo behavior. This PR is subject to the same bad behavior.


Keavon:
Did you intend to close this PR?

mobile-bungalow:
Yep, I wanted to wait until I was assigned the task and also there was a regression after merging I wanted to fix before opening the PR.

Keavon:
Ok, I assigned you now (no need to have that be a blocker though, it's just for tracking purposes). In the future, you can just mark the PR as a draft while it's awaiting your finishing touches then reopen it and request a review upon that time. It's fine for PRs, as long as they aren't permanently abandoned (closing is best in that case), to sit as drafts since we will just ignore them when looking at the PR queue. No worries though!

Keavon:
!build

Keavon:
Lookin' good!

Some QA feedback to improve the UX:

- [ ] If just the anchor is selected, switching from Sharp to Smooth should *average* the angles of the two handle vectors. Right now it just picks one at random and makes the other match its angle.
- [ ] The same as above should happen if *both* handles are selected (with or without the anchor, also). Right now it's grayed out in that case.
- [ ] If one of the two handles is selected, but also the anchor is selected, it should do the current behavior for when that handle is selected but no anchor is selected. Right now it's grayed out.

In summary for the three bullet points above, it should accept any combination of the (up to) three control points within the same manipulator group, never graying it out in that case.

Once you have that implemented as described above, the next step would be making it work for multiple manipulator groups. As long as each manipulator group has the same state (solid or smooth), it should show that state. If they differ across manipulator groups, or there are none selected, I suppose we gray out the widget in that case.

mobile-bungalow:
All the features above should be working now, for one and multiple points.

Keavon:
!build

Keavon:
Waiting on the fix for a crash found by @mobile-bungalow as mentioned on Discord.

Keavon:
Fantastic. Ready for QA and review now?

mobile-bungalow:
yup

Keavon:
!build

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -50,6 +50,51 @@ pub struct ManipulatorPointInfo<'a> {
...
+const ANCHOR_OUT: u8 = 0b101;
+const ANCHOR_IN: u8 = 0b110;
+const ALL: u8 = 0b111;
+
+#[derive(Clone, Eq, PartialEq)]
+pub struct ManipulatorGroupData {
+	pub selection_type: u8,
+	pub group_id: ManipulatorGroupId,
+	pub layer_id: Vec<LayerId>,
+}
```
TrueDoctor, line 66: Using Bitflags here will most likely yield worse performance here:
The struct consists of 64bit aligned fields and one 8bit aligned field -> 48bits of padding will be added anyway so we could just as well store 3 bool values on the struct directly and still use the same amount of meromy/cache lines (same memory throughput) but no longer need to spend cycles on the bitshift logic.
I think it would be beneficial to just use bools instead here to improve maintainability.
(and let me know if you have any questions about the performance aspects)

0HyperCube, line 66: We already store the selected points in `SelectedLayerState`.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -201,6 +246,106 @@ impl ShapeState {
...
+					},
+				});
+			}
+			_ => {
+				warn!("The current selection type is invalid");
+			}
+		};
+
+		Some(())
+	}
```
0HyperCube, line 347: We already have a function for changing to and from sharp called `flip_sharp`. Perhaps this could be extended to support set sharp to remove code duplication?

0HyperCube (changes requested):

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -201,6 +246,106 @@ impl ShapeState {
 		Some(())
 	}
 
+	/// Reinsert a control point at the same point to make its handles recalculate their positions based on a new constraint.
+	pub fn blink_manipulator_group(&self, group_data: &ManipulatorGroupData, responses: &mut VecDeque<Message>, document: &Document) -> Option<()> {
```
Keavon, line 250: "blink" is sort of a confusing name, I still don't really have an understanding of why that word was chosen because it doesn't convey the meaning I'd expect in this context (it doesn't happen across multiple frames so there's no "blinking" in any sense of the word I'd intuitively understand it).

editor/src/messages/tool/common_functionality/shape_editor.rs
mobile-bungalow: I might be able to factor `flip_sharps` smoothing code into a separate function to reduce duplication. But I think the functions don't have enough control flow in common for unifying them into one function to make sense.

0HyperCube: They both have the code for removing or creating the handles and detecting if a manipulator group is already sharp. These should probably be separate functions which can be called for both behaviours.

editor/src/messages/tool/common_functionality/shape_editor.rs
mobile-bungalow: renamed and retooled this function to `smooth_selected_groups`.

editor/src/messages/tool/common_functionality/shape_editor.rs
mobile-bungalow: All selected point state is now taken as needed from the `SelectedLayerState` and processed as needed.

editor/src/messages/tool/common_functionality/shape_editor.rs
Keavon: That sounds like a much more sensible name, thank you.

editor/src/messages/tool/common_functionality/shape_editor.rs
mobile-bungalow: all the Shared code between the two functions has been moved into a separate, shared, function.

Keavon (approved):
Everything looks great now! Thanks for your patience with the code review process, it definitely helped to ensure we're committing high-quality code (both in terms of what you wrote and the surrounding parts you're maintaining through its evolution).

PR #1417, opened 2023-09-06
Move DeleteLayer to come before SelectionChanged on DeletedSelectedLayers

### LINKED ISSUE #1393: The Path tool's box selection crashes after a layer was deleted
## Describe the crash

Selecting rectangle in deleted folder causes panic

## Steps to reproduce
- Create a new document
- Create a rectangle
- Select path tool
- ~~Put rectangle in folder (CTRL-G)~~ (not necessary)
- Delete ~~folder~~ rectangle manually
- Select area where the rectangle was
- *panic*

graphite-panic

## Stacktrace
```
panicked at 'called `Option::unwrap()` on a `None` value', editor/src/messages/tool/common_functionality/overlay_renderer.rs:48:92

rs/<@https://editor.graphite.rs/index.70018820.js:25:3480
handleJsMessage@https://editor.graphite.rs/index.70018820.js:17:46128
li/e<@https://editor.graphite.rs/index.70018820.js:17:82971
ei/t.wbg.__wbg_call_4c73e4aecced6a7d/<@https://editor.graphite.rs/index.70018820.js:17:78318
qr@https://editor.graphite.rs/index.70018820.js:17:51476
ei/t.wbg.__wbg_call_4c73e4aecced6a7d@https://editor.graphite.rs/index.70018820.js:17:78280
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[809]:0x2d7a39
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[3545]:0x488fa4
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[3430]:0x482626
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[5382]:0x502f0e
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[5108]:0x4fdc83
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[238]:0xa5ef3
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[1251]:0x367525
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[1867]:0x3d9680
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[302]:0x17ec82
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[2302]:0x41846c
@https://editor.graphite.rs/graphite_wasm_bg.c4523f9b.wasm:wasm-function[3204]:0x47215d
onMouseUp@https://editor.graphite.rs/index.70018820.js:17:54942
ji/action/<@https://editor.graphite.rs/index.70018820.js:17:103328
action@https://editor.graphite.rs/index.70018820.js:17:103372
Async*ji/<@https://editor.graphite.rs/index.70018820.js:25:195
ji@https://editor.graphite.rs/index.70018820.js:25:139
_h@https://editor.graphite.rs/index.70018820.js:45:307873
Kt@https://editor.graphite.rs/index.70018820.js:16:18159
xh@https://editor.graphite.rs/index.70018820.js:45:308056
Mh@https://editor.graphite.rs/index.70018820.js:45:308120
p@https://editor.graphite.rs/index.70018820.js:45:308488
Lt@https://editor.graphite.rs/index.70018820.js:16:15593
Tt@https://editor.graphite.rs/index.70018820.js:16:15228
promise callback*Mt@https://editor.graphite.rs/index.70018820.js:16:15042
Kt/l.ctx</<@https://editor.graphite.rs/index.70018820.js:16:18342
Kt/l.ctx<@https://editor.graphite.rs/index.70018820.js:16:18395
kh/<@https://editor.graphite.rs/index.70018820.js:45:308691
async*M@https://editor.graphite.rs/index.70018820.js:16:11293
Wt/<@https://editor.graphite.rs/index.70018820.js:16:17457
Tt@https://editor.graphite.rs/index.70018820.js:16:15390
Kt@https://editor.graphite.rs/index.70018820.js:16:18690
zh@https://editor.graphite.rs/index.70018820.js:45:308788
@https://editor.graphite.rs/index.70018820.js:45:308864
```

eolculnamo2:
Fixes #1393

Let me know if there is some other underlying issue that needs to be addressed. I'm just skipping most of the body of render_subpath_overlays if the document.generate_transform_relative_to_viewport errors (in this case, from its inability to find the deleted folder)


Keavon:
I went to reproduce the issue as described in #1393 and found that it doesn't actually need to be inside a folder. It happens also if you drag a box selection with the Path tool after deleting the layer. So, based on the title of this PR (but having not looked closely at your code), I ask if the solution you've taken requires it be specific to a folder? If that's the case, you may want to test if it's still an issue without involving the folder step.

Thanks!

eolculnamo2:
Thanks for that! This change fixes the issue, but there's still some weirdness I'd like to figure out. 

The Err which is unwrapped up the stack comes from this function which seems to only be able to Err if the let Ok(folder) condition is met.

```
	pub fn multiply_transforms(&self, path: &[LayerId]) -> Result<DAffine2, DocumentError> {
		let mut root = &self.root;
		let mut trans = self.root.transform;
		for id in path {
			if let Ok(folder) = root.as_folder() {
				root = folder.layer(*id).ok_or_else(|| DocumentError::LayerNotFound(path.into()))?;
			}
			trans = trans * root.transform;
		}
		Ok(trans)
	}
```

Will keep you posted

editor/src/messages/tool/common_functionality/overlay_renderer.rs
```diff
@@ -45,8 +45,9 @@ impl OverlayRenderer {
 	}
 
 	pub fn render_subpath_overlays(&mut self, selected_shape_state: &SelectedShapeState, document: &Document, layer_path: Vec<LayerId>, responses: &mut VecDeque<Message>) {
-		let transform = document.generate_transform_relative_to_viewport(&layer_path).ok().unwrap();
-		if let Ok(layer) = document.layer(&layer_path) {
+		let transform = document.generate_transform_relative_to_viewport(&layer_path).ok();
+		let layer = document.layer(&layer_path);
+		if let (Ok(layer), Some(transform)) = (layer, transform) {
```
0HyperCube, line 50: This sort of works around the issue without really solving it. The core issue is that we appear to handle the `DeleteSelectedLayers` message after the `SelectionChanged` message when handling `DeleteSelectedLayers`. This means that the selection is updated before the layer is actually deleted.

0HyperCube (changes requested):

editor/src/messages/tool/common_functionality/overlay_renderer.rs
0HyperCube: The solution for this would be to modify the implementation of the `DeleteSelectedLayers` message to ensure that `DeleteLayer` is called before `SelectionChanged`.

eolculnamo2: Thanks, I had a feeling that might be the case.

Is the current solution any better?

RBird111:
I've tried the proposed changes and the crash still occurs. It looks like the source is of the crash is the `bezier-rs` library.

Specifically in the `Subpath` module:
1. [Inside of the iterator implementation](https://github.com/GraphiteEditor/Graphite/blob/b9027883a8535abcaf15c51ce3491b8d36734cbb/libraries/bezier-rs/src/subpath/mod.rs#L53C2-L53C2) a call to next on an empty `Subpath` results in a subtract with overflow error. We can kick the can down road by having it end iteration in that case

```rust
	fn next(&mut self) -> Option<Self::Item> {
		if self.subpath.is_empty() {
		    return None;
		}
		let len = self.subpath.len() - 1
		    + match self.subpath.closed {
			true => 1,
			false => 0,
		    };
		if self.index >= len {
		    return None;
		}
		let start_index = self.index;
		let end_index = (self.index + 1) % self.subpath.len();
		self.index += 1;

		Some(self.subpath[start_index].to_bezier(&self.subpath[end_index]))
	}
```

2. [Inside the function `subpath_to_svg`](https://github.com/GraphiteEditor/Graphite/blob/b9027883a8535abcaf15c51ce3491b8d36734cbb/libraries/bezier-rs/src/subpath/core.rs#L143). With the above fix in place a crash occurs in this function after attempting to index into an empty subpath so we can silently exit the function before that happens
```rust
	/// Write the curve argument to the string (the d="..." part)
	pub fn subpath_to_svg(&self, svg: &mut String, transform: glam::DAffine2) -> std::fmt::Result {
		if self.is_empty() {
		    return Ok(());
		}
		let start = transform.transform_point2(self[0].anchor);
		write!(svg, "{SVG_ARG_MOVE}{},{}", start.x, start.y)?;
		for bezier in self.iter() {
		    bezier.apply_transformation(|pos| transform.transform_point2(pos)).write_curve_argument(svg)?;
		    svg.push(' ');
		}
		if self.closed {
		    svg.push_str(SVG_ARG_CLOSED);
		}
		Ok(())
	}
```

These two bandaids will stop the reported crashes but an error does start appearing in the console that I can't track down:
`Error: <path> attribute d: Expected moveto path command ('M' or 'm'), "Z".`

Also I don't know anything about image editing so I'm not sure what a bezier or subpath are but from the way the iterator is implemented and how `Subpath` is being blindly indexed into there's an assumption that there should always be at least one element even in the case of an 'empty' iterator. Something is invalidating that assumption and it's probably the core problem.

eolculnamo2:
https://github.com/GraphiteEditor/Graphite/assets/27943776/e671b261-7d02-4d6d-af26-1496065bfe55

Thanks for trying it out... The fix seems to work when I try it locally -- am I missing something that you're trying? Here's a video that demos this. It shows it breaking with the current code then I make the change in the PR , wait for it to build, then show it without breaking.

0HyperCube (approved):
Looks good.

0HyperCube:
@RBird111 
1. We haven't really considered the cause of empty subpaths. The user probably shouldn't be able to create them (as it would be confusing), but we should also try not to panic where possible.
2. The svg path element defines an arbitrary set of curves, with a syntax like `M10,0 L10,10` (to move to (10,0) and draw a line from there to (10,10)). When dealing with an empty subpath, we likely produce an invalid path, which we should fix.

Keavon:
Hi @eolculnamo2, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

Keavon:
Hi @eolculnamo2, we're grateful of your past code contribution to Graphite, but unless we hear from you by the **end of this weekend** with agreement about relicensing your code to include the MIT license, we will begin steps to remove your code contributions **so that you will no longer be a Graphite code contributor**. We'd much rather avoid this outcome, so please respond now.

PR #1423, opened 2023-09-19
Make 'Close All Documents' not confirm if all open documents are already saved

toadkarter:
**TASK** 
From #code-todo-list ([link here](https://discord.com/channels/731730685944922173/881073965047636018/1150153442833223802))

**ADDITIONS** 
After pressing "Close All", the "Save" dialogue no longer pops if none of the documents have unsaved changes. Some outdated doctests in color.rs were also updated.

**COMMENTS**
Previously, pressing "Close All" would immediately trigger a DialogMessage asking if the user wishes to save their work before closing all documents. I have replaced this with a new PortfolioMessage which checks if there are any unsaved documents currently open. If there are, we pop the previous DialogMessage. If not, we just run the usual "CloseAllDocuments" message without the dialog.

I have decided to call the new message the same as the corresponding DialogMessage for consistency, but if you think this is confusing I would be happy to rename it. 

Please note that I have not added a keyboard shortcut to this action as one did not exist in the original build. 

If you would like me to make any changes or if I have not approached this task correctly please don't hesitate to let me know!

### DIFF
--- a/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
+++ b/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
@@ -71,8 +71,8 @@ impl LayoutHolder for MenuBarMessageHandler {
 						},
 						MenuBarEntry {
 							label: "Close All".into(),
-							shortcut: action_keys!(DialogMessageDiscriminant::CloseAllDocumentsWithConfirmation),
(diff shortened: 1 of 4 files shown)


Keavon:
!build

Keavon (approved):
Code looks clean and from a QA angle it all works as expected. Nice job!

PR #1427, opened 2023-09-25
Fix crash when a cycle is introduced into the graph

### LINKED ISSUE #1392: [Crash Report] Creating cycle with nodes causes crash
## Describe the Crash
Feeding the a layer (or any other) node into itself causes the editor to crash.
image

As noted in the error log, a cycle is detected.
But this should not cause the editor to crash and data to be lost.

## Steps to reproduce
- Create a new document
- Feed the output of the layer node into the input (See image)
- *Crash*

## Stacktrace
```
ns/<@https://editor.graphite.rs/index.0536bdd5.js:25:3480
handleJsMessage@https://editor.graphite.rs/index.0536bdd5.js:17:46021
ai/e<@https://editor.graphite.rs/index.0536bdd5.js:17:82856
Qr/t.wbg.__wbg_call_4c73e4aecced6a7d/<@https://editor.graphite.rs/index.0536bdd5.js:17:78203
Kr@https://editor.graphite.rs/index.0536bdd5.js:17:51369
Qr/t.wbg.__wbg_call_4c73e4aecced6a7d@https://editor.graphite.rs/index.0536bdd5.js:17:78165
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[805]:0x2cf630
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[3504]:0x47c4cc
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[3391]:0x475cd3
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[5321]:0x4f576b
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[822]:0x2d6759
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[822]:0x2d696c
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[822]:0x2d696c
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[240]:0xbb33a
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[830]:0x2dc139
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[522]:0x237c92
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[1889]:0x3db83e
@https://editor.graphite.rs/graphite_wasm_bg.360fb6cb.wasm:wasm-function[5814]:0x4fd462
Hr@https://editor.graphite.rs/index.0536bdd5.js:17:50725
i@https://editor.graphite.rs/index.0536bdd5.js:17:50186
```

Full error log: https://pastebin.com/HgY0xZk4

toadkarter:
Closes #1392

**EXPECTED BEHAVIOUR**
Editor no longer crashes when a cycle is introduced into the node graph.
An error message in the browser output appears when this occurs instead.

**TESTING CRITERIA**
1. Open the editor with a new document.
2. Switch to node graph view.
3. Create a node graph with a cycle in it, as per the screenshot below. 
4. Notice that the editor no longer crashes. 
5. Notice that in the browser output an error message appears saying that a cycle has been detected.

image

**COMMENTS** 
- The main fix here is removing the `panic!` macro in the `topological_sort()` function and instead having that function return a `Result` type. 
- As you can see, there are quite a few changes to the code throughout as this error needs to be propagated to every place where this function is called. 
- I have also amended the tests that call this function to take into account the new return type. I have written a new test that tests the behaviour of `topological_sort()` being called on a graph with a cycle in it, with the expected outcome being an error. 
- Hopefully this is in line with what is expected but I would be more than happy to make any changes.

### DIFF
--- a/node-graph/graph-craft/src/graphene_compiler.rs
+++ b/node-graph/graph-craft/src/graphene_compiler.rs
@@ -8,7 +8,7 @@ use crate::proto::{LocalFuture, ProtoNetwork};
 pub struct Compiler {}
 
 impl Compiler {
-	pub fn compile(&self, mut network: NodeNetwork) -> impl Iterator<Item = ProtoNetwork> {
+	pub fn compile(&self, mut network: NodeNetwork) -> Result<impl Iterator<Item = ProtoNetwork>, String> {
 		println!("flattening");
 		let node_ids = network.nodes.keys().copied().collect::<Vec<_>>();
 		for id in node_ids {
@@ -17,15 +17,23 @@ impl Compiler {
 		network.remove_redundant_id_nodes();
 		network.remove_dead_nodes();
 		let proto_networks = network.into_proto_networks();
-		proto_networks.map(move |mut proto_network| {
-			proto_network.resolve_inputs();
-			proto_network.generate_stable_node_ids();
-			proto_network
-		})
+
+		let proto_networks_result: Result<Vec<ProtoNetwork>, String> = proto_networks
+			.map(move |mut proto_network| {
+				proto_network.resolve_inputs()?;
+				proto_network.generate_stable_node_ids();
+				Ok(proto_network)
+			})
+			.collect();
+
+		match proto_networks_result {
+			Ok(vec) => Ok(vec.into_iter()),
+			Err(error) => Err(error),
+		}
 	}
 	pub fn compile_single(&self, network: NodeNetwork) -> Result<ProtoNetwork, String> {
 		assert_eq!(network.outputs.len(), 1, "Graph with multiple outputs not yet handled");
-		let Some(proto_network) = self.compile(network).next() else {
+		let Some(proto_network) = self.compile(network)?.next() else {
 			return Err("Failed to convert graph into proto graph".to_string());
 		};
 		Ok(proto_network)
--- a/node-graph/graph-craft/src/graphene_compiler.rs
+++ b/node-graph/graph-craft/src/graphene_compiler.rs
@@ -18,18 +18,15 @@ impl Compiler {
 		network.remove_dead_nodes();
 		let proto_networks = network.into_proto_networks();
 
-		let proto_networks_result: Result<Vec<ProtoNetwork>, String> = proto_networks
+		let proto_networks_result: Vec<ProtoNetwork> = proto_networks
 			.map(move |mut proto_network| {
 				proto_network.resolve_inputs()?;
 				proto_network.generate_stable_node_ids();
 				Ok(proto_network)
 			})
-			.collect();
+			.collect::<Result<Vec<ProtoNetwork>, String>>()?;
 
-		match proto_networks_result {
-			Ok(vec) => Ok(vec.into_iter()),
-			Err(error) => Err(error),
-		}
+		Ok(proto_networks_result.into_iter())
 	}
 	pub fn compile_single(&self, network: NodeNetwork) -> Result<ProtoNetwork, String> {
 		assert_eq!(network.outputs.len(), 1, "Graph with multiple outputs not yet handled");
--- a/node-graph/graph-craft/src/proto.rs
+++ b/node-graph/graph-craft/src/proto.rs
@@ -371,7 +371,7 @@ impl ProtoNetwork {
 			}
 		}
 		self.reorder_ids()?;
-		return Ok(());
+		Ok(())
 	}
 
 	fn replace_node_id(&mut self, outwards_edges: &HashMap<u64, Vec<u64>>, node_id: u64, compose_node_id: u64, skip_lambdas: bool) {
@@ -494,11 +494,8 @@ impl ProtoNetwork {
 		self.inputs = self.inputs.iter().filter_map(|id| new_positions.get(id).copied()).collect();
 		self.output = *new_positions.get(&self.output).unwrap();
 
-		if order.len() != self.nodes.len() {
-			return Err("Assertion failed: order.len() == self.nodes.len()".to_string());
-		}
-
-		return Ok(());
+		assert_eq!(order.len(), self.nodes.len());
+		Ok(())
 	}
 }
 
--- a/node-graph/graph-craft/src/proto.rs
+++ b/node-graph/graph-craft/src/proto.rs
@@ -694,8 +694,7 @@ mod test {
 	#[test]
 	fn topological_sort() {
 		let construction_network = test_network();
-		let sorted = construction_network.topological_sort().unwrap();
-
+		let sorted = construction_network.topological_sort().expect("Error when calling 'topological_sort' on 'construction_network.");
 		println!("{:#?}", sorted);
 		assert_eq!(sorted, vec![14, 10, 11, 1]);
 	}
@@ -711,8 +710,8 @@ mod test {
 	#[test]
 	fn id_reordering() {
 		let mut construction_network = test_network();
-		construction_network.reorder_ids().unwrap();
-		let sorted = construction_network.topological_sort().unwrap();
+		construction_network.reorder_ids().expect("Error when calling 'reorder_ids' on 'construction_network.");
+		let sorted = construction_network.topological_sort().expect("Error when calling 'topological_sort' on 'construction_network.");
 		println!("nodes: {:#?}", construction_network.nodes);
 		assert_eq!(sorted, vec![0, 1, 2, 3]);
 		let ids: Vec<_> = construction_network.nodes.iter().map(|(id, _)| *id).collect();
@@ -725,9 +724,9 @@ mod test {
 	#[test]
 	fn id_reordering_idempotent() {
 		let mut construction_network = test_network();
-		construction_network.reorder_ids().unwrap();
-		construction_network.reorder_ids().unwrap();
-		let sorted = construction_network.topological_sort().unwrap();
+		construction_network.reorder_ids().expect("Error when calling 'reorder_ids' on 'construction_network.");
+		construction_network.reorder_ids().expect("Error when calling 'reorder_ids' on 'construction_network.");
+		let sorted = construction_network.topological_sort().expect("Error when calling 'topological_sort' on 'construction_network.");
 		assert_eq!(sorted, vec![0, 1, 2, 3]);
 		let ids: Vec<_> = construction_network.nodes.iter().map(|(id, _)| *id).collect();
 		println!("{:#?}", ids);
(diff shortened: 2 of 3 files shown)


node-graph/graph-craft/src/graphene_compiler.rs
```diff
@@ -8,7 +8,7 @@ use crate::proto::{LocalFuture, ProtoNetwork};
 pub struct Compiler {}
 
 impl Compiler {
-	pub fn compile(&self, mut network: NodeNetwork) -> impl Iterator<Item = ProtoNetwork> {
```
TrueDoctor, line 11: I think it would make more sense to return `Result<Iterator>` instead

toadkarter: Thanks, will make that change!

node-graph/graph-craft/src/graphene_compiler.rs
```diff
@@ -17,15 +17,23 @@ impl Compiler {
 		network.remove_redundant_id_nodes();
 		network.remove_dead_nodes();
 		let proto_networks = network.into_proto_networks();
```
TrueDoctor, line 19:
```suggestion
		let proto_networks = network.into_proto_networks()?;
```
You should be able to just use the `?` operator and then wrap the proto_networks in an Ok()

node-graph/graph-craft/src/proto.rs
```diff
@@ -687,17 +694,25 @@ mod test {
 	#[test]
 	fn topological_sort() {
 		let construction_network = test_network();
-		let sorted = construction_network.topological_sort();
+		let sorted = construction_network.topological_sort().unwrap();
```
TrueDoctor, line 697: This should be an `.expect`

TrueDoctor: that just makes it a bit easier to debug when tests fail

node-graph/gstd/src/gpu_nodes.rs
```diff
@@ -392,7 +392,7 @@ pub struct BlendGpuImageNode<Background, B, O> {
 }
 
 #[node_macro::node_fn(BlendGpuImageNode)]
-async fn blend_gpu_image(foreground: ImageFrame<Color>, background: ImageFrame<Color>, blend_mode: BlendMode, opacity: f32) -> ImageFrame<Color> {
+async fn blend_gpu_image(foreground: ImageFrame<Color>, background: ImageFrame<Color>, blend_mode: BlendMode, opacity: f32) -> Result<ImageFrame<Color>, String> {
 	let foreground_size = DVec2::new(foreground.image.width as f64, foreground.image.height as f64);
```
TrueDoctor, line 396: At this point we probably don't want to return a result because this is part of the node system which would force the user to deal with all of the error management themselves. In this case I think it would make most sense to throw a console warning and then return an empty image frame

toadkarter: Thanks very much for your comments, I will incorporate these! Quick question though on this one - when you say console message do you mean a `log::debug!` macro call, or something more specific where we send an explicit `ErrorDialog` to the frontend?

TrueDoctor: sending an explict dialouge to the frontend would of course be ideal but that does require extra work. For now you can use a `log::error!`

Keavon: I'd rather we don't annoy the user with a dialog anytime they create a cycle, that connection should just be silently rejected.

TrueDoctor: Yes in that case the dialogue won't be displayed because the cycle never exists. This is intended to cover cases where the user loads an invalid document for example

toadkarter: Makes sense - thanks both!

node-graph/graph-craft/src/graphene_compiler.rs
toadkarter: I  could be wrong but I think_ `proto_networks_result` is the one that returns the Result, so I added the question mark operator to that expression instead of `proto_networks` - the match statement on the bottom then can just be wrapped in Ok() like you said

node-graph/graph-craft/src/proto.rs
toadkarter: Thanks, I've added descriptive error messages with `expect()` in my tests instead of using `unwrap()`

node-graph/gstd/src/gpu_nodes.rs
toadkarter: I've updated this one - it works but I'm not sure if it's not 100% idiomatic Rust, so let me know if you would like me to change it. I tried using a match statement or chaining functions but ultimately we have an expression there that has to return a different type if it's an error, so I just used a simple if statement.

toadkarter:
Ack, I think the CI pipeline might be failing (can't seem to create a Docker container?) I've put my changes in but if the tests could be re-run just to make sure all is ok that would be much appreciated.

Keavon:
That CI failure is for the cargo-deny license checker, it looks like some URL it depends on is currently unavailable and giving a 503. It'll probably resolve itself soon, but even so we can just bypass that CI check if necessary.

node-graph/gstd/src/gpu_nodes.rs
```diff
@@ -64,7 +64,7 @@ impl<T: GpuExecutor> Clone for ComputePass<T> {
 }
 
 #[node_macro::node_impl(MapGpuNode)]
-async fn map_gpu<'a: 'input>(image: ImageFrame<Color>, node: DocumentNode, editor_api: graphene_core::application_io::EditorApi<'a, WasmApplicationIo>) -> ImageFrame<Color> {
+async fn map_gpu<'a: 'input>(image: ImageFrame<Color>, node: DocumentNode, editor_api: graphene_core::application_io::EditorApi<'a, WasmApplicationIo>) -> Result<ImageFrame<Color>, String> {
```
TrueDoctor, line 67: Same argument as for the BlendGpu node should return an empty image frame as well

node-graph/gstd/src/gpu_nodes.rs
```diff
@@ -442,7 +442,22 @@ async fn blend_gpu_image(foreground: ImageFrame<Color>, background: ImageFrame<C
...
+			image: Image {
+				data: Vec::<Color>::new(),
+				width: 0,
+				height: 0,
+			},
+			transform: DAffine2::ZERO,
+		};
+	}
+
+	let proto_networks = proto_networks_result.unwrap().collect();
```
TrueDoctor, line 459: You can use a let else block here instead
```rust
let Ok(proto_networks) = compiler.compile(network.clone() else {return ImageFrame::empty()};
```

node-graph/gstd/src/gpu_nodes.rs
toadkarter: Got it, thanks - should I do something similar with `compile_gpu` and `create_compute_pass_descriptor` or are those ok as they (as far as I understand) are mostly called within the user-facing functions like the ones I have amended?

node-graph/gstd/src/gpu_nodes.rs
toadkarter: Thanks, that's super helpful. I think we still need to call `collect()` on the result so I have done something along the lines of the below, which seems to work:

```
let Ok(proto_networks_result) = compiler.compile(network.clone()) else {return ImageFrame::empty()};
let proto_networks = proto_networks_result.collect();
```

node-graph/gstd/src/gpu_nodes.rs
TrueDoctor: Yes, basically all nodes

toadkarter: Cool - in that case I think my latest commit should be all of the nodes accounted for. 

`create_compute_pass_descriptor` is not a node as far as I can see and `compile_gpu` seems to be a compilation process that probably should be able to fail, much like the `compile` method in `graphene-compiler.rs`. It looks like, from the TODO on it, that `compile_gpu` is intended to be moved to `graph-craft`, I can have a look at doing that maybe on a different PR if this is still intended.

node-graph/graph-craft/src/proto.rs
```diff
@@ -370,7 +370,8 @@ impl ProtoNetwork {
 				self.replace_node_id(&outwards_edges, node_id, compose_node_id, true);
 			}
 		}
-		self.reorder_ids();
+		self.reorder_ids()?;
+		return Ok(());
```
TrueDoctor, line 374: You should be able to just write `Ok(())` here instead of adding an explicit return right? Btw. did you run `cargo clippy` on the code?

node-graph/graph-craft/src/proto.rs
```diff
@@ -491,7 +494,11 @@ impl ProtoNetwork {
 		self.inputs = self.inputs.iter().filter_map(|id| new_positions.get(id).copied()).collect();
 		self.output = *new_positions.get(&self.output).unwrap();
 
-		assert_eq!(order.len(), self.nodes.len());
+		if order.len() != self.nodes.len() {
+			return Err("Assertion failed: order.len() == self.nodes.len()".to_string());
+		}
+
+		return Ok(());
```
TrueDoctor, line 501: I think this should actually still be an assert because a failure of this function indicates a bug in our code and not user error right?

TrueDoctor:
The code is starting to get there, I think after my last few points are addressed, this is good to merge, thanks for the persistence, I hope you learned a thing or two about error handling :D

toadkarter:
> The code is starting to get there, I think after my last few points are addressed, this is good to merge, thanks for the persistence, I hope you learned a thing or two about error handling :D

Awesome, and thanks very much for your review, it's been incredibly helpful :) When I make those tweaks would you like me to squash all the commits or will I keep them separate?

Keavon:
When we hit the merge button on this PR, GitHub squashes it all into a single commit to master named after this PR's title.

toadkarter:
> When we hit the merge button on this PR, GitHub squashes it all into a single commit to master named after this PR's title.

Got it, thanks!

node-graph/graph-craft/src/proto.rs
toadkarter: I think that's right, and we don't need the return on the "Ok()" here either. This has now been fixed.

node-graph/graph-craft/src/proto.rs
toadkarter: Yup, and I've just run `cargo clippy` - nothing else seems to be showing up. This return statement has now been removed!

TrueDoctor (approved):

PR #1432, opened 2023-10-08
Node graph select targets

TrueDoctor:
Stores the input with which a monitor node was called which basically allows us to infer the monitor nodes data's position in viewport space


Keavon:
@TrueDoctor what's the status of this? It got closed when I deleted the `full-document-graph` branch.

PR #1433, opened 2023-10-08
Migrate the Select tool to the document graph

0HyperCube:
TODO:
- [x] Fix with multiple transform nodes
- [x] Fix deep select (new implementation ~10 lines - although it probably doesn't match the previous behaviour as I haven't really tested it)
- [x] Don't select artboards with select tool.

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -1117,11 +1117,11 @@ impl DocumentMessageHandler {
 			space += 1;
 			if layer_node.has_children(&self.metadata()) {
 				path.push(layer_node.to_node());
-				if self.layer_metadata(path).expanded {
-					structure.push(space);
-					self.serialize_structure(folder, structure, data, path);
-					space = 0;
-				}
+				//if self.layer_metadata(path).expanded {
+				structure.push(space);
+				self.serialize_structure(layer_node, structure, data, path);
+				space = 0;
+				//}
 				path.pop();
 			}
 		}
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -375,7 +375,7 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
 				let bounds = match bounds {
 					ExportBounds::AllArtwork => self.all_layer_bounds(&render_data),
 					ExportBounds::Selection => self.selected_visible_layers_bounding_box(&render_data),
-					ExportBounds::Artboard(id) => self.document_legacy.metadata.bounding_box_document(id),
+					ExportBounds::Artboard(id) => self.metadata().bounding_box_document(id),
 				}
 				.unwrap_or_default();
 				let size = bounds[1] - bounds[0];
@@ -636,7 +636,7 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
 			RenderRulers => {
 				let document_transform_scale = self.navigation_handler.snapped_scale();
 
-				let ruler_origin = self.document_legacy.metadata.document_to_viewport.transform_point2(DVec2::ZERO);
+				let ruler_origin = self.metadata().document_to_viewport.transform_point2(DVec2::ZERO);
 				let log = document_transform_scale.log2();
 				let ruler_interval = if log < 0. { 100. * 2_f64.powf(-log.ceil()) } else { 100. / 2_f64.powf(log.ceil()) };
 				let ruler_spacing = ruler_interval * document_transform_scale;
@@ -964,10 +964,13 @@ impl DocumentMessageHandler {
 	pub fn network(&self) -> &NodeNetwork {
 		&self.document_legacy.document_network
 	}
+	pub fn metadata(&self) -> &document_legacy::document_metadata::DocumentMetadata {
+		&self.document_legacy.metadata
+	}
 
 	/// Remove the artwork and artboard pan/tilt/zoom to render it without the user's viewport navigation, and save it to be restored at the end
 	pub(crate) fn remove_document_transform(&mut self) -> DAffine2 {
-		let old_artwork_transform = self.document_legacy.metadata.document_to_viewport;
+		let old_artwork_transform = self.metadata().document_to_viewport;
 		self.document_legacy.metadata.document_to_viewport = DAffine2::IDENTITY;
 		DocumentLegacy::mark_children_as_dirty(&mut self.document_legacy.root);
 
@@ -1130,11 +1133,11 @@ impl DocumentMessageHandler {
 
 	fn serialize_structure(&self, folder: LayerNodeIdentifier, structure: &mut Vec<u64>, data: &mut Vec<LayerId>, path: &mut Vec<LayerId>) {
 		let mut space = 0;
-		for layer_node in folder.children(&self.document_legacy.metadata) {
+		for layer_node in folder.children(&self.metadata()) {
 			data.push(layer_node.to_node());
 			info!("Pushed child");
 			space += 1;
-			if layer_node.has_children(&self.document_legacy.metadata) {
+			if layer_node.has_children(&self.metadata()) {
 				path.push(layer_node.to_node());
 				if self.layer_metadata(path).expanded {
 					structure.push(space);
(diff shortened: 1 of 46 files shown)


0HyperCube:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

0HyperCube:
!build

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -115,137 +109,93 @@ impl LayoutHolder for SelectTool {
...
+	}
+
+	fn flip_widgets(&self) -> impl Iterator<Item = WidgetHolder> {
+		[(FlipAxis::X, "Horizontal"), (FlipAxis::Y, "Horizontal")].into_iter().map(|(flip_axis, name)| {
+			IconButton::new("Flip".to_string() + name, 24)
+				.tooltip("Flip ".to_string() + name)
+				.on_update(move |_| DocumentMessage::FlipSelectedLayers { flip_axis }.into())
+				.widget_holder()
+		})
+	}
```
Keavon, line 151: This is using the FlipHorizontal icon for both horizontal and vertical now.

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -115,137 +109,93 @@ impl LayoutHolder for SelectTool {
...
+		widgets.push(self.deep_selection_widget());
+
+		// Pivot
+		if self.tool_data.selected_layers_count > 0 {
+			widgets.push(Separator::new(SeparatorType::Related).widget_holder());
+			widgets.push(self.pivot_widget());
+		}
+
+		// Align
+		if self.tool_data.selected_layers_count >= 2 {
```
Keavon, line 175: We'd like to preserve these being grayed out instead of hiding them so the user knows it's possible.

Keavon:
How easy would it be to make the Select tool not select the artboard? I also notice the "Empty Stack" node is appearing as a layer in the (still-to-be-rewritten) layer panel, but if it's a real quick change, it'd be nice to temporarily remedy that until I rewrite the layer panel in the next few weeks.

editor/src/messages/tool/tool_messages/select_tool.rs
0HyperCube: `(FlipAxis::Y, "Horizontal")` - typo

0HyperCube:
@Keavon 
> How easy would it be to make the Select tool not select the artboard? As soon as I click the artboard, it breaks the rendering of the artwork (but the shape-hover overlays show the current positions after navigating).

If you drag the artboard, that inserts a transform node. However this causes the graph to no longer compile as we can't implement transform node for the artboard.

> I also notice the "Empty Stack" node is appearing as a layer in the (still-to-be-rewritten) layer panel, but if it's a real quick change, it'd be nice to temporarily remedy that until I rewrite the layer panel in the next few weeks.

It doesn't - are you sure that it isn't the artboard layer (which is completely white by default)?

Keavon:
> If you drag the artboard, that inserts a transform node. However this causes the graph to no longer compile as we can't implement transform node for the artboard.

I see, yes that makes sense. Do you agree that preventing the Select tool from interacting with artboards would be the right solution here? (And then only the Artboard tool can do that.)

> It doesn't - are you sure that it isn't the artboard layer (which is completely white by default)?

Oh yes, you're right. that'd the artboard layer. It shows up there in a new document before you draw any layers. I did just test that by changing the artboard color and I see the thumbnail updates accordingly.

0HyperCube:
> I see, yes that makes sense. Do you agree that preventing the Select tool from interacting with artboards would be the right solution here? (And then only the Artboard tool can do that.)

Fixed @Keavon

Keavon:
!build

Keavon:
This looks great! Only a few little things I'm noticing:

- I can't reproduce it reliably, but thrice I've noticed that as I multi-select (shift-clicking one shape then another in the viewport), it ends up making both shapes disappear permanently from the render until I Ctrl+Z it. We'll have to keep watching this one since I have no idea how to make it happen and it's sort of rare.
- The pivot widget should also be grayed out, not hidden, just like the alignment and flip widgets.

I think that's all that I directly notice pertaining to the Select tool, so this should be ready to merge pretty soon then! In fact, I'll just push the pivot widget change now so it's in good shape for a review also by @TrueDoctor and then we'll be good to merge.

After this, I think the most obvious problems users will encounter in this branch is that artboards should be fed into from a layer stack, rather than living at the base of a layer stack; and that selecting a layer in the node graph doesn't show the upstream nodes in the Properties panel. But those are both fine to fix in a separate PR and are probably somewhat approachable by new contributors. Thank you so much for this!

Keavon:
!build

Keavon:
!build

TrueDoctor:
I'm not quite happy with the overall approach of relying on node names for special behaviour. We could potentially add a dedicated node tag (or just in general some sort of tagging system) to make the system a bit more robust.
Another issue with the current approach is that we still don't support Flip nodes or other user-custom transform nodes right? But I can address that in a followup pr.

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -139,20 +135,8 @@ impl MessageHandler<DocumentMessage, (u64, &InputPreprocessorMessageHandler, &Pe
...
-									responses.add(LayerChanged { affected_layer_path: path.clone() });
-									self.layer_range_selection_reference = path.clone();
-
-									if *is_selected {
-										responses.add(AddSelectedLayers {
-											additional_layers: vec![path.clone()],
-										});
-									}
+								DocumentResponse::CreatedLayer { .. } => {
+									unimplemented!()
```
TrueDoctor, line 139: Could you add a comment explaining why this will never occur / if should be implemented at some point change it to a `todo!` comment?

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1130,17 +1111,17 @@ impl DocumentMessageHandler {
...
 			space += 1;
-			if layer_node.has_children(&self.document_legacy.metadata) {
+			if layer_node.has_children(&self.metadata()) {
 				path.push(layer_node.to_node());
-				if self.layer_metadata(path).expanded {
-					structure.push(space);
-					self.serialize_structure(folder, structure, data, path);
-					space = 0;
-				}
+				//if self.layer_metadata(path).expanded {
```
TrueDoctor, line 1120: Can this be removed?

editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
```diff
@@ -503,29 +519,21 @@ impl MessageHandler<GraphOperationMessage, (&mut Document, &mut NodeGraphMessage
...
 			GraphOperationMessage::TransformSetPivot { layer, pivot } => {
 				let bounds = LayerBounds::new(document, &layer);
 				if let Some(mut modify_inputs) = ModifyInputsContext::new_layer(&layer, document, node_graph, responses) {
 					modify_inputs.pivot_set(pivot, bounds);
 				}
 
-				let pivot = pivot.into();
-				responses.add(Operation::SetPivot { layer_path: layer, pivot });
+				//let pivot = pivot.into();
+				//responses.add(Operation::SetPivot { layer_path: layer, pivot });
```
TrueDoctor, line 536:
```suggestion
```

TrueDoctor (commented):
21 files left to review, will continue asap

0HyperCube:
> I'm not quite happy with the overall approach of relying on node names for special behaviour. We could potentially add a dedicated node tag (or just in general some sort of tagging system) to make the system a bit more robust.

I have previously mentioned that the approach of relying on node names and input indexes is not robust.

>  Another issue with the current approach is that we still don't support Flip nodes or other user-custom transform nodes right? But I can address that in a followup pr.

> I can't reproduce it reliably, but thrice I've noticed that as I multi-select (shift-clicking one shape then another in the viewport), it ends up making both shapes disappear permanently from the render until I Ctrl+Z it. We'll have to keep watching this one since I have no idea how to make it happen and it's sort of rare.

Is there any error logs? Does it still happen when you save and then reload? Do the layers still appear in the layer tree? If you drag a connection between two nodes in the graph UI then does anything change (this causes the layer tree to be recomputed)?

editor/src/messages/portfolio/document/document_message_handler.rs
0HyperCube: We need to add expanded state to the new layer system.

Keavon:
!build

Keavon (approved):
QA looks great!

PR #1436, opened 2023-10-19
A few minor lints and docs

nyurik:
* Added required packages to compile on Debian-style linux
* Inlined some format args, and removed some `&` in args (they cause about 6% slowdown that compiler cannot inline)
* a few spelling mistakes

### DIFF
--- a/node-graph/gcore/src/vector/style.rs
+++ b/node-graph/gcore/src/vector/style.rs
@@ -14,7 +14,7 @@ const OPACITY_PRECISION: usize = 3;
 
 fn format_opacity(name: &str, opacity: f32) -> String {
 	if (opacity - 1.).abs() > 10_f32.powi(-(OPACITY_PRECISION as i32)) {
-		format!(r#" {}-opacity="{:.precision$}""#, name, opacity, precision = OPACITY_PRECISION)
+		format!(r#" {name}-opacity="{opacity:.OPACITY_PRECISION$}""#)
 	} else {
 		String::new()
 	}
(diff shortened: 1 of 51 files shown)


node-graph/gcore/src/vector/style.rs
```diff
@@ -14,7 +14,7 @@ const OPACITY_PRECISION: usize = 3;
 
 fn format_opacity(name: &str, opacity: f32) -> String {
 	if (opacity - 1.).abs() > 10_f32.powi(-(OPACITY_PRECISION as i32)) {
-		format!(r#" {}-opacity="{:.precision$}""#, name, opacity, precision = OPACITY_PRECISION)
+		format!(r#" {name}-opacity="{opacity:.OPACITY_PRECISION$}""#)
```
Keavon, line 17: Just commenting on this one so I can double-check it's correct.

nyurik: This one got auto-fixed with clippy, but yes, i believe you can inline the width too (I wrote the clippy lint for it, so I would take it with a grain of salt :D )

```
cargo clippy --workspace --allow-dirty --fix --benches --tests --bins -- -A clippy::all -W clippy::uninlined_format_args
```

Keavon: I'm just unfamiliar with that Rust string formatter syntax. But it sounds like you're the expert on that if you wrote the lint for it!

Keavon (approved):
Awesome, thanks for the cleanup!

Keavon:
!build

node-graph/gcore/src/vector/style.rs
nyurik: famous last words :) I did look into it at some point, learning it in depth, but not the most fun topic to learn tbh :)

Keavon:
Thanks for the contribution! Please let me know if you'd like any suggestions on a next task if you're interested in contributing further. Cheers!

PR #1437, opened 2023-10-21
Add more details to Graphene concept documentation

0HyperCube:


node-graph/README.md
```diff
@@ -6,20 +6,31 @@ Graphite is an image editor which is centred around a node based editing workflo
...
 	pub name: String,
-	// A NodeInput::Node { node_id, output_index } specifies an input from another node.
-	// A NodeInput::Value { tagged_value, exposed } specifies a constant value. An exposed value is visible as a dot in the node graph UI.
-	// A NodeInput::Network(Type) specifies a node will get its input from outside the graph, which is resolved later.
+	/// The inputs to a node, which are either:
+	/// - From other nodes within this graph [`NodeInput::Node`],
+	/// - A constant value [`NodeInput::Value`],
+	/// - A [`NodeInput::Network`] which specifies that this input is from outside the graph, which is resolved in the graph flattening step.
 	pub inputs: Vec<NodeInput>,
+	/// TODO: what is this?
```
TrueDoctor, line 20: Usually compose nodes are automatically inserted to link two nodes together. When the node specifies a manual composition input it opts out of this behaviour and calls the previous node itself

TrueDoctor: (This is a replacement for the short circuit system)

node-graph/README.md
```diff
@@ -54,34 +72,49 @@ pub fn multiply_opacity(document_node: &DocumentNode, node_id: NodeId, _context:
...
 }
 ```
 
-The `eval` function can only take one input. To support more than one input, the node struct can contain references to other nodes (it is the references that implement the `Node` trait). If the input is a constant, then it will reference a node that simply evaluates to a constant. If the input is a node, then the relevant proto-node will be referenced. To evaluate the opacity multiplier input, you can pass in `()` (because no input is required to calculate the opacity multiplier) which returns an `f64`. This is because of the generics we have applied: `N: Node<'i, (), Output = f64>` A helper function to create a new node struct is also defined here.
+The `graphene_core::value::CopiedNode` is a node that, when evaluated, copies `10f32` and returns it.
 
-This process can be made more concise using the `node_fn` macro, which can be applied to a function like `image_opacity` with an attribute of the name of the node:
+## Creating a new protonode
+
+Instead of manually implementing the `Node` trait with complex generics, one can use the `node_fn` macro, which can be applied to a function like `image_opacity` with an attribute of the name of the node:
```
TrueDoctor, line 117: should we mention the other node_fn like macros as well here?

node-graph/graph-craft/src/document.rs
```diff
@@ -255,9 +272,13 @@ impl NodeInput {
 
 #[derive(Clone, Debug, PartialEq, Hash, DynAny)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+/// Represents the implementation of a node, which can be a nested [`NodeNetwork`], a proto [`NodeIdentifier`] or extract
 pub enum DocumentNodeImplementation {
+	/// A nested [`NodeNetwork`] that is flattened by the [`NodeNetwork::flatten`] function.
 	Network(NodeNetwork),
+	/// A protonode identifier which can be found in `node_registry.rs`.
 	Unresolved(NodeIdentifier),
+	/// TODO: what?
```
TrueDoctor, line 281: This is used for macro functionality and allows you to get the input node as a `DocumentNode` supplied to your node

node-graph/README.md
0HyperCube: `node_new` and `node_impl`?

node-graph/interpreted-executor/src/dynamic_executor.rs
```diff
@@ -123,11 +132,13 @@ impl BorrowTree {
 		self.nodes.get(&id).cloned()
 	}
 
+	/// Evaluate the output node of the [`BorrowTree`]
 	pub async fn eval<'i, I: StaticType + 'i, O: StaticType + 'i>(&'i self, id: NodeId, input: I) -> Option<O> {
 		let node = self.nodes.get(&id).cloned()?;
 		let output = node.eval(Box::new(input));
 		dyn_any::downcast::<O>(output.await).ok().map(|o| *o)
 	}
+	/// Evaluate the output node of the [`BorrowTree`] and cast it to a tagged value
```
TrueDoctor, line 141: This is required to ensure no borrowed data can escape the Node Graph

node-graph/README.md
0HyperCube: That explains why the `manual_composition` in the document network leads to a `ProtoNodeInput::ShortCircut` which is what was confusing to me. Perhaps the document graph concept and proto graph concept should be named the same thing @TrueDoctor?

TrueDoctor (commented):
Looks pretty good to me, thanks for the effort

node-graph/README.md
TrueDoctor: Yeah, `node_new` can be used if you need to specify the implementation manually and `node_impl` can be used if you want to add multiple impl blocks which would otherwise create conflicting new functions

node-graph/README.md
TrueDoctor: Yeah, that seems reasonable I'm not sure what my reason was for not renaming that as well in my pr

0HyperCube: ProtoNodeInput::Network is the same as ProtoNodeInput::ManualComposition (formerly ProtoNodeInput::ShortCircuit) so I changed them both to ProtoNodeInput::ManualComposition.

TrueDoctor: Network = ManualComposition sounds strange, are you sure about that?

TrueDoctor: If we remove the Network type we would also need to remove the proto network inputs but I'm not sure that that would be a good idea. Having the inputs allows us to do network specialization without having to recompile the entire network similar to the concept of parameterized shaders

0HyperCube: There was no difference in the implementation. They both just specify that the protonode's input will be given to it as an argument to `.eval(x)`.

0HyperCube: I'm not really sure I understand how merging two identical enum variants relates to making a shader parameterised (presumably through uniforms in the graphics api which are used in the shader glsl code).

TrueDoctor: Yes, the idea would be to have the network input as a placeholder which is replaced before the graph execution and actually make it impossible to execute a network with graph inputs present

TrueDoctor: But that is also something that we could add back in the future if that is needed and just remove it for now to make the code less confusing

node-graph/graph-craft/src/document.rs
```diff
@@ -177,6 +203,11 @@ pub enum NodeInput {
 	/// Input that is provided by the parent network to this document node, instead of from a hardcoded value or another node within the same network.
 	Network(Type),
 
+	/// A Rust source code string. Allows us to insert literal Rust code. Only used for GPU compilation.
+	/// We can use this whenever we spin up Rustc. Sort of like inline assembly, but because our language is Rust, it acts as inline Rust.
+	Inline(InlineRust),
+
+	// TODO: Update, remove, or transplant this comment to a new place (ShortCircuit was renamed to ManualComposition, and may also have some relationship to Network?)
```
Keavon, line 210: @TrueDoctor @0HyperCube I'm wondering about this comment and what should become of it. I'm also not sure what the relationship is between the former ShortCitcuit and with Network. It seems that, from this enum at least, ShortCircuit wasn't just renamed to ManualComposition but was instead entirely removed?

Maybe we should move the explanation in this comment to another place where ManualComposition does occur? Or otherwise just delete it?

0HyperCube: `manual_composition` is now specified on the document node. This reduces invalid state, as it makes it impossible for a document node to have 2 ShortCircuit inputs (aka manual_composition inputs).

The concept explained here is similar to the one on the documentation for `pub manual_composition: Option<Type>`. As such you can merge them however you choose @Keavon.

node-graph/graph-craft/src/document.rs
```diff
@@ -40,18 +43,80 @@ fn return_true() -> bool {
...
+	///                      │   CacheNode   │    │       F       │
+	///                      │               ├───►│               │───► RESULT (END)
+	/// ┌───────────────┐    ├───────────────┤    └───────────────┘
+	/// │               │◄───┤               │
+	/// │       G       │    │  Cached Data  │
+	/// │               ├───►│               │
+	/// └───────────────┘    └───────────────┘
+	/// ```
+	///
+	/// Now, the call from `F` directly reaches the `CacheNode` and the `CacheNode` can decide whether to call `G.eval(input_from_f)` in the event of a cache miss or just return the cached data in the event of a cache hit.
```
Keavon, line 102: @TrueDoctor @0HyperCube I've further extended the explanation by merging in the flowcharts from `enum NodeInput::ShortCircuit` and adding more context. But I want to make sure I haven't written any inaccuracies, so I'd appreciate if you can both proofread this for correctness. Thanks!

node-graph/graph-craft/src/document.rs
```diff
@@ -40,18 +43,80 @@ fn return_true() -> bool {
...
 	pub inputs: Vec<NodeInput>,
+	/// Manual composition is a way to override the default composition flow of one node into another.
+	///
+	/// Through the usual node composition flow, the upstream node providing the primary input for a node is evaluated before the node itself is run.
+	/// - Abstract example: upstream node `G` is evaluated and its data feeds into the primary input of downstream node `F`, just like function composition where function `F` is evaluated and its result is fed into function `F`.
+	/// - Concrete example: a node that takes an image as primary input will get that image data from an upstream node that produces image output data and is evaluated first before being fed downstream.
+	///
+	/// This is achieved by automatically inserting `ComposeNode`s, which run the first node with the overall input and feed the output into the second node.
+	/// The `ComposeNode` is basically a function composition operator: the parentheses in `F(G)` or circle operator in `G ∘ F`.
```
TrueDoctor, line 60:
```suggestion
	/// This is achieved by automatically inserting `ComposeNode`s, which run the first node with the overall input and feed the output into the second node.
	/// The `ComposeNode` is basically a function composition operator: the parentheses in `F(G(x))` or circle operator in `(F ∘ G)(x)`.
```

node-graph/graph-craft/src/document.rs
```diff
@@ -40,18 +43,80 @@ fn return_true() -> bool {
...
+	/// ```text
+	///              [ComposeNode]         [ComposeNode]
+	/// ┌───────────────┐  |  ┌───────────────┐  |  ┌───────────────┐
+	/// │               │◄─|──┤               │◄─|──┤               │◄─── EVAL (START)
+	/// │       G       │  |  │PassThroughNode│  |  │       F       │
+	/// │               ├──|─►│               ├──|─►│               │───► RESULT (END)
+	/// └───────────────┘  |  └───────────────┘  |  └───────────────┘
+	/// ```
+	///
+	/// Without manual composition, `ComposeNode`s would be automatically inserted (during the graph rewriting step) between nodes where the arrows are drawn in the diagram above.
```
TrueDoctor, line 81: It is technically not inserted between the other nodes but I'm not sure if that is relevant

Keavon: Could you please reword it to make the statement not make an inaccurate claim? I'm not sure what happens if that isn't technically the case.

node-graph/graph-craft/src/document.rs
TrueDoctor: Besides the two comments this looks pretty good to me

node-graph/graph-craft/src/document.rs
TrueDoctor: The compose node is not inserted in between the two other nodes, it is in of it self a manual composition nodes which calls the two other nodes in sucsession.
image
(my touchscreen decided to not want to work hence the weird drawing)

PR #1439, opened 2023-10-22
Fix the Path tool's smooth/sharp buttons

0HyperCube:
The smooth button and the position in the path tool bar at the top were broken following the document graph migration.


Keavon:
@0HyperCube if you could also please review my changes here, I'd appreciate it: https://github.com/GraphiteEditor/Graphite/pull/1439/commits/1c3d419619f4d30976270a457067cb426c4305d4

Then you can go ahead and merge after if you think all is in order.

Keavon (approved):

PR #1441, opened 2023-10-22
Fix deleting layers

0HyperCube:
Fix deleting layers by selecting and pressing the delete key when using the select tool.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -121,7 +121,6 @@ pub struct NodeGraphMessageHandler {
 	pub layer_path: Option<Vec<LayerId>>,
 	pub network: Vec<NodeId>,
 	has_selection: bool,
-	graph_open: bool,
 	#[serde(skip)]
 	pub widgets: [LayoutGroup; 2],
 }
@@ -348,15 +347,15 @@ impl NodeGraphMessageHandler {
 			return false;
 		}
 
-		let mut first_input_node: Option<NodeInput> = None;
+		let mut reconnect_to_input: Option<NodeInput> = None;
 
 		if reconnect {
 			// Check whether the being-deleted node's first (primary) input is a node
 			if let Some(node) = network.nodes.get(&deleting_node_id) {
 				// Reconnect to the node below when deleting a layer node.
-				let index = if node.name == "Layer" { 7 } else { 0 };
-				if matches!(&node.inputs.get(index), Some(NodeInput::Node { .. })) {
-					first_input_node = Some(node.inputs[index].clone());
+				let reconnect_from_input_index = if node.name == "Layer" { 7 } else { 0 };
+				if matches!(&node.inputs.get(reconnect_from_input_index), Some(NodeInput::Node { .. })) {
+					reconnect_to_input = Some(node.inputs[reconnect_from_input_index].clone());
 				}
 			}
 		}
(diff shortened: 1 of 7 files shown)


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -360,8 +361,10 @@ impl NodeGraphMessageHandler {
 		if reconnect {
 			// Check whether the being-deleted node's first (primary) input is a node
 			if let Some(node) = network.nodes.get(&deleting_node_id) {
-				if matches!(node.inputs.first(), Some(NodeInput::Node { .. })) {
-					first_input_node = Some(node.inputs[0].clone());
+				// Reconnect to the node below when deleting a layer node.
+				let index = if node.name == "Layer" { 7 } else { 0 };
```
Keavon, line 365: It's unclear to me what these magic numbers mean.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -787,7 +790,11 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
 			NodeGraphMessage::SetNodeInput { node_id, input_index, input } => {
 				if let Some(network) = document.document_network.nested_network_mut(&self.network) {
 					if let Some(node) = network.nodes.get_mut(&node_id) {
-						node.inputs[input_index] = input
+						let structure_changed = node.inputs[input_index].as_node().is_some() || input.as_node().is_some();
```
Keavon, line 793: Is it theoretically possible this indexing could go out of bounds and panic? Do you think `.get()` would be a worthwhile defensive programming measure?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -121,6 +121,7 @@ pub struct NodeGraphMessageHandler {
 	pub layer_path: Option<Vec<LayerId>>,
 	pub network: Vec<NodeId>,
 	has_selection: bool,
+	graph_open: bool,
```
Keavon, line 124: Don't we already have a boolean in the Rust code storing if the graph overlay is open?

Keavon (approved):

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: We cannot access the portfolio message manager from within the graph message handler whilst creating the actions. This causes the node graph delete shortcut to override other delete actions from the path tool.

0HyperCube: I am unsure why the graph open boolean got moved from the node graph message handler in the first place as we now have duplicate state.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: It is the input index, allowing "the node below when deleting a layer node" (from the comment above).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
Keavon: I'm pretty sure it was always in the portfolio message handler since back when I originally wrote it.

From a user perspective, it's basically a question a whether it's more useful/intuitive for the graph overlay state to follow the active document tab or be global to the application. It's the same question as the active tool and the working colors, both of which are global. For consistency with those, that's why I also picked to have the overlay be global; however I don't necessarily feel strongly about keeping it that way. If it's a significant code simplification to have it follow the document, I'm okay with that change. But we should only keep the state in a single boolean in the code. Would you mind refactoring it to remove the redundancy?

Keavon (approved):
💯

PR #1444, opened 2023-10-24
Switch build system to Vite

### LINKED ISSUE #1385: Switch build system from Parcel to Vite
We switched from Webpack to Parcel, but it's been causing lots of problems. We need to switch to Vite instead.

mdonoughe:
This seems like it works, but it generates a lot of warnings about unscoped global styles. I didn't check if there's a way to disable those.

I had to clean up some imports because I was getting errors from importing things that don't exist.

Closes #1385


Keavon:
Ooh exciting, thank you for helping with this! I'll aim to take a closer look later tonight or tomorrow morning.

0HyperCube:
!build

0HyperCube (commented):
Looks good to me. Friendly reminder to @Keavon to have a look. Thoughts:
- https://9a746686.graphite.pages.dev/third-party-licenses.txt just redirects to the main page.
- There is a random spinning arc that appears when loading the page during development.

Keavon:
!build

Keavon:
!build

Keavon (approved):
I fixed up a few tiny regressions, including the license file not making it into the output as @0HyperCube pointed out. Things all seem to be working very nicely now, as far as I can tell!

The only problem I'm bothered by is that, in dev mode, the DOM is polluted by Svelte's `s-CSSHASH` class names on all the elements. Perhaps the previous Parcel build never operated in dev mode so those were omitted? We can live with it for the moment but aim to find a fix going forward.

This is really valuable, and paves the way towards reenabling other web tooling like linting and autoformatting and removing the awkwardness of running multiple terminal commands in parallel for compiling the Wasm into the build. Future work, of course... this is all good to merge now! Thank you for your help!

Issue #1394, opened 2023-08-22
Tracking Issue: Document graph migration MVP


TrueDoctor:
- [x] Pen tool doesn't place more than one point
- [x] Occasionally drawing with rectangle tool doesn't work/is offset
- [x] Thumbnails go over the entire node graph and break the click targets 
- [x] Thumbnails should display the left input, not the result of the composite
- [x] Placing on top of other artwork
- [x] Bounding boxes and hit testing
- [x] Rulers
- [x] Dragging in/pasting/importing images (#1418)
- [x] Each drawing tool selecting the layer it just created
- [x] Automatically insert the empty stack nodes (so the user doesn't accidentally disconnect and get an error). (#1445)
- [x] Preview vector data (#1446)
- [x] Grouping (with <kbd>Ctrl</kbd><kbd>G</kbd>) and ungrouping and the necessary graph rearrangement (#1176)
- [x] Migrate demo artwork (#1459)
- [x] Migrate tests to node graph so CI passes again (#1456 #1459)
- [x] Drag an image into the graph and fix popup when dragging into document. (#1459)
- [x] Naive placement of layers in artboards (#1459)
- [x] Opacity and blend mode settings on layers (#1459)
- [x] Fix performance regressions that make it slow to interact with the editor (#1477)
- [x] Fix collapsing/expanding layers (currently crashes) and hiding with the eye icon (currently does nothing) (#1481)
- [x] Fix bug where the artboard isn't centered when creating a new document (instead it starts in the top left) (#1473)
- [x] Fix wireframe view mode (#1473)
- [x] Fix bug where layer click targets are offset by the Transform node from rendered location (#1473)
- [x] G/R/S shortcuts (#1473)
- [x] Fix bug where node graph doesn't change after switching document tabs
    - [x] Fix bug where connection links get tangled upon switching document tabs
- [x] Fix the tremendous volume of JS console warning/error spam (#1484)
    - [x] Clean up all the FrontendMessages being sent to the JS but lacking a subscribed handler (#1484)
- [x] Fix bug with how a Transform node feeding into an Artboard node works [(video)](https://files.keavon.com/-/OutstandingColossalEskimodog/capture.mp4) (#1484)
    - [x] Fix bug where a Transform node feeding into an Artboard node causes the artboard's selection outline to be misaligned from the artboard itself (#1484)
- [x] #1485
- [x] <kbd>Ctrl</kbd><kbd>A</kbd> should not select the artboard
- [x] #1488
[...]

PR #1446, opened 2023-10-29
Improve previewing node data

### LINKED ISSUE #1442: Unable to rotate document
Following the work on footprints by @TrueDoctor, rotating the document with <kbd>alt</kbd> + <kbd>left mouse drag</kbd> no longer works as expected.

0HyperCube:
Vector data, vector data without the cull or transform node, numbers, bools, and optional colours can now be previewed.

Part of #1394 also closes #1442, which was also due to the render node.


node-graph/gstd/src/wasm_application_io.rs
```diff
@@ -288,77 +289,129 @@ fn decode_image_node<'a: 'input>(data: Arc<[u8]>) -> ImageFrame<Color> {
...
-				let context = canvas.get_context("2d").unwrap().unwrap().dyn_into::<CanvasRenderingContext2d>().unwrap();
-				let image_data = web_sys::ImageData::new_with_u8_clamped_array_and_sh(array, pixmap_size.width(), pixmap_size.height()).expect("Failed to construct ImageData");
-				context.put_image_data(&image_data, 0.0, 0.0).unwrap();
+// Render with the data node taking in Footprint.
+impl<'input, 'a: 'input, T: 'input + GraphicElementRendered, F: 'input + Future<Output = T>, Data: 'input, Surface: 'input, SurfaceFuture: 'input> Node<'input, WasmEditorApi<'a>>
+	for RenderNode<Data, Surface, Footprint>
+where
+	Data: Node<'input, Footprint, Output = F>,
+	Surface: Node<'input, (), Output = SurfaceFuture>,
+	SurfaceFuture: core::future::Future<Output = Arc<SurfaceHandle<HtmlCanvasElement>>>,
```
TrueDoctor, line 363: What is the reason for no longer using the macro here?

0HyperCube: It doesn't work with the third generic argument (used to allow the upstream node to take either a `Footprint` or `()`)

TrueDoctor: I guess I am still confused why we need this, when do we not have a Footprint?

0HyperCube: If the user creates a rectangle node, then a fill node, then connects that to the output.

0HyperCube: rectangle -> fill -> output graph

TrueDoctor: Ah, so the issue is the missing `CullNode` ?

0HyperCube: Also if the user wishes to preview a colour node:
Previewing a colour node

TrueDoctor: In both cases I would prefer if we still used the IntoNode to do the conversion for us instead of making the render node overly generic to improve code reuse

0HyperCube: Could you elaborate on what you mean by using the into node? What is converted into what? Who inserts the into node?

TrueDoctor: image
basically this thing, this does the conversion and the render node only gets a proper output

0HyperCube: This `IntoNode` still requires a node that has a `Footprint` for input. We want to render nodes that have `()` for input.

We also do not want to convert to a graphic group. `bool`s can't be converted into a graphic group, yet they should still be previewed.

TrueDoctor: what should the preview for a bool look like? Also does it make sense to connect it to the render node at all in that case?

TrueDoctor: If we don't connect the data to the render node, the graphite editor can decide how to display the data

0HyperCube: boolean preview
It makes sense to connect to the boolean node to the render node because it is rendered into the document.

0HyperCube: > If we don't connect the data to the render node, the graphite editor can decide how to display the data

I'm not sure I understand the advantage of having the RenderNode render only some types of data, and the editor render other types of data. Would it not be easier to render everything in one place?

TrueDoctor: Ideally I'd like the render node to be as lightweight as possible and move as much of the functionality out of it as possible to make it reusable and composable. I don't really think about displaying bools as "rendering" per se, that feels more like a debug output. If the user were to for example connect a "spreadsheet" to the output I think it would make more sense to show a spreadsheet UI from which the user can copy paste data, search for values etc. rather than rendering that to an SVG.

Keavon: How about wrapping the render node, which produces only graphical content, with another node which also includes "debug"-style non-graphical data type visualizations? That might help to keep it lightweight as you'd desire while still preserving this PR's cool behavior as it's currently implemented.

From a product design perspective, it remains an open question what the desired final behavior should look like for users viewing non-graphical data. But this current approach is quite a decent stand-in that provides useful value to the user in absence of any other fancy UIs like the spreadsheet panel you mentioned. Having something at all serves as a good angle to imagine what further improvements could look like. So I'd definitely like the current behavior to be merged, although I'll leave the code structuring up to the two of you as you're both more knowledgeable than me about the code.

TrueDoctor: But to reiterate my point about the non-use of the macro, shouldn't it be possible to describe one of the use cases using the `node_fn` macro and then add the second implementation using the `node_impl` macro?

TrueDoctor: I guess the way I would go about this is to just reuse the monitor nodes for this and by clicking the monitor node we could allow the user to preview the data that flows through it.  This would allow us to not have to channel everything through the render node which would be an architecturally much cleaner solution. (with the current solution this would mess up caching etc.)

TrueDoctor: > This leads to an error as the PhantomData is not properly handled.

Yeah, I've since tried that out as well although I have to say that I'm confused as to why the implementations are in conflict with each other without using the parameter phantom data, could be that the rust type system is not powerful enough or that I misunderstand something about the setup

0HyperCube: > I guess the way I would go about this is to just reuse the monitor nodes for this

This wouldn't work for previewing a node that the user just placed into the graph without connecting to the output (e.g. both examples above) as the monitor node would never be run.

0HyperCube: I have made it so the executor renders things in the case of a node being previewed. I have kept the current behaviour in the render node so it works if the user decides to connect their bool to the output node (which they might do if they didn't see the preview button).

0HyperCube: Perhaps? So when doing typing for the protonodes, if it found a node with no matching implementation, it would check to see if the given input is a Footprint and if there is an implementation with input of `()`?

Keavon (approved):
Awesome 💯

PR #1450, opened 2023-11-06
Fix frontend webmanifest so installing a PWA isn't broken

tomayac:
This makes the app installable as a PWA:

Screenshot 2023-11-06 at 09 21 56

### DIFF
--- a/frontend/public/site.webmanifest
+++ b/frontend/public/site.webmanifest
@@ -16,6 +16,15 @@
         }
     ],
     "theme_color": "#ffffff",
-    "background_color": "#ffffff",
-    "display": "standalone"
+    "background_color": "#222222",
+    "display": "standalone",
+    "screenshots": [
+        {
+            "src": "https://static.graphite.rs/content/index/gui-demo-valley-of-spires__2.avif",
+            "sizes": "1920x1080",
+            "type": "image/avif",
+            "form_factor": "wide",
+            "label": "Graphite Editor on desktop."
+        }
+    ]
 }
--- a/frontend/public/site.webmanifest
+++ b/frontend/public/site.webmanifest
@@ -1,6 +1,8 @@
 {
     "name": "Graphite",
     "short_name": "Graphite",
+    "start_url": "/",
+    "id": "/",
     "icons": [
         {
             "src": "./android-chrome-192x192.png",
--- a/frontend/public/site.webmanifest
+++ b/frontend/public/site.webmanifest
@@ -17,14 +17,5 @@
     ],
     "theme_color": "#ffffff",
     "background_color": "#222222",
-    "display": "standalone",
-    "screenshots": [
-        {
-            "src": "https://static.graphite.rs/content/index/gui-demo-valley-of-spires__2.avif",
-            "sizes": "1920x1080",
-            "type": "image/avif",
-            "form_factor": "wide",
-            "label": "Graphite Editor on desktop."
-        }
-    ]
+    "display": "standalone"
 }


Keavon:
!build

Keavon:
Could you also please change `background_color` to `#222222`? Thanks for this!

tomayac:
> Could you also please change `background_color` to `#222222`? Thanks for this!

Done. Also added a screenshot for richer install UI.

Keavon:
!build

Keavon:
Thanks, but the screenshot will get out of date as we keep building the app and the effort of maintaining it outweighs the tiny benefit it provides to users. So I'd prefer you remove that, but I appreciate the idea.

tomayac:
This is the install UI with screenshot:

Screenshot 2023-11-06 at 10 29 48

Keavon:
I saw by testing it in the build link, yep! I wasn't even aware Chrome displayed an image when installing, interesting. But I don't think it really helps the user with much of anything, as they already are in the app when installing it. For what it's worth, we don't really intend to fully support PWAs because we will soon have a proper desktop app that runs the Rust backend code natively. At that time, we might remove PWA support, or keep it around without putting effort into supporting it.

Keavon:
Alright, merged! Thank you for noticing this issue and helping fix it!

tomayac:
Thanks for accepting the PR!

One other use case for screenshots are PWA stores like https://store.app/. Your listing there will be evergreen with screenshots from the manifest, so whenever you refresh them, the store will automatically get them, too (and you won't have to think of submitting new screenshots via each store's UI).

Side track with full disclosure: I work on the Chrome browser on a team that helps developers realize their app ideas fully on the web. We [feature Graphite](https://developer.chrome.com/fugu-showcase/#editor.graphite.rs) on our page for a while now. Since you're using Wasm already, I don't expect you will gain much, if any, performance gains by going Rust-native (I assume with something like Tauri). Is there more motivation for choosing a "proper desktop app" over what the web delivers today?

Keavon:
Thanks, I saw that in your GH bio, very cool! I did notice over the past couple months our analytics showed a few visitors from the Fugu showcase which made me aware of that only recently :)

And yes, Tauri is the plan. Wasm is fast but my understanding is that it still provides a roughly 30% overhead compared to native, if what I recall reading is approximately accurate. It will also help us with a more native-feeling desktop app experience with a customized window frame. And then we can support additional features that only work in a native app, like our goal of bundling Stable Diffusion and other ML models with either [Candle](https://github.com/huggingface/candle) or [Burn](https://github.com/burn-rs/burn). Ideally something nice and portable so it can run dependency-free for easy installation by our users. Unfortunately using the WGPU backend, and then doing the inference through WebGPU, is far too slow and would probably run out of VRAM— but I did recently run across the existence of the [WebNN](https://github.com/webmachinelearning/webnn/blob/main/explainer.md) API working group, so that will be an exciting development in the future to allow our web version to be as fully-featured as possible compared to the Tauri desktop version.

On the subject of your role, there are web API limitations we've run into that prevent us from doing what we'd like compared to a native app. Would discussing those with you and identifying limitations/pain points be something you're interested in? I don't have the time or knowledge to make formal proposals or RFCs but these are APIs, or extensions to existing ones, that really need to be built for a Graphite-style web app to work as well as a productivity desktop app. Most of these are actually just related to accessing user input. I can also send some obscure GitHub issue discussions (about the lacking web API features) your way which need way more traction than they (haven't) had.

tomayac:
> Thanks, I saw that in your GH bio, very cool! I did notice over the past couple months our analytics showed a few visitors from the Fugu showcase which made me aware of that only recently :)

That's great to hear! 

> And yes, Tauri is the plan. Wasm is fast but my understanding is that it still provides a roughly 30% overhead compared to native, if what I recall reading is approximately accurate.

In the following, I'm honestly interested in hearing your thoughts and decision process; it's not meant to become a "sales pitch" for the web. 

My questions at this stage would probably be: (i) is this actually accurate? and (ii) is this indeed the performance bottleneck for Graphite? and finally (iii) compared to going to a URL, is downloading the Tauri desktop app actually the better experience for users, especially since it means you'd still need to deal with [cross-platform issues](https://tauri.app/v1/references/webview-versions/) (unlike with Electron for example).

> It will also help us with a more native-feeling desktop app experience with a customized window frame.

Did you see the [Window Controls Overlay API](https://web.dev/articles/window-controls-overlay)?

> And then we can support additional features that only work in a native app, like our goal of bundling Stable Diffusion and other ML models with either [Candle](https://github.com/huggingface/candle) or [Burn](https://github.com/burn-rs/burn). 

Did you see [Stable Diffusion in the browser](https://websd.mlc.ai/)?

> Ideally something nice and portable so it can run dependency-free for easy installation by our users. Unfortunately using the WGPU backend, and then doing the inference through WebGPU, is far too slow and would probably run out of VRAM

My colleague @beaufortfrancois would probably be interested in hearing your WebGPU feedback. 

> — but I did recently run across the existence of the [WebNN](https://github.com/webmachinelearning/webnn/blob/main/explainer.md) API working group, so that will be an exciting development in the future to allow our web version to be as fully-featured as possible compared to the Tauri desktop version.

I'm not an expert in the area, but from the feedback I heard, people were missing support for transformers in WebNN.

> On the subject of your role, there are web API limitations we've run into that prevent us from doing what we'd like compared to a native app. Would discussing those with you and identifying limitations/pain points be something you're interested in? I don't have the time or knowledge to make formal proposals or RFCs but these are APIs, or extensions to existing ones, that really need to be built for a Graphite-style web app to work as well as a productivity desktop app. Most of these are actually just related to accessing user input. I can also send some obscure GitHub issue discussions (about the lacking web API features) your way which need way more traction than they (haven't) had.

I'd be _very_ interested in learning more here. Happy to be linked to threads or discuss here, or also hop on a VC if this would be easier.

Keavon:
Sorry for the delay in responding @tomayac, I'll get back to you pretty soon!

Keavon:
Hi @tomayac, sorry for the delay! Happy new year, by the way :)

> I'd be very interested in learning more here. Happy to be linked to threads or discuss here, or also hop on a VC if this would be easier.

Could we please schedule a VC some time next week, or whenever you're free? You can email me (my email address is listed at <https://keavon.com>) to set something up. I've collected my notes about use cases where the browser APIs are lacking for a desktop app-like experience such as Graphite and I'd love to share them with you and hear your thoughts. Thanks :)

tomayac:
Thanks for getting back! I have sent two invites for the coming week. Hope one of them works. Please accept/deny, and we can take it from there. Looking forward to hearing your thoughts!

PR #1454, opened 2023-11-14
Store Input for Monitor Nodes

TrueDoctor:
Stores the input with which a node was evaluated in the monitor node.


TrueDoctor:
> Code looks good - although I'm not quite sure about the reason for this change?

This was originally intended to implement the node graph selection targets. with the input, you can determine the final output position in viewport space for the data. And that also works with arbitrary user-implemented transform nodes. 
Afaik you have currently implemented a different workaround to solve that issue but I still had the changes locally and thought it might make sense to merge them for when we might need that functionality again

PR #1455, opened 2023-11-14
Refactor Graphite dependency management

TrueDoctor:
Use workspace dependencies throughout the projects to minimize the number of different versions of crates used in the project


0HyperCube (approved):

node-graph/interpreted-executor/Cargo.toml
```diff
@@ -13,7 +13,7 @@ quantization = ["graphene-std/quantization"]
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-graphene-core = { path = "../gcore", features = ["async", "std"] }
+graphene-core = { path = "../gcore", features = ["std"] }
```
Keavon, line 16: Should the other deps in this file (lower down) be switched over to `workspace` too?

Keavon:
`cargo check`:

```
error: failed to load manifest for workspace member `C:\Users\Keavon\Projects\Graphite\editor`

Caused by:
  failed to load manifest for dependency `graphite-document-legacy`

Caused by:
  failed to load manifest for dependency `graphene-std`

Caused by:
  failed to load manifest for dependency `compilation-client`

Caused by:
  failed to load manifest for dependency `future-executor`

Caused by:
  failed to read `C:\Users\Keavon\Projects\Graphite\node-graph\future-executor\Cargo.toml`

Caused by:
  The system cannot find the path specified. (os error 3)
```

PR #1473, opened 2023-11-25
Fix transforms, Brush tool, and G/R/S

0HyperCube:
Transforms are now less broken.

Also I fixed the brush tool and GRS system.


Keavon (approved):
Fantastic improvements in this one, thank you! Getting close to feature parity with prod 🎉.

PR #1474, opened 2023-11-25
Added more refined Readme file

DsChauhan08:
Created a better design and more condensed form of README file. Looking to add a small a few more changes if you like.


Keavon:
Hi, thank you for your interest and efforts! But unfortunately the changes you've proposed feel very much like an overzealous marketing team which isn't something we want to convey. We also want to keep our readme pretty short. It'll definitely get improvements going forward, and feel free to suggest areas where you find it lacking.

PR #1476, opened 2023-11-25
Avoid unnecessary graph sends to the frontend

0HyperCube:
Avoid unnecessarily sending the graph when the graph UI is closed. Probably increases performance a little.


Keavon (approved):
LGTM!

PR #1484, opened 2023-12-03
Reduce console spam

TrueDoctor:


Keavon:
@0HyperCube @TrueDoctor Great work fixing all these transform issues! We still have one regression that I notice: folders now have no bounds so clicking them shows just the pivot and no rectangle for the transform cage.

Keavon:
Oh I didn't realize that! I'll make a note of it in the tracking issue then.

Keavon (approved):
Great collaborative effort on this @0HyperCube and @TrueDoctor! Good to see the code simplified and bugs solved in the end.

PR #1487, opened 2023-12-04
Improve text transforms

### LINKED ISSUE #1485: Text appears in the wrong place
Text now appears in the wrong place due to #1484. The `TransformSet` in viewport space now sets the transform in document space.

0HyperCube:
Closes #1485

- Set text in viewport
- Insert within artboard
- Fix freehand after adding text
- Fix adding in freehand and splines to transformed artboard.


Keavon (approved):
Looks like this fixes all the issues I encountered yesterday! Thanks.

PR #1494, opened 2023-12-09
Serve the demo artwork in each build

0HyperCube:
Instead of importing the demo artwork from a GitHub link, which means the dev has to share the same file used by prod even though they aren't compatible, now we copy the demo artwork files into each deployed build's distributables.


Keavon:
Temporarily blocked on https://github.com/nguyenbatranvan/vite-multiple-assets/issues/8 which is an issue with a plugin that can allow us to have multiple `public`/assets folders. That seems cleaner because I'd ideally still like to define as much as possible purely in Rust so there are a few dependencies on the JS code. But if that isn't resolved in a few days, your current approach is a solid one too so we'll roll with that. But in the next few days I'll see if things change in that issue linked above.

Keavon (approved):

PR #1497, opened 2023-12-09
Fixes for removing artboards; white infinite canvas background

0HyperCube:
- Fix crash when drawing on a deleted artboard. 
- Fixes a warning when deleting a layer.
- ~~Fixes the clear all artboards button~~
- White background when there are no artboards.

Part of #1394

### DIFF
--- a/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
+++ b/editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
@@ -254,7 +254,7 @@ impl LayoutHolder for MenuBarMessageHandler {
 				no_active_document,
 				MenuBarEntryChildren(vec![vec![MenuBarEntry {
 					label: "Clear Artboards".into(),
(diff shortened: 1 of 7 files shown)


editor/src/messages/portfolio/menu_bar/menu_bar_message_handler.rs
```diff
@@ -254,7 +254,7 @@ impl LayoutHolder for MenuBarMessageHandler {
 				no_active_document,
 				MenuBarEntryChildren(vec![vec![MenuBarEntry {
 					label: "Clear Artboards".into(),
-					action: MenuBarEntry::create_action(|_| DialogMessage::RequestComingSoonDialog { issue: None }.into()),
+					action: MenuBarEntry::create_action(|_| GraphOperationMessage::ClearArtboards.into()),
```
Keavon, line 257: This was disabled because it crashes (as it wasn't ported to document nodes). And with this PR (at least on my computer) this still does crash. So I'll revert this line for now so it doesn't hold up the rest of the PR which I'm merging now.

Keavon (approved):

PR #1498, opened 2023-12-09
Folder bounds

### LINKED ISSUE #1508: Selected folders incorrectly have a zero-size bounding box
The Select Tool needs to show the bounding box for its transform cage when a layer is selected. Currently this appears zero-sized, so we just see a pivot widget floating somewhere in space.

0HyperCube:
**Blocked on the magic pivot system #1530**

- The bounding box of a layer is now computed based of it and any of its descendants (with the appropriate relative transform applied)
- Transform nodes on folders correctly transform around the pivot
- Everything else is broken.

transformissue.webm

For the transform node in the footprint system - the footprint is incorrect in cases where there is a transform node with a scale if the bounding box is not 1x1.

This is because the transform for the footprint in the transform node is computed using the pivot transform assuming that the bounding box of the layer is 1x1. We then call the upstream nodes with this incorrect footprint. Then we get the actual data and compute the real pivot based on the bounding box of the layer.

I noticed this whilst trying to debug transforming folders - I made the folder's pivot be the centre of its bounding box (which it should presumably be so that it transforms around the actual middle) and I can no longer properly transform the inner layer due to this bug.

The issue is that the transformations with the select tool are based of a monitor node which stores the footprint that is passed into the layer. If the footprints are incorrect then the transformations will be incorrect (causing the to transform incorrectly). 

Closes #1508

### DIFF
--- a/document-legacy/src/document_metadata.rs
+++ b/document-legacy/src/document_metadata.rs
@@ -279,10 +279,16 @@ impl DocumentMetadata {
 
 	/// Get the bounding box of the click target of the specified layer in the specified transform space
 	pub fn bounding_box_with_transform(&self, layer: LayerNodeIdentifier, transform: DAffine2) -> Option<[DVec2; 2]> {
-		self.click_targets
-			.get(&layer)?
-			.iter()
-			.filter_map(|click_target| click_target.subpath.bounding_box_with_transform(transform))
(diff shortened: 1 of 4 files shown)


Keavon:
Broken out from #1394.

PR #1499, opened 2023-12-10
Make menu lists searchable

### LINKED ISSUE #1503: Menu lists like font selection should be searchable

0HyperCube:
Searchable font list as suggested by @1oonie. All of the other menus are also searchable.

Closes #1503

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -1180,7 +1180,7 @@ impl DocumentMessageHandler {
 							.icon(DocumentMode::SelectMode.icon_name())
 							.on_update(|_| DialogMessage::RequestComingSoonDialog { issue: Some(330) }.into()),
 						MenuListEntry::new(format!("{:?}", DocumentMode::GuideMode))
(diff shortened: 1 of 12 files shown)


Keavon:
I'm presently working through these and checking them off as I go:

- [x] Left/Right/Up/Down keys need to preventDefault when they move the menu highlight or open/close a submenu
- [x] Spacebar/Enter to activate a submenu entry throws a console error (but clicking works fine, and also it's fine for non-submenus)
- [ ] Deleting all text from the search needs to remove the search bar and return to how it started
- [ ] Fix CSS for search box
- [ ] Escape should close the search, not the whole menu
- [ ] Make it so opening a submenu deselects the search bar text so, for example, you can activate a submenu entry with the spacebar without typing a space in the previous menu's search box
- [ ] Make searching in a submenu possible by opening the search box in that submenu while moving focus away from the previous menu's search box
- [ ] Remove the FontInput component
- [ ] Finish code review

0HyperCube:
@Keavon It is probably not necessary to add these features now - I think we could merge this as a non-breaking UX improvement and then add the rest of these improvements later.

Keavon (approved):

PR #1501, opened 2023-12-10
Fix nudging and nudge resizing

### LINKED ISSUE #1317: Make arrow key nudge-resizing operate in document space

0HyperCube:


Keavon:
Could you please make this operate in document space, not viewport space? So it should always nudge everything by 1px (or 10px if <kbd>Shift</kbd> is held) regardless of viewport zoom. Then you might encounter the issue described in #1317, which occurred upon switching from viewport to document space. I could also just edit that issue to describe the request to change to being relative to document space, and we proceed to merge this as-is for now (since having it at all is preferable to delaying until that can be fixed into its ideal form).

Keavon:
I decided it's a better use of your time @0HyperCube to merge this now so you can choose to make nudging operate in document space at a time of your choice instead of needing to do it now, or it can be left to the community to work on if you judge that you have bigger fish to fry compared to this quality-of-life improvement/papercut. So I updated #1317 to describe the improvement of switching it to document space (instead of its previous focus, which was a bug resulting from switching it to document space, as that's not as relevant yet). Feel free to tackle that now, or whenever, or leave it for the community— it's no longer critical path as a part of the document graph refactor tracking issue.

PR #1502, opened 2023-12-10
Fix deleting all artboards without crashing

0HyperCube:
Delete all artboards now doesn't crash.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
@@ -1,6 +1,5 @@
 use super::{resolve_document_node_type, VectorDataModification};
 use crate::messages::prelude::*;
-use crate::messages::tool::common_functionality::graph_modification_utils::is_artboard;
 
 use bezier_rs::Subpath;
 use document_legacy::document::Document;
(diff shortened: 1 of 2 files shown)


Keavon:
The functionality I had in mind was removing the artboards but not the artwork contained within. Basically transplanting all the artwork to an infinite canvas. Would that be a change you can make, please?

editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
```diff
@@ -1,5 +1,6 @@
 use super::{resolve_document_node_type, VectorDataModification};
 use crate::messages::prelude::*;
+use crate::messages::tool::common_functionality::graph_modification_utils::is_artboard;
```
Keavon, line 3: This unused import can be removed.

0HyperCube:
> The functionality I had in mind was removing the artboards but not the artwork contained within. Basically transplanting all the artwork to an infinite canvas. Would that be a change you can make, please?

If you have placed artwork as a child of an artboard, it will be deleted when the parent artboard is deleted. However you could move the arwork to outside of the artboard if you wish to avoid this.

0HyperCube:
If you decide you want deleting a group or artboard not to delete child layers, then you are welcome to open a separate issue for that and then implement it when you have time.

Keavon:
> If you decide you want deleting a group or artboard not to delete child layers, then you are welcome to open a separate issue for that and then implement it when you have time.

Sounds good, I'll merge this now as-is then!

Keavon (approved):

PR #1506, opened 2023-12-11
Fix blend mode and opacity

0HyperCube:
Note that it doesn't work if the opacity is driven from the graph. However if we sampled the opacity from the monitor node then it wouldn't work for images (as the opacity applied to an image is not stored).


Keavon:
UX feedback for some edge cases:

- [ ] If an individual node (rather than a layer) is selected, we want the widgets disabled instead of showing a dash.
    - [ ] However it would be good for a selected Opacity or Blend Mode node to each show its value in its respective widget.
- [ ] The backwards probing logic should stop when it reaches another layer, rather than following it up that layer's node chain

Keavon (approved):

PR #1513, opened 2023-12-15
New node: Copy to Points

Keavon:
Part of #912


node-graph/gcore/src/vector/vector_data.rs
```diff
@@ -102,6 +102,10 @@ impl VectorData {
 	pub fn local_pivot(&self, normalized_pivot: DVec2) -> DVec2 {
 		self.transform.transform_point2(self.layerspace_pivot(normalized_pivot))
 	}
+
+	pub fn cloned_subpaths(&self) -> Vec<bezier_rs::Subpath<ManipulatorGroupId>> {
+		self.subpaths.clone()
+	}
```
0HyperCube, line 108: I'm not sure why we need this function as subpaths is public anyway?

0HyperCube (approved):

node-graph/gcore/src/vector/vector_data.rs
Keavon: Thanks, that was leftover dead code.

PR #1536, opened 2023-12-24
Make CopyToPoints node resolution aware

TrueDoctor:

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -157,6 +157,7 @@ async fn copy_to_points<FP: Future<Output = VectorData>, FI: Future<Output = Vec
 	points: impl Node<Footprint, Output = FP>,
 	instance: impl Node<Footprint, Output = FI>,
 ) -> VectorData {
+	// TODO: https://github.com/GraphiteEditor/Graphite/pull/1536#discussion_r1436422419
 	let points = self.points.eval(footprint).await;
 	let instance = self.instance.eval(footprint).await;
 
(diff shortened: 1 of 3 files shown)


node-graph/gcore/src/vector/vector_nodes.rs
```diff
@@ -144,12 +146,20 @@ fn generate_bounding_box(vector_data: VectorData) -> VectorData {
...
 
 #[node_macro::node_fn(CopyToPoints)]
-fn copy_to_points(points: VectorData, instance: VectorData) -> VectorData {
+async fn copy_to_points<FP: Future<Output = VectorData>, FI: Future<Output = VectorData>>(
+	footprint: Footprint,
+	points: impl Node<Footprint, Output = FP>,
+	instance: impl Node<Footprint, Output = FI>,
+) -> VectorData {
+	let points = self.points.eval(footprint).await;
+	let instance = self.instance.eval(footprint).await;
```
0HyperCube, line 161: Should we include instance's transform? e.g. if instance is a rectangle with a transform applied to scale it - should the scale be applied to the subpath before it is instanced? Currently I feel it is slightly confusing just to ignore the transform of the instance.

Keavon: Yes, I agree it should. Also, same for gradients— it should apply per-instance and not to the graphic after the instance is created (unless the gradient is added later than this node). Although that might be tricky with the SVG format if we don't throw things in groups, which then gets difficult to put a gradient over the top of. So I'm not sure... 🤔.

Keavon (approved):

PR #1558, opened 2024-01-10
Redesign transforms

### LINKED ISSUE #1180: When text is modified in the Properties panel, it shifts around
Editing the text in the properties panel causes the position of the text in the document to shift (we need to update the position when the text changes, as the position is relative to the centre of the text). This only happens if the text has a scale other than 1x1 in X and Y.

### LINKED ISSUE #1530: Add origin to the Transform node (and visualize it with an overlay)
Design and implement the new pivot system containing a pivot stored in layer space (?) instead of normalised bounding box (so a pivot of (0.5, 0.5) would no longer be in the middle of the bounding box as it is currently). Note that this will break things such as rotating around the pivot, transformations applied by tools, etc. Careful thought should be applied as to how exactly the pivot should be stored.

"Pivot" concept is split into "reference point" and "pivot"
- Reference point: after the graph has rendered, this is the point from which we measure X and Y location
- Pivot: an Option<> which starts out as None, it must be set (either by a Set Pivot node or a widget next to anything, like Scale or Rotation, which requires it)

Also we likely don't use the nine-square widget for pivot to ensure people think of it only as the reference point concept

So the Transform node changes like so:
- "Pivot" becomes "Reference Point"
- Rotation and Scale get a "Pick Pivot" widget which lets you place a pivot in the artwork, it's required if the pivot is None
- Also: add skew? And non-affine?

Goals: don't have shift caused by upstream changes, thus we can't use the bounding box as a pivot reference, instead we have to explicitly have the user set any desired pivots

---

As part of combining issue #534 into here, here's the description that was part of it:

- Folders have transforms?
- Showing skew in properties panel
- Non-affine perspective transforms (dragging each corner of the bounding box to an arbitrary place, except concave)
- Inconsistent behaviour between lines / polylines and other shapes (render_index in simple_shape)
- Allowing outline width to be scaled / not as desired
- How to display layer & folder(if applicable) transforms in properties panel (i.e. do we show layer transform or bounding box, and how is it transformed etc.)
- Clear reference for contributes / artists

TrueDoctor:
Closes #1530, closes #1180


TrueDoctor:
@Keavon can you test if the behaviour  for folders is sensible? (the placing of the pivot marker is still broken, should be placed based on the selections bounding box). But I think the math is right at least.
Next steps would include fixing the displaying of the bounding box and automatically placing the pivot in the correct spot

Keavon:
What should I look for? It seems like you'll need to complete those next steps before anything is functional enough for me to answer your question. Maybe we can do a call and you can walk me through answering exactly what you're wondering?

PR #1576, opened 2024-01-13
New node: Morph

### LINKED ISSUE #1556: Feature Request: Blendshapes Node
Hello,

it would be pretty awesome to have a "Blendshapes" node. I found a video of how this works in Houdini and I think it could be implemented pretty much the same way in Graphite:
https://www.youtube.com/watch?v=yRUu4zi5dBQ

A requirement for the input shapes would probably be that they have the same number of points.
I think the node could also blend the fill and stroke of the shapes, similar to how in the video the color of the different meshes are blended.

For reference this is my original message in Discord where I asked if this was possible:

"Another cool feature that I think might work with vector graphics would be sliders to change proportions. I think in 3D that is usually done with morph targets.

When there is support for editing keyframes I imagine that making e.g. length and width sliders could work like this:
- Have 4 keyframes for a body part: "short and thin", "long and thin", "short and thick", "long and thick"
- Calculate two temporary keyframes by interpolating between "short and thin"/"long and thin" and "short and thick"/"long and thick" with the desired length value
- Interpolating between those temporary keyframes with the desired width value"

0HyperCube:
Closes #1556

example of node


Keavon (approved):
This is such an unexpectedly awesome feature! I'm really impressed. Now I'm itching to add a basic timeline to do animation in Graphite.

PR #1577, opened 2024-01-13
Display graph errors in the viewport

### LINKED ISSUE #1269: Display node graph errors to the user
We used to have this enabled but it was spammy. Now that (at least ideally) we have fewer node graph errors, reenabling this would be good. And even better to display a warning in the footer bar so the user can click it to view the error instead of having it show up automatically. This bonus step would require some frontend changes (hence the Web label on this issue).

0HyperCube:
Closes #1269


Keavon (approved):
Could you also please make it clear the click targets, so we don't see the previously updated shape outlines appear when hovering with the Select tool?

PR #1579, opened 2024-01-14
SVG import

### LINKED ISSUE #625: Importing editable SVGs
We won't be able to support every feature of the SVG spec, but working on supporting this will help us move towards adding more features that SVG provides that we've forgotten about so far. Before we transition to the node system, we can probably reach a happy middleground of supporting most common SVG features. Overly complex SVGs might not end up looking the same once imported, but we can't strive for perfection quite yet.

0HyperCube:
Closes #625

SVG import of graphite logo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1195,6 +1195,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d7439c3735f405729d52c3fbbe4de140eaf938a1fe47d227c27f8254d4302a5"
 
 [[package]]
+name = "data-url"
+version = "0.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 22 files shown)


TrueDoctor:
does it make sense to couple this in some way with the `usvg` export function I added for rust based svg rendering?
By moving the `usvg` -> `GraphicElement` and `GraphicElement` -> `usvg` conversions closer to each other we might make it easier to catch subtle errors

0HyperCube:
> does it make sense to couple this in some way with the `usvg` export function I added for rust based svg rendering? By moving the `usvg` -> `GraphicElement` and `GraphicElement` -> `usvg` conversions closer to each other we might make it easier to catch subtle errors

We convert a usvg to a document graph which then produces the GraphicElement. It isn't really possible to move the code for generating the graph from an SVG into gcore as it requires the `ModifyInputsContext` stuff from the editor to handle the generation of the node graph.

Keavon (approved):

PR #1580, opened 2024-01-15
Save work periodically to reduce loss from crashes

### LINKED ISSUE #929: Save work periodically to reduce loss from crashes
Try and find more times it would make sense to save, since it does likely take a little while to serialize and write the document which should ideally not cause annoying freezes for the user. We should aim to hopefully never make the user lose more than, say, 30 seconds of work.

Also, closing/reloading the page sometimes doesn't save correctly (maybe it doesn't have enough time if the editor is being slow or the document is big?). This should help in that case. Investigating the actual cause would also be nice.

Alternatively, now that we have moved toward the full node graph, it may be feasible to take a VS Code style approach where everything you do is saved in real time but the `*` isn't updated until you download the file. It might even be better to remove the `*` entirely, since whether you download the file isn't really that useful, except when #747 is in use (in a compatible browser and enabled by the user).

Comeza:
Closes #929

### DIFF
--- a/editor/src/messages/portfolio/portfolio_message_handler.rs
+++ b/editor/src/messages/portfolio/portfolio_message_handler.rs
@@ -72,13 +72,19 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
 					document.process_message(message, responses, document_inputs)
 				}
 			}
-			PortfolioMessage::AutoSaveActiveDocument { force } => {
+			PortfolioMessage::AutoSaveActiveDocument => {
 				if let Some(document_id) = self.active_document_id {
 					if let Some(document) = self.active_document_mut() {
-						if !document.is_auto_saved() || force {
-							document.set_auto_save_state(true);
-							responses.add(PortfolioMessage::AutoSaveDocument { document_id });
-						}
+						document.set_auto_save_state(true);
+						responses.add(PortfolioMessage::AutoSaveDocument { document_id });
+					}
+				}
+			}
+			PortfolioMessage::AutoSaveAllDocuments => {
+				for (document_id, document) in self.documents.iter_mut() {
+					if !document.is_auto_saved() {
+						document.set_auto_save_state(true);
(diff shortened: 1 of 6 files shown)


Keavon:
Thanks for the PR! Please comment in the issue so GitHub lets me assign it to you. Looking forward to landing this!

Keavon:
Thanks, I assigned it to you. Once you've finished implementing the feature and it seems to work in a production-ready way, please mark the PR as ready for review and also ping me so I notice and come to take a look with QA testing and code review. Welcome!

TrueDoctor:
@Keavon are you d'acorde  with the approach taken here? Currently this saves the document after 30 s of inactivity. Should we add a second longer timer which forces a save if no inactivity was sufficiently long?

Keavon:
I was thinking that maybe a good time to save would be, if 30 seconds have elapsed, when the history stack changes. Does that seem feasible?

0HyperCube:
!build

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -76,8 +76,16 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
...
 					if let Some(document) = self.active_document_mut() {
 						document.set_auto_save_state(true);
+						responses.add(PortfolioMessage::AutoSaveDocument { document_id });
+					}
+				}
+			}
+			PortfolioMessage::AutoSaveAllDocuments => {
+				for (document_id, document) in self.documents.iter_mut() {
+					if !document.is_auto_saved() {
+						document.set_auto_save_state(true);
```
Keavon, line 86: Isn't this line redundant, since it happens in `AutoSaveDocument`?

Keavon: Disrgard.

Keavon (approved):
Sorry this took so abnormally long to review!

PR #1581, opened 2024-01-17
Path tool: sliding point insertion

### LINKED ISSUE #1578: Path tool sliding point insertion
For the Path tool, instead of double clicking to insert a point like it is now, we want it so instead you click *once* and it shows a sliding dot (similar to Shift+V or GG in Blender to slide a point) and it lets you move the point where the split happens, then clicking again applies it (or right clicking to cancel, or maybe moving the mouse far away cancels it).

Nikita-str:
Closes #1578


Keavon:
!build

Nikita-str:
seems like done, check it when you can(or I can attach record screen before it)

about cargo-deny: after merging master into it's response ok, do I need to commit merge? (idk will it affect on file changes of the PR and therefore don't sure about will it be convenient to view changes ://)

Nikita-str:
@Keavon I forgot to mention you heh

Keavon:
!build

Keavon:
Thanks! A few quick notes to give you something to work on before I go to bed (I'll be more thorough on my next review):
- In addition to right click to cancel, <kbd>Esc</kbd> also needs to cancel.
- Double clicking currently also converts the new point to sharp upon its creation (which is the behavior for double clicking an existing point, so it seems both are happening). Preventing that is important so the user can just double click on the curve to add a point, without choosing to slide it, like before.
- I'd appreciate if the sliding point can be drawn as an overlay (a yellow anchor point instead of the regular blue ones, and no handles), and not actually committed to the shape until you click again. That means this instanced light bulb shouldn't appear and slide along; it should appear only once you click to confirm:
  
  https://github.com/GraphiteEditor/Graphite/assets/4388688/0039c113-adc1-4948-9a33-77f99493df4e
- Good job adding the hints to the footer bar. You can call those "Insert" and "Cancel".

This is looking very nice, thank you for helping add this!

Nikita-str:
@Keavon 
did what you said (maybe I didn't correct understood the last one)
review it again please

Keavon:
!build

Keavon:
Looking good! Only two little issues that I notice now. One is that we shouldn't be reusing the yellow pivot crosshair, please update the anchor point square overlay render code to take a color override (probably as an `Option` so the normal usages can avoid having to specify the default blue). The other is that it seems I can't slide a point close to the edges ends of a segment, it's locked between something like 10% on either end.

https://github.com/GraphiteEditor/Graphite/assets/4388688/8e82c667-1f32-4b5c-b744-9d4c2da83764

Nikita-str:
About sliding a point too close to the edges -- it was done intentionally (and you can scale up and then move it closer), because I think that user don't want to make two point practically in the same place, but if it's incorrect assumption I can remove that behavior. So it's incorrect?

Keavon:
Ah, good thinking! Yes, we don't want the user to put a point exactly on top of the anchor. That's nice to hear it adapts as you zoom in. How do you define it currently? Is it based on a percentage like my 10% estimate or some viewport pixel length? (If the latter, what happens if you're zoomed out a lot?) Can you add that magic number to `consts.rs` and try a value that's more like 1/10 the current value?

Nikita-str:
Oh, I don't know about `consts.rs` so it was a local const.
It's 7.5% for line that have length 400 and if a line is longer than that value becomes less, and if a line is shorter than the value becomes more but no more than 25% (on each side)
[code](https://github.com/GraphiteEditor/Graphite/pull/1581/files#diff-8593d87a3c27b6171bdc417c89c854490dc72ccc1765c5bbf9973e5f8364eb89R116)

Keavon:
Maybe I misinterpreted your previous reply, but does it adapt based on viewport zoom? From your explanation and the code, it looks like it doesn't. That would be very helpful to add, and you could probably simplify it like this:

Allow the user to move it to within 2px (put this value in `consts.rs`) of either anchor point. If the user is zoomed out enough that there isn't 2px on either side, don't enter point insertion mode to begin with.

0HyperCube:
> Ah, good thinking! Yes, we don't want the user to put a point exactly on top of the anchor. That's nice to hear it adapts as you zoom in. How do you define it currently? Is it based on a percentage like my 10% estimate or some viewport pixel length? (If the latter, what happens if you're zoomed out a lot?) Can you add that magic number to `consts.rs` and try a value that's more like 1/10 the current value?

Might it be useful to position two anchors ontop of each other if you needed that for some procedural effect or something?

Nikita-str:
@0HyperCube we can remove "too close to the sides" behavior for example when ctrl down, should I do it?

@Keavon done
(sorry for time lapse)

Instead of additional `Option` argument I [use](https://github.com/GraphiteEditor/Graphite/pull/1581/files#diff-fc1f68d21385d829e2a2a8eadca8f208e900ee9c3262a58fcc79ae5ff4251733R153) `enum` because:
with `Option` it can be done as next:
```rust
fn _(.., selected: bool, color: Option<&str>) {
    let color = color.unwrap_or(if selected { COLOR_OVERLAY_BLUE } else { COLOR_OVERLAY_WHITE });
}
```
and here it seems that both params `selected` and `color` is responsible for the same thing, for me it seems incorrect.
OR
with `Option` it also can be done as next:
```rust
fn _(.., selected: bool, selected_color: Option<&str>) {
    let color = if selected { selected_color.unwrap_or(COLOR_OVERLAY_BLUE) } else { COLOR_OVERLAY_WHITE });
}
```
but what if we will need another unselected color?
```rust
fn _(.., selected: bool, selected_color: Option<&str>, unselected_color: Option<&str>) {
let color = if selected { selected_color.unwrap_or(COLOR_OVERLAY_BLUE) } else { unselected_color.unwrap_or(COLOR_OVERLAY_WHITE) })
}
```
isn't it too many arguments for such thing?

Keavon:
I would worry that placing a point exactly on top of another point could be a footgun, and if the user really needs to achieve that, it can be done by dragging and snapping the point with the Path tool. I think having the viewport-relative pixel distance is the best balance of simplicity and safety, at least for now but we can always improve it if we learn that users find it troublesome (that's another nice part of having the pixel distance in `consts.rs`).

Regarding the colors thing, I see where you're going with that thought and it seems like a sensible abstraction, however I don't think we really need an abstraction for that because I don't think we're likely to need to generalize it any further in the future. My suggestion was basically: the old function draws blue squares, but if the caller wants to override that, they can pass a color. Instead of the function using `COLOR_OVERLAY_BLUE`, it uses `color.unwrap_or(COLOR_OVERLAY_BLUE)`. I'd prefer the simplicity of that, since I don't think "Selected"/"Temporary"/"Unselected" models where we're headed in the future with this system.

Thanks!

Nikita-str:
@Keavon redid with `Option`

Keavon:
!build

Keavon:
When inserting along a straight segment, can you please make it so the inserted point is sharp instead of smooth (having handles)?

Nikita-str:
Done
Also do correct insertion on quadratic segments

Meanwhile why there often used `handle = Some(self.anchor)` instead of `handle = None`? (`SetManipulatorPosition` instead of `RemoveManipulatorPoint`)
For performance reason?
`cubic(A, B, C, C) != quadratic(A, B, C)` or it can be neglected?
image

Keavon:
Thanks for fixing that!

I'm not entirely sure that I understand your question, but I think you may have answered it yourself in observing that `cubic(A, B, C, C) != quadratic(A, B, C)`. Do we support quadratic and linear forms of Beziers, but our editing tools generally work with cubic (and "deleting a handle" means setting it to 0 length and hiding it to the user but the curve remains cubic).

If that doesn't answer your question, would you be able to rephase it?

Keavon (approved):

PR #1582, opened 2024-01-17
Remove old subpath code

0HyperCube:
From code todo list. This code was unnecessary after our migration to the node graph and bézier-rs types.


Keavon:
Wow, zero lines of additions? So this code was just... hanging out without even needing a refactor for removal? Great find!

Keavon (approved):

PR #1583, opened 2024-01-17
Restore select tool functionality and shallow select mode

0HyperCube:
Restore select tool functionality:
- Shallow select
- <kbd>alt</kbd> click drag to duplicate
- <kbd>ctrl</kbd> <kbd>D</kbd> duplicate
- <kbd>alt</kbd> to centre
- Snapping when constrained or centred

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -398,12 +398,29 @@ impl MessageHandler<DocumentMessage, DocumentInputs<'_>> for DocumentMessageHand
 				responses.add(FrontendMessage::UpdateDocumentLayerStructure { data_buffer });
 			}
 			DuplicateSelectedLayers => {
-				// TODO: Reimplement selected layer duplication
-				// self.backup(responses);
-				// self.layer_range_selection_reference = None;
-				// for path in self.selected_layers_sorted() {
-				// 	responses.add(DocumentOperation::DuplicateLayer { path: path.to_vec() });
-				// }
+				self.backup(responses);
+				for layer_ancestors in self.metadata.shallowest_unique_layers(self.selected_nodes.selected_layers(&self.metadata)) {
+					let Some(layer) = layer_ancestors.last().copied() else { continue };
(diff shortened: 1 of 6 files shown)


Keavon:
I can't believe you managed to remain code negative while adding back all this functionality— including Deep/Shallow selection which used to be that hideous mess of hundreds of lines of highly indented code. This is working remarkably well in general! I just have a few notes for how we can further improve the UX:

- [ ] Could you please make the duplicated copy (the one being dragged) from <kbd>Alt</kbd>-drag appear above the original in the layer stack instead of beneath?
- [ ] And for Shallow Select, could you please make that ignore artboards?
- [ ] There's also one more overall behavioral change we need from shallow selection mode, it needs to follow this algorithm described in this drawing I made:  
  
  image
  
  - Circles are folders
  - Rectangles are layers with graphical content
  - The gray circle and two gray rectangles are the three selections in this example
  - The two lighter faded gray rectangles are children of the selected folder
  - The orange arrows point from which graphical content is clicked in the viewport to which item ends up getting selected (or, if <kbd>Shift</kbd> is held, added to the selection)
      - If <kbd>Shift</kbd> is held and something is added to the selection, any sub-layers of that are deselected
- [ ] As a bonus, we'd ideally like to also visualize what is about to be selected in shallow select mode by showing, upon hover, an overlay of the bounding box for the layer that will be selected upon clicking.

---

I'm more than happy to merge this in any state and continue those in future PRs. I'm just nothing them here or I can transfer them to a separate issue if you'd prefer we save those enhancements for later.

Keavon (approved):
The code looks good!

0HyperCube:
I'm not going to try and interpret this...
gibberish

Keavon:
Thanks!

Keavon:
And one of us can move the Deep/Shallow Select dropdown to the user preferences at any point in the future. I'm personally ambivalent about whether it's part of the tool options or the preferences, but I respect your opinion about that as you mentioned in Discord. But I'd say it's moderately low priority to change at this moment so feel free to submit a PR anytime now or later in the future.

Issue #972, opened 2023-01-15
Make draggable widgets set only one undo/redo state upon committing a value


Keavon:
Draggable number input sliders, color pickers, typeable text fields, and any other widgets we may have now or in the future should be able to send two types of values to the backend. An "update" and a "commit". The former should preview the change like it does now, while the latter should set a history state. Reference the text input widgets for something conceptually related (but not fully implemented end-to-end like this issue describes) regarding an update vs. a commit of a value change. That should use the same system.

Currently we set a history point for every change which wipes out history with tons of small changes caused by single-pixel drag movements in the widgets.

Keavon:
@zhiyuang please leave a comment in this issue.

zhiyuang:
The root cause is that right now always trigger `Document.StartTransaction` to save a history step. I have raised a PR to split update and commit. And then we only trigger `Document.StartTransaction` on commit action, not update action.

Keavon:
Thanks @zhiyuang. And for future reference, no need to give an elaborate comment, I just need any comment at all so GitHub lets me assign you the issue.

PR #1584, opened 2024-01-23
Split widget callbacks into update and commit so only the latter adds a history state

zhiyuang:
Closes https://github.com/GraphiteEditor/Graphite/issues/972

try to fix ColorButton and NumberInput widgets.

### DIFF
--- a/editor/src/messages/layout/layout_message_handler.rs
+++ b/editor/src/messages/layout/layout_message_handler.rs
@@ -48,17 +48,29 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
 		use LayoutMessage::*;
 		#[remain::sorted]
 		match message {
-			CommitLayout { layout_target, widget_id, value } => {
-				let layout = if let Some(layout) = self.layouts.get_mut(layout_target as usize) {
-					layout
-				} else {
+			ResendActiveWidget { layout_target, widget_id } => {
+				// Find the updated diff based on the specified layout target
+				let Some(diff) = (match &self.layouts[layout_target as usize] {
+					Layout::MenuLayout(_) => return,
+					Layout::WidgetLayout(layout) => Self::get_widget_path(layout, widget_id).map(|(widget, widget_path)| {
+						// Create a widget update diff for the relevant id
+						let new_value = DiffUpdate::Widget(widget.clone());
+						WidgetDiff { widget_path, new_value }
+					}),
+				}) else {
+					return;
+				};
+				// Resend that diff
+				self.send_diff(vec![diff], layout_target, responses, &action_input_mapping);
+			}
+			SendLayout { layout, layout_target } => self.diff_and_send_layout_to_frontend(layout_target, layout, responses, &action_input_mapping),
+			WidgetValueCommit { layout_target, widget_id, value } => {
+				let Some(layout) = self.layouts.get_mut(layout_target as usize) else {
 					warn!("CommitLayout was called referencing an invalid layout. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
 					return;
 				};
 
-				let widget_holder = if let Some(widget_holder) = layout.iter_mut().find(|widget| widget.widget_id == widget_id) {
-					widget_holder
-				} else {
+				let Some(widget_holder) = layout.iter_mut().find(|widget| widget.widget_id == widget_id) else {
 					warn!("CommitLayout was called referencing an invalid widget ID, although the layout target was valid. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
 					return;
 				};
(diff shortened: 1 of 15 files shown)


0HyperCube:
!build

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -48,6 +48,21 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+					warn!("CommitLayout was called referencing an invalid layout. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
+					return;
+				};
+
+				if layout.iter_mut().find(|widget| widget.widget_id == widget_id).is_none() {
+					warn!("CommitLayout was called referencing an invalid widget ID, although the layout target was valid. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
+					return;
+				}
+
+				responses.add(DocumentMessage::StartTransaction);
```
0HyperCube, line 64: It appears this is mistakenly adding undo commits when modifying the tool options (e.g. dragging the weight and stroke colour with the line tool at the top).

tool options

0HyperCube (commented):
Thanks for this amazing editing improvement.

editor/src/messages/layout/layout_message_handler.rs
zhiyuang: oh, i see. I missed this condition, maybe I need to do something like on_update callback

0HyperCube:
!build

Keavon:
I'll give this a detailed code review tomorrow, bedtime now though! Thanks @0HyperCube for the review so far.

frontend/src/components/widgets/WidgetSpan.svelte
```diff
@@ -76,31 +84,31 @@
 	{#each widgets as component, index}
 		{@const checkboxInput = narrowWidgetProps(component.props, "CheckboxInput")}
 		{#if checkboxInput}
-			<CheckboxInput {...exclude(checkboxInput)} on:checked={({ detail }) => updateLayout(index, detail)} />
+			<CheckboxInput {...exclude(checkboxInput)} on:checked={({ detail }) => commitAndUpdateLayout(index, detail)} />
 		{/if}
 		{@const colorInput = narrowWidgetProps(component.props, "ColorButton")}
 		{#if colorInput}
-			<ColorButton {...exclude(colorInput)} on:value={({ detail }) => updateLayout(index, detail)} />
+			<ColorButton {...exclude(colorInput)} on:value={({ detail }) => updateLayout(index, detail)} on:start={() => commitLayout(index)} />
```
Keavon, line 91: Can you please help me understand why we are committing the layout upon the start of the user's dragging? Why doesn't this happen upon completion of the dragging?

zhiyuang: it also seems unreasonable to me first. After tried myself, I try to understand like this: for undo history, we always need to commit a undo history and keep a snapshot before we begin new changes. Then upon the start of dragging means commit a undo history before user begin new dragging changes. 

Actually we commit the layout of a last status, not a layout changed by this widget. Maybe we could consider change a name of this.

Keavon: But doesn't that mean that we don't save a history step upon completing the drag? So if you change a number from 1 to 10 by dragging its NumberInput slider, that value of 10 never gets saved as a history step? So the user could move on to other things, and if they undo several times, the history jumps back two operations at once (the operation that moved from 1 to 10, and whatever the other operation was adjacent to that).

zhiyuang: > But doesn't that mean that we don't save a history step upon completing the drag? 

saving a history step upon completing the drag is a responsibility belongs to the next widget's commit_layout or next action that trigger `Document.startTransaction`. That's why common widgets use `updateAndCommitLayout`.

zhiyuang: > So if you change a number from 1 to 10 by dragging its NumberInput slider, that value of 10 never gets saved as a history step?

if user's next action is to add a path using pen tool. During we adding a path, we will trigger `Document.StartTransanction` then save a history step. (Our existing history related code have handled this properly I think)

Keavon: I see! I think part of my confusion is because I don't have a full understanding of how `StartTransaction`/`AbortTransaction`/`CommitTransaction` work (since the original author of that never documented them). I did test this all out and it seems to work well so I trust that you're right about this. I'll keep going with my code review.

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -48,6 +48,89 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+
+				let widget_holder = if let Some(widget_holder) = layout.iter_mut().find(|widget| widget.widget_id == widget_id) {
+					widget_holder
+				} else {
+					warn!("CommitLayout was called referencing an invalid widget ID, although the layout target was valid. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
+					return;
+				};
+
+				#[remain::sorted]
+				match &mut widget_holder.widget {
```
Keavon, line 67: I'm not a big fan of having this match statement for all the widget types exist twice (also on lines 163-291). Would you mind breaking it out into a function and combining the two use cases into one so it does the update or commit logic depending on the function's arguments?

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -48,6 +48,89 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+					warn!("CommitLayout was called referencing an invalid layout. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
+					return;
+				};
+
+				let widget_holder = if let Some(widget_holder) = layout.iter_mut().find(|widget| widget.widget_id == widget_id) {
+					widget_holder
+				} else {
+					warn!("CommitLayout was called referencing an invalid widget ID, although the layout target was valid. `widget_id: {widget_id}`, `layout_target: {layout_target:?}`",);
+					return;
+				};
```
Keavon, line 64: These can be simplified with `let else` syntax.

editor/src/messages/layout/layout_message_handler.rs
```diff
@@ -48,6 +48,89 @@ impl<F: Fn(&MessageDiscriminant) -> Vec<KeysGroup>> MessageHandler<LayoutMessage
...
+
+				#[remain::sorted]
```
Keavon, line 66: I'm not a big fan of having this match statement for all the widget types exist twice (also on lines 163-291). Would you mind breaking it out into a function and combining the two use cases into one so it does the update or commit logic depending on the function's arguments?

zhiyuang: sure, I could refactor it. But not that quick, I will find some time later today.

frontend/src/components/widgets/WidgetSpan.svelte
0HyperCube: `CommitTransaction` is a noop. `StartTransaction` clones the document and stores it in a list. `AbortTransaction` pops a document from the list and makes it current.

Keavon: That's interesting that `CommitTransaction` is a noop.  Is there a reason we have it? Do you think we should we just remove it, then find more intuitive names for the remaining two?

Keavon (approved):

frontend/src/components/widgets/WidgetSpan.svelte
0HyperCube: You're welcome to modify it.

Issue #971, opened 2023-01-15
Undo/redo last point extended using the Pen tool


Keavon:
Each point drawn with the pen tool while it performs its role of extending a vector shape's path should be integrated into the history so undo/redo can step through line extension steps.

auto-kad:
I'll work on this

Keavon:
@zhiyuang please leave a comment in this issue.

zhiyuang:
Assign to me 😃

PR #1587, opened 2024-01-26
Properly handle pen tool undo and redo

zhiyuang:
Closes https://github.com/GraphiteEditor/Graphite/issues/971

refer to Figma


0HyperCube:
!build

Keavon:
This is a phenomenal improvement that was much needed! I spent 30 seconds testing it but I'll do a closer look tomorrow. I should hopefully get to reviewing both of your PRs tomorrow (I've been rather consumed by some code I've been working on which has been where all my attention has gone this past week).

0HyperCube (approved):
Thanks for this great improvement - the code looks good.

From a UX perspective I'm not sure how we want to deal with undo after drawing a complete path and confirming. Inkscape deletes the whole path. Figma switches back to the pen tool and does undo point by point. This PR currently just does undo point by point without selecting the path tool.

Keavon:
I'm also unsure. I was slightly surprised when it happened while I was testing it since that's not what I'm used to, but when I thought about it, I realized that might actually be desired. I'd suggest we merge as-is and think about the UX more over time and take into consideration the pros/cons throughout its usage and any potential future user feedback.

PR #1588, opened 2024-01-27
Usability improvements to the NumberInput component

### LINKED ISSUE #313: Polish interactivity of the NumberInput widget

Nikita-str:
Part of #313
TODO: File an issue for these completed features

done (features from 7th to 14th inclusively from the issue):
* Up/Down arrows inc/dec by 1, with `Shift` by 10, with `Ctrl`(if value isn't Integer) by 0.1
* Holding mouse down on the arrow buttons change value multiple times
* `Ctrl+scroll` on hover to inc/dec value 
* Hover mouse over then `Ctrl`+{`C` \ `V` \ `/`(`value = 1/value`) \ `Backspace`(revert to default) \ `Enter` \ `-` (change sign of value)}. **MAYBE** not all of them should use `Ctrl`

Unit conversion(6th feature) and cleaning up floating point imprecision(15th feature) need to be discussed before doing them


0HyperCube:
!build

Keavon:
Here's an incomplete set of notes from my QA testing and code review. I'm not done. But I'm leaving them here. If you'd to take any of these on, feel free, or you can wait till I'm done (and I might fix some of them myself).

- Up/down while editing text: only do that if the text cursor is already at the left/right side
- Swap Ctrl/Shift modifiers for 0.1x and 10x to match with dragging NumberInputs fields.
- Ctrl+scroll over input, when text isn't being edited: update without entering and remaining in text editing mode
- Paste on hover over box: missing history state (and does this use Accel so it's Cmd on Mac?)
- We don't want Ctrl key modifier for `Backspace`, `-`, `/`, etc.

Keavon:
Thanks for contributing the effort that went into this. Ultimately it required more rework to be compatible with the subsequent alterations and avoidance of regressions which I didn't quite have time for, and it's gotten out of sync with the current state of things. But I appreciate your eagerness to help!

PR #1590, opened 2024-01-28
Bezier-rs: Use nonzero winding order for Poisson-disk insideness test

0HyperCube:
I see ~20ms performance improvement but it may just be random error.

### DIFF
--- a/libraries/bezier-rs/src/bezier/solvers.rs
+++ b/libraries/bezier-rs/src/bezier/solvers.rs
@@ -592,9 +592,14 @@ impl Bezier {
 	///
 	/// Cast a ray to the left and count intersections.
 	pub fn winding(&self, target_point: DVec2) -> i32 {
-		let extrema = self.get_extrema_t_list();
+		let [x_extrema_t, y_extrema_t] = self.unrestricted_local_extrema();
+		let x_extrema_t = x_extrema_t.map(|t| t.filter(|&t| t > 0. && t < 1.));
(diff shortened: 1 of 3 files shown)


Keavon:
I just did a performance profiling to compare the speed of the build link above and dev (so both running with release optimizations) and I'm seeing an average of 200ms on dev and 250ms on release with the same artwork where I'm dragging around an anchor point. This also fits with my observations that the framerate felt higher on dev before I ran the profiler. So unfortunately this seems to slow it down, not speed it up as was hoped. Which is odd, since just by glancing at the code it does look like you generally simplified things and I'd expect it to run faster. Do you think there are any ways to further optimize this? mI also wonder why my test with a ~50ms performance drop is in the opposite direction of your observed ~20ms perf improvement.

0HyperCube:
image
Dev is broken anyways.

Keavon:
That one point coming out on the left side of your image appears to be a "lucky" rare point that passes the two-direction check. I never managed to run into that during my tests but it's, of course, a possibility as you just found. I suppose it's now just a question of whether we want to trade performance for correctness (or, if we can figure out why it was faster on your machine but slower on mine and fix it for all, that'd be even better).

Keavon:
With your updates, it looks like my new test has this going from 700ms per frame (on master) down to 650ms per frame (in this branch), so your optimizations helped make this a net positive! And the switch to the nonzero fill rule is also appreciated since that lets us match the SVG default. Thanks for sticking with this and finding ways to make it even faster!

Keavon (approved):
💯

Issue #1550, opened 2024-01-07
Add ability to break a closed curve to be open


Keavon:
Ctrl+Delete, for example, could delete an anchor without reconnecting it. This would need to change segments so the new opening is the start and end points.

Keavon:
@zhiyuang please leave a comment in this issue.

zhiyuang:
Assign to me 😃

PR #1593, opened 2024-02-02
Support for deleting points to break path

zhiyuang:
Closes https://github.com/GraphiteEditor/Graphite/issues/1550

Only handle the closed path case and support only delete one anchor right now. 

Select a anchor, then Ctrl + Delete to break

### DIFF
--- a/editor/src/messages/input_mapper/default_mapping.rs
+++ b/editor/src/messages/input_mapper/default_mapping.rs
@@ -177,8 +177,8 @@ pub fn default_mapping() -> Mapping {
 		entry!(PointerMove; refresh_keys=[Alt, Shift, Control], action_dispatch=LineToolMessage::PointerMove { center: Alt, lock_angle: Control, snap_angle: Shift }),
 		//
 		// PathToolMessage
-		entry!(KeyDown(Delete); modifiers=[Control], action_dispatch=PathToolMessage::BreakPath),
-		entry!(KeyDown(Backspace); modifiers=[Control], action_dispatch=PathToolMessage::BreakPath),
+		entry!(KeyDown(Delete); modifiers=[Shift, Control], action_dispatch=PathToolMessage::BreakPath),
+		entry!(KeyDown(Backspace); modifiers=[Shift, Control], action_dispatch=PathToolMessage::BreakPath),
 		entry!(KeyDown(Lmb); action_dispatch=PathToolMessage::DragStart { add_to_selection: Shift }),
 		entry!(PointerMove; refresh_keys=[Alt, Shift], action_dispatch=PathToolMessage::PointerMove { alt: Alt, shift: Shift }),
 		entry!(KeyDown(Delete); action_dispatch=PathToolMessage::Delete),
--- a/editor/src/messages/input_mapper/default_mapping.rs
+++ b/editor/src/messages/input_mapper/default_mapping.rs
@@ -177,6 +177,8 @@ pub fn default_mapping() -> Mapping {
 		entry!(PointerMove; refresh_keys=[Alt, Shift, Control], action_dispatch=LineToolMessage::PointerMove { center: Alt, lock_angle: Control, snap_angle: Shift }),
 		//
 		// PathToolMessage
+		entry!(KeyDown(Delete); modifiers=[Control], action_dispatch=PathToolMessage::BreakPath),
+		entry!(KeyDown(Backspace); modifiers=[Control], action_dispatch=PathToolMessage::BreakPath),
 		entry!(KeyDown(Lmb); action_dispatch=PathToolMessage::DragStart { add_to_selection: Shift }),
 		entry!(PointerMove; refresh_keys=[Alt, Shift], action_dispatch=PathToolMessage::PointerMove { alt: Alt, shift: Shift }),
 		entry!(KeyDown(Delete); action_dispatch=PathToolMessage::Delete),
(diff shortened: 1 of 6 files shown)


Keavon:
I'm very glad to see this, thank you! I'll review it in detail ASAP, although I'm behind on my reviews because I've been focusing on some things with short-term deadlines. Most likely this weekend though I'll be able to do do the remaining reviews. Thanks for your patience and continued work on these while you're waiting on them to get merged, I promise the PRs won't get abandoned!

Keavon:
!build

zhiyuang:
thanks, take your time. I'm still diving into the codebase and contributing some codes during the process 😆

Keavon:
Just some quick comments about the behavior:

Eventually we'll have a menu, similar to Blender's <kbd>X</kbd> hotkey menu, that will let the user choose how to delete or dissolve the selection. But for now we'll rely on hotkeys. What I had in mind was that <kbd>Ctrl</kbd><kbd>Delete</kbd> would delete the selected point but also the adjacent segments. However the behavior you implemented, where it breaks it but keeps the adjacent segments, is still useful and I think assigning it to <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>Delete</kbd> would be the best, until we eventually get around to building the <kbd>X</kbd> hotkey menu (which will require some frontend work and refactoring from me to build context menus).

Also, when breaking these points, it'd be nice to have their other handle be removed (equivalent to selecting it and hitting <kbd>Delete</kbd>). So instead of this being what you see after breaking the point and separating them a bit:

image

I'd like to see this instead:

image

Lastly, it would be useful if it was also possible to break/delete an anchor that's part of an already open shape. Doing this would result in two separate subpaths.

Hopefully those suggestions would be easy enough to tweak. This is all looking good and is very useful, so thank you for doing it!

zhiyuang:
thanks for the detailed comments. I will try to change the hotkeys and delete handles first. 

For delete an anchor for open shape, I also thought about it. After deleting the anchor on an open curve, would we expect two break parts to be in one layer or two separate layers? I have tried in Inscape, they will create a new layer to draw the new break part.

Keavon:
For the moment I haven't made up my mind which is likely to be better. The idea of subpaths is that a layer can contain multiple of them. They're used, for example, to create a donut shape where there's an outer circle as one subpath and an inner circle as the other subpath (and since they have opposite winding directions, the inner one is drawn as a hole). It's probably easier for you to make them subpaths so I'd suggest you go with that option. This also would make it possible for people to actually create subpaths which isn't possible right now through any other means. But it splitting into two layers would also be acceptable. We may also want to eventually add a menu action that can split subpaths into individual layers. Or perhaps another to combine multiple selected layers into subpaths of one layer.

Keavon:
Please ping me when you've had the chance to complete each of the suggestions I made and I'll be ready to jump back into reviewing and merging this. Thanks!

zhiyuang:
- [x] Key mapping change to `Ctrl` `Shift` `Delete`
- [x] Remove break point's one-side handle
- [ ] Break an open path

I had a initial try to handle breaking an open path yesterday, I chose to break them as subpaths in a layer. But met a weired issue that both subpaths are always moved together. I will spend more time today to see how to solve this and do some tests. Will let you know once done.

Keavon:
Could you post a video of how both subpaths were being moved together?

zhiyuang:
https://github.com/GraphiteEditor/Graphite/assets/32867472/997b4aac-4adc-498d-a1cc-ab674dc7b4b3

Keavon:
Ah, yes, gotcha. Pay close attention to which anchor points are selected (when they're filled with blue), we have it so when you click on a shape with the Path tool, it becomes fully selected by default so it's easily draggable. But now that we are actually using multiple subpaths within a layer, we would like to update that behavior so it only selects all the anchors of the subpath you click on, not all anchors from all subpaths. However, if you prefer to work on that in a separate PR, that could be deferred (the choice is yours).

Keavon:
There is one more bullet point to add to your list, by the way, if you don't mind implementing it too:
- [ ] <kbd>Ctrl</kbd><kbd>Delete</kbd> should delete the anchor and both its adjacent segments

Thanks :)

zhiyuang:
Oh, you give good hints on the issue. I could spend more time on this PR. And add a new to-do in the list 😄  .

- [x] Key mapping change to `Ctrl` `Shift` `Delete`
- [x] Remove break point's one-side handle
- [ ] Break an open path (in progress)
- [ ] `Ctrl` `Shift` should delete the anchor and both its adjacent segments

zhiyuang:
@Keavon do we have an existing function to check if the mouse click on a subpath?

Keavon:
I'll have to ask @0HyperCube if he knows, I'm not sure.

0HyperCube:
> @Keavon do we have an existing function to check if the mouse click on a subpath?

@zhiyuang for the select tool, we use `ClickTarget::intersect_rectangle`.

zhiyuang:
- [x] Key mapping change to `Ctrl` `Shift` `Delete`
- [x] Remove break point's one-side handle
- [x] Break an open path (support break at multi points simultaneously)
- [x] `Ctrl` `Shift` should delete the anchor and both its adjacent segments (support break at multi points simultaneously)

@Keavon Break path related to-dos should have been finished, you could have a code review when you have time. Subpath selection issue I could open another PR to fix, let this PR stays on path break related changes.

Keavon:
!build

zhiyuang:
This is a case has been broken after I refactor some code. Have been fixed.

zhiyuang:
And I find Ctrl+Delete also has this issue

PR #1596, opened 2024-02-04
Fix floating point error with Poisson-disk points being sampled outside shape

0HyperCube:
Fix floating point error in fish sampling.


Keavon (approved):
Thanks for helping debug this with me!

PR #1597, opened 2024-02-05
Fix Select tool's broken undo

zhiyuang:
As a follow up of https://github.com/GraphiteEditor/Graphite/pull/1587. I find it affects normal select tool undo behavior.


Keavon:
!build

PR #1598, opened 2024-02-08
DropdownInput preview support and ColorButton history improvements

zhiyuang:
As follow ups of https://github.com/GraphiteEditor/Graphite/pull/1584#issuecomment-1926521666


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -894,7 +894,7 @@ pub fn black_and_white_properties(document_node: &DocumentNode, node_id: NodeId,
 pub fn blend_properties(document_node: &DocumentNode, node_id: NodeId, _context: &mut NodePropertiesContext) -> Vec<LayoutGroup> {
 	let backdrop = color_widget(document_node, node_id, 1, "Backdrop", ColorButton::default(), true);
 	let blend_mode = blend_mode(document_node, node_id, 2, "Blend Mode", true);
-	let opacity = number_widget(document_node, node_id, 3, "Opacity", NumberInput::default().mode_range().min(0.).max(100.).unit("%"), true);
+	let opacity = number_widget(document_node, node_id, 3, "Opac", NumberInput::default().mode_range().min(0.).max(100.).unit("%"), true);
```
Keavon, line 897: What's the reason for this change? (And below on line 2268.)

frontend/src/components/widgets/WidgetSpan.svelte
```diff
@@ -66,9 +66,15 @@
 	}
 
 	function widgetValueCommitAndUpdate(index: number, value: unknown) {
+		console.log("widget value commit and udpdate");
```
Keavon, line 69: Please remove all the logs before merging.

zhiyuang:
sry,  it's still work in progress, not ready even for reviewing. I should open this as a draft pr...

Keavon:
Ah, yes, good to open PRs early but just please mark them as a draft. I usually don't poke around draft PRs. But just make sure to ping me once you're ready for review, and switch it to non-draft. Thanks :)

zhiyuang:
@Keavon Actually, I opened this pr actually would like to confirm two user experience behaviors. 

> Also, same for TextInput and TextAreaInput that should send updates as the user types but only commits when that currently happens.

Does it mean user type in TextInput in right panel, would like the text change according to in the canvas as user typing.(Right now it only changes when the TextInput is unfocus.)

> So if you edit a color, once you release the mouse after dragging any of the sliders, it should send a commit at that time.

Does it means these sliders? (Right now these sliders do not save any history steps separately, only save a history step after close Color Panel)
image

Keavon:
> Does it mean user type in TextInput in right panel, would like the text change according to in the canvas as user typing.(Right now it only changes when the TextInput is unfocus.)

If I'm interpreting your question correctly, I think you're asking if text created with the Text tool should be re-rendered as the user types in the Properties panel? My answer to that is: maybe, but only if the performance isn't too bad. There is also #1589 currently open so let's avoid implementing that right now in case it causes conflicts. I'm mostly just interested in having the text widget send its real-time text but letting the user of the text, like the Properties panel for the Text node, decide if it wants to use the real time version or the committed version (which is what is used right now).

> Does it means these sliders? (Right now these sliders do not save any history steps separately, only save a history step after close Color Panel)

image

- A, B, C: Should update when the user is finished dragging
- D: Should update when the user commits a new typed value by hitting <kbd>Enter</kbd>, clicking out of the box, or closing the entire color picker popover by moving the mouse far away from it
- E: Since this can be dragged or typed in, the previous two bullet points apply to this
- F: Should update immediately upon being clicked or picking a color with the eyedropper

All of the above should send the history update as described, but duplicates should be avoided if the user committed a change that resulted in a value that's the same as when they started to edit that value.

zhiyuang:
@Keavon could have a code review when you have time.

This PR is mainly about Dropdown preview ability and Color Button related history actions (All actions in below image). 
image

For TextInput widget, I haven't handled in this PR.  I found unless  we really need send updates on every typing by the users, current text widget has satisfied enough right now. Maybe we could return to this if we meet some concrete issues in future 🤔

editor/src/messages/layout/utility_types/widgets/input_widgets.rs
```diff
@@ -67,6 +67,10 @@ pub struct DropdownInput {
 
 	pub tooltip: String,
 
+	#[derivative(Default(value = "false"))]
+	#[serde(rename = "previewOnHover")]
+	pub preview_on_hover: bool,
```
Keavon, line 72: I'm pretty sure this shouldn't be necessary. It can work just like your previous PR where NumberInput was set to send both a preview value and a committed value. Then the backend code which wants this widget to send preview values would just handle those temporary values directly from the backend, whereas those which only care about the final committed value do nothing with the preview values. The approach I'm suggesting here would allow this widget to have a more standardized API alongside NumberInput widgets and others, and would be generally cleaner.

frontend/wasm/src/editor_api.rs
```diff
@@ -285,6 +285,13 @@ impl JsEditorHandle {
 		Ok(())
 	}
 
+	/// Update the value of a given UI widget, and commit it to the history
+	#[wasm_bindgen(js_name = widgetValueRevert)]
+	pub fn widget_value_revert(&self) -> Result<(), JsValue> {
+		self.dispatch(DocumentMessage::Undo);
```
Keavon, line 291: This makes me nervous. But it shouldn't be necessary anymore once you make the change I suggested above.

Keavon:
I'm marking this as a draft while awaiting the requested changes. Please un-mark as draft and also ping me when ready for a second review. Thanks!

Keavon:
Hi, I'm just checking up on how things are going with this feedback round. Thanks :)

zhiyuang:
sry for the late, a little busy these two weeks. I'll look forward to see if I could find some time next week.

Keavon:
Sure thing, thank you for the update.

zhiyuang:
!build

Keavon:
!build

frontend/src/components/widgets/inputs/DropdownInput.svelte
```diff
@@ -24,11 +24,21 @@
...
 	$: watchSelectedIndex(selectedIndex);
 	$: watchActiveEntry(activeEntry);
+	$: watchOpen(open);
+
+	function watchOpen(open: boolean) {
+		if (open) {
+			initialSelectedIndex = selectedIndex;
+		} else {
+			initialSelectedIndex = undefined;
+		}
```
Keavon, line 40:
```suggestion
		initialSelectedIndex = open ? selectedIndex : undefined;
```

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1026,6 +1025,7 @@ impl DocumentMessageHandler {
 
 	/// Copies the entire document into the history system
 	pub fn backup(&mut self, responses: &mut VecDeque<Message>) {
+		debug!("start transaction");
```
Keavon, line 1028:
```suggestion
```

Keavon:
!build

Keavon:
I rebased this for you, and now strangely I'm getting fatal console errors when I hover over the dropdown menu entries. Can you please help investigate? I committed some debug instrumentation which points out that it's getting a value of -1 where it should be able to fit as a `u64`, which is weird since it also gets sent from the JS code as a number that's 0 or greater. Maybe you can help figure out the issue? Thanks.

zhiyuang:
it's related to these codes, if I comment these out, it could work. https://github.com/GraphiteEditor/Graphite/pull/1598/commits/d6e357f5bd0d37103034ec7f1043f977fd815621

zhiyuang:
but I don't know how fix this yet. I'm not familiar with MenuList. It should be related to this PR: https://github.com/GraphiteEditor/Graphite/pull/1499

Keavon:
Ah, that's a good hint, thank you for investigating and narrowing it down to that! If you'd like me to take it from here, I can do that in the next couple days. Or if you'd like to keep digging and find the solution, please feel free to do that as well.

frontend/src/components/floating-menus/MenuList.svelte
```diff
@@ -346,9 +352,9 @@
 		highlighted = newHighlight;
 
 		// Interactive menus should keep the active entry the same as the highlighted one
-		if (interactive && newHighlight?.value !== activeEntry?.value && newHighlight) {
-			dispatch("activeEntry", newHighlight);
-		}
+		// if (interactive && newHighlight?.value !== activeEntry?.value && newHighlight) {
+		// 	dispatch("activeEntry", newHighlight);
+		// }
```
zhiyuang, line 357: for the preview case for blend mode. The activeEntry actually is the one being hovered, not the one being highlighted. So in this case it could be different. Maybe @0HyperCube could help have a look here.

zhiyuang:
@Keavon updated, pls check

Keavon (approved):

Keavon:
Pardon the very long delay, this required rebuilding some mental context and I was quite busy with things unrelated to this. Thank you for helping make this happen, and your patience in my finally returning to merge it.

PR #1599, opened 2024-02-09
Add shaking a dragged node free from its link

### LINKED ISSUE #904: Shake a node to un-wire its connection through its primary input/output
We need some kind of robust shake detection algorithm in JS that has very minimal false positives/negatives.

pwbh:
I have added a simple shake detection. Resolves #904.

~~It seems to work well enough, although I can change it to work against some time delta which will check the differences between values in each "aggressive" movement back and forth, it probably will be more accurate, but please test this see how it feels first :)~~


0HyperCube:
!build

pwbh:
Actually I am seeing that fast movement to left or right is triggering the shake check, I will improve it to use deltas between movement left and right and only trigger if the movement has been both to left and right in interval of some threshold, this should produce more accurate results.

Keavon:
Yes, following what I mentioned in my comment in the issue, you need to take a much more sophisticated approach specifically designed to actually detect a user's intentional pointer shake over a series of samples spanning time, probably about half a second. That means recording every sample over that time in a running queue (discarding any that are older than the sample) and performing analysis on them each frame. I suggested a metric which optimizes for detecting (beyond some threshold) a high ratio of direction change over displacement, meaning the user is keeping the mouse in a relatively similar position but shaking a lot.

pwbh:
@Keavon I have updated the implementation, feels much more accurate 😄

Keavon:
!build

Keavon:
I gave it a quick test and it seems to have a very high rate of false positives. I also gave a quick glance at the code and I don't see a division sign anywhere, meaning I'm not sure if you took my suggestion to use an algorithm based on ratio of direction change over average displacement. Can you please carefully re-read my suggested algorithm and precisely implement that? (I also noticed that you seem to be using only the X axis, which I don't think is robust compared to considering both axes together. Direction change can be measured as the cross product of the vectors of mouse deltas in neighboring samples.)

pwbh:
> I gave it a quick test and it seems to have a very high rate of false positives. I also gave a quick glance at the code and I don't see a division sign anywhere, meaning I'm not sure if you took my suggestion to use an algorithm based on ratio of direction change over average displacement. Can you please carefully re-read my suggested algorithm and precisely implement that? (I also noticed that you seem to be using only the X axis, which I don't think is robust compared to considering both axes together. Direction change can be measured as the cross product of the vectors of mouse deltas in neighboring samples.)

It returns a lot of false positives because of the alert message, it felt better when I tested it against trigger with `console.log` instead, but I'll tweak it to reflect your comment anyway, thanks.

pwbh:
@Keavon just an FYI, but before going with your suggestions I'd like to try to tweak this implementation tomorrow morning, I just realized its missing a few things (for example I am not checking whether a movement occurred in both directions, and not only in one before checking for difference in captured positions), I'll test it again and if those changes won't achieve the desired effect, I'll reimplement it by your suggestion.

Keavon:
!build

Keavon:
!build

Keavon:
I'll look more closely at the code for shake detection, but it behaves reasonably well at the moment. Please proceed with implementing the actual shake removal now which should call into the backend (and I think that will require changes to the backend to implement the actual removal, you can reference the deletion code which does the job of reconnecting the link.)

Keavon:
Checking on your status.

pwbh:
hey hey hey, will try to finish it up in the coming week, sorry for delay was working on cpuple of personal stuff.

PR #1618, opened 2024-02-18
Use f64 not f32 throughout graph; add Vector2 value node

0HyperCube:
- Using doubles in graph.
- Vector2 value node as suggested by @Ezbaze.


Keavon (approved):
💯

PR #1619, opened 2024-02-18
Fix multiple dragged nodes snapping back sometimes

0HyperCube:
Fix dragging a selection of nodes on one axis (in screen pixels) resetting the selection as reported by @Ezbaze and @Keavon.


Keavon (approved):
Great job catching this really subtle cause of the symptom.

PR #1632, opened 2024-02-28
Add Mirror node

### LINKED ISSUE #1453: Add the Mirror node
This node should take in vector data, a boolean for "Mirror X", and a boolean for "Mirror Y". The centre of transformation should probably be around the pivot (indicated in the document with a blue circle inside another blue circle). 
pivot
If both mirror x and y are active, 4 copies of the vector data will be output. The default values should be true for mirror x and false for mirror y. This is similar to the mirror node in blender.

A guide to the node system can be found https://github.com/GraphiteEditor/Graphite/blob/master/node-graph/README.md - if you have any questions please ask in the discord chat.

Steps:
- Define a new node that just outputs the vector data that has been input (to ensure nodes are set up properly). This is similar to the circular repeat node.
- Compute the pivot position of the vector data using the appropriate function. You can log this to ensure that it is correct.
- Clone the subpaths in the vector data and mirror all of the positions around the computed pivot. You may need to modify the bézier_rs library and the VectorData struct.

skoriop:
This will add the Mirror node:
- [x] Define a new node that just outputs the vector data that has been input (to ensure nodes are set up properly). 
- [x] Compute the pivot position of the vector data using the appropriate function.
- [x] Clone the subpaths in the vector data and mirror all of the positions around the computed pivot.

This will close #1453 when finished.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/document_node_types.rs
@@ -2813,6 +2813,18 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 			properties: node_properties::image_color_palette,
 			..Default::default()
 		},
+		DocumentNodeDefinition {
+			name: "Mirror",
+			category: "Vector",
+			inputs: vec![
+				DocumentInputType::value("Vector Data", TaggedValue::VectorData(graphene_core::vector::VectorData::empty()), true),
+				DocumentInputType::value("Mirror X", TaggedValue::Bool(true), false),
+				DocumentInputType::value("Mirror Y", TaggedValue::Bool(false), false),
+			],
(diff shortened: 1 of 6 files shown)


Keavon:
Your code looks nice! However I need to get @TrueDoctor's comment about the addition of the `pivot` field, because this addition may likely conflict with his ongoing work to revamp the transform system. I'm not sure how to proceed with that so I'll wait for him to reply here.

Keavon:
Blocked on #1558. We'll have to let this wait a bit. Hopefully @TrueDoctor can complete that issue in the next week or two.

Keavon:
I think I'll close this for now. Thanks for your efforts, and I'm sure you learned a lot from this that'll be helpful knowledge as you work on the animation project this summer. But with the awkwardness of how transforms are currently (not) dealt with, it's rather hard to approach this at the moment. And there are lots of merge conflicts so closing for the moment is the easiest way to proceed. Thanks though!

PR #1636, opened 2024-02-29
Change devcontainer image to Debian to support ARM platforms

haikalvidya:
This pull request aims to facilitate running the development container on ARM-based processors by switching the base image to Debian. Previously, the image was specified as "mcr.microsoft.com/devcontainers/universal:2", which might not be compatible with ARM architectures based on this information (https://github.com/devcontainers/images/blob/main/src/universal/README.md). 

The decision to opt for the Debian base image is grounded in the documentation of the Graphite, which suggests compatibility with Debian distributions. Although there are two options available (base:ubuntu and base:debian), choosing Debian aligns better with the documented requirements, ensuring a smoother development experience. And I've confirmed it runs fine on the Github Codespace and of course ARM-based machine.

more information about the images : https://github.com/devcontainers/images/blob/main/src/base-debian/README.md

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,5 +1,5 @@
 {
-	"image": "mcr.microsoft.com/devcontainers/universal:2",
+	"image": "mcr.microsoft.com/devcontainers/base:debian",
 	"features": {
 		"ghcr.io/devcontainers/features/rust:1": {
 			"profile": "default"


Keavon:
Thanks for researching this!

> The decision to opt for the Debian base image is grounded in the documentation of the Graphite, which suggests compatibility with Debian distributions.

I'm curious if you could clarify what you mean by this? What Graphite documentation mentions Debian?

haikalvidya:
> Thanks for researching this!
> 
> > The decision to opt for the Debian base image is grounded in the documentation of the Graphite, which suggests compatibility with Debian distributions.
> 
> I'm curious if you could clarify what you mean by this? What Graphite documentation mentions Debian?

ooohh, this one https://graphite.rs/volunteer/guide/getting-started/ mention ```On Debian-based Linux distributions, you may need to install the following packages:```

Keavon:
Gotcha, thanks for explaining. That mention in the getting started docs just means any Debian-based Linux distro (where most commonly that means Ubuntu), there's no particular preference for the base Debian distro or anything. However your claim about Debian being lighter than Ubuntu seems like a valid point in favor of the former. But that issue you found on Codespaces 2-core machines makes me wonder, does that happen more with the Debian image compared to the (potentially even lighter) image we're using currently which was recommended by the Codespaces docs?

Keavon:
The relevant issue: https://github.com/devcontainers/images/issues/545

haikalvidya:
The image(photo) that I send is from master codebase of Graphite, so the image was the one that currently on codebase right now. And about that relevant issue on image `mcr.microsoft.com/devcontainers/universal:2` install [oryx](https://github.com/microsoft/Oryx) that not support arm-based machine, and I think on Graphite development we dont need that.

Keavon:
@TrueDoctor @0HyperCube do you have any thoughts about which OS to use for this? Does Debian seem like a good choice compared to `universal`?

Keavon (approved):

PR #1637, opened 2024-03-02
Fix layers losing their names upon being grouped

### LINKED ISSUE #1644: Layers lose their names when grouped

moOsama76:
Make layers reserve their visible names when grouped together (CTRL+G)

Closes #1644


editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -419,7 +420,13 @@ impl MessageHandler<DocumentMessage, DocumentInputs<'_>> for DocumentMessageHand
 
 					let id = NodeId(generate_uuid());
 					let insert_index = -1;
-					responses.add(GraphOperationMessage::NewCustomLayer { id, nodes, parent, insert_index });
+					responses.add(GraphOperationMessage::NewCustomLayer {
+						id,
+						nodes,
+						parent,
+						insert_index,
+						alias: String::from(""),
```
TrueDoctor, line 428: You can use `String::new()` instead, which does not allocate memory

moOsama76: Sure

moOsama76: done

0HyperCube: @TrueDoctor I'm pretty sure `String::from("")` won't allocate - it looks like it just transmutates the string into a byte slice and then converts it to a Vec which will not allocate if the size is 0.

0HyperCube (approved):
Works well - thanks for this fix.

Keavon:
@moOsama76 please resolve the merge conflict (and in the future, it'll be easier if you make sure you are always rebasing your ongoing work with master on a frequent basis to reduce the conflicts that arise at the end). Thanks!

Keavon:
Actually I went ahead and did it for you since it was faster.

editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
```diff
@@ -689,6 +695,8 @@ impl MessageHandler<GraphOperationMessage, GraphOperationHandlerData<'_>> for Gr
 						})
 						.unwrap_or_default();
 
+					modify_inputs.document_network.nodes.get_mut(&id).unwrap().alias = alias.clone();
```
Keavon, line 698: Please avoid this unwrap, perhaps using `if let`.

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -180,10 +180,10 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
 						};
 
 						let node = layer.to_node();
+						let previous_alias = active_document.network().nodes.get(&node).unwrap().alias.clone();
```
Keavon, line 183: Please avoid this unwrap, perhaps defaulting to an empty string.

editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
Keavon: I made this change, but please look at the code to learn from the updated code for your future reference.

editor/src/messages/portfolio/portfolio_message_handler.rs
Keavon: I made this change, but please look at the code to learn from the updated code for your future reference.

Keavon:
!build

moOsama76:
Thank you, I was going to fix but my partition get full and I had some cache problems

editor/src/messages/portfolio/document/document_message_handler.rs
TrueDoctor: @0HyperCube  In debug mode it allocates, when compiling with optimizations the same assembly is produced so I guess it is just good practice to use `String::new()` to guarantee it won't allocate + speed up debug builds but will most likely not make difference in production: https://godbolt.org/z/9r5o7jrs4

PR #1638, opened 2024-03-02
Implement Path tool aborting with Esc/RMB

devsargam:
# Fixes:
Screenshot 2024-03-02 at 11 07 06


Keavon:
!build

Keavon:
Sorry to say, this does not work. Sending an abort message to the tool is used for telling a tool that its entire purpose is not needed anymore (like when switching away from a tool, to another tool). You will need to actually implement this as functionality on the tool. There are plenty of examples in other tools which do this which you can reference which should be of help. Most importantly though, please be sure to actually test your code in practice (since if you test this one, you will see that it does nothing). Good luck :)

devsargam:
Extremely sorry for raising pr without testing

Keavon:
I'm going to temporarily close this. Please reopen it when you're ready.

PR #1640, opened 2024-03-02
Update Polygon tool to draw stars/ngons so they're tilted upright

haikalvidya:
I found out that polygon and star polygon is not facing upwards, which is not inline from the icon in the toolbar that the polygons are facing upwards, so user expected that polygon would be facing it upward, but when I use it, it is facing sideways.

image

So i update the angel of polygon on this PR.


Keavon:
!build

Keavon (approved):

Keavon:
Thanks for this! Very helpful.

PR #1643, opened 2024-03-02
Add clear artboards button to undo history

### LINKED ISSUE #1641: The action "Clear artboards" is not counted when undoing
I discovered that the action that came before "Clear Artboards" also undoes when "Clear Artboards" is completed and then undone.

https://github.com/GraphiteEditor/Graphite/assets/76188139/4169e633-a4d5-42c2-9fda-f3affd1ee0e4

haikalvidya:
This will fix action "Clear artboards" is not counted when undoing

Closes #1641


0HyperCube (approved):
Works well - thanks for this fix.

Issue #1633, opened 2024-02-28
Avoid layers being reordered when dealing with groups


Keavon:
- Grouping (with <kbd>Ctrl</kbd><kbd>G</kbd>):
    - [x] #1627:
        **External consistency:** Grouping the selected layers causes the group to be inserted at the top of the layer stack. It should go just above where the layer selected layers were located.
        The current incorrect behavior:
        | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/fa51a964-ee79-43f4-8e0e-c75f17f1fcf7) | -> <kbd>Ctrl</kbd><kbd>G</kbd> -> | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/6228cc38-78f6-4c57-acb8-772dcece6b13) |
        | - | - | - |
    - [x] #1645
        **Internal consistency:** Grouping layers should also preserve the order that they were sorted once those layers end up in the new group
        The current incorrect behavior:
        | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/378dae38-f131-45e7-83c6-4aa3bdebac2e) |  -> <kbd>Ctrl</kbd><kbd>G</kbd> -> | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/2edbfdd0-5223-4149-876b-cafa34a9a650) |
        | - | - | - |
- Ungrouping (with <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>G</kbd>):
    - [ ] #1672
        **External consistency:** Ungrouping should also preserve order by placing the content in the hierarchy where the group used to be.
        The current incorrect behavior:
        | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/30beb10b-b252-4111-a8ac-a3054430c54b) | -> <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>G</kbd> -> | ![image](https://github.com/GraphiteEditor/Graphite/assets/4388688/b6b38eaf-4c03-45dd-9204-d79bcd45a9a4) |
        | - | - | - |
    - [x] #1645
        **Internal consistency:** While ungrouping, the content in the group itself should also be have its order preserved amongst its grouped siblings.
        The current incorrect behavior:
[...]

PR #1645, opened 2024-03-03
Fix internal consistency of layer ordering when grouping

moOsama76:
Grouping layers preserve the order that they were sorted once those layers end up in the new group

Part of #1633

### DIFF
--- a/editor/src/messages/portfolio/portfolio_message_handler.rs
+++ b/editor/src/messages/portfolio/portfolio_message_handler.rs
@@ -170,23 +170,11 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
 						.metadata()
 						.shallowest_unique_layers(active_document.selected_nodes.selected_layers(active_document.metadata()));
 
-					let mut get_last_elements: Vec<_> = binding.iter().map(|x| x.last().expect("empty path")).collect();
+					let get_last_elements: Vec<_> = binding.iter().map(|x| x.last().expect("empty path")).collect();
 
-					let selected_to_vector = active_document
-						.selected_nodes
-						.selected_layers(active_document.metadata())
-						.filter(|_x| true)
-						.map(|x| x.clone())
-						.collect::<Vec<_>>();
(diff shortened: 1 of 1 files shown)


0HyperCube:
!build

Keavon:
!build

moOsama76:
just leaving a comment

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -171,14 +171,27 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
...
-							continue;
-						};
+						.shallowest_unique_layers(active_document.selected_nodes.selected_layers(active_document.metadata()));
+
+					let mut get_last_elements: Vec<_> = binding.iter().map(|x| x.last().expect("empty path")).collect();
+
+					let selected_to_vector = active_document
+						.selected_nodes
+						.selected_layers(active_document.metadata())
+						.filter(|_x| true)
```
0HyperCube, line 183: Redundant filter

editor/src/messages/portfolio/portfolio_message_handler.rs
```diff
@@ -171,14 +171,27 @@ impl MessageHandler<PortfolioMessage, (&InputPreprocessorMessageHandler, &Prefer
...
+						.map(|x| x.clone())
+						.collect::<Vec<_>>();
```
0HyperCube, line 185: I don't see why it is necessary to collect this and reallocate then sort the other list such that it is in the same order as this one. A more efficient method would probably be just to do something like`active_document.metadata.all_layers().filter(|layer| get_last_elements.contains(&layer)).collect()` and then not do sorting.

moOsama76: Yes I understand

0HyperCube (approved):

Keavon (approved):

moOsama76:
this also fixes the Internal consistency of ungrouping (a sub-issue of #1633 )

PR #1648, opened 2024-03-03
Add corner rounding to the Rectangle node

### LINKED ISSUE #1642: Support rounded corners with the Rectangle node
Currently it is possible to create a rectangle by inserting a rectangle node in the node graph ("Node Graph" button in top right corner then right click in graph and select "Rectangle"). However it is not possible to make the corners round.

An option for the border radius should be added to the properties (on the right when the rectangle node is selected) that allows for creating a rounded rectangle. Note that it is not possible to create a perfect arc using bézier curves, so an approximation will need to be used (like for the circle currently).

Here is the a screenshot of the current rectangle node for reference:
Current rectangle node

skoriop:
This PR will add a border radius property that will allow the creation of rounded rectangles.

Closes #1642


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler/node_properties.rs
```diff
@@ -1508,7 +1508,7 @@ pub fn rectangle_properties(document_node: &DocumentNode, node_id: NodeId, _cont
 
 		LayoutGroup::Row { widgets }
 	};
-	vec![operand("Size X", 1), operand("Size Y", 2)]
+	vec![operand("Size X", 1), operand("Size Y", 2), operand("Border Radius", 3)]
```
Keavon, line 1511: Please name this Corner Radius.

Keavon:
!build

Keavon:
Looks good. Can you also please add a checkbox parameter, that's enabled by default, called "Clamped" which prevents going below 0 and above half the adjacent dimension(s)? So a 50x100 rectangle, if clamped, would have a corner with a radius of 25x50 (similar to the behavior of CSS).

Keavon:
And could you please add the all-corners/individual-corners toggle between setting a scalar and vector4 value? Thanks.

Keavon:
I'll mark this as a draft until the per-corner feature is ready. Please let us know on Discord if you need any help making the generic inputs (scalar or vec4) work. If you prefer, instead, to merge this as-is and add the per-corner feature in a later PR, that's also acceptable (just mention your preference here and we can merge it sooner). Please ping me once you're ready for review again.

otdavies:
!build

otdavies:
Crash found, when you have clamp enabled negative scale causes the editor to crash: 
chrome_Hm06xgHHHs

skoriop:
> Can you please add the logic so, when switching from Uniform to Individual, it sets the four values to the value that existed; and when switching from Individual to Uniform, it sets the single value to the first of the four values?

Done, along with fixing the negative scale crash

Keavon:
!build

Keavon:
One last little issue, if you switch back from Individual to Uniform, it does now display the first item as requested but it doesn't actually apply it. You have to click that number and press enter to apply it.

Keavon:
Oh, and more more issue with Clamped mode, sorry. If I set a single corner's radius to something big but the others are 0 or something small, the big rounded corner should extend all the way to the edge, not only 50%.

image


Expectation:

image

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -32,18 +34,34 @@ fn ellipse_generator(_input: (), radius_x: f64, radius_y: f64) -> VectorData {
...
 	size_x: SizeX,
 	size_y: SizeY,
+	is_individual: IsIndividual,
+	corner_radius: CornerRadius,
+	clamped: Clamped,
 }
 
 #[node_macro::node_fn(RectangleGenerator)]
-fn square_generator(_input: (), size_x: f64, size_y: f64) -> VectorData {
+fn square_generator<T: Any>(_input: (), size_x: f64, size_y: f64, is_individual: bool, corner_radius: T, clamped: bool) -> VectorData {
```
0HyperCube, line 46: Could I recommend something like:
```rs
trait CornerRadius {
	fn generate(self, size: DVec2, clamped: bool) -> super::VectorData;
}
impl CornerRadius for f64 {
	fn generate(self, size: DVec2, clamped: bool) -> super::VectorData {
		let clamped_radius = if clamped { self.clamp(0., size.x.min(size.y).max(0.) / 2.) } else { self };
		super::VectorData::from_subpaths(vec![Subpath::new_rounded_rect(size / -2., size / 2., [clamped_radius; 4])])
	}
}
impl CornerRadius for [f64; 4] {
	fn generate(self, size: DVec2, clamped: bool) -> super::VectorData {
		let clamped_radius = if clamped {
			self.map(|x| x.clamp(0., size.x.min(size.y).max(0.) / 2.))
		} else {
			self
		};
		super::VectorData::from_subpaths(vec![Subpath::new_rounded_rect(size / -2., size / 2., clamped_radius)])
	}
}
```
Avoiding reliance on the Any trait. Then in the implementation you could just call `corner_radius.generate(DVec2::new(size_x, size_y), clamped)`.

editor/src/messages/portfolio/document/node_graph/document_node_types.rs
```diff
@@ -2253,6 +2260,9 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 				DocumentInputType::none(),
 				DocumentInputType::value("Size X", TaggedValue::F64(100.), false),
 				DocumentInputType::value("Size Y", TaggedValue::F64(100.), false),
+				DocumentInputType::value("Individual Corner Rounding", TaggedValue::Bool(false), false),
+				DocumentInputType::value("Corner Radius", TaggedValue::F64(0.), false),
```
0HyperCube, line 2264: It is somewhat confusing to have the state encoded both in the type of the tagged value (either `TaggedValue::F64(_)` or `TaggedValue::F64Array4(_)`) and in this bool `"Individual Corner Rounding"`. Perhaps, in order to reduce chance of data inconsistency, you could remove this redundancy.

0HyperCube (changes requested):

editor/src/messages/portfolio/document/node_graph/document_node_types.rs
TrueDoctor: A common way of doing this would be to use an enum:
```rust
enum CornerRounding {
   Individual([f64; 4),
   Uniform(f64),
)
```

0HyperCube: I don't think that is necessary as the typing is already stored in the TaggedValue enum @TrueDoctor. Introducing a new type adds complexity if we want the input to be exposed to the graph as we would have to add a convert node for the f64.

Keavon: +1 to @0HyperCube's reasoning.

node-graph/gcore/src/vector/generator_nodes.rs
skoriop: Thanks, that seems much cleaner

skoriop:
> One last little issue, if you switch back from Individual to Uniform, it does now display the first item as requested but it doesn't actually apply it. You have to click that number and press enter to apply it.

I've added a `Message::Batched` variant (as suggested by @TrueDoctor on Discord) and modified the `on_update` callback function of the radio input to ensure that it also updates the corner radius input.


> Oh, and more more issue with Clamped mode, sorry. If I set a single corner's radius to something big but the others are 0 or something small, the big rounded corner should extend all the way to the edge, not only 50%.

For this, I modified the clamping code for the individual rounding. It first finds the largest corner radius, clamps the two corner radii adjacent to it and finally clamps the last corner radius (which is diagonally opposite to the largest corner radius).

Keavon:
!build

Keavon:
> For this, I modified the clamping code for the individual rounding. It first finds the largest corner radius, clamps the two corner radii adjacent to it and finally clamps the last corner radius (which is diagonally opposite to the largest corner radius).

Seems like a decent algorithm you devised, although I think it's more important to remain consistent with user expectations from CSS (especially if Graphite gets used as a tool for web design mockups in the future). Could you please update it to use this algorithm?:

https://drafts.csswg.org/css-backgrounds/#corner-overlap

After that, we're all good to merge this!

skoriop:
Yup, I've updated that part.

Keavon:
!build

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -52,24 +52,16 @@ impl CornerRadius for f64 {
...
-			(0..4).for_each(|i| {
-				if radius[i] > max {
-					(max, index) = (radius[i], i);
+			// https://drafts.csswg.org/css-backgrounds/#corner-overlap
+			let mut f: f64 = 1.;
+			for i in 0..4 {
+				let l = if i % 2 == 0 { size.x } else { size.y };
+				let s = self[i] + self[(i + 1) % 4];
+				if l < s {
+					f = f.min(l / s);
```
Keavon, line 61: Please use descriptive variable names, not letters. (Except `i` which is the standard iterator variable.)

skoriop: Sure!

skoriop: Cool, I've renamed that to `adjacent_corner_radius_sum` now.

Keavon (approved):

PR #1651, opened 2024-03-05
Fix artboard deletion

### LINKED ISSUE #1505: Document > Clear Artboards should not delete its children
Instead of deleting its children, it should ungroup those children from the artboard and then delete just the artboard, leaving the artwork behind. The code for ungrouping with Ctrl+Shift+G can probably be referenced to achieve a similar outcome here.

We also want the same for deleting an artboard from the Artboard tool, where the artwork remains behind.

(However, when deleting from the Layers panel, we do want the current behavior of deleting its children since that's also how deleting folders in the Layers panel works, so no change there.)

A continuation of #1502

skoriop:
This PR ensures that:
- [x] `Document > Clear Artboards` does not delete its children
- [x] `Artboard Tool > Delete` does not delete its children

Closes #1505.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
@@ -523,6 +523,42 @@ impl<'a> ModifyInputsContext<'a> {
 		let new_input = if is_artboard_layer { node.inputs[0].clone() } else { node.inputs[1].clone() };
 		let deleted_position = node.metadata.position;
 
+		let new_input_artboard_layer = node.inputs[1].clone();
+		if is_artboard_layer {
+			if let Some(new_input_id) = new_input.as_node() {
+				let mut final_layer_node_id = new_input_id.clone();
+				while let Some(input_id) = self
+					.document_network
+					.nodes
+					.get(&final_layer_node_id)
+					.and_then(|input_node| input_node.inputs.get(1).and_then(|x| x.as_node()))
+				{
+					final_layer_node_id = input_id;
+				}
+				{
+					if let Some(final_layer_node) = self.document_network.nodes.get_mut(&final_layer_node_id) {
+						final_layer_node.inputs[1] = new_input_artboard_layer.clone();
+					}
+				}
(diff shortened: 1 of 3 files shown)


Keavon:
!build

Keavon:
If I draw several artboards, selecting one with the Artboard tool and deleting it causes the others to disappear (from getting disconnected in the graph). I think this is still the case in master so you didn't cause a regression, but since this is related to that bug I'd appreciate it if you'd help investigate and fix that here too. (Or feel free to say no, it can also be a followup PR.)

Keavon:
!build

editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
```diff
@@ -510,17 +510,62 @@ impl<'a> ModifyInputsContext<'a> {
 		});
 	}
 
-	fn delete_layer(&mut self, id: NodeId, selected_nodes: &mut SelectedNodes) {
+	fn delete_layer(&mut self, id: NodeId, selected_nodes: &mut SelectedNodes, mut is_artboard_layer: bool) {
```
Keavon, line 513: The function signature shouldn't have this be `mut`, since we don't want the caller to care about the implementation detail. You can redeclare the variable to be mut in the body of the function.

Keavon (changes requested):
- [x] The code you've added in `graph_operation_message_handler.rs` is pretty complicated to read, that is, how it works isn't obvious from reading it without having to analyze it pretty closely. Could you add some comments to each line/block of lines explaining what each line/block does? Something to summarize their purposes. Thanks.
- [ ] Document > Clear Artboards crashes. To reproduce: create a new document with the default artboard; use the Artboard tool to draw a new artboard; click Clear Artboards and you'll see it crashes.

Keavon:
I'm marking this as draft for the moment, please mark it as ready for review again when you're ready, and also ping me so I notice it's ready again. Thanks.

Keavon (approved):
Nice work on this one! It's definitely a valuable addition to Graphite :)

PR #1655, opened 2024-03-07
Fix orphaned child layers left behind when a group is ungrouped or deleted

### LINKED ISSUE #1654: Disconnected layers get left behind when deleting/ungrouping their parents
The `delete_layer` function is failing to fully remove all child nodes associated with the parent layer, resulting in incomplete deletion of layer elements.

If you group multiple layers, then either ungroup or delete the group, all but one of the children will be left behind as orphans.

haikalvidya:
this is resolve `delete_layer` function that failing to fully remove all child nodes associated with the parent layer.

Closes #1654

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/graph_operation_message_handler.rs
@@ -680,8 +680,6 @@ impl MessageHandler<GraphOperationMessage, GraphOperationHandlerData<'_>> for Gr
 				alias,
 			} => {
 				trace!("Inserting new layer {id} as a child of {parent:?} at index {insert_index}");
-				debug!("Inserting new layer {id} as a child of {parent:?} at index {insert_index}");
-				// debug!("The Node is {:#?}", nodes);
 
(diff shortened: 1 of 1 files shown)


0HyperCube (approved):
Looks good - thanks for this fix.

Keavon:
!build

Keavon:
I may have assumed something different when I renamed this PR. But what I assumed this PR does, now that I'm trying it at the build link above, doesn't seem to work. I draw an ellipse layer, select the layer, and hit delete; the nodes that fed into that layer are still left behind. Is that expected? Or can you please direct me to what this PR is intended to do if I misinterpreted its purpose? Thanks.

0HyperCube:
@Keavon This works fine for me. Note that if you have the node graph open, the selected node rather than layer will be deleted (which in your case is just the layer node). If you close the node graph and then delete the layer, it will work as expected. This is the same as on the main branch.

This PR fixes an issue whereby when deleting a group containing >=2 layers, the nodes only for the top most child would be deleted (as we followed the primary flow from the group which led through only the top most child layer). However this has now been resolved by also following the primary flow of all child layers.

Keavon:
Ah, that makes sense. Thanks for clarifying and pardon my confusion! This looks nice.

PR #1656, opened 2024-03-08
Create a history step when importing SVG files

milan-sedivy:
Generally the issue is that the actual import and creation of a new SVG is handled through `GraphOperationMessageHandler` which doesn't have a "syncish" way of backuping data (the only way how you could backup the data from there is by rewriting the `DocumentMessage::Backup` message and fixing code around and after that you would call it, but that leads to an async backup of data meaning the "step before the import" will likely not get saved in time).

So my approach was to add a `DocumentMessage::ImportSvg` ... which pretty much tries to create a `usvg::Tree` the same way as `GraphOperationMessage::NewSvg` does... with the exception that it doesn't care about the result... only cares about the succesful unwraping `Ok(_)` of the `Option<Tree>` and then proceeds to send the original message (and calls self.backup(responses) before)

Now ofcourse the idea of just deserializing the tree and sending that crossed my mind but after investigating serde I realized it would just be a reimplementation of their `Tree::to_string` method which seemed pointless.
I'll look into what is your branching strategy and I'll submit a PR so that you can take a look at it
Maybe just to add... of course an option would be to just `self.backup(...) `on every `GraphOperationMessage` received by `DocumentMessageHandler`... but that would lead to undesired behaviour with some of the other tools that work with vector graphics

### DIFF
--- a/editor/src/messages/portfolio/document/document_message.rs
+++ b/editor/src/messages/portfolio/document/document_message.rs
@@ -10,8 +10,8 @@ use graphene_core::raster::Image;
 use graphene_core::vector::style::ViewMode;
 use graphene_core::Color;
 
+use glam::DAffine2;
 use serde::{Deserialize, Serialize};
-
 #[impl_message(Message, PortfolioMessage, Document)]
 #[derive(PartialEq, Clone, Debug, Serialize, Deserialize)]
 pub enum DocumentMessage {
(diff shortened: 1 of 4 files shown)


0HyperCube (approved):
Looks good - thanks for the code improvement.

milan-sedivy:
> Looks good - thanks for the code improvement.

Thanks for the suggestion/feedback, really appreciate it!

Keavon (approved):

PR #1658, opened 2024-03-09
Allow cancellation of pivot drag in Select tool

milan-sedivy:
Added Abort - tested in multiple scenarios:
- ESC/RMB when DraggingPivot
- Change tool when DraggingPivot
- Tested other actions such as resizing that might be affected by the change
- Tested if rotations around pivot match after aborting


Keavon:
!build

Keavon (approved):

PR #1660, opened 2024-03-09
Don't save undo/redo history when aborting transactions

### LINKED ISSUE #1659: Aborted Transactions are saved in Redo history
Aborting a transaction is being "undone" with a simple `undo`. This results in it being stored within Redo history, potentially erasing previous history steps if length was reached and leads to unwanted behavior such as recreating something that was aborted.

milan-sedivy:
For now solved with a simple split of undo/redo into a "forgettable" version and non-forgettable.  
Fixes #1659

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -1074,15 +1074,14 @@ impl DocumentMessageHandler {
 	}
 
 	pub fn undo_with_history(&mut self, responses: &mut VecDeque<Message>) {
-		let Some(network) = undo(responses) else { return };
+		let Some(previous_network) = self.undo(responses) else { return };
 
-		let previous_network = std::mem::replace(&mut self.network, network);
 		self.document_redo_history.push_back(previous_network);
 		if self.document_redo_history.len() > crate::consts::MAX_UNDO_HISTORY_LEN {
 			self.document_redo_history.pop_front();
 		}
 	}
-	pub fn undo(&mut self, responses: &mut VecDeque<Message>) {
+	pub fn undo(&mut self, responses: &mut VecDeque<Message>) -> Option<NodeNetwork> {
 		// Push the UpdateOpenDocumentsList message to the bus in order to update the save status of the open documents
 		responses.add(PortfolioMessage::UpdateOpenDocumentsList);
 		//If there is no history return and don't broadcast SelectionChanged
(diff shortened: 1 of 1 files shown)


0HyperCube (approved):

Keavon:
!build

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1073,30 +1073,40 @@ impl DocumentMessageHandler {
...
 		if self.document_redo_history.len() > crate::consts::MAX_UNDO_HISTORY_LEN {
 			self.document_redo_history.pop_front();
 		}
 	}
-
-	pub fn redo(&mut self, responses: &mut VecDeque<Message>) {
+	pub fn undo(&mut self, responses: &mut VecDeque<Message>) -> Option<NodeNetwork> {
 		// Push the UpdateOpenDocumentsList message to the bus in order to update the save status of the open documents
 		responses.add(PortfolioMessage::UpdateOpenDocumentsList);
+		//If there is no history return and don't broadcast SelectionChanged
```
Keavon, line 1087:
```suggestion
		// If there is no history return and don't broadcast SelectionChanged
```

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1073,30 +1073,40 @@ impl DocumentMessageHandler {
...
+		responses.add(BroadcastEvent::SelectionChanged);
 
-		let Some(network) = self.document_redo_history.pop_back() else { return };
+		let previous_network = std::mem::replace(&mut self.network, network);
+		Some(previous_network)
+	}
+	pub fn redo(&mut self, responses: &mut VecDeque<Message>) -> Option<NodeNetwork> {
+		// Push the UpdateOpenDocumentsList message to the bus in order to update the save status of the open documents
+		responses.add(PortfolioMessage::UpdateOpenDocumentsList);
+		//If there is no history return and don't broadcast SelectionChanged
```
Keavon, line 1098:
```suggestion
		// If there is no history return and don't broadcast SelectionChanged
```

Keavon:
Redo history is still wiped out (as before), but now it doesn't put the *wrong* history state into the redo history. E.g. previously, you could drag a shape upwards and abort, and that upward location where you aborted would be stored in one step of redo history. Now it simply loses all redo history, which is a separate issue.

Issue #1657, opened 2024-03-09
All drag actions should be cancellable with Esc/RMB


Keavon:
This issue tracks all drag actions that don't currently implement cancellation with <kbd>Esc</kbd>/right click. An example of this desired behavior is the Rectangle tool, which lets you draw a rectangle by dragging it and hit <kbd>Esc</kbd> or right click to cancel the action. Another example is the Select tool which lets you drag a layer around on the canvas and similarly cancel it to restore it to its pre-drag location.

- [x] Select tool: moving the pivot (#1658)
- [x] Artboard tool: resizing an artboard (#1664)
- [x] Artboard tool: drawing a new artboard (`#1664`)
- [x] Navigate tool: zooming by click-dragging (`#1664`)
- [x] Navigate tool: tilting by <kbd>Alt</kbd>-click-dragging (`#1664`)
- [ ] Fill tool: filling should be cancellable after Lmb is pressed but before it's released
- [x] Gradient tool: drawing a new (or replacement) gradient line (`#1664`)
- [x] Gradient tool: reposition the endpoint of an existing gradient line (`#1664`)
- [x] Path tool: moving an anchor/handle (or multiple), including the toggling with <kbd>Alt</kbd> between smooth/sharp that might have occurred during the drag (`#1664`)
- [x] Freehand tool: drawing the trace, it should be canceled (`#1664`)
- [x] Brush tool: drawing the trace, it should be canceled (`#1664`)

milan-sedivy:
Hi Keavon, I would like to work on this

milan-sedivy:
Currently done: 
 - Artboard tool
 - Navigation tool
 
 I'll probably do the rest today as well (having a more hectic day than anticipated).

PR #1661, opened 2024-03-09
Make Artboard tool respond to Abort keys

milan-sedivy:
Part of #1657 

This is the last tool that I will do in a separate pull request as most of the logic is easy to implement.

Implemented 3 aborts:
- When resizing
- When drawing
- When moving (I've seen it implemented with other tools)


Keavon:
I moved these changes into #1664 and cleaned up `fn actions()` to reuse the common code. That issue can now supersede this one.

PR #1664, opened 2024-03-09
Fix cancellation of drag actions in all tools where it's missing

### LINKED ISSUE #1663: Some tools send DocumentMessage::Undo instead of DocumentMessage::AbortTransaction when Aborting
Example of a tool with this behavior is "SelectTool".

DocumentMessage::Undo stores the history step, which can lead to the deletion of a previously stored redo step see #1659 and related Pull request for more information.

milan-sedivy:
Part of #1657, #1663 

A bigger PR focused on fixing multiple tools at once along with other issues discovered.


milan-sedivy:
@Keavon everything but the filltool (and artboardtool which is in a separate PR) should be fixed by this PR. I plan on maybe refactoring the solution in ArtBoardTool based on what I learnt from other tools (nothing serious, just removing a long pattern match with a MessageDiscriminant)

Keavon (approved):
I'm thrilled that you were able to implement this with so few lines of code. Simplicity is key!

PR #1666, opened 2024-03-10
Make the Fill tool's fill click operation cancellable

### LINKED ISSUE #1657: All drag actions should be cancellable with Esc/RMB

milan-sedivy:
This is the last tool that needed a cancellable transaction.

It is implemented in a way that it introduces a "Fake drag" the reason for this is simple, emitting `DocumentMessage::AbortTransaction` currently works as an undo step. Thus if the user would not perform an operation before trying to cancel the transaction it would result in undoing his current history.

This could and should be reimplemented once a Document can be in different states such as "Transaction In Progress"

Closes #1657

### DIFF
--- a/editor/src/messages/input_mapper/default_mapping.rs
+++ b/editor/src/messages/input_mapper/default_mapping.rs
@@ -253,6 +253,9 @@ pub fn default_mapping() -> Mapping {
 		// FillToolMessage
 		entry!(KeyDown(Lmb); action_dispatch=FillToolMessage::FillPrimaryColor),
 		entry!(KeyDown(Lmb); modifiers=[Shift], action_dispatch=FillToolMessage::FillSecondaryColor),
+		entry!(KeyUp(Lmb); action_dispatch=FillToolMessage::PointerUp),
+		entry!(KeyDown(Escape); action_dispatch=FillToolMessage::Abort),
+		entry!(KeyDown(Rmb); action_dispatch=FillToolMessage::Abort),
 		//
 		// BrushToolMessage
 		entry!(PointerMove; action_dispatch=BrushToolMessage::PointerMove),
--- a/editor/src/messages/input_mapper/default_mapping.rs
+++ b/editor/src/messages/input_mapper/default_mapping.rs
@@ -254,8 +254,8 @@ pub fn default_mapping() -> Mapping {
 		entry!(KeyDown(Lmb); action_dispatch=FillToolMessage::FillPrimaryColor),
 		entry!(KeyDown(Lmb); modifiers=[Shift], action_dispatch=FillToolMessage::FillSecondaryColor),
 		entry!(KeyUp(Lmb); action_dispatch=FillToolMessage::PointerUp),
-		entry!(KeyDown(Escape); action_dispatch=FillToolMessage::Abort),
 		entry!(KeyDown(Rmb); action_dispatch=FillToolMessage::Abort),
+		entry!(KeyDown(Escape); action_dispatch=FillToolMessage::Abort),
 		//
 		// BrushToolMessage
 		entry!(PointerMove; action_dispatch=BrushToolMessage::PointerMove),
--- a/editor/src/messages/tool/tool_messages/fill_tool.rs
+++ b/editor/src/messages/tool/tool_messages/fill_tool.rs
@@ -59,7 +59,10 @@ impl<'a> MessageHandler<ToolMessage, &mut ToolActionHandlerData<'a>> for FillToo
 
 impl ToolTransition for FillTool {
 	fn event_to_message_map(&self) -> EventToMessageMap {
-		EventToMessageMap::default()
+		EventToMessageMap {
+			tool_abort: Some(FillToolMessage::Abort.into()),
+			..Default::default()
+		}
 	}
 }
 
@@ -83,11 +86,12 @@ impl Fsm for FillToolFsmState {
 		let ToolMessage::Fill(event) = event else {
 			return self;
 		};
-		let Some(layer_identifier) = document.click(input.mouse.position, &document.network) else {
-			return self;
-		};
+
 		match (self, event) {
 			(FillToolFsmState::Ready, color_event) => {
+				let Some(layer_identifier) = document.click(input.mouse.position, &document.network) else {
+					return self;
+				};
 				// TODO: Use a match statement here instead of if-else
 				let color = if color_event == FillToolMessage::FillPrimaryColor {
 					global_tool_data.primary_color
--- a/editor/src/messages/tool/tool_messages/fill_tool.rs
+++ b/editor/src/messages/tool/tool_messages/fill_tool.rs
@@ -10,8 +10,9 @@ pub struct FillTool {
 #[impl_message(Message, ToolMessage, Fill)]
 #[derive(PartialEq, Eq, Clone, Debug, Hash, Serialize, Deserialize, specta::Type)]
 pub enum FillToolMessage {
-	// General messages
+	// Standard messages
 	Abort,
+
 	// Tool-specific messages
 	PointerUp,
 	FillPrimaryColor,
@@ -87,6 +88,7 @@ impl Fsm for FillToolFsmState {
 		};
 		match (self, event) {
 			(FillToolFsmState::Ready, color_event) => {
+				// TODO: Use a match statement here instead of if-else
 				let color = if color_event == FillToolMessage::FillPrimaryColor {
 					global_tool_data.primary_color
 				} else {
@@ -117,7 +119,7 @@ impl Fsm for FillToolFsmState {
 			])]),
 			FillToolFsmState::Filling => HintData(vec![HintGroup(vec![
 				HintInfo::mouse(MouseMotion::Rmb, ""),
-				HintInfo::keys([Key::Escape], "Cancel Fill").prepend_slash(),
+				HintInfo::keys([Key::Escape], "Cancel").prepend_slash(),
 			])]),
 		};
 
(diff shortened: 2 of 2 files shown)


editor/src/messages/input_mapper/default_mapping.rs
```diff
@@ -253,6 +253,9 @@ pub fn default_mapping() -> Mapping {
 		// FillToolMessage
 		entry!(KeyDown(Lmb); action_dispatch=FillToolMessage::FillPrimaryColor),
 		entry!(KeyDown(Lmb); modifiers=[Shift], action_dispatch=FillToolMessage::FillSecondaryColor),
+		entry!(KeyUp(Lmb); action_dispatch=FillToolMessage::PointerUp),
+		entry!(KeyDown(Escape); action_dispatch=FillToolMessage::Abort),
+		entry!(KeyDown(Rmb); action_dispatch=FillToolMessage::Abort),
```
Keavon, line 258:
```suggestion
		entry!(KeyDown(Rmb); action_dispatch=FillToolMessage::Abort),
		entry!(KeyDown(Escape); action_dispatch=FillToolMessage::Abort),
```

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -10,7 +10,10 @@ pub struct FillTool {
 #[impl_message(Message, ToolMessage, Fill)]
 #[derive(PartialEq, Eq, Clone, Debug, Hash, Serialize, Deserialize, specta::Type)]
 pub enum FillToolMessage {
+	// General messages
+	Abort,
```
Keavon, line 14:
```suggestion
	// Standard messages
	Abort,

```

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -90,6 +115,10 @@ impl Fsm for FillToolFsmState {
 				HintInfo::mouse(MouseMotion::Lmb, "Fill with Primary"),
 				HintInfo::keys_and_mouse([Key::Shift], MouseMotion::Lmb, "Fill with Secondary"),
 			])]),
+			FillToolFsmState::Filling => HintData(vec![HintGroup(vec![
+				HintInfo::mouse(MouseMotion::Rmb, ""),
+				HintInfo::keys([Key::Escape], "Cancel Fill").prepend_slash(),
```
Keavon, line 120:
```suggestion
				HintInfo::keys([Key::Escape], "Cancel").prepend_slash(),
```

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -71,17 +85,28 @@ impl Fsm for FillToolFsmState {
...
-		let fill = Fill::Solid(color);
-
-		responses.add(DocumentMessage::StartTransaction);
-		responses.add(GraphOperationMessage::FillSet { layer: layer_identifier, fill });
-		responses.add(DocumentMessage::CommitTransaction);
-
-		FillToolFsmState::Ready
+		match (self, event) {
+			(FillToolFsmState::Ready, color_event) => {
+				let color = if color_event == FillToolMessage::FillPrimaryColor {
```
Keavon, line 90:
```suggestion
				// TODO: Use a match statement here instead of if-else
				let color = if color_event == FillToolMessage::FillPrimaryColor {
```

Keavon:
!build

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -90,6 +117,10 @@ impl Fsm for FillToolFsmState {
 				HintInfo::mouse(MouseMotion::Lmb, "Fill with Primary"),
 				HintInfo::keys_and_mouse([Key::Shift], MouseMotion::Lmb, "Fill with Secondary"),
 			])]),
+			FillToolFsmState::Filling => HintData(vec![HintGroup(vec![
+				HintInfo::mouse(MouseMotion::Rmb, ""),
+				HintInfo::keys([Key::Escape], "Cancel").prepend_slash(),
+			])]),
```
Keavon, line 123:
```suggestion
			FillToolFsmState::Filling => HintData(vec![HintGroup(vec![HintInfo::mouse(MouseMotion::Rmb, ""), HintInfo::keys([Key::Escape], "Cancel").prepend_slash()])]),
```

Keavon:
There are a few bugs I found:
- If I click and drag outside the shape then release, I get stuck in the mousedown state (notice the hints bar) and neither left clicking nor right clicking nor <kbd>Escape</kbd> outside the shape fix the stuck state (I have to left click to apply, or right click to abort, inside the shape)
- If I hold down the left mouse button, then hit <kbd>Ctrl</kbd><kbd>Z</kbd>, then abort with Rmb, this causes it to go back to a previous history step (it undoes the creation of a layer I made immediately before this, for example).
  
  If this is true for all the abortable/cancellable tools you've worked on, let's skip fixing it in this specific PR.

Keavon (changes requested):
There are a few bugs I found:
- If I click and drag outside the shape then release, I get stuck in the mousedown state (notice the hints bar) and neither left clicking nor right clicking nor <kbd>Escape</kbd> outside the shape fix the stuck state (I have to left click to apply, or right click to abort, inside the shape)
- If I hold down the left mouse button, then hit <kbd>Ctrl</kbd><kbd>Z</kbd>, then abort with Rmb, this causes it to go back to a previous history step (it undoes the creation of a layer I made immediately before this, for example).
  
  If this is true for all the abortable/cancellable tools you've worked on, let's skip fixing it in this specific PR.

milan-sedivy:
> There are a few bugs I found:
> 
> * If I click and drag outside the shape then release, I get stuck in the mousedown state (notice the hints bar) and neither left clicking nor right clicking nor Escape outside the shape fix the stuck state (I have to left click to apply, or right click to abort, inside the shape)
> * If I hold down the left mouse button, then hit CtrlZ, then abort with Rmb, this causes it to go back to a previous history step (it undoes the creation of a layer I made immediately before this, for example).
>   If this is true for all the abortable/cancellable tools you've worked on, let's skip fixing it in this specific PR.

The CTRL+Z+ESC/RMB will **probably** be true for all tools but I will look into an easy hotfix (before the whole transaction system is refactored so that it doesn't rely on undos).

As for the other behavior I will check and see what I can do to fix it.

milan-sedivy:
I noticed a new TODO (or maybe old?) i.e.

// TODO: Use a match statement here instead of if-else

In that given case i think it's preferable to keep it as an if-else rather than match, because of the fact that match would just add another arm and I think that down the line it might lead to uncaught issues.

milan-sedivy:
@Keavon 
> * If I hold down the left mouse button, then hit CtrlZ, then abort with Rmb, this causes it to go back to a previous history step (it undoes the creation of a layer I made immediately before this, for example).
>   If this is true for all the abortable/cancellable tools you've worked on, let's skip fixing it in this specific PR.

As suspected this is something that is affecting all tools that either call Undo or AbortTransaction. (Not just the tool I worked on)

I was hoping for a quick hotfix through a "hack" in `default_mapping.rs` but unfortunately to implement the hack I would have to refactor that file. This definitely needs a separate issue/PR
So currently there are two approaches I suggest:
- (the hack) Fix keymappings so that more modifiers => bigger priority when chosing an `action_dispatch` this would allow us to make a `DocumentMessage::Noop` and send that when you use a forbidden combination (CTRL + Z + LMB)
- a better approach would be to refactor StartTransaction/CommitTransaction/AbortTransaction so that they actually do what the names suggest. That would allow us to move from AbortTransaction using **undo** under the hood.

Keavon:
Ok, thanks for investigating. Can you move your findings about the undo bug to another issue which can be home for that future PR?

The to-do I added is a suggestion to better model the invariants. Else doesn't imply filling with the secondary color. If there's something other than filling with primary or secondary color, it'll still enter the else arm. A match statement better models the problem and is this cleaner.

milan-sedivy:
> Ok, thanks for investigating. Can you move your findings about the undo bug to another issue which can be home for that future PR?
> 
> The to-do I added is a suggestion to better model the invariants. Else doesn't imply filling with the secondary color. If there's something other than filling with primary or secondary color, it'll still enter the else arm. A match statement better models the problem and is this cleaner.

Hmm good point didn't think about that.

As of the PR it's already done (with an issue)

milan-sedivy (commented):
It has been fixed

Keavon:
!build

Keavon (approved):

PR #1667, opened 2024-03-10
Improve the Select tool's input hints

milan-sedivy:
Previously SelectTool had the same KeyHints no matter what state, this was caused by the fact that it was done in a hackish way by using `fn standard_tool_messages` instead of `fn update_hints`. This PR addresses this by a small change to the code and allows to differentiate between states.

out


Keavon (approved):
Feel free to read my diff if you'd like to pick up on a few Rust code idiomatic-ness tips.

PR #1675, opened 2024-03-11
Block the undo shortcut while dragging with a tool

### LINKED ISSUE #1674: Performing an undo right before cancelling a tool's drag action results in a doubled undo
AbortTransaction is only possible while holding down LMB however if CTRL + Z is pressed while LMB is held down then pressing ESC / RMB to Abort leads to a second undo.

This is expected from the way the code takes care of AbortTransaction right now (it proceeds with an undo).

There are two possible ways on how to tackle this:
1) Create a keymapping CTRL + Z + LMB that sends a NOOP message instead of UNDO
2) Refactor Start/Commit/Abort Transaction messages.

milan-sedivy:
Big thanks to @0HyperCube

This is a fix that is presented in the issue as 1) i.e. prevent it by sending a NOOP instead of UNDO when CTRL + Z + LMB is pressed.

Closes #1674

### DIFF
--- a/editor/src/messages/input_mapper/default_mapping.rs
+++ b/editor/src/messages/input_mapper/default_mapping.rs
@@ -48,6 +48,8 @@ pub fn default_mapping() -> Mapping {
 		// NORMAL PRIORITY
 		// ===============
 		//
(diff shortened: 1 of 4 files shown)


Keavon:
!build

Keavon (approved):

PR #1676, opened 2024-03-11
Add nondestructive vector editing

### LINKED ISSUE #134: Make tools use primitive vector shape generator nodes
Blocked on https://github.com/GraphiteEditor/Graphite/issues/1162

- Spline tool
- Line tool
- Rectangle tool
- Ellipse tool
- Polygon tool

### LINKED ISSUE #999: Make the Pen tool not show the in-progress segment until committed
### Current behavior:
The extending segment is updated while the mouse moves to show the shape as it will appear. This is undesirable because it can get in the way of viewing something you're tracing over.

https://user-images.githubusercontent.com/4388688/215305080-fff17869-3366-4cf0-8a0e-5a044a9a740e.mp4

### Desired behavior:
The extending segment is updated only after releasing the mouse cursor to commit the new segment. Only the thin blue overlay should show the extension preview.

https://user-images.githubusercontent.com/4388688/215305489-ca9cbc12-1b1a-4503-bef6-8046bbf62ff9.mp4

It should apply to the Pen tool only.

### LINKED ISSUE #1162: Vector Edit node
Adjust the "Shape" node to support taking in an existing path and modifying it. If there is no input given, it creates a new shape. If stable id generation is added to the text tool, this will allow the path of text to be modified in a non destructive way.

Required in order to implement #134

### LINKED ISSUE #1180: When text is modified in the Properties panel, it shifts around

### LINKED ISSUE #1736: Filled but open shapes don't have click targets for their fill
Environment information:
- OS: PopOS `22.04` `x86_64`
- Rust version: `1.79.0-nightly`
- Node version: `19.8.1`

Steps to replicate:
- Select `pencil tool` and create a circle like shape
- Select `fill tool` and attempt to fill it by clicking inside with a color different from background

Actual result:
Shape is not filled

Expected result:
Shape is filled with selected color

Pencil shape can be filled by clicking on the stroke of the shape, there's an ongoing effort on #1711 

---
I can use the `Pen tool` to create shapes and if I attempt to fill them, they are filled. Using the `Pencil` however the fill does not show up visually but it shows up in the `Node graph`

image

Editing the fill from the shape properties does allow me to change the color:
image

### LINKED ISSUE #1800: Remove the pivot from transforms
This is a simplification before we implement the updated model for transform origins in the future.

The draggable yellow crosshair is still kept as part of the Select tool's transform cage, which affects <kbd>G</kbd>/<kbd>R</kbd>/<kbd>S</kbd> but isn't part of the graph's render process.

0HyperCube:
Closes #1162, closes #1800, closes #1736, closes #1180, closes #134, closes #1180, closes #999 (addresses the pen tool not the spline tool).

This MR is a very early work in progress.

ToDo:
- [x] Selecting only attached verts when clicked
- [x] Using alt to toggle the colinears
- [ ] ~~Fix booleans~~ Impossible - I will port some other (working) boolean impl at another time.
- [x] Remove dead codes
- [x] Clean up and document the new files
- [x] Dissolving many points
- [x] Not using handles with shift aligning anchors.
- [x] Create some new bugs
- [ ] Resolve the new bugs


Keavon:
!build

0HyperCube:
Closing as I can't be bothered to fix the merge conflicts @Keavon.

Keavon:
@0HyperCube if you're ever blocked by something like merge conflicts, please let me know! I'm here to help make things run smoothly for you in any moment I'm able to help. Merge conflicts are almost always within my range of capabilities, so just give me a ping any moment that you'd rather I deal with them for you and I'd be most happy to. It's my goal to make sure you're happy with the work you're doing, so if there's busywork I can take on for you, please don't hesitate to throw it my way.

0HyperCube:
[non_destructive_spline.webm](https://github.com/GraphiteEditor/Graphite/assets/78500760/5e141260-22e3-4770-9a1e-55955e986e35)

Keavon:
It's amazing to see this all coming together. I looked pretty hard and couldn't find many regressions at all, so I commend you on your thorough testing already. Here are just 3 (or maybe 4) rather minor issues I was able to find.

- Select tool: I can no longer double-click a shape to make it change to the Path tool for editing its points I think this is a regression?
- Freehand tool: I see you added smoothing, nice! But could that be made nondestructive by making it still draw a polyline like before, but having it insert the *Spline from Points* node in the node chain so the smoothing can be disabled later on if the user chooses? Having the automatic insertion of that smoothing node be optional with a tool option checkbox could also be helpful. But either or both aspects of this could wait for later on, no need to do that in this PR.
- Path tool: Inserting a point causes a subsequent edge to become a separate path stroke:  

https://github.com/GraphiteEditor/Graphite/assets/4388688/a03b529f-67f0-4d44-a45a-055cee30d73f
- Path tool: deleting a point (with <kbd>Ctrl</kbd><kbd>Delete</kbd>) causes the resulting shape's blue overlay hover outline to be different from the actual shape (but this might actually just be a symptom of the previous bullet point, as that's occurring here too):  

https://github.com/GraphiteEditor/Graphite/assets/4388688/fed4a028-47e0-4799-96c6-7e1f0fad6e55

None of these are important enough to block merging this in the current state, so if you're not feeling down to fix these right now, no worries. This is a massive improvement now that we can modify shapes, even generated ones, non-destructively!

Keavon (approved):

PR #1678, opened 2024-03-11
Place new folders in the correct stack order relative to other selected layers

moOsama76:
Closes new folder creation (sub-issue of #1633 )


Keavon:
!build

Keavon:
We'd like it to be placed just above the least-nested layer/folder that has a selection:

image

Right this this goes at the very top which isn't desired. I see in your code, you're using an insert index of `0` which will instead need to be calculated accordingly.

moOsama76:
Yes sure, sorry about that

Keavon:
Please see my diff (49ed33732b0b93fe62a11efe97385328614dd32d) to learn from my little cleanup so you can write more idiomatic and performant Rust in the future :)

Keavon (approved):

PR #1681, opened 2024-03-12
Fix doubled translation from dragging layers sharing an upstream Transform node

### LINKED ISSUE #1529: Dragging layers sharing a transform node applies multiple translations
Create a document containing two rectangles. Enter the node graph and add a layer node, using the data from the stroke node of one of the rectangles. Select all of the rectangles and drag them using the select tool. Observe how the translation is applied several times. When you drag around several layers we apply a transformation to the first transform node found for each layer; however because there are several layers sharing the same transform node, the translation is applied several times.

Solutions:
- In select_tool.rs, instead of applying a delta to the current transform, store the transform at the start of the drag and compute the new transform based on that (similar to how dragging the selection bounds works)
- In graph_operation_message_handler.rs, check for cases where the node is used by several other nodes, and if that is the case then create a new transform node for each layer separately before applying the operation.

https://github.com/GraphiteEditor/Graphite/assets/78500760/8011ef59-a055-4e29-87a8-9df5d73351e6

shipp02:
Hello,

This fixes above issue, demo:

https://github.com/GraphiteEditor/Graphite/assets/51503822/70b7fccb-8520-45d5-a599-0b4f8ce24ecf

Closes #1529

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -18,6 +18,7 @@ use graph_craft::document::{NodeId, NodeNetwork};
 use graphene_core::renderer::Quad;
 
 use std::fmt;
+use std::iter::zip;
 
 #[derive(Default)]
 pub struct SelectTool {
@@ -253,6 +254,8 @@ struct SelectToolData {
 	drag_start: ViewportPosition,
 	drag_current: ViewportPosition,
 	layers_dragging: Vec<LayerNodeIdentifier>,
+	old_layers: Vec<Vec<LayerNodeIdentifier>>,
+	old_transforms: Vec<DAffine2>,
 	layer_selected_on_start: Option<LayerNodeIdentifier>,
 	select_single_layer: Option<LayerNodeIdentifier>,
 	has_dragged: bool,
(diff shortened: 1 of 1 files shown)


Keavon:
@0HyperCube you're better than me at understanding this part of the code and the implications of this change. Could you please see if this looks good to go? Thanks.

Keavon:
@shipp02 please leave a comment in the original issue so I can assign it to you.

Keavon:
I finally got a chance to do some QA testing on this. The first thing I noticed is that Dragging individual layers is completely broken. Just draw two layers and try dragging each of them on after the other. This is probably caused by your choice to store the transform in the Select tool's struct. I'd avoid attempting to persist state since this should be a stateless problem as far as I can tell.

0HyperCube:
@Keavon persisting state is necessary to implement the first of my suggested solutions from the issue:
> In select_tool.rs, instead of applying a delta to the current transform, store the transform at the start of the drag and compute the new transform based on that (similar to how dragging the selection bounds works)

shipp02:
Thank you for confirming my approach!

 I was in the middle of finals hence the lack of a response from me in the
last week.

I will fix the bug @Keavon found while he was running QA.

Keavon:
@shipp02 just checking if you've been able to continue addressing this.

shipp02:
I was able to reproduce the regression you mentioned but wasn't sure how to fix it.

Keavon:
I'm going to close this for now but feel free to reopen it if you're able to continue making progress.

PR #1698, opened 2024-03-22
Fix hidden state missing from copy/paste, causing hidden layers to reappear when pasted or grouped

haikalvidya:
This is related to [this](https://discord.com/channels/731730685944922173/881073965047636018/1219836949750218762)

This change makes copy-and-paste bring the node's visible state to the copy buffer. This way, when you paste, the state stays the same.


Keavon (approved):
Thanks ;)

Issue #1679, opened 2024-03-12
Crash when Pen tool's in-progress point snaps along an angle with its previous anchor


Keavon:
**Describe the Crash**
This crash happens immediately upon hoving the in-progress anchor point over the previous anchor point while snapping is enabled and angle constraint (<kbd>Shift</kbd> to snap 15° or <kbd>Ctrl</kbd> to lock angle) is active:

**Steps To Reproduce**
Describe precisely how the crash occurred, step by step, starting with a new editor window.
1. Ensure snapping is enabled
2. Use the Pen tool and click to start drawing
3. Hold <kbd>Shift</kbd> (Snap 15°) or <kbd>Ctrl</kbd> (Lock Angle)
4. Move the pointer so the new point snaps to the already drawn anchor which the in-progress segment is coming from
5. Immediate crash when it snaps

**Additional Details**

https://github.com/GraphiteEditor/Graphite/assets/4388688/6044dab6-7aa8-4a4f-a607-5eeb7223d91d

**Browser and OS**
Chrome 122, Windows 10 or 11

**Stack Trace**
```
panicked at C:\Users\Keavon\.cargo\registry\src\index.crates.io-6f17d22bba15001f\glam-0.25.0\src\f64\dvec2.rs:478:9:
assertion failed: other_len_sq_rcp.is_finite()

Error
    at http://localhost:8080/src/io-managers/panic.ts:7:24
    at callCallback (http://localhost:8080/src/wasm-communication/subscription-router.ts:24:9)
    at Object.handleJsMessage (http://localhost:8080/src/wasm-communication/subscription-router.ts:32:5)
    at http://localhost:8080/src/wasm-communication/editor.ts?t=1710214358055:33:19
    at http://localhost:8080/wasm/pkg/graphite_wasm.js?t=1710214358055:2217:37
    at handleError (http://localhost:8080/wasm/pkg/graphite_wasm.js?t=1710214358055:375:18)
    at imports.wbg.__wbg_call_5830eb5e103aa772 (http://localhost:8080/wasm/pkg/graphite_wasm.js?t=1710214358055:2216:67)
    at js_sys::Function::call2::h0ed55ef12986170a (http://localhost:8080/wasm/pkg/graphite_wasm_bg.wasm:wasm-function[11510]:0x864b2b)
[...]

PR #1701, opened 2024-03-24
Fix crash when Pen tool's in-progress point snaps along an angle with its previous anchor

Sahilgill24:
#1679


0HyperCube (approved):
Looks good - thanks for this fix.

PR #1702, opened 2024-03-26
Add layer locking feature

### LINKED ISSUE #1127: Add layer locking
Add a 🔒 to layers to prevent them from being selected in the viewport. It should still allow selection in the Layer Tree panel, and updating of values in the Properties panel, but shouldn't allow viewport-related interaction (using the tool and certain actions accessible from hotkeys and/or the menu bar).

haikalvidya:
This will close this issue ticket https://github.com/GraphiteEditor/Graphite/issues/1127.

Adding Layer locking feature, the  behavior of this feature:
1. Make the layer cannot be drag
2. Disabled transform cage when selected object
3. Can't be modified with G/R/S, or with nudging
4. User still can click through the locking layer to select other object behind it

Closes #1127


Keavon:
!build

Keavon:
- If I select a locked layer and a non-locked layer:
  - In the Path tool, I can click and drag the non-locked layer and they both move, and the same for resizing
  - In the Path tool, the yellow pivot widget shows the bounding box calculated based on the locked and non-locked layers, instead it should calculate based on the bounding box of just the non-locked layers; and if ALL the selected layers are locked, no pivot shown be shown at all
- If a parent is locked and so is its child, the icon is incorrectly showing unlocked for the child
- We want to exclude the selected layers from all of these actions:
  image
  That means both the code which checks if those icons (pivot basis, alignment, flipping, boolean ops) should be disabled (filter out the locked layers when the selected layers are counted) and the message handlers for alignment and flipping.

Keavon:
!build

Keavon:
!build

frontend/src/components/panels/Layers.svelte
```diff
@@ -424,10 +428,10 @@
 						<IconButton
 							class={"status-toggle"}
 							classes={{ inactive: !listing.entry.parentsUnlocked }}
-							action={(e) => (toggleLayerVisibility(listing.entry.id), e?.stopPropagation())}
+							action={(e) => (toggleLayerLock(listing.entry.id), e?.stopPropagation())}
 							size={24}
-							icon={listing.entry.parentsUnlocked ? "PadlockLocked" : "PadlockUnlocked"}
-							hoverIcon={listing.entry.parentsUnlocked ? "PadlockUnlocked" : "PadlockLocked"}
+							icon={(listing.entry.parentId && listing.entry.parentsUnlocked) || !listing.entry.unlocked ? "PadlockLocked" : "PadlockUnlocked"}
+							hoverIcon={(listing.entry.parentId && listing.entry.parentsUnlocked) || !listing.entry.unlocked ? "PadlockUnlocked" : "PadlockLocked"}
```
Keavon, line 434: I don't think we actually need the first part of the `||` clause? Just the change to `!listing.entry.unlocked` (my mistake for copy/pasting it to use `listing.entry.parentsUnlocked` when I first wrote this, thanks for the fix). Or does that first part of the `||` clause do something I'm not realizing?

haikalvidya: yeah, I think so, `listing.entry.parentsUnlocked` is just used for the classes, for marker inactive classes on that icon button.

Keavon: Thanks for the change. Sorry to nitpick but can you also remove the `!` and reverse the ternary clauses?

haikalvidya: Sure, I think that make easier to understand the operation

Keavon: Thanks.

Keavon:
Ok, with that tidied up, I've checked off the items in my checklist above. So there's still a few small things needing improvements but it's very close now!

Keavon:
!build

Keavon (approved):
Fantastic work! This is a really nice feature to have and it seems to be well integrated.

Issue #1703, opened 2024-03-26
Customising grid colour


0HyperCube:
Users should be able to customise the colours of the grid to allow for increased contrast with the background colour.

Here is the current grid settings.
![grid settings](https://github.com/GraphiteEditor/Graphite/assets/78500760/05fafd4c-2ff3-4475-ae7e-2bba6d717318)

A field called colour should be added here. See `grid_overlays.rs`.

Issue #1704, opened 2024-03-26
Grid as dots


0HyperCube:
Users should be able to set a rectangular grid to be displayed as dots instead of as lines.

Here is the current grid settings.
![grid settings](https://github.com/GraphiteEditor/Graphite/assets/78500760/05fafd4c-2ff3-4475-ae7e-2bba6d717318)

A toggle for displaying as dots should be added. See `grid_overlays.rs`.

PR #1709, opened 2024-03-27
Dot grid

adamgerhant:
Closes [#1704](https://github.com/GraphiteEditor/Graphite/issues/1704)

Adds a dot type grid option. Snapping and spacing is the same as the rectangular grid.

Changes:
- Adds `grid_overlay_dot`function to draw circles using the overlay context
- Adds `GridType::Dot` with similar implementation to `GridType::Rectangle`
- UI changes for selecting the Dot option
- Replaces matches! macro for selecting index to a match statement
- Adds circle function to [utility_types.rs](https://github.com/GraphiteEditor/Graphite/pull/1709/files#diff-fc1f68d21385d829e2a2a8eadca8f208e900ee9c3262a58fcc79ae5ff4251733)

dots

### DIFF
--- a/editor/src/messages/portfolio/document/overlays/grid_overlays.rs
+++ b/editor/src/messages/portfolio/document/overlays/grid_overlays.rs
@@ -79,10 +79,34 @@ fn grid_overlay_isometric(document: &DocumentMessageHandler, overlay_context: &m
 	}
 }
 
+fn grid_overlay_dot(document: &DocumentMessageHandler, overlay_context: &mut OverlayContext, spacing: DVec2) {
+	let origin = document.snapping_state.grid.origin;
+	let Some(spacing) = GridSnapping::compute_rectangle_spacing(spacing, &document.navigation) else {
+		return;
+	};
+	let document_to_viewport = document.metadata().document_to_viewport;
+	let bounds = document_to_viewport.inverse() * Quad::from_box([DVec2::ZERO, overlay_context.size]);
+
+	let cmp = |a: &f64, b: &f64| a.partial_cmp(b).unwrap();
+	let min_x = bounds.0.iter().map(|&corner| corner.x).min_by(cmp).unwrap_or_default();
+	let max_x = bounds.0.iter().map(|&corner| corner.x).max_by(cmp).unwrap_or_default();
+	let min_y = bounds.0.iter().map(|&corner| corner.y).min_by(cmp).unwrap_or_default();
+	let max_y = bounds.0.iter().map(|&corner| corner.y).max_by(cmp).unwrap_or_default();
+	for line_index in 0..=((max_x - min_x) / spacing.x).ceil() as i32 {
+		let x_pos = (((min_x - origin.x) / spacing.x).ceil() + line_index as f64) * spacing.x + origin.x;
+		for line_index in 0..=((max_y - min_y) / spacing.y).ceil() as i32 {
+			let y_pos = (((min_y - origin.y) / spacing.y).ceil() + line_index as f64) * spacing.y + origin.y;
+			let circle_pos =  DVec2::new(x_pos, y_pos);
+			overlay_context.circle(document_to_viewport.transform_point2(circle_pos), 1., Some(COLOR_OVERLAY_GRAY), Some(COLOR_OVERLAY_GRAY));
+		}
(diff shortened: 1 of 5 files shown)


adamgerhant:
Added grid color select to close #1703. 

Changes:
- Adds support for prefixed hexed codes
- Adds grid_color field to GridSnapping 
- Color selector UI

Note: blocked by #1710 since initial color is not correct.

editor/src/messages/portfolio/document/overlays/grid_overlays.rs
```diff
@@ -135,9 +171,16 @@ pub fn overlay_options(grid: &GridSnapping) -> Vec<LayoutGroup> {
 				RadioEntryData::new("isometric")
 					.label("Isometric")
 					.on_update(update_val(grid, |grid, _| grid.grid_type = GridType::ISOMETRIC)),
+				RadioEntryData::new("dot")
+					.label("Dot")
+					.on_update(update_val(grid, |grid, _| grid.grid_type = GridType::DOT)),
```
0HyperCube, line 176: The dot grid works well, but I think that it should not be displayed as a seperate grid type (as it is merely an alternative rendering method for the rectangle grid). Probably a check box if the rectangle grid enabled would work better.

node-graph/gcore/src/raster/color.rs
```diff
@@ -831,6 +860,26 @@ impl Color {
...
+			return None;
+		}
+   		let color_str = &color_str[1..];
+		let r = u8::from_str_radix(&color_str[0..2], 16).ok()?;
+		let g = u8::from_str_radix(&color_str[2..4], 16).ok()?;
+		let b = u8::from_str_radix(&color_str[4..6], 16).ok()?;
+		let a = u8::from_str_radix(&color_str[6..8], 16).ok()?;
+
+		Some(Color::from_rgba8_srgb(r, g, b, a))
+	}
```
0HyperCube, line 881: Please reuse the existing function above to reduce code duplication. e.g. `colour_string.strip_prefix("#").and_then(|value|Color::from_rgba_str(value)`. This will also ensure that it is actually a `#`.

node-graph/gcore/src/raster/color.rs
```diff
@@ -847,6 +896,23 @@ impl Color {
...
+		if color_str.len() != 7 {
+			return None;
+		}
+		let color_str = &color_str[1..];
+		let r = u8::from_str_radix(&color_str[0..2], 16).ok()?;
+		let g = u8::from_str_radix(&color_str[2..4], 16).ok()?;
+		let b = u8::from_str_radix(&color_str[4..6], 16).ok()?;
+
+		Some(Color::from_rgb8_srgb(r, g, b))
+	}
```
0HyperCube, line 914: Again `colour_string.strip_prefix("#").and_then(|value|Color::from_rgb_str(value)` will suffice.

0HyperCube (changes requested):
Thanks for your effort so far.

adamgerhant:
I believe all changes should have been made. I removed the Dot grid type, and instead added a Boolean field that tracks whether the grid should be displayed with dots. This also makes it easy to add a dot grid for the Isometric type, if there are plans for that. Just remove the condition for displaying the checkbox, create the display function, and use it accordingly.

I removed the prefixed color methods and used concatenation and remove_prefix accordingly. Currently this PR is blocked by the Color bug, since it starts at `#999999` instead of `#cccccc`. An easy fix could be to replace the hex constant with a Color.

Keavon:
!build

Keavon:
Square 1px pixel grid aligned (no antialiasing) dots would already be the desired appearance. Just be aware that a dotted line might drift if the ratio is a non-integer fraction between the document space and viewport space, so some cleverness might be required to enable the avoidance of antialiasing by staying perfectly grid aligned. Maybe several lines per row/column with different screenspace pixel spacing intervals as the error accumulates between the non-integer fraction ratio and the ratio you need to maintain screenspace grid alignment.

Keavon:
I'm putting this back to draft status till you have time again to deal with the performance suggestions in my previous comment. But don't focus on this till your other PR is done, since that's the much higher priority feature. Thanks :)

Keavon:
!build

PR #1712, opened 2024-03-29
Generalize layers as merge nodes to enable adjustment layers

### LINKED ISSUE #1510: Make all nodes stackable and generalize layers as stacked Merge nodes
Currently, the Layer node performs a compositing operation with the bottom and left data. But we also want other adjustment-style nodes (with a single primary input and output of the same type, like blur, HSV, grayscale, Imaginate, etc.) to support being layers also.

To make this change, we need to tweak the current definition of a layer. The current functionality of the Layer node should become the Composite node, which is a standard-looking node. Then all nodes should be given a boolean flag as metadata indicating whether that node should be drawn as a layer. If drawn as a layer, its left-to-right primary input/output becomes bottom-to-top primary input/output. Other exposed input parameters are drawn on the left (for now it's ok to just support one input until I work out the styling for multiple). We can add a "Convert to Layer"/"Convert to Node" toggle up near where we have the "Make Hidden" and "Preview" buttons.

Here's an example from the mockup of what a standard Composite layer (which is today's functionality of layers) and a Hue/Saturation adjustment node look like when they are both acting as layers:

image

The image icon indicates it's a Composite node which has the features of what layers already do today. Thus these will be the most common form of nodes. But making this generalization will allow other nodes, like Hue/Saturation as shown in the mockup above, to stacked vertically in the graph. This is especially important for Imaginate's img2img feature now that the rasterize-the-layers-below approach is gone as part of the unified document graph refactor.

Q&A:
- What about nodes with more than one parameter (secondary) input? -> Eventually the styling may be adjusted as needed to support multiple. For now we can just assume zero or one.
- What about nodes with returned (secondary) outputs? -> The design might or might not accommodate them, but for now we just don't show those outputs so there is no way to connect them. If a node is converted from a node into a layer, we disconnect those.

adamgerhant:
Closes #1510


Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
```diff
@@ -103,10 +103,73 @@ impl MessageHandler<GraphOperationMessage, GraphOperationMessageData<'_>> for Gr
 					modify_inputs.brush_modify(strokes);
 				}
 			}
+			GraphOperationMessage::MoveUpstreamSiblingsToChild { new_parent, mut upstream_sibling_ids } => {
```
Keavon, line 106: Should the message signature have this be `mut` because the intention is for this to mutate its caller's data?

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
```diff
@@ -114,70 +123,115 @@ impl<'a> ModifyInputsContext<'a> {
...
+	///			-> Post node input_index: 0
+	///		↑		if skip_layer_nodes == 0, return (Post node, Some(Layer1), 1)
+	///	->	Layer1	input_index: 1
+	///		↑		if skip_layer_nodes == 1, return (Layer1, Some(Layer2), 0)
+	///	->	Layer2	input_index: 2
+	///		↑
+	///	->	NLN
+	///		↑		if skip_layer_nodes == 2, return (NLN, Some(Layer3), 0)
+	///	->	Layer3	input_index: 3
+	///				if skip_layer_nodes == 3, return (Layer3, None, 0)
```
Keavon, line 136: Please don't use tab characters in ASCII art. Can you please fix up the formatting for this so it's nice and readable?

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
```diff
@@ -523,161 +616,18 @@ impl<'a> ModifyInputsContext<'a> {
 		});
 	}
 
-	pub fn resize_artboard(&mut self, location: IVec2, dimensions: IVec2) {
+	pub fn resize_artboard(&mut self, mut location: IVec2, mut dimensions: IVec2) {
```
Keavon, line 619: Do these need to mutate the data given by their caller?

Keavon (approved):
Superb work on this!

PR #1717, opened 2024-03-30
Fix "contrain" typo

smarizvi110:
Fix typo 'contrain' to true 'constrain'

Found this typo propagated in many places when going through code for snapping.


Keavon (approved):
Thanks

PR #1724, opened 2024-04-05
Add alignment buttons popover option for alignment relative to artboard

### LINKED ISSUE #1720: Alignment button options (align to artboard)
This popover menu is currently "coming soon".

image

We'd like to have a checkbox (off by default) called "To Artboard" which makes the alignment of all selected layers work in relation to the artboard rather than their collective bounding box.

There has also been [a request](https://discord.com/channels/731730685944922173/1134981447539761273/1216556761910345868) for letting the user choose a specific layer's bounding box to align all other layers to (meaning that chosen layer doesn't move, but the others do).

Potentially one more feature that would be handy is a way to choose between "deep" and "shallow" treatment of nested layers, where only the selected folders get aligned vs. each of their contents get aligned.

nico-mayora:
Closes #1720.

~~This issue is still very much WIP--note the numerous `debug!` calls :^), but wanted to get this PR going as I have run into a small issue:~~

~~Note `editor/src/messages/tool/tool_messages/select_tool.rs:173`: I've made it so that if `to_artboard` is `true`, the alignment options are again enabled for just one selected object. In practise, this doesn't work--ticking the checkbox won't update the alignment buttons automatically. We need to reselect the object for the effect to take place. Is there a way this automatic refresh can be done?~~
### Solved!


editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -168,10 +188,37 @@ impl LayoutHolder for SelectTool {
...
-		widgets.extend(self.alignment_widgets(disabled));
-		widgets.push(PopoverButton::new("Align", "Coming soon").disabled(disabled).widget_holder());
+		widgets.extend(self.alignment_widgets(disabled, align_to_artboard, align_link_selected));
+		widgets.push(
+			// TODO: This button should be hidden when the canvas is infinite.
+			PopoverButton::new("Align", "Change alignment properties")
+				.options_widget(vec![
+					LayoutGroup::Row {
+						widgets: vec![
+							CheckboxInput::new(false)
```
0HyperCube, line 202: In order for the UI to display check mark correctly, you need to pass the current value of the input (here `tool_data.align_to_artboard`) into the CheckboxInput::new() function.

Same for the below input.

nico-mayora: Ah, I see. Where would be best to initialise the variables to true and false though?

0HyperCube: The default value is based on the default function for the select tool state (automatically derived at the moment).

Keavon:
!build

Keavon:
QA feedback:
- [ ] A dash is shown in the dropdown menu until it is open. Only upon opening the menu is the dash replaced with the first item.
- [ ] The list is not always populated with the current selection.
- [ ] Let's use a three-way radio input to pick which alignment mode to use. "Selection" / "Layer" / "Artboard" can be the three choices.

nico-mayora:
> QA feedback:
> 
>     * [ ]  A dash is shown in the dropdown menu until it is open. Only upon opening the menu is the dash replaced with the first item.
> 
>     * [ ]  The list is not always populated with the current selection.
> 
>     * [ ]  Let's use a three-way radio input to pick which alignment mode to use. "Selection" / "Layer" / "Artboard" can be the three choices.

I've pushed fixes for points 1 and 3, but can't manage to reproduce 2. Would you mind elaborating a bit more?

nico-mayora:
@Keavon
> QA feedback:
> 
>     * [ ]  A dash is shown in the dropdown menu until it is open. Only upon opening the menu is the dash replaced with the first item.
> 
>     * [ ]  The list is not always populated with the current selection.
> 
>     * [ ]  Let's use a three-way radio input to pick which alignment mode to use. "Selection" / "Layer" / "Artboard" can be the three choices.

I've pushed fixes for points 1 and 3, but can't manage to reproduce 2. Would you mind elaborating a bit more?

Keavon:
I have some changes I need to push to your branch, but I seemingly can't right now. The below feedback should be taken into account after you've been able to pull the changes that I still need to push for you. See my Discord message about this push permission issue.

- [ ] I've decided it's better to have only two radio button modes: Layers and Artboard. Layers mode is a combined form of the current Selection and Layer modes. The dropdown menu should have a default (top) entry "Combined Bounds of Selection", with the selected layers available to choose if 2+ are selected currently. It should always switch back to "Combined Bounds of Selection" when the selected layer is lost. For example, if the user picks a layer from that menu, then closes the menu and deselects it, the menu should reset next time the user opens the alignment popover or clicks one of the alignment buttons. But it should try to memorize the layer chosen by the user while that menu is closed, allowing the user to deselect that layer then re-select it as long as the popover wasn't re-opened or the alignment buttons used because those activities will necessitate that it resets back to the default "Combined Bounds of Selection" option. If for some reason the popover is open while the chosen layer is deselected (perhaps by keyboard shortcut like <kbd>Ctrl</kbd><kbd>Shift</kbd><kbd>A</kbd> to deselect), it should immediately switch back to "Combined Bounds of Selection" because the menu is actively open. Essentially, the chosen value should *only* be preserved in the event of deselection while the menu is closed and the user isn't doing alignment actions.
- [ ] If there are no artboards (infinite canvas is in use), Artboard alignment mode should be disabled (and if that's currently the active mode, it should be reset back to Selection alignment mode using the same rules as above where it only resets once the popover is viewed or the alignment buttons are used, allowing the user to create an artboard again and not lose the state which would be frustrating).
- [ ] Artboard alignment should be relative to the parent artboard of each layer, not the last-selected artboard
- [ ] This PR breaks the options bar for the Select tool, making it empty upon loading the editor until a layer is selected
- [ ] To reproduce the issue I mentioned above where the `-` or other incorrect layers shows in the layers dropdown, here is one of many ways you can perform some steps to run into these problems (but just select and deselect various layers, you'll immediately run into it since it's extremely easy to trigger):
    - Draw some shapes and refresh the web page
    - Select one layer
    - Switch to Layer alignment mode
    - Click in the canvas to deselect the layer
    - Use <kbd>Ctrl</kbd><kbd>A</kbd> to select all layers
    - View the alignment list again and it should be a dash now, despite having more than one layer selected
    - Keep clicking around (or multi-selecting in the Layers panel) to select and deselect various layers and you'll see that it starts showing a dash, or only one layer that isn't the selected layer, or other assortments of the wrong results

nico-mayora:
> Artboard alignment should be relative to the parent artboard of each layer, not the last-selected artboard

What if the selection spans several artboards and "Preserve relative positions" is checked?

Keavon:
Good question. Let's say it preserves relative positions of each per-artboard group of layers.

PR #1725, opened 2024-04-06
Fix shallow/deep selection

### LINKED ISSUE #1723: Fix Deep/Shallow selection regression
At some point not that long after @0HyperCube helpfully rebuilt Deep/Shallow Select in 208e907fa4b561f28bc874211a43f4a762777157, it seems to have broken as a result of a regression ().

image

Deep Select only selects the shallow group, not the deepest layer in the nested groups. Shallow Select's double clicking causes the double-clicked layer to disappear.

Maybe someone can bisect the Git history to find when that occurred (it was observed to be occurring prior to March, so Jan-Feb are the months to look). This would be a good issue for someone to find and fix.

Tiger3018:
Fix: remove folder (introduced in #1595, `GraphicElementRendered::add_click_targets`, thanks to @0HyperCube ) when clicked (`DocumentMessageHandler::click`), except for shallow select.

This should also fix other tools, like Fill tool which hook Lmb hints.

Todos:

* ~shallow select is not working when long-interval clicks. (Should be same as <https://71e86a06.graphite.pages.dev/>)~

  * This is caused by the detection of double click. ALL double click works fine, except for one state:
  
    Only double click at the first shallow selection will also deepen one level (one -> shallow select, one -> deepen one level). This behavior is reproduced in the previous fine PR(71e86a06), so won't fix in this PR.

* shallow hover is showed same as deep hover. (*Whether to implemented?*)

Closes #1723

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -10,7 +10,7 @@ use crate::messages::portfolio::document::node_graph::NodeGraphHandlerData;
 use crate::messages::portfolio::document::overlays::grid_overlays::{grid_overlay, overlay_options};
 use crate::messages::portfolio::document::properties_panel::utility_types::PropertiesPanelMessageHandlerData;
 use crate::messages::portfolio::document::utility_types::clipboards::Clipboard;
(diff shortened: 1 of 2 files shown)


Amq2705 (approved):

Keavon (approved):

Keavon:
@Tiger3018 thanks for this. Please leave a comment saying "Done" in the issue #1723.

PR #1727, opened 2024-04-10
Fix interactive outlining of layers within the Select tool's box selection

### LINKED ISSUE #1707: Don't outline hovered layers when box selecting
Currently when box selecting with the select tool, the hovered layer is outlined. This is confusing as other layers that are also part of the box selection are not outlined. 
confusing outlining when box selection

To fix, the hovered layer should not be outlined when box selecting. Perhaps all the layers within the current box selection should instead be highlighted. The outline comes from `overlay_context.outline` in the `SelectToolMessage::Overlays` handler.

shyamjayakannan:
Closes #1707 

@0HyperCube, @Keavon  please let me know if this makes it behave as intended or if changes are required and I may have misunderstood what was required to be done.

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -437,12 +437,12 @@ impl Fsm for SelectToolFsmState {
 				// Update pivot
 				tool_data.pivot.update_pivot(document, &mut overlay_context);
 
-				// Check if the tool is in Selection Box mode
+				// Check if the tool is in box selection mode
 				if matches!(self, Self::DrawingBox { .. }) {
-					// get the updated box bounds
+					// Get the updated selection box bounds
 					let quad = Quad::from_box([tool_data.drag_start, tool_data.drag_current]);
 
-					// Outline layers to be selected
+					// Draw outline visualizations on the layers to be selected
 					for layer in document.intersect_quad(quad, &document.network) {
 						overlay_context.outline(document.metadata().layer_outline(layer), document.metadata().transform_to_viewport(layer));
 					}
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -407,13 +407,6 @@ impl Fsm for SelectToolFsmState {
 					overlay_context.outline(document.metadata().layer_outline(layer), document.metadata().transform_to_viewport(layer));
 				}
 
-				// Get the layer the user is hovering over
-				let click = document.click(input.mouse.position, &document.network);
-				let not_selected_click = click.filter(|&hovered_layer| !document.selected_nodes.selected_layers_contains(hovered_layer, document.metadata()));
-				if let Some(layer) = not_selected_click {
-					overlay_context.outline(document.metadata().layer_outline(layer), document.metadata().transform_to_viewport(layer));
-				}
-
 				// Update bounds
 				let transform = document
 					.selected_nodes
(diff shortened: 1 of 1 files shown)


editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -432,9 +425,25 @@ impl Fsm for SelectToolFsmState {
...
+					}
+
+					// Update the selection box
+					overlay_context.quad(quad);
+				} else {
+					// Get the layer the user is hovering over
+					let click = document.click(input.mouse.position, &document.network);
+					let not_selected_click = click.filter(|&hovered_layer| !document.selected_nodes.selected_layers_contains(hovered_layer, document.metadata()));
+					if let Some(layer) = not_selected_click {
+						overlay_context.outline(document.metadata().layer_outline(layer), document.metadata().transform_to_viewport(layer));
```
Tiger3018, line 445: To avoid multiple `overlay_context.outline(...)` in this if .... else .... brace, a mutable Vec<layer_identifier> outside this if/else brace might be useful.

Tiger3018 (changes requested):

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -432,9 +425,25 @@ impl Fsm for SelectToolFsmState {
 				// Update pivot
 				tool_data.pivot.update_pivot(document, &mut overlay_context);
 
-				// Update dragging box
+				// Check if the tool is in Selection Box mode
 				if matches!(self, Self::DrawingBox { .. }) {
-					overlay_context.quad(Quad::from_box([tool_data.drag_start, tool_data.drag_current]));
+					// get the updated box bounds
+					let quad = Quad::from_box([tool_data.drag_start, tool_data.drag_current]);
```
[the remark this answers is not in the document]
shyamjayakannan: I have replaced it with `tool_data.selection_quad()`

editor/src/messages/tool/tool_messages/select_tool.rs
shyamjayakannan: @Tiger3018 hope the changes are what you require. Thanks for replying

Keavon: Sorry I wasn't able to code review this until now. @shyamjayakannan please revert the change requested by @Tiger3018 because it's less efficient and adds complexity than the good solution you had already chosen. Thanks.

Keavon:
!build

editor/src/messages/tool/tool_messages/select_tool.rs
Tiger3018: This code change is just fit to my suggestion earlier. However, towards the solution of `let mut Vec`, a simple `let xx = if` can also do the same thing while keeping the efficiency and simplicity. See [this answer](https://stackoverflow.com/a/59917403).

The second situation can be wrapped by `vec![]`. Because it should only have one layer result in the near future, I don't have a idea if this could introduce code latency @Keavon , since I have encountered multiple codes in Graphite like this.

0HyperCube: @Tiger3018 the linked answer only works if all branches give the same type (which here it wouldn't easily). Allocating on the heap does have some cost (although not particularly significant).

shyamjayakannan: @Keavon I have reverted the changes as requested. Thanks for the reply

Keavon:
!build

Keavon (approved):

PR #1740, opened 2024-05-07
Loosen the Graphene type system to allow contravariant function arguments

### LINKED ISSUE #1741: RFC: Contravariance for node inputs with automatic coercion:
**RFC: Contravariance for node inputs with automatic coercion:**
*Current Status:*
Currently the types supplied to functions are treated as invariant.
If node A is chained with node B and B expects an input of `T → V`, the type of node A has to be exactly `T → V`. 
*Suggested Changes:*
Node inputs should be allowed to be **contravariant**. That means if the signature of A can be `T' → V` as long as T' ∝  T (With ∝ denoting the subtyping relation) = as long as T is more general than T'.
We then need to define a subtyping relation to have `()` to be the root of that relation.  In practice that means that any node which expects `T → V` can also take a node with `() -> V` as input.
*Implementation:*
For the implementation we have to consider two things: the type checking and the type coercion.
It should be pretty easy to extend the type checking to support contravariant function arguments.
Type coercion is a bit tricker, when using rustc for compilation we could just express this relationship using traits but due to the dynamic execution that is infeasible as it would lead to a type explosion for the node implementations.
Instead we'll have to insert explicit nodes to perform the conversion which work much like the cull node does currently.
*Performance Impact:*
When compiling with rustc this does not cause any runtime overhead. When linking the graph dynamically, it add the cost of calling an extra node. The compilation overhead is expected to be minimal.
*Open questions:*
How do we communicate this to the user / developer? When a chain A - B - C of nodes is created and A, C are resolution aware and B is not resolution aware, the typesystem would complain that the link between A and B is invalid even though node B itself is the problem, somewhat obfuscating the error.
*Alternatives:*
* Manually placing conversion nodes as necessary and adding more tooling to fix special cases like the bottom of a layer stack.
* Forcing everything to be resolution aware (this complicates caching, has a negative impact on performance and would be confusing to reason about)

TrueDoctor:
This partially implements the contravariant function arguments RFC by permitting to call nodes with any input if the node expects to receive `()` as an input

Closes #1741

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/document_node_types.rs
+++ b/editor/src/messages/portfolio/document/node_graph/document_node_types.rs
@@ -326,7 +326,7 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 			category: "Structural",
 			implementation: DocumentNodeImplementation::Network(NodeNetwork {
 				imports: vec![NodeId(0), NodeId(0)],
-				exports: vec![NodeOutput::new(NodeId(2), 0)],
+				exports: vec![NodeOutput::new(NodeId(1), 0)],
 				nodes: [
 					DocumentNode {
 						name: "Load Resource".to_string(),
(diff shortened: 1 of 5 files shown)


Keavon (approved):

Keavon:
- **Load Image** regression (works in master), however this works when the **Transform** node isn't included (you just have to zoom into the 1x1 image at the artboard origin):
  image
- **Noise Pattern** has the same regression:
  image
- **Text** doesn't have regressions, but oddly just... doesn't work when you add it from the node graph, but it does work when you add it from the Text tool and give it the same options. This is also the case in master, but I'm mentioning it here regardless.
  image

Keavon:
I think I'll merge this now, though, and let you solve those regressions in a followup PR. Since I'd like to do a release today.

PR #1750, opened 2024-05-10
Node network subgraph editing

### LINKED ISSUE #1172: Creating user-defined subgraphs
- Double-clicking nodes (except those based on proto nodes) enters their subgraph, replacing what's currently viewed in the node graph
- A breadcrumb trail shows which node you're in relative to the document's root network
- The user can edit the nodes that make up a network-based document node
- The current "Output" node is generalized to support multiple graph exports, and an "Input" node is added to support (multiple) graph imports
- All the built-in nodes are given sane X/Y coordinates so users viewing those node internals aren't left with a mess
- Question for @TrueDoctor: what do we do with `DocumentNodeImplementation::Extract` nodes— just make them not double clickable as with `DocumentNodeImplementation::ProtoNode`?
- Do we need a way to edit the network's document node's type signature and Properties panel entries?

Future work (for another issue and PR):
- A way to edit a document node definition, not just a document node instance
    - This has no parent network, so unlike editing an instance, the currently-hooked-up imported/exported data has no effect on the wire colors inside the graph for generic data flow (it'll just be gray)
    - A way to define additional metadata and Properties panel layout for the definition
    - A way to save an entire definition in a nice compact and human-readable file format so we can replace the long document node definitions file with a folder each containing its own definition file in whatever format we choose (RON?)

### LINKED ISSUE #1764: Visualize data type wire and connector colors based on actual data
Currently we hard-code the colors for the data types of nodes and their output wires. This should be made dynamic based on the actual concrete data in usage.

### LINKED ISSUE #1765: Connector input index skips hidden inputs
When disconnecting an input, the input index is determined by finding the index clicked connector from all visible connectors. This means that hidden inputs are not accounted for, and hiding an input shifts the index of all inputs below up by one. For example, when the "Dimensions" input is exposed but the "Location" is hidden for the Artboard node, it is impossible to disconnect the "Dimensions" input. Another example is when trying to disconnect the combine/split channel nodes. To solve this the index of input could be added to `FrontendGraphInput`, and it would be stored it as a data attribute on the port.

Another important consideration is that the index when connecting is correctly determined, its only broken when disconnecting.

adamgerhant:
Closes #1172, closes #1765, and closes #1764. Discussion in [this Discord thread](https://discord.com/channels/731730685944922173/1238015208534966313).

Double-clicking nodes (except those based on proto nodes) enters their subgraph, replacing what's currently viewed in the node graph.

Todo
- [x] A breadcrumb trail shows which node you're in relative to the document's root network
- [x] Determine input/output type from compiled node network in order to display nodes without document definition. 
- [x] Remove "Imports" field for node network, and instead store import index in each `NodeInput::Network`. This allows multiple nodes to use the same import field, and enables creation of the "Imports" node
- [x] Create a UI only "Imports" node, used for customizing the network import index for any node
- [x] Create a UI only "Exports" node, used for customizing the exports of a node
- [x] Save root node for each network, which is an `Option<NodeId>` default node that `DocumentMetadata` will use to display/update/add layers
- [x] Seperate NodeGraphMessages and GraphOperationMessages to ensure that tool and panel operations only affect document network, while graph UI operations only affect the selected network.
- [x] Add restrictions for selecting across different levels of nested networks

Issues:
- [x] Previews disappear when entering a nested network 
- [x] Properties panel doesn't show information for selected layers in the document network when entering a nested network.
- [x] Ensure selected nodes are always in the same network
- [x] Auto node connection when dragging it onto a link creates an infinite loop that freezes the program
- [x] Auto node connection does not work between import and some node
- [x] Can't select import/export nodes
- [x] Root to export connection is doubled
- [x] Hide "Preview" for node not in current network, but show "Make Hidden"
- [x] Properties panel does not work when not in the document network
- [x] Fix deleting node outside of nested network
- [x] Remove placeholder and replace with type name from definition if it exists (Boolean operation), if not then use name of data type.
- [x] Set root node to None if the primary export should be used as root
- [x] Can't move artboard in layer panel
- [x] Drawing artboard when there is a layer adds the layer to primary input, not secondary
- [x] New artboards are not added in layer stack, they are offset to left
- [x] Deleting layer does not delete children
- [x] Export input type not correct when previewing ToArtboard protonode
- [x] Swapping inputs for isolated nodes breaks them
- [x] Copy/cut and paste is broken in document network
- [x] Fix "Preview"/"Stop Preview" buttons
- [x] Errors don't display on node graph
- [x] Auto reconnect does not work when when deleted node is between import node and network node
- [x] Deleting node between import and export causes crash since it tries to reconnect, which is currently not supported.
- [x] Deleting layer inside of nested network randomly deletes children
- [x] Copy pasting node exposes all inputs
- [x] Add colors for internal types such as "GraphicElement"
- [x] Swap node inputs for ConstructLayer and AddArtboardNode
- [x] Fix thumbnails not generating when entering and exiting a network
- [x] Copy and pasting nodes with network input retains their input. 
- [x] Disconnecting between combine and split channels node doesn't work #1765
- [x] Copy and pasting protonodes does not resolve default tagged value from compiled network. They are instead set to None. Will require a rework for copying nodes (`map_ids`)
- [x] Inserting a non manual composition node, such as the "Multiply" node between the imports and some node crashes (can't reproduce)
- [x] Deleting split channels that is connected to combine channels crashes
- [x] Fix demo artwork

Deferred:
- [ ] Add positions for all `DocumentNodeDefinition` internal networks
- [ ] Reset transform in Graph.svelte when entering nested network - will fix in next PR when moving node graph state to Rust
- [ ] Resolved types for the Import node (which is derived from the parent node) is not saved for hidden node inputs. Nodes with manual composition also may have strange behavior #1767 

Change to consider:
- If network cannot be rendered, remove all types from node graph, since they are from the last render and do not represent the current state


Keavon (approved):
Whoo, what a herculean effort you've put in with this one! What a valuable addition to making the node graph so much more complete now.

PR #1752, opened 2024-05-13
Update Git Attributes for JSON syntax highlighting in `*.graphite` files

spenserblack:
This is just the type of small PR that I always compulsively do when I encounter a project with a new filetype :laughing:

This will apply the JSON syntax highlighting to *some* `*.graphite` files on GitHub (syntax highlighting is disabled for large minified files). It marks *most* `*.graphite` files as binary and generated, hiding diffs locally and in GitHub PRs. It reverts this for the graphite test files, since they seem to be manually written.

Currently, whether or not syntax highlighting is applied to `test_files/*.graphite` seems to almost be random (linguist's Bayesian classifer timing out perhaps :shrug:), but you can see in my fork that syntax highlighting is always applied.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,3 +4,6 @@
 # The `eol=lf` sets the conversion to an LF line ending
 # https://git-scm.com/docs/gitattributes
 * text=auto eol=lf
+
+# Adds syntax highlighting to Graphite files on GitHub and minimizes diffs both locally and on GitHub
+*.graphite binary linguist-generated linguist-language=JSON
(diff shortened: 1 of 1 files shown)


spenserblack:
> You mention that your fork's repo files show the JSON syntax highlighting but when I navigate to https://github.com/spenserblack/Graphite/blob/master/demo-artwork/isometric-fountain.graphite I don't see syntax highlighting. Did you mean somewhere else or something?

Yeah, I was referring to the [test files](https://github.com/spenserblack/Graphite/blob/fcc05213dadd6dde20d36854426742d289978422/node-graph/graphene-cli/test_files/levels.graphite). Unfortunately, because the "real" Graphite files resemble minified JSON, I don't believe they will get syntax highlighting, except possibly for very small files. So I think this is mostly helpful for development/debugging.

Keavon:
I see, makes sense! Thanks for this contribution.

Keavon (approved):

PR #1754, opened 2024-05-18
Fix off by one subpath and unrelated crash

0HyperCube:
Fix my off by one subpath creation and also an unrelated crash from broken delete points code.


0HyperCube:
I don't have time to debug further.

0HyperCube:
I cannot get any crashes with this last commit.

Keavon (approved):
This works well, thank you!

PR #1755, opened 2024-05-20
New node: Rasterize

### LINKED ISSUE #1270: Node to rasterize vector graphics
Using the arbitrary canvas targets for rendering introduced in #1256, we can easily repurpose that to render SVG content and read back the rendered canvas data. This lets us avoid needing to switch to Vello quite so soon.

TrueDoctor:
Closes #1270


Keavon:
- [ ] We need the default to be `X: 0 px`, `Y: 0 px`, `W: 100 px`, `H: 100 px`, `Resolution: 100%`
- [ ] We need the `Resolution` to not update when `W` and `H` are changed:

https://github.com/GraphiteEditor/Graphite/assets/4388688/86c47d1a-af17-4b81-990c-3abd72c42822
- [ ] The aspect ratio gets broken when the `W` or `H` changes, until manually editing and committing a change to the `Resolution` box even if the value is the same; however shifting the `X` and `Y` values doesn't update it:

https://github.com/GraphiteEditor/Graphite/assets/4388688/b5b0ce4b-6a83-464d-8745-7fd918aa83ce

Keavon (approved):

PR #1759, opened 2024-05-26
Add boolean operations

### LINKED ISSUE #1091: Boolean operations node
The legacy boolean operations were unstable, and are no longer compatible with the migration of vector data to nodes (#1065) so they had to be removed. But they're very important in the artist's workflow, so re-adding them as a node is high priority.

In order of importance, this should:
- [x] Either use web-sys to call Paper.js, or finish the attempted Paper.js Rust port in [this branch](https://github.com/GraphiteEditor/Graphite/tree/paperjs-boolean-ops)
- [x] Introduce a vector operations node operating on vector data in the graph
- [x] Hook up the boolean buttons in the Select tool's options bar so they take the selected vector layers and combine them into one graph set up so they feed into the boolean operations node

Each of these would be suitable for a single or separate PRs depending on the author's approach.

Keavon:
### *Boolean operations: where life imitates art*

image

https://github.com/GraphiteEditor/Graphite/assets/4388688/ee495016-59b6-412e-b484-a5c93200d787

Closes #1091

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2329,6 +2329,7 @@ dependencies = [
  "serde_json",
  "tokio",
  "url",
+ "usvg",
  "vello",
  "vulkan-executor",
  "wasm-bindgen",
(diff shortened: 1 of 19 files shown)


TackerTacker:
That's awesome! Thank you. 
Literally changes the tool from useless to usable for me.
Any word on when this will be live?

Keavon:
Indeed! It's been vitally needed to be considered a real editor.

Unfortunately, by a rare occurrence, our self-hosted CI/CD server is down and we'll need @TrueDoctor to wake up and fix it. Once that is fixed (hopefully within 24 hours), this change should be live on https://dev.graphite.rs and then we'll probably aim to do a release in the next few days to https://editor.graphite.rs .

Keavon:
Ok, it's live now on https://dev.graphite.rs (@TackerTacker).

PR #1760, opened 2024-05-26
Restore functionality of Imaginate

### LINKED ISSUE #1511: Reimplement the Imaginate tool
Following the migration to the unified node graph, the previous Imaginate functionality has changed. We no longer rasterize the area below the Imaginate frame now that compositing is part of the single document graph.

For now, users can use the Imaginate feature by inserting the Imaginate node in the graph inline with the flow of raster data, like so:

image

The Imaginate tool should be rebuilt to work based on the new graph-based layer compositing system with a draggable frame. Until #1510 is done, it can work left-to-right as a node, but we ideally want it to be its own layer that stacks vertically so it can modify its contents underneath itself in the stack.

The Imaginate node also needs to be able to rasterize vector data fed into it, since it only supports raster for the moment. Deciding precisely how it handles its frame bounds in different circumstances is still up for design discussion.

Keavon:
Closes #1511


TrueDoctor:
I think this can probably be closed, most of the relevant parts of the code have been merged in other prs, the only thing worth extracting would be the cloudflare api interaction, Thoughts @Keavon ?

Keavon:
I plan to extract the API parts once the deadlock is fixed, then I'll either close or reuse this PR for the API stuff.

PR #1763, opened 2024-05-27
Fix tool drawing start point offset when an artboard has a transform

0HyperCube:
Removes offset to the initial point when drawing with a tool, especially the Freehand tool, as reproduced with these steps:
- Move the artboard away from the origin by dragging with artboard tool
- Draw with the Freehand tool and notice how the first point has the artboard transform offset

A regression introduced by #1712.

### DIFF
--- a/editor/src/messages/tool/tool_messages/artboard_tool.rs
+++ b/editor/src/messages/tool/tool_messages/artboard_tool.rs
@@ -157,7 +157,11 @@ impl ArtboardToolData {
 		};
 
 		let center = from_center.then_some(bounds.center_of_transformation);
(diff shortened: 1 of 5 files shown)


Keavon (approved):
Thank you!

Keavon (changes requested):
Wait, sorry, I just tested the build link and it doesn't seem to fix the bug.


https://github.com/GraphiteEditor/Graphite/assets/4388688/48bbd2e9-1675-4736-8354-6a74e6ae2ebd

Keavon (approved):
This works great now, thank you for the amendments to fix both aspects of the bug! 💯

PR #1776, opened 2024-06-05
Run Raw-rs tests only on CI

elbertronnie:
Created a new feature `raw-rs-tests`. Tests inside raw-rs will only run if this feature is enabled. Workflow files are updated accordingly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: 🧪 Run Rust tests
         run: |
-          mold -run cargo nextest run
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
 
       - name: 🧪 Run Rust tests
         run: |
-          mold -run cargo nextest run
+          mold -run cargo nextest run --all-features
```
Keavon, line 90: @TrueDoctor is this necessary? Why wasn't it already set?

Keavon (approved):

PR #1780, opened 2024-06-11
Add npm to recommended packages to install

LukeFranceschini:
You need npm to run the next step in the getting started instructions, so it should be on the list of packages to be installed.

### DIFF
--- a/website/content/volunteer/guide/getting-started/_index.md
+++ b/website/content/volunteer/guide/getting-started/_index.md
@@ -20,13 +20,13 @@ git clone https://github.com/GraphiteEditor/Graphite.git
 On Debian-based (Ubuntu, Mint, etc.) Linux distributions, you may need to install the following packages:
 
 ```sh
-sudo apt install libgtk-3-dev libsoup2.4-dev libjavascriptcoregtk-4.0-dev libwebkit2gtk-4.0-dev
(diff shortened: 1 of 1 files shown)


Keavon:
Don't you think this is already sufficiently covered by the first paragraph of that page?

> Graphite is built with Rust and web technologies. Install the latest LTS version of [Node.js](https://nodejs.org/) and stable release of [Rust](https://www.rust-lang.org/), as well as [Git](https://git-scm.com/).

That list of Linux-specific dependencies is for other weird things that really ought to just be included by the tooling (and is on other OSs). Adding `npm` to that list of system dependencies seems out of place. I don't think it would be appropriate to mix in `npm` there, since the reader should have already followed the steps above to install it; or may already have npm installed; or may prefer a different way to install it like through `nvm`.

LukeFranceschini:
Oh hmm, I thought that LTS meant the one that's in the package manager for my system (apt on Ubuntu), so I  just used `sudo apt install nodejs` directly. I guess I should have clicked through the link and followed those specific instructions instead of just installing the base `nodejs` package. I'll try using the website instructions instead and go from there.

I haven't used Node.js (or really any javascript framework) before, so thanks for your patience and understanding! : )

Keavon:
No worries. I've found that, while it's the least convenient OS to install npm on, using nvm (node version manager) is the last inconvenient way.

PR #1791, opened 2024-06-18
Fix PathStyle::clear_fill() doctest in style.rs

LukeFranceschini:
This should allow `cargo test` to pass successfully when run in `graphite/node-graph/gcore` (previously it failed on `is_some()` because `clear_fill()` doesn't return an `Option`).


Keavon (approved):
Thanks for this!

PR #1792, opened 2024-06-19
Fix bounding boxes of VectorData with rotations in several nodes

LukeFranceschini:
The VectorData::bounding_box() function does not apply the VectorData's transform, and bounding_box_with_transform() must be used instead of applying the transform to the computed bounding box. It may make sense in the future to move the current version of bounding_box to a private method bounding_box_no_transform() and have the public bounding_box call bounding_box_with_transform so that external code doesn't need to know about how the VectorData stores it's data.

This fixes the following nodes:
- Repeat
- Circular Repeat
- Bounding Box

Also add a test to the bounding box node for this situation.

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -76,7 +76,9 @@ fn repeat_vector_data(vector_data: VectorData, direction: DVec2, angle: f64, ins
 	// Repeat the vector data
 	let mut result = VectorData::empty();
 
-	let Some(bounding_box) = vector_data.bounding_box() else { return vector_data };
+	let Some(bounding_box) = vector_data.bounding_box_with_transform(vector_data.transform) else {
+		return vector_data;
+	};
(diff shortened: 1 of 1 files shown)


Keavon:
@0HyperCube transforms are your area of expertise so I'd love if you could take a moment to review this, thanks!

0HyperCube (approved):
Thanks for this fix @LukeFranceschini. In regard to making the bounding box with no transform private, this would break much of the editor. For example the select tool takes the layer space bounding box and then transforms that into view space (so when selecting a rotated shape, the bounding box is also rotated).

This change is definitely desirable @Keavon as currently the bounding box node doesn't produce a proper bounding box for an input with a rotation transform (as we take the bounding box in layer space and then transform just two of the corners to form an incomplete AABB).

Furthermore the repeat node with a direction of zero and an angle currently rotates around the layer space centre without any transform which breaks as soon as any translation is applied before the repeat. This means that the centre of rotation is not where it is expected.

Keavon:
!build

PR #1794, opened 2024-06-21
Add layer node chains, import/export edge connectors, and refactor graph editing to go thru a NodeNetworkInterface

### LINKED ISSUE #1891: Implement layer chains
Nodes connected only to one another in a primary flow, connecting into a layer's left input, should be encapsulated by the layer's left "glass" boxed area.

### LINKED ISSUE #1892: Abstract all graph manipulations through a node network API

### LINKED ISSUE #1893: Move network imports/exports from the "pseudo" nodes to the graph left/right edges
When panning, it will stick to the edges and the wire will grow/shrink to reach between the graph and the sticky import/export connectors; however they will get pushed offscreen when the nodes move offscreen as well.

It should look like this.

image

adamgerhant:
Closes #1892, closes #1891, closes #1893

Moves `NodeNetwork` to a private field and creates an API for modifying the network. This API contains the various functions for manipulating the network, automatically organizing it, and caching.

Todo
- [x] Zoom to fit all is broken
- [x]  Add UI distinction between primary selection and all other selections
- [x] Fix lock/hide icons
- [x] Can't move layers
- [ ] Shift selecting multiple layers does not visually expand the selection box to include all layers.

Crashes
- [x] If artboard is not a layer and a shape is drawn
- [x] Deleting all nodes and a shape is drawn
- [x] Moving layer to bottom of stack in layer panel
- [x] Select layer, ctrl c, ctrl v, move layer, ctrl z, memory access out of bounds crash
- [ ] Toggling layer to node in nested network

Deferred:
- Move wire paths to Rust to fix disconnecting, display lines as vertical/horizontal
- Remove node_graph_to_viewport and derive with calculate_offset_transform from the node_graph_ptz. 
- Automatically snap the viewport to keep export side bars aligned with the grid
- Add x offset for top of child stack (be able to move stacks left from their parent, but not overlap the wire)
- Automatically add nodes into chain if they are vertically aligned, within 3 spaces, and no additional primary outputs
- Only snap layers on the x axis when converting to stack position
- Store document metadata structure as transient metadata
- Fix deleting previewed nodes
- Add vec of network types to NodeNetwork, rather than storing type for each NodeInput::Network
- Drag selection in node graph runs document and checks for artwork intersections on every frame 
- "To Artboard" has incorrect displayed types


TrueDoctor:
Getter methods in rust should not use the `get_` prefix

TrueDoctor:
I'm noticing a massive decrease in responsiveness while panning around I think the weird behaviour is likely caused by a combination of bad performance and the cache invalidation logic

adamgerhant:
That would have to be due to the Svelte rendering processes, since the wasm portion of each frame when panning accounts for just 0.2 ms in both dev and this branch. Currently dev is taking ~200ms a frame when panning in Red Dress, while my branch is taking ~120ms.

Keavon:
It appears to me that the color change for selected layers when either panel becomes the active or inactive panel isn't working anymore.

Keavon (approved):
Yet another Herculean development completed. You're on fire! This edited like 1/8 of our entire codebase 🤯. Way to go!

PR #1799, opened 2024-06-29
Simplify build process by auto-installing npm packages and simplify the contributor guide

0HyperCube:
- Highlight how manually installing the correct version of wasm-bindgen prevents it from being re-downloaded each build.
- Clarify that the linux packages are in addition to the dependencies listed above them.


Keavon:
This is helpful, thanks!

Keavon (approved):

TrueDoctor:
In my gpu restoration pr I bumped `wasm-bindgen` to version `0.2.92`  does that cause any issues (other than updating the docs)?

PR #1802, opened 2024-07-04
Integrate Vello for vector rendering

### LINKED ISSUE #1094: Replace SVG renderer with Vello
Currently, an SVG string is dumped into the web page and the browser renders it. This issue involves replacing that approach with a `<canvas>` element drawn onto by the Rust backend, with rendering of the vector scene provided by [Vello](https://github.com/linebender/vello). It should likely be implemented as a node where Vello takes the vector scene and produces a texture to be drawn. There are also additional places in the Graphite code where JS is currently used to rasterize the SVG scene, that code should be removed. This extends to various other supporting tech like blob URLs which will also be no longer necessary, meaning a good amount of JS code may be cleaned up as a result.

TrueDoctor:
Closes #1094


editor/src/messages/dialog/preferences_dialog/preferences_dialog_message_handler.rs
```diff
@@ -45,6 +45,15 @@ impl PreferencesDialogMessageHandler {
 				})
 				.widget_holder(),
 		];
+		let use_vello = vec![
+			TextLabel::new("Input").min_width(60).italic(true).widget_holder(),
```
0HyperCube, line 49: This probably shouldn't be under input. Also it might be worth writing "experimental" or something in the UI.

Keavon: Yeah, I'd recommend:
"Renderer" and "Vello (experimental)"

TrueDoctor: > This probably shouldn't be under input. Also it might be worth writing "experimental" or something in the UI.

Yeah already have exactly that change locally :D

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1141,26 +1142,13 @@ impl MessageHandler<DocumentMessage, DocumentMessageData<'_>> for DocumentMessag
...
 				} else {
 					self.node_graph_to_viewport.insert(self.node_graph_handler.network.clone(), transform);
-
-					responses.add(FrontendMessage::UpdateNodeGraphTransform {
-						transform: Transform {
-							scale: transform.matrix2.x_axis.x,
-							x: transform.translation.x,
-							y: transform.translation.y,
-						},
-					})
```
0HyperCube, line 1160: You broke panning in the node graph overlay.

PR #1804, opened 2024-07-06
Fix multi-subpath boolean operations

### LINKED ISSUE #1803: Fix boolean operations with several subpaths
When one of the layers in the boolean operation has several subpaths, the result is not correct. This is due to Paper.js returning an incorrect result; for example:
```js
console.info(booleanOperation("M0,0 L100,0 L100,100 L0,100Z", "M50,50 L60,50 L60,150 L50,150Z M70,50 L80,50 L80,150 L70,150Z", "unite"));
```
boolean result
Red is the first path, green is the second path, white is the result (note the added curves in the union).

0HyperCube:
Closes #1803.


Keavon:
I love it when the fix is simple like this! Good job investigating the root of the problem.

Keavon (approved):

PR #1806, opened 2024-07-06
Attempt to fix the segfault in tests

0HyperCube:
We have been getting segfaults in the testing [e.g. in this PR](https://github.com/GraphiteEditor/Graphite/actions/runs/9819112631/job/27112356540#step:13:5126). I think it is probably related to how we now run the node graph to test that the demo artwork renders. It appears that the demo art test returns ok `test dispatcher::test::check_if_demo_art_opens ... ok` before a segfault `SIGSEGV [   1.771s] graphite-editor dispatcher::test::check_if_demo_art_opens`.

The `NodeContainer` has a raw pointer that is freed on drop, yet can be cloned. This means that if the cloned instance is dropped then the original may still contain a dangling pointer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: 🧪 Run Rust tests
         run: |
-          mold -run cargo nextest run --all-features
+          mold -run cargo nextest run --all-features -j 1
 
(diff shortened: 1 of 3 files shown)


Keavon:
Superseded by #1814.

PR #1807, opened 2024-07-07
Fix 'Zoom with Scroll' not working after a page refresh

imor:
This fixes a bug in which the preferences dialog's _Zoom with Scroll_ option reverted to the default behavior after a page refresh. 

Steps to reproduce:

1. Open File -> Preferences and select _Zoom with Scroll_.
2. Open a document and try zooming with the mouse scroll wheel without pressing the Ctrl key and observe that the zoom works.
3. Refresh the page.
4. Open a document (if the document you opened the last time was empty, you'd need to open a new document, otherwise the old document will open itself, in which case you don't need to open another document).
5. Try zooming with the mouse scroll wheel again and observe that the zoom doesn't work (it pans instead). Try with pressing the Ctrl key and observe that it works.
6. Open File -> Preferences and observe that the Zoom with Scroll option is still selected.

The bug is due to the _Zoom with Scroll_ option not affecting the key mapping when deserialized preferences are first loaded. Sending a `PreferencesMessage::ModifyLayout { ... }` fixes this.

### DIFF
--- a/editor/src/messages/preferences/preferences_message_handler.rs
+++ b/editor/src/messages/preferences/preferences_message_handler.rs
@@ -40,7 +40,7 @@ impl MessageHandler<PreferencesMessage, ()> for PreferencesMessageHandler {
 					responses.add(PortfolioMessage::ImaginatePreferences);
 					responses.add(PreferencesMessage::ModifyLayout {
 						zoom_with_scroll: self.zoom_with_scroll,
-					})
+					});
 				}
 			}
 			PreferencesMessage::ResetToDefaults => {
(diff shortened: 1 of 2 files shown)


Keavon:
Nice find and fix! Thank you for this. We should be good to merge shortly, I just have a code stylistic nitpick: can you please put a semicolon at the end of your message dispatch to match the others?

editor/src/messages/preferences/preferences_message_handler.rs
```diff
@@ -38,6 +38,9 @@ impl MessageHandler<PreferencesMessage, ()> for PreferencesMessageHandler {
 					responses.add(PortfolioMessage::ImaginateServerHostname);
 					responses.add(PortfolioMessage::ImaginateCheckServerStatus);
 					responses.add(PortfolioMessage::ImaginatePreferences);
+					responses.add(PreferencesMessage::ModifyLayout {
+						zoom_with_scroll: self.zoom_with_scroll,
+					})
```
Keavon, line 43:
```suggestion
					});
```

Keavon:
!build

imor:
Thanks for adding the semicolon. I'm fixing the failing tests. It's taking a bit of time setting up testing tools (cargo-nextest, mold etc.) for the first time.

Keavon:
You didn't break the tests, it's a race condition we have which #1806 is currently trying to solve. I'll just ignore CI on this one and merge it, then CI will probably break a few times on master which I'll just have to keep re-running until it eventually gets lucky with a pass.

PR #1812, opened 2024-07-07
Drag upstream nodes

adamgerhant:
Adds the ability to automatically drag upstream nodes. Fixes bug when deselecting a node with shift+click.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message.rs
@@ -67,6 +67,7 @@ pub enum NodeGraphMessage {
 	MoveSelectedNodes {
 		displacement_x: i32,
 		displacement_y: i32,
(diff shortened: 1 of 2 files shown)


Keavon:
Can you please adjust this to ignore layer bottom inputs? Dragging a layer should only move it and its left inputs, not those below it.

adamgerhant:
I made the change, but wouldn’t that mostly be done by the layer/stack locking. I thought this was useful because if there are a bunch of overlapping layers you can drag and first layer and all layers below it would also move.

Keavon:
Can you also please make it de-conflict the amplified movement that we currently get when a node/layer is selected while another upstream node/layer is also selected?

https://github.com/GraphiteEditor/Graphite/assets/4388688/52e22bd1-b90a-4a13-b388-0c2e43f578b9

Keavon (approved):

PR #1818, opened 2024-07-12
Fix snapping bug where snapped-together shapes couldn't be moved

### LINKED ISSUE #1817: Snapped-together layers get permanently stuck together and unmoveable while snapping is enabled
This is a recent regression.

In this video, there also appears to be another regression I also discovered while recording this which is that rectangles don't snap together when translated with the Select tool when they overlap edges, however when scaled it does work.

https://github.com/GraphiteEditor/Graphite/assets/4388688/75234359-5df8-4f97-b3ae-2835a630d088

I've disabled snapping by default in a17ed68008ab9d16f929fac2d23fc98367d7e151 as a mitigation for this high-severity bug. It should be reenabled by reverting the change to the `snapping_enabled` field in `misc.rs` and in the five demo art files.

0HyperCube:
Closes #1817


Keavon (approved):
Thanks!

Issue #1140, opened 2023-04-17
Dragging images into a fresh editor


Keavon:
If a user drags an image into the welcome screen (shown when zero documents are open), it should create a new document with an artboard matching the dimensions of the image, and the image placed at 100% scale with the top left corner at the document's origin coordinates. Pasting an image from the clipboard should have the same effect.

Additionally, when the welcome screen is not shown, users should be able to drag an image into the tab bar (the dark horizontal bar occupied by the tabs, including the tabs themselves).

All this should also work for .graphite files.

Related: we'd like to support URLs (similar to <https://editor.graphite.rs/#demo/red-dress>) where we can open a bitmap or SVG as a new document. Care should be taken to ensure security for potentially malicious files that could be loaded and opened by the editor (someday stealing account info, for example, if the user is logged into Graphite?).

abumsoo:
Hi! I can work on this if it's not being worked on already

Keavon:
Sure, go for it! Just please be careful about which places in the frontend that the drop handler should exist. When on the welcome screen (no docs open), it should be everywhere. When a document is open, it needs to exclude parts related to the currently open document (so at least the viewport) but still allow it being dropped in the tab bar and other editor areas. In the layer panel would probably be fine for now but eventually that should probably be inserted into the currently open document equivalent to dropping it in the viewport of the document (feel free to work on that change too if you'd like).

You can also reuse this functionality to open .graphite files dropped into the editor.

abumsoo:
Currently, trying to open an non-graphite image file by clicking on 'Open Document' gives an error. Do we want 'Open Document' to have the same behavior as what's described in this issue?

abumsoo:
Understood. I think it does filter by `.graphite` files as I had to manually change the filter to "All Files". File > Import seems to be disabled on the welcome screen (if that's what you mean by empty editor).

Keavon:
Yeah, you'd have to reenable File > Import on the welcome screen once that functionality is added.

abumsoo:
I would like to remove myself from this task as I wasn't able to make much progress and I will be busy for the next couple of weeks. I don't want to keep someone else from taking on this task. But maybe I'll pick it up again in a few weeks if it's still here.

Keavon:
Sounds good @abumsoo, thanks for the update.

PR #1842, opened 2024-07-20
Allow dragging an image into the welcome screen

abumsoo:
Fixes part of #1140 

So far, this enables the ability to drag an image into the welcome screen to open a new document with the same dimensions and paste the image onto a new layer. 

I think I will add tests soon, but would appreciate some feedback in the meantime. I'm not sure I like how I named the new message `OpenImageAsBackground`. I also reused a lot of the code from `PasteImage`.

Todo:
- [ ] enable dragging into tab bar
- [ ] allow .graphite files to be dropped in
- [ ] allow dropping into layer panel

### DIFF
--- a/editor/src/messages/portfolio/document/document_message.rs
+++ b/editor/src/messages/portfolio/document/document_message.rs
@@ -93,6 +93,9 @@ pub enum DocumentMessage {
 		image: Image<Color>,
 		mouse: Option<(f64, f64)>,
 	},
+	PasteImageAsBackground {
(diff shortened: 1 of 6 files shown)


TrueDoctor:
A general way to avoid code duplication with the message system is to either use functions, or have one message dispatch another one. You could for example add a PasteImageWithSize message which is called by both the PasteImage and PasteImageAsBackround message

TrueDoctor:
!build

PR #1846, opened 2024-07-23
Sandbox node graph execution on native targets and attempt recovery from panics on Wasm

TrueDoctor:
This implements panic catching for native targets using the rust std `catch_unwind` function. As this does not work on wasm yet: https://github.com/rust-lang/rust/issues/118168, I've implemented a hacky workaround which should give the user the ability to still save their work when the editor crashes and recover some of their work

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2449,6 +2449,7 @@ dependencies = [
  "serde",
  "serde_json",
  "specta",
+ "spin",
  "thiserror",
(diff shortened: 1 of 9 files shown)


TrueDoctor:
Note to test this you can enable the vello feature which currently just calls a panic, this will of course need to be removed before merging

Keavon (approved):

PR #1847, opened 2024-07-23
Fix Layers panel UI not updating on selection change and expand/collapsing

### LINKED ISSUE #1848: Update LayerPanel after Layer selection changed, including opacity and group expand/collapse

TrueDoctor:
Also fixes a panic on empty layer.

Closes #1848


Keavon:
It looks like expanding/collapsing layers in the Layers panel still isn't updating. Right now you have to, for example, rename a layer to get it to send the UI update to show the changed expansion state.

Keavon (approved):

PR #1856, opened 2024-07-24
Serialize documents with images in linear space instead of sRGB

### LINKED ISSUE #1851: Images dragged in the editor get washed out every time the document is reloaded (gamma issue)

TrueDoctor:
Closes #1851 

Breaking: This changes the document format for images and will break existing save files


Keavon:
Users having images in a save file is unlikely, because raster isn't well-supported yet and those images would have gotten progressively more and more washed out anyways over time. This breaking change is acceptable.

Keavon (approved):

PR #1874, opened 2024-07-29
Apply opacity and blend modes to individual layers in Vello

### LINKED ISSUE #1873: Fix Vello rendering of opacity and blend modes (see Isometric Fountaintain)

TrueDoctor:
Closes #1873 

image
image

### DIFF
--- a/node-graph/gcore/src/graphic_element/renderer.rs
+++ b/node-graph/gcore/src/graphic_element/renderer.rs
@@ -348,6 +348,20 @@ impl GraphicElementRendered for VectorData {
 		use crate::vector::style::GradientType;
 		use vello::peniko;
 
+		let multiplied_transform = transform * self.transform;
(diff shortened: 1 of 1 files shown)


TrueDoctor:
> Your screenshot of 🌵 still appears broken.
@Keavon 
This is the same behaviour as in Firefox. If this is a violation of svg spec we should investigate if we should report this as an error upstream

Keavon:
Ok, digging into this further, those black shapes appearing are expected and Chrome is indeed rendering this wrong to make them not appear. This is a useful test case, but it'll be good to retire this demo artwork very soon.

Keavon:
There's one weird rendering artifact now (tested in Chrome):

https://github.com/user-attachments/assets/1462e21a-0a26-4efc-be37-336d09fd758e

Keavon (approved):

PR #1875, opened 2024-07-29
Apply correct stroke styling with Vello

### LINKED ISSUE #1862: Apply correct stroke styles in Vello
We currently basically only translate the stroke weight into vello. The other options should be ported as well

TrueDoctor:
Closes #1862

### DIFF
--- a/node-graph/gcore/src/graphic_element/renderer.rs
+++ b/node-graph/gcore/src/graphic_element/renderer.rs
@@ -327,9 +327,10 @@ impl GraphicElementRendered for VectorData {
 
 	fn bounding_box(&self, transform: DAffine2) -> Option<[DVec2; 2]> {
 		let stroke_width = self.style.stroke().map(|s| s.weight()).unwrap_or_default();
+		let miter_limit = self.style.stroke().map(|s| s.line_join_miter_limit).unwrap_or(1.);
 		let scale = transform.decompose_scale();
 		// We use the full line width here to account for different styles of line caps
-		let offset = DVec2::splat(stroke_width * scale.x.max(scale.y));
+		let offset = DVec2::splat(stroke_width * scale.x.max(scale.y) * miter_limit);
 		self.bounding_box_with_transform(transform * self.transform).map(|[a, b]| [a - offset, b + offset])
 	}
 
--- a/node-graph/gcore/src/graphic_element/renderer.rs
+++ b/node-graph/gcore/src/graphic_element/renderer.rs
@@ -328,7 +328,8 @@ impl GraphicElementRendered for VectorData {
 	fn bounding_box(&self, transform: DAffine2) -> Option<[DVec2; 2]> {
 		let stroke_width = self.style.stroke().map(|s| s.weight()).unwrap_or_default();
 		let scale = transform.decompose_scale();
-		let offset = DVec2::splat(stroke_width * scale.x.max(scale.y) / 2.);
+		// We use the full line width here to account for different styles of line caps
+		let offset = DVec2::splat(stroke_width * scale.x.max(scale.y));
 		self.bounding_box_with_transform(transform * self.transform).map(|[a, b]| [a - offset, b + offset])
 	}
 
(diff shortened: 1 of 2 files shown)


node-graph/gcore/src/graphic_element/renderer.rs
```diff
@@ -423,10 +424,26 @@ impl GraphicElementRendered for VectorData {
...
+			};
 			let stroke = kurbo::Stroke {
 				width: stroke.weight,
 				miter_limit: stroke.line_join_miter_limit,
-				..Default::default()
+				join,
+				start_cap: cap,
+				end_cap: cap,
+				dash_pattern: stroke.dash_lengths.into(),
+				dash_offset: stroke.dash_offset,
```
Keavon, line 446: There's a mismatch in behavior with how dash_offset affects dash_pattern: 

https://github.com/user-attachments/assets/220484a6-e428-402f-9f0e-7e98f8c4c64f

Please also change the properties layout code such that:
- You fix a bug I just found with `dash_offset` where negative values should be valid, but currently it's getting clamped to 0 when you try to type a negative value.
- Dash Offset is grayed out (`.disabled(true)` is set on its widget builder pattern) when Dash Lengths is empty.
- Miter Limit is grayed out when Line Join isn't set to Miter.

node-graph/gcore/src/graphic_element/renderer.rs
```diff
@@ -423,10 +424,26 @@ impl GraphicElementRendered for VectorData {
...
+			let cap = match stroke.line_cap {
+				LineCap::Butt => Cap::Butt,
+				LineCap::Round => Cap::Round,
+				LineCap::Square => Cap::Square,
+			};
+			let join = match stroke.line_join {
+				LineJoin::Miter => Join::Miter,
+				LineJoin::Bevel => Join::Bevel,
+				LineJoin::Round => Join::Round,
+			};
```
Keavon, line 438: The shape is getting cropped when a miter join is used. Be aware that the user can create arbitrarily sharp points with arbitrarily high Miter Limit values. The cropping thing will always need to be able to handle that.

Keavon (changes requested):

node-graph/gcore/src/graphic_element/renderer.rs
TrueDoctor: I did apply a gap of one line width, to the transform, If that is not sufficient we'll most likely have to add some sort of magic constant to account for the oversizing

node-graph/gcore/src/graphic_element/renderer.rs
TrueDoctor: It might be that this is an upstream `vello` issue, I don't see how my code would break this in any way

TrueDoctor: Added the properties panel improvements

node-graph/gcore/src/graphic_element/renderer.rs
Keavon: Definitely not a constant, since as I said, it can be arbitrarily sharp. It'll require some kind of equation to calculate.

node-graph/gcore/src/graphic_element/renderer.rs
Keavon: If you think that's a Vello issue, can you please ask them about it?

node-graph/gcore/src/graphic_element/renderer.rs
TrueDoctor: This is fixed and resolved

node-graph/gcore/src/graphic_element/renderer.rs
TrueDoctor: https://xi.zulipchat.com/#narrow/stream/197075-gpu/topic/Dashed.20Line.20misbehaves.20when.20used.20with.20dash-offset/near/455494308

Keavon:
The stroke needs to be expanded to avoid cropping when there's an Opacity node on the shape (as long as it's less than 100%).

Keavon (approved):

PR #1878, opened 2024-07-30
Refactor document node type lookup function to fix performance degradation over time

### LINKED ISSUE #1913: Performance leak when many layers are in a scene
It initially accounts for half the frame time, and gets worse and worse over time until you reload the editor.

TrueDoctor:
Closes #1913

### DIFF
--- a/.github/workflows/ci-and-dev-build.yml
+++ b/.github/workflows/ci-and-dev-build.yml
@@ -89,7 +89,7 @@ jobs:
 
       - name: 🧪 Run Rust tests
         run: |
-          mold -run cargo nextest run --all-features
+          mold -run cargo test --all-features
 
   # miri:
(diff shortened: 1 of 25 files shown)


TrueDoctor:
This is now only bound on mem copys for sending the data from the node runtime to the frontend. This indicates that we might need a diff based api to fully solve this issue

TrueDoctor:
The document type lookup which took previously about 50% of the frame time on the Sakura artwork has now basically disappeard from the flamegraph:
image

TrueDoctor:
Noticed a bug where when opening the same document twice and then closing one of them, types disappear

TrueDoctor:
Fixed

Keavon (approved):
Incredible improvement! And nice attention to detail on the doc comments, too.

PR #1885, opened 2024-08-02
Yarn test

0HyperCube:
Testing yarn package manager.


TrueDoctor:
I've also heard people like pnpm do you have any insight into your choice to try yarn?

0HyperCube:
I tried yarn simply because I had worked on a project with it previously.

PR #1887, opened 2024-08-02
Use a coarse bounding box to avoid a detailed check for intersection when clicking artwork

0HyperCube:
A simple bounds check for the intersection which improves performance slightly.


node-graph/gcore/src/graphic_element/renderer.rs
```diff
@@ -25,9 +25,15 @@ use vello::*;
 pub struct ClickTarget {
 	pub subpath: bezier_rs::Subpath<PointId>,
 	pub stroke_width: f64,
```
Keavon, line 27: I think it would be wise to make these private and force their manipulation to go through the API in order to keep the bounding box from getting outdated.

TrueDoctor (approved):

PR #1894, opened 2024-08-04
Add quick measure overlays with Alt pressed

0HyperCube:
Part of #1896

Some edge cases ~~to consider (I'm not sure what anyone else does here since this isn't in Inkscape)~~ **(explained in the newly filed issue above)**:
- [ ] Rotated bounding box
- [ ] Rotated hover target
- [ ] Intersecting bounding boxes between the selected and hovered
- [ ] How we choose between [horizontal then vertical] or [vertical then horizontal] for drawing the lines (currently we always do horizontal then vertical from the selected layer).
- [ ] How we do text when the document is rotated >90

demo

### DIFF
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -260,4 +260,26 @@ impl OverlayContext {
 			.fill_text(text, pos.x + padding, pos.y - padding - metrics.font_bounding_box_descent())
 			.expect("draw text");
 	}
+
+	pub fn angle_text(&self, text: &str, pos: DVec2, direction: DVec2, padding: f64, pivot: Pivot) {
+		self.render_context.translate(pos.x, pos.y).expect("translate");
+		let angle = -direction.angle_to(DVec2::X);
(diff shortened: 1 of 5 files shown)


Keavon:
I filed an issue with detailed diagrams for each scenario: #1896

To answer your questions:

> - Rotated bounding box
> - Rotated hover target

It should operate based on the AABB (in document space). We also want those bounding boxes (for both the selected and hovered shapes) to be drawn when holding Alt to make it clear what's being measured against what (so even if the shapes aren't rectangles, the measurement lines in the issue's diagrams actually connect to something instead of terminating confusingly in empty space).

> - Intersecting bounding boxes between the selected and hovered
> - How we choose between [horizontal then vertical] or [vertical then horizontal] for drawing the lines (currently we always do horizontal then vertical from the selected layer).

See the diagrams in the issue.

> - How we do text when the document is rotated >90

The measurement lines and AABBs for the shapes should all get tilted along with the viewport navigation PTZ tilt, but the actual text labels should remain upright for easier readability.

Keavon (approved):

PR #1908, opened 2024-08-06
Restore the Imaginate node with the full node graph architecture (but a flaky deadlock remains)

TrueDoctor:


Keavon (approved):
Hopefully, as soon as possible, we can fix the deadlock issue, then I can polish up the remaining features and add more server backends once you're able to help fix the freeze. But it makes sense to merge this now. Thanks for helping me to revive this much-delayed and important feature.

PR #1909, opened 2024-08-07
Make noise generation resolution aware

TrueDoctor:
Implements footprint based noise generation. I originally intended to also fix the Mandelbrot node in this pr but it turns out  that the node itself was never broken and it only stopped working due to an regression somewhere else which has since been fixed.

# Changes:
- Remove dimension parameter
  The dimensions are new inferred from the footprint
- Add clip parameter
  The user can now choose if they want to constrain the noise 
# Alternatives considered
- Make dimension optional
 We could instead make the dimension parameter optional, which would eliminate the need for a transform node after the noise node. This might provide a nicer to use api and could be implemented in a follow-up pr.
 
 We now only render what is visible on the screen and also support clipping to the viewport itself:
 
image

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/document_node_types.rs
+++ b/editor/src/messages/portfolio/document/node_graph/document_node_types.rs
@@ -776,48 +776,10 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 			identifier: "Noise Pattern",
 			node_template: NodeTemplate {
 				document_node: DocumentNode {
-					implementation: DocumentNodeImplementation::Network(NodeNetwork {
-						exports: vec![NodeInput::node(NodeId(1), 0)],
-						nodes: vec![
-							DocumentNode {
-								inputs: vec![
-									NodeInput::network(concrete!(()), 0),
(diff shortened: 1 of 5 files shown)


Keavon:
Nice! That's also great to see the Mandelbrot node has gotten fixed already too. My observation is that we'll want to remove the "Clip" parameter so it's always boundless, and always have it be infinite, then use clipping masks or other nodes to actually constrain the bounds as the user desires. We'll also want to inverse the Scale parameter, I think.

TrueDoctor:
> Nice! That's also great to see the Mandelbrot node has gotten fixed already too. My observation is that we'll want to remove the "Clip" parameter so it's always boundless, and always have it be infinite, then use clipping masks or other nodes to actually constrain the bounds as the user desires. We'll also want to inverse the Scale parameter, I think.

Doing this would hurt the performance as you would generate a bunch of useless information which only later gets clipped. This would even evaluate the texture if the artboard is not even in view.
> We'll also want to inverse the Scale parameter, I think.
So using `1./scale` ? We could instead rename scale to frequency, as we now just pass the value through instead of dividing it by some magic constant

Keavon (approved):

PR #1911, opened 2024-08-07
Add grid snapping to graph imports/exports; improve layer panel drag into/between insertion; better preserve graph space on reordering

adamgerhant:
Reworks the default positioning system when moving nodes in the layer panel, mostly by ensuring that node gap positioning is correctly maintained. Each layer is the height of that layer plus the height to the next layer below. Fixes all edge cases when moving layers or different groups of selected layers to any other location. Also improves the previewing system and import/export edges

Todo:
- Fix crash when deleting all nodes inside a network
- Create space when moving layers to different stacks

### DIFF
--- a/editor/src/consts.rs
+++ b/editor/src/consts.rs
@@ -1,5 +1,7 @@
 // Graph
 pub const GRID_SIZE: u32 = 24;
+pub const EXPORTS_TO_EDGE_PIXEL_GAP: u32 = 120;
+pub const IMPORTS_TO_EDGE_PIXEL_GAP: u32 = 120;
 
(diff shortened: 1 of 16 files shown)


Keavon:
One more edge case when snapping needs to get engaged: scrolling with the horizontal mouse scroll wheel or horizontal scroll bar.

Keavon (approved):
Ready to merge after you fix that edge case.

Keavon:
One more very minor, harmless edge case I found: when you open a demo artwork, then open the graph, its export is in the wrong place before you pan for the first time.

PR #1921, opened 2024-08-10
Add colors to all nodes in a graph, even if disconnected, and properly display hidden network imports

### LINKED ISSUE #1767: Hidden inputs not added to BorrowTree inputs_source_map
Resolved types for the Import node (which is derived from the parent node) are not available for hidden node inputs. This is because although the hidden `Type`'s are stored in the `TypingContext`, the `document_node_types(&self) -> ResolvedDocumentNodeTypes` function only gets types for inputs in the `BorrowTree` inputs_source_map. To fix this, all inputs need to be added to the inputs_source_map. I believe this will require a rework to how exposed inputs are handled. Is there even a reason to consider whether the input is hidden when compiling into proto nodes? Hiding an input is a UI only concept, the value node still has the same functionality.

This image shows the error. The Type is a `(Concrete(())->Concrete(bool))`, and is correctly passed to the next node, but it is displayed as General
ApplicationFrameHost_Tip9Bq6YOi

I think this is purely visual, and doesn't affect any functionality. There is also some strange behavior where the last type becomes the type for the first hidden. In this example the location input has resolved type Concrete(())->Concrete(bool).

adamgerhant:
- Fixes input types not being displayed. 
- Reworks how types are determined by iterating to the next protonode.
- Displays types for nodes disconnected from network.  
- Fixes types being cleared when undoing/redoing
- Fixes types not being updated when connected to a different type.
- Fixes #1767

Deferred:
- Connect disconnect groups of nodes to the export in order to determine types for the connections between nodes in the disconnected group, rather than selecting at random. These types should be added to the `ResolvedDocumentNodeTypes`.

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -575,7 +575,6 @@ impl NodeNetworkInterface {
 		} else if let graph_craft::document::DocumentNodeImplementation::ProtoNode(protonode) = &node.implementation {
 			let node_id_path = &[network_path, &[*node_id]].concat();
 			let primary_output_type = self.resolved_types.types.get(node_id_path).map(|ty| ty.output.clone()).or_else(|| {
-				log::warn!("no compiled output type found for {:?} {:?}", node_id_path, &node.implementation);
 				let Some(node_io_hashmap) = NODE_REGISTRY.get(protonode) else {
 					log::error!("Could not get hashmap for proto node: {protonode:?}");
 					return None;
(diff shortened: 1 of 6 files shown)


editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -464,6 +500,104 @@ impl NodeNetworkInterface {
...
+				let mut all_node_io_types = node_io_hashmap.keys().collect::<Vec<_>>();
+				all_node_io_types.sort_by_key(|node_io_types| {
+					let mut hasher = DefaultHasher::new();
+					node_io_types.hash(&mut hasher);
+					hasher.finish()
+				});
+				let Some(node_types) = all_node_io_types.first() else {
+					log::error!("Could not get node_types from hashmap");
+					return None;
+				};
```
TrueDoctor, line 592: instead of doing a sort and then taking the first value, you could just use a min_by_key here

Keavon (approved):

TrueDoctor (approved):

Keavon:
I am getting this when I add the `Extract` node:

> error	frontend/wasm/src/lib.rs:71
> panicked at editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs:1626:60:
> Primary output should always exist

TrueDoctor:
Yeah, that is fine for now
will fix when we use that node again

Keavon:
I'm pretty sure there are other nodes for which this same crash would occur, probably anything without a primary output? I will hold off on merging for now. But if it's only that node, we can just remove it from the catalog (which should happen anyways).

PR #1925, opened 2024-08-12
Add a profiling action to CI which comments on PRs with notable demo art performance variances

TrueDoctor:


TrueDoctor:
We unfortunetly can't fully test this yet as the ci tries to compare against master, but master does not have the necessary benchmark setup yet. And the ci only shows significant changes  so we'll need to actually have changes to test against

Keavon (approved):

PR #1936, opened 2024-08-14
Fix all Clippy warnings

Keavon:


TrueDoctor (approved):

TrueDoctor:
To run perf on a criterion benchmark, you have to use the full path to the benchmark binary, e.g.: 
`/home/dennis/Projects/rust/Graphite/target/release/deps/compile_demo_art-c71307254ef46547 --profile-time 5 --bench`
With the profile-time and --bench flags. The help command no longer seems to work in the newer version of criterion

Issue #1889, opened 2024-08-03
Click targets get offset after paint with the boolean operation node


0HyperCube:
By creating two layers, applying a boolean operation, then panning the document, the click targets become offset for the original layers.

This occurs because the boolean layer node has a Memoize Impure node inside of it, so the click targets are not updated when the document is panned.

Click targets are stored in layer space. In order to compare with the mouse it is necessary to calculate the transform from layer to viewport. We do this using `transform_to_document` which in tern uses the footprint. Footprints are always stored in viewport space. By not reprocessing the monitor node that stores the footprint, the click targets become offset.

The easiest fix is to remove the memoize impore node from the boolean node. Keavon claims that without this Memoize impure node, the performance is too bad. I'm not sure how else to approach this.

adamgerhant:
Why the Footprint transform used instead of using the current document_to_viewport transform?
Instead of  `.map(|(footprint, transform)| footprint.transform * transform)`, `.map(|(_, transform)| self.document_to_viewport * transform)` can be used, which solved the issue.

PR #1945, opened 2024-08-23
Improve layer positioning in graph upon reordering; improve history system; add selection history

### LINKED ISSUE #706: Layer selection history
The Forward/Backward mouse navigation buttons should enable the user to move through selection history (similar to VS Code's text selection), selecting different sets of layers as they were throughout the past. Layers which don't exist anymore, but were selected in the past, presumably would just not end up being selected. (And if zero of layers that were selected during a selection history state in the past end up existing now, filter that history state out.)

[A note](https://stackoverflow.com/questions/57102502/preventing-mouse-fourth-and-fifth-buttons-from-navigating-back-forward-in-browse) about capturing the Forward/Backward mouse buttons.

### LINKED ISSUE #1910: Node graph better organization
I'm using graphite with multiple artboards in a single project, and the node graph is starting to get crowded :

An overview of a cluttered node graph

At this point, manual organization of nodes is tedious, and needs to be done regularly to account for new nodes.

### Automatically placement is messy, some nodes are placed on top of each other :

New nodes aren't placed in the most convenient of manners, and often need to be manually moved before being able to work with them :

Nodes being placed on top of each other

### New tools to add ?

Some additional features might make working with the node graph more convenient, such as :

- A way to group nodes into collapsible folders
- A way to quickly hide/show nodes based on selection

___

I found two related issues : #1820 and #1819 

#1820 lists the following shortcut which is indeed very useful : `Drag to move node (shift to drag row)`

### LINKED ISSUE #1943: Aborting a drag with a tool destroys redo history
Draw some things, undo a bit, redo to see how those can be redone, undo back again so there is still redo history, then draw and right click while drawing to abort. Notice that you can't redo anymore.

adamgerhant:
- Improves the automatic positioning of nodes in the graph when layers are reordered.
- Reworks transactions with start/end points to ensure redo history is maintained when aborting. Closes #1943
- Adds selection history alongside document undo/redo history (<kbd>Alt</kbd><kbd>[</kbd> and <kbd>Alt</kbd><kbd>]</kbd>, also the mouse forward/back buttons in Chrome). Closes #706

Also closes #1910.

Todo:
- [x] Alt+click and drag to duplicate selection
- [x] Improve the undo/redo system to fix <https://github.com/GraphiteEditor/Graphite/issues/1943>
- [x] Deleting should shift all nodes up by replacing single node selection with all upstream nodes
- [x] Adding a layer should create space first
- [x] Grouping artboard crashes
- [x] Cannot select with artboard tool
- [x] Selecting layer with non layer node between layer and export crashes

Deferred:
- Get footprints from RenderOutput to allow caching for any layer. Caching currently causes click targets to be desynced since the transform is not updated when panning.
- Prevent rendering the entire network when a layer is selected
- Solve #1889

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -354,8 +354,9 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
 					}
 					// Abort a box selection
 					if self.box_selection_start.is_some() {
-						responses.add(NodeGraphMessage::SelectedNodesSet { nodes: Vec::new() });
 						self.box_selection_start = None;
+						responses.add(NodeGraphMessage::SelectedNodesSet { nodes: Vec::new() });
+						responses.add(FrontendMessage::UpdateBox { box_selection: None });
 						return;
 					}
 					// Abort dragging a wire
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -484,7 +484,6 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
 				}
 
 				if let Some(clicked_id) = clicked_id {
-					responses.add(DocumentMessage::StartTransaction);
 					let Some(selected_nodes) = network_interface.selected_nodes(selection_network_path) else {
 						log::error!("Could not get selected nodes in PointerDown");
 						return;
@@ -533,6 +532,7 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
 						responses.add(NodeGraphMessage::SelectedNodesSet { nodes: updated_selected })
 					}
 
+					responses.add(DocumentMessage::StartTransaction);
 					return;
 				}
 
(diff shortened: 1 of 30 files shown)


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -42,22 +42,26 @@ pub struct NodeGraphMessageHandler {
...
 	shift_without_push: bool,
 	disconnecting: Option<InputConnector>,
 	initial_disconnecting: bool,
 	/// Node to select on pointer up if multiple nodes are selected and they were not dragged.
 	select_if_not_dragged: Option<NodeId>,
-	/// The start of the dragged line that cannot be moved, stored in node graph coordinates
+	/// The start of the dragged line (cannot be moved), stored in node graph coordinates
 	pub wire_in_progress_from_connector: Option<DVec2>,
-	/// The end point of the dragged line that can be moved, stored in node graph coordinates
+	/// The end point of the dragged line (cannot be moved), stored in node graph coordinates
```
Keavon, line 55: Should this say "cannot" or "can"? It previosly said "can".

Keavon:
Can you please add *Layer* > *Previous Selection (<kbd>Alt</kbd><kbd>[</kbd>)* and *Layer* > *Next Selection (<kbd>Alt</kbd><kbd>]</kbd>)* directly under (in the same section as) *Layer* > *Deselect All*?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
adamgerhant: it should say cannot because you cannot move the wire from the output port

Keavon (approved):

Issue #1854, opened 2024-07-24
Allow renaming documents by clicking on the document tab


TrueDoctor:
This task will be done by @Keavon and it isn't open to community contributions because of some subtle trickiness involved.

![image](https://github.com/user-attachments/assets/7cd1fd52-8373-4a49-a91f-52082473f515)

It should be possible to rename a document by double-clicking the tab and renaming it in place

PR #1948, opened 2024-08-31
Allow renaming tabs

vdawg-git:
Closes https://github.com/GraphiteEditor/Graphite/issues/1854

### DIFF
--- a/editor/src/messages/layout/utility_types/layout_widget.rs
+++ b/editor/src/messages/layout/utility_types/layout_widget.rs
@@ -7,6 +7,7 @@ use crate::messages::input_mapper::utility_types::input_keyboard::KeysGroup;
 use crate::messages::input_mapper::utility_types::misc::ActionKeys;
 use crate::messages::prelude::*;
 
+use derivative::*;
 use std::sync::Arc;
 
 #[repr(transparent)]
@@ -435,9 +436,12 @@ impl LayoutGroup {
 	}
 }
 
-#[derive(Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize, specta::Type)]
+#[derive(Debug, Clone, Derivative, serde::Serialize, serde::Deserialize, specta::Type)]
+#[derivative(PartialEq)]
 pub struct WidgetHolder {
 	#[serde(rename = "widgetId")]
+	// Required to prevent spurious updates from the popover (the id changes even if the content doesn't).
+	#[derivative(PartialEq = "ignore")]
 	pub widget_id: WidgetId,
 	pub widget: Widget,
 }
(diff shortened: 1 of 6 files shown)


Visual-Dawg:
The renaming works, however some other part of the app sometimes throws an error after a document got renamed:
```js
graphite_wasm.js:1255 error	
frontend/wasm/src/editor_api.rs:114While handling FrontendMessage ""UpdatePropertyPanelOptionsLayout"", 
JavaScript threw an error: JsValue(TypeError: Cannot use 'in' operator to search for 'columnWidgets' in undefinedTypeError: Cannot use 'in' operator to search for 'columnWidgets' in undefined    
at http://localhost:8080/src/wasm-communication/messages.ts:753:27    
at Array.reduce (<anonymous>)    
at http://localhost:8080/src/wasm-communication/messages.ts:752:42    
at Array.forEach (<anonymous>)    
at patchWidgetLayout (http://localhost:8080/src/wasm-communication/messages.ts:751:16)    
at http://localhost:8080/src/components/panels/Properties.svelte?t=1725132032461:286:4    
at callCallback (http://localhost:8080/src/wasm-communication/subscription-router.ts:25:9)    at Object.handleJsMessage (http://localhost:8080/src/wasm-communication/subscription-router.ts:33:5)    at http://localhost:8080/src/wasm-communication/editor.ts:18:19    
at http://localhost:8080/wasm/pkg/graphite_wasm.js:2388:37)
```

vdawg-git:
Sorry, in bed yesterday I noticed that the renaming should be cancel-able. 

The PR notifications remembered me to implement that

0HyperCube (commented):
Thanks for this contribution.; it works well so far.

I added some logging but I'm not able to reproduce your issue with the widget diffing in the properties. If you can reproduce it then please send the debug info.

Keavon:
It may be a few days before I can commit some changes that you can rely on to fix this so it behaves the same as layer renaming. I need to focus on other code priorities but I'll try to do that soon.

TrueDoctor:
@Keavon what is your status on this?

Keavon:
I haven't had time yet. It's probably a couple hours of work, but this feature isn't a priority since it adds no new functionality, so it hasn't been where I've put my focus in the past days. I'll try to avoid letting this rot for too long though.

Keavon:
Actually, on further thought, the hard part isn't what's done by this PR— it's the generalizing of the existing components. It requires unifying how the TextInput and TextLabel can behave like one another to support the exact renaming behavior seen currently, and as of yet exclusively, for layers in the Layers panel. Since, unfortunately, this PR doesn't take that approach, it isn't suitable for merging in its current state. But once I'm able to unify the behaviors of those two components and replace how it's used in the Layers panel, adding it to the document tab and sending that to the backend is about 10 minutes of work and would probably only use a few lines of code present in this PR. So it wouldn't actually make much sense to keep this PR around and reuse it when I have time for that, whenever that might be. The reality is that this is just a tricky frontend task that seems simple but isn't, so I don't suggest having it as a community-contributed task. Hopefully you got to learn a bit about the backend <-> frontend communication in this process, and sorry to waste your time with this @vdawg-git.

vdawg-git:
All good and thank you for the detailed explanation

PR #1949, opened 2024-09-01
Switch attribute-based vector data from referencing PointIds to indexes in the points table

### LINKED ISSUE #1888: Segment domain point reference as index
Currently it is very slow to iterate over the bézier segments in some VectorData (which we do a lot). This is because for each segment we iterate over, we must find the start and end position by resolving the PointId. This is done via linear search; it is slow when there are a large number of points and segments e.g. in the red dress demo artwork.

Instead of storing a PointId, the segment domain should store a usize index. This speeds up the common usecase of iterating over some segments, at the cost of having to update the indexes whenever a point is inserted or deleted. This tradeoff seems worthwhile given that we are already running into performance limitations of the current approach.

In order to facilitate procedural editing, the system of point ids will be kept.  Optimisations for the path modification node can be done seperately.

0HyperCube:
Closes #1888. Also fixes a few other issues e.g. dragging a handle after extending from the start of a path and some warnings.

### DIFF
--- a/node-graph/gcore/src/vector/vector_data/attributes.rs
+++ b/node-graph/gcore/src/vector/vector_data/attributes.rs
@@ -99,6 +99,7 @@ impl PointDomain {
 		let mut keep = self.id.iter().map(&f);
 		self.positions.retain(|_| keep.next().unwrap_or_default());
 
+		// TODO(TrueDoctor): Consider using a prefix sum to avoid this Vec allocation (https://github.com/GraphiteEditor/Graphite/pull/1949#discussion_r1741711562)
 		let mut id_map = Vec::with_capacity(self.ids().len());
 		let mut new_index = 0;
 		for id in self.ids() {
--- a/node-graph/gcore/src/vector/vector_data/attributes.rs
+++ b/node-graph/gcore/src/vector/vector_data/attributes.rs
@@ -646,7 +646,6 @@ impl<'a> Iterator for StrokePathIter<'a> {
 
 				break;
 			};
-			println!("At {point_index} start {current_start} value {val:#?}");
 
 			let mut handles = self.vector_data.segment_domain.handles()[val.segment_index];
 			if val.start_from_end {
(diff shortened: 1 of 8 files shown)


node-graph/gcore/src/vector/vector_data/attributes.rs
```diff
@@ -531,37 +560,122 @@ impl super::VectorData {
...
+				groups.push(bezier_rs::ManipulatorGroup {
+					anchor: self.vector_data.point_domain.positions()[point_index],
+					in_handle,
+					out_handle: None,
+					id: self.vector_data.point_domain.ids()[point_index],
+				});
+
+				break;
+			};
+			println!("At {point_index} start {current_start} value {val:#?}");
```
Keavon, line 649:
```suggestion
```

Keavon (approved):
The code looks great as far as I can tell, I like the nice small, atomic helper functions. I'll defer to @TrueDoctor for comment on the architectural implications about using indexes but I think that's probably good! I wasn't able to find any regressions in my QA testing but I also perhaps didn't know precisely what to look for.

node-graph/gcore/src/vector/vector_data/attributes.rs
```diff
@@ -95,9 +95,25 @@ impl PointDomain {
...
+				new_index += 1;
+			} else {
+				id_map.push(usize::MAX); // A placeholder for invalid ids. This is checked after the segment domain is modified.
+			}
+		}
+
+		let update_index = |index: &mut usize| *index = id_map[*index];
+		segment_domain.start_point.iter_mut().for_each(update_index);
+		segment_domain.end_point.iter_mut().for_each(update_index);
+
```
TrueDoctor, line 116: By building a prefix sum over the keep vector we should be able to avoid this allocation. But we can do that when benchmarking shows that this is a bottleneck

TrueDoctor:
!build-profiling

TrueDoctor:
image
We seem to spend a bunch of time writing to stdout?

TrueDoctor (approved):

PR #1952, opened 2024-09-02
Add path-bool library

TrueDoctor:
Closes #1759


Keavon (approved):
I'll do code cleanup as a followup PR so we can merge this now.

PR #1954, opened 2024-09-02
Raw-rs: Flip and rotate image based on camera orientation

elbertronnie:
- [x] Add Orientation Tag
- [x] Transform image based on camera orientation


Keavon:
For performance, it would be good to originally write to the image array with the transform, instead of the performance impact of swapping the pixel positions. That way the array doesn't need to be traversed another time. Is that feasible?

Keavon:
I'll merge this as-is so you can work on the pipeline architectural changes we discussed in our design meeting.

Keavon (approved):

PR #1957, opened 2024-09-04
Store network metadata information in the Node Network

adamgerhant:
Stores all network metadata as value inputs to cloned nodes. This will allow the new node graph rendering process to have access to all the information.


Keavon:
@adamgerhant do we close this? I don't recall what the status of this PR is, besides it's old and has a lot of merge conflicts. Did #2105 supersede this, or does this have a different focus from that? Is there anything this had set out to accomplish that is still needed to be accomplished today?

Keavon:
I'm going to assume the answer to my previous comment's question is yes, but let me know otherwise if this should be reopened.

PR #1942, opened 2024-08-20
Refactor the node macro and simply most of the node implementations

TrueDoctor:
Closes #1881

(Also fixes #1591)

PR #1965, opened 2024-09-07
Make the node graph use a document node's default type

0HyperCube:
Resolves an issue where random image frames where added.

I also added the TypeSource enum which is displayed in the input and output tooltips in order to help debug how types are generated.

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -171,7 +171,7 @@ impl NodeNetworkInterface {
 		layers
 	}
 
-	pub fn chain_width(&mut self, node_id: &NodeId, network_path: &[NodeId]) -> u32 {
+	pub fn chain_width(&self, node_id: &NodeId, network_path: &[NodeId]) -> u32 {
 		if self.number_of_inputs(node_id, network_path) > 1 {
 			let mut last_chain_node_distance = 0u32;
 			// Iterate upstream from the layer, and get the number of nodes distance to the last node with Position::Chain
@@ -414,7 +414,7 @@ impl NodeNetworkInterface {
 	/// Converts all node id inputs to a new id based on a HashMap.
 	///
 	/// If the node is not in the hashmap then a default input is found based on the compiled network, using the node_id passed as a parameter
-	pub fn map_ids(&self, mut node_template: NodeTemplate, node_id: &NodeId, new_ids: &HashMap<NodeId, NodeId>, network_path: &[NodeId]) -> NodeTemplate {
+	pub fn map_ids(&mut self, mut node_template: NodeTemplate, node_id: &NodeId, new_ids: &HashMap<NodeId, NodeId>, network_path: &[NodeId]) -> NodeTemplate {
 		for (input_index, input) in node_template.document_node.inputs.iter_mut().enumerate() {
 			if let &mut NodeInput::Node { node_id: id, output_index, lambda } = input {
 				if let Some(&new_id) = new_ids.get(&id) {
@@ -461,7 +461,7 @@ impl NodeNetworkInterface {
 	}
 
 	/// Try and get the [`Type`] for any [`InputConnector`] based on the `self.resolved_types`.
-	fn node_type_from_compiled(&self, input_connector: &InputConnector, network_path: &[NodeId]) -> Option<(Type, TypeSource)> {
+	fn node_type_from_compiled(&mut self, input_connector: &InputConnector, network_path: &[NodeId]) -> Option<(Type, TypeSource)> {
 		let (node_id, input_index) = match *input_connector {
 			InputConnector::Node { node_id, input_index } => (node_id, input_index),
 			InputConnector::Export(export_index) => {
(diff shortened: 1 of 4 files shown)


editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -453,114 +454,143 @@ impl NodeNetworkInterface {
...
+			log::error!("Could not get current network in input_type");
+			return None;
+		};
+		let Some(node) = current_network.nodes.get(&node_id) else {
+			log::error!("Could not get node {node_id} in input_type");
+			return None;
+		};
+		// If the input_connector is a NodeInput::Value, return the type of the tagged value.
+		if let Some(value) = node.inputs.get(input_index).and_then(|input| input.as_value()) {
+			return Some((value.ty(), TypeSource::TaggedValue));
```
adamgerhant, line 490: Im not sure if the type of the tagged value should directly be returned. It should be in the resolved types.

0HyperCube: It may be in the resolved types (unless the node is disconnected). However it will always resolve to the same type so we may as well just return here.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -453,114 +454,143 @@ impl NodeNetworkInterface {
...
-							}
-						});
-					if let Some((input_connector, _)) = downstream_connection {
-						Some(self.input_type(&input_connector, &node_id_path))
-					}
-					// Nothing is connected to the import
-					else {
-						Some(concrete!(()))
+	/// Try and get the [`DocumentNodeDefinition`] for a node
+	pub fn get_node_definition(&self, network_path: &[NodeId], node_id: NodeId) -> Option<&DocumentNodeDefinition> {
```
adamgerhant, line 458: Im not sure how True Doctors PR to automatically generate definitions for all nodes fits into this, but this function will also have to return those definitions.

0HyperCube: If you mean #1942, then that will be fine since the node definitions are added to the list used by the function here. Therefore no change will be necessary.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -453,114 +454,143 @@ impl NodeNetworkInterface {
...
+		};
+		// If the input_connector is a NodeInput::Value, return the type of the tagged value.
+		if let Some(value) = node.inputs.get(input_index).and_then(|input| input.as_value()) {
+			return Some((value.ty(), TypeSource::TaggedValue));
+		}
+		let node_id_path = [network_path, &[node_id]].concat().clone();
+		match &node.implementation {
+			DocumentNodeImplementation::Network(nested_network) => {
+				// Attempt to resolve where this import is within the nested network (it may be connected to the node or directly to an export)
+				let node_input_connection = nested_network
```
adamgerhant, line 496: use cached outward_wires instead of looping through all nodes/exports

editor/src/messages/portfolio/document/utility_types/network_interface.rs
adamgerhant: I believe that the plan for disconnected node groups is to execute them seperately in order to resolve all their types/thumbnails

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -453,114 +454,143 @@ impl NodeNetworkInterface {
...
-									if *import_index == input_index {
-										return type_from_node(node.1, network_node_input_index);
-									}
-								}
-							}
+				(input_type, TypeSource::RandomProtonodeImplementation)
+			}
+			DocumentNodeImplementation::Network(network) => {
+				for (child_id, child_node) in &network.nodes {
+					for (network_node_input_index, input) in child_node.inputs.iter().enumerate() {
```
adamgerhant, line 556: Outward_wires can also be used here

adamgerhant:
Another change to consider is using the nested type when determining color. I think this is why the monitor node is grayed out in the artboard node

0HyperCube:
> Another change to consider is using the nested type when determining color. 

The type in the tooltip is the same as the type used for the colour so this is already implemented @adamgerhant.

The issue with the monitor node is that it only the ArtboardGroup type is given a colour (not the Artboard type):
https://github.com/GraphiteEditor/Graphite/blob/d9ae01bf90f07c43edfe3e509a1a5c799168f229/editor/src/messages/portfolio/document/node_graph/utility_types.rs#L34

editor/src/messages/portfolio/document/utility_types/network_interface.rs
0HyperCube: I didn't refactor this existing code origionally because it requires the input_type having a mutable reference to the network interface. This necessitated rewriting the frontend graph input code.

adamgerhant (approved):
guess_type_from_node should eventually be considered an error once all nodes have definitions and default input types

Keavon:
This is what the graph looks like when you draw a shape and edit a point:

capture

Issue #1482, opened 2023-11-28
Stretch and tilt ruler increments when canvas is tilted


Keavon:
If the user tilts the canvas in the viewport, the scale of the increments becomes inaccurate. We want to essentially tilt the measuring system with the canvas. So for example, if the canvas is viewed with a 45° tilt, the increments become √2 = 1.414... wider.

Additionally, the horizontal and vertical ticks (for the vertical and horizontal rulers, respectively) should themselves become tilted to help visualize the fact that they are measuring tilted lines that, if conceptually extended into the document, would remain axis-aligned with document space. It's important that they actually align, so for example the increment ticks at `0` for both rulers would up pass through the document origin (if the ticks were extended beyond the rulers). The tick marks would be allowed to tilt up to 45°, at which point the axes would switch and the tilt would become less angled until 90° is reached whereupon the ticks are not tilted but the X and Y rulers mark the dimensions of the Y and X axes, respectively. This behavior of course continues with additional canvas tilt until a full rotation at 360° brings things back to how they started.

Care should be taken to ensure the ticks and numbers generally don't overlap, so the number positioning and tick density would need to intelligently adjust to avoid overlap depending on the canvas zoom level and tilt angle.

This is mostly a web, so Svelte/CSS/JS and figuring out some math and logic is required. Plumbing through the canvas tilt, thus extending upon the existing origin point and scale information sent by the backend, would be all the necessary Rust knowledge. [This PR](https://github.com/GraphiteEditor/Graphite/pull/1479) can provide a reference for what Rust code needs to change, and feel free to also reach out on Discord to ask for help with this aspect if you're unfamiliar with Rust.

HarmanHans:
Hey, I would like to try and contribute to this! I'm quite new to open source and Rust, so not sure how to best go about things. Thanks!

Keavon:
Hi @HarmanHans, I'd recommend we discuss this on Discord so you can get a better feel for what's involved and how that fits with your experience, or if there might be other better suited issues to take as a first task. Feel free to ping me in the #general channel.

HarmanHans:
Ok, sounds good!

GiacomoViolini:
Hello! This is my first contribution. Would it be possible to have a review?

jsjgdh:
Fixed in PR #3844

PR #1966, opened 2024-09-07
Set integer ruler intervals when zoomed in

### LINKED ISSUE #1927: Set integer ruler increments in the Node Graph
When the node graph is open, the ruler increments should be rounded to a whole number. This may require calculating an offset and/or custom tick spacing to ensure the ticks are aligned when the zoom is changed.
chrome_EGt9ReU4b0

hydrobeam:
## Summary

Closes #1927 

Adds a check for when the interval would be too small and then applies `.ceil()` to make it a whole number. Now, the progression of intervals is:

...200, 100, 50, 25, 12.5,7(6.25), 4(3.125), 2(...),1

## Discussion

I chose to start rounding at 12.5 since 12.5 is a fairly legible number. It only started looking weird below that. A slight issue is that 7 is used when 6.25 is rounded up, which isn't a clean number for the ruler :shrug:. I couldn't think of a clean way to get it to 8 or 6 without messing up the other numbers (i thought about something like: `ruler_interval += ruler_interval % 2` to force evenness, but then `ruler_interval` bottoms out at 2, which felt too big). 

## Results

When toggling the node graph display, the ruler now looks like this:
image

And for very high zoom percentages (> 60000%), the ruler tops out at an interval of 1. 
image

Previously, the long decimals would overlap each other and look bad:
image

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -797,10 +797,14 @@ impl MessageHandler<DocumentMessage, DocumentMessageData<'_>> for DocumentMessag
 				let ruler_origin = document_to_viewport.transform_point2(DVec2::ZERO);
 				let log = ruler_scale.log2();
 				let mut ruler_interval: f64 = if log < 0. { 100. * 2_f64.powf(-log.ceil()) } else { 100. / 2_f64.powf(log.ceil()) };
-				// when the interval becomes too small, force it to be a whole number.
-				// progression of intervals is: ...200,100,50,25,12.5,7(6.25),4(3.125),2(1.5625),1
-				if ruler_interval < 12.5 {
-					ruler_interval = ruler_interval.ceil();
+
+				// when the interval becomes too small, force it to be a whole number, then to powers of 10.
+				// progression of intervals is: ...,100,50,25,12.5,6(6.25),4(3.125),2(1.5625),1.0,0.1,0.01,...
+				if ruler_interval < 1.0 {
+					ruler_interval = 10.0_f64.powf(ruler_interval.log10().ceil());
+				} else if ruler_interval < 12.5 {
+					// round to nearest even number
+					ruler_interval = 2.0 * (ruler_interval / 2.0).round();
 				}
 				let ruler_spacing = ruler_interval * ruler_scale;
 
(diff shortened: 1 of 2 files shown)


Keavon:
!build

Keavon:
This is great! Nice work on implementing it in such a simple equation.

You're right that it's a bit weird for the progression to jump from increments of 12.5 to 7 to 4. Can you add a special case to turn the progression from the first line into the second line here?:
```
...200, 100, 50, 25, 12.5, 7 (6.25), 4 (3.125), 2 (1.5625), 1
...200, 100, 50, 25, 12.5, 6 (6.25), 4 (3.125), 2 (1.5625), 1
```

I'm also thinking that it would be great if, at some point when you zoom in more, we start to get the decimals for `.0`, `.1`, `.2`, ..., `.9` and continuing that at additional orders of magnitude (`.01`, `.001`, etc.). But that doesn't necessarily have to prevent this PR from merging, as it's already very useful in its current form. Would you like to take a crack at that though?

hydrobeam:
I wasn't sure about going lower than 1.0 before since going below 1 unit means you're already insanely zoomed in, but I've added it in [74f6647](https://github.com/GraphiteEditor/Graphite/pull/1966/commits/74f6647a47b50397e7599fcee2c1e1e02f3d2a77), also modified the progression to go to 6 and not 7.

Showcase of sub 1.0 scales:

https://github.com/user-attachments/assets/b3a506b7-cdc7-41a9-ab61-9a19a585cc97

The max zoom% seems to be 1million%, so you can't really go infinitely deep.

Also, I needed to add the `.toFixed()` stuff in `RulerInput.svelte` because there were some weird floating point shenanigans going on that I couldn't resolve:

image

So I opted to truncate the strings depending on how many decimal slots were expected.

Keavon:
!build

Keavon:
This is perhaps a separate issue but I'm bringing it up as a concern, even if it's not something you fix in this PR:

If the labeled increments aren't divisible by 5, we end up forcing sub-increments divided by 5 and 10 anyways, causing a misalignment. This is captured at 1500x document zoom and we see it's choosing to display the labels in increments of 6. Each gray bar is set at 1px increments in document space. We can see how they misalign because of the 6 bars between each labeled number, but only 5 minor (10 micro) increments.

capture

I'm open to suggestions about the right solution here, but I think it would basically mean either forcing the labeled numbers to always be increments of 5, or changing the number of minor and micro increments to divide evenly into the distances between the labeled increments.

Keavon (approved):
Thank you for this feature improvement! If you'd like to follow this up with the fix to the issue I mentioned above, you could mention the same issue so both this and that PR would be the ones that close the same issue (I don't think filing a separate issue makes sense). I'll also direct your attention to #1482 and #1969 as other ruler-related improvements that you might be interested in working on later. Cheers!

PR #1967, opened 2024-09-07
Fix transform bugs with primitive shape

moOsama76:
Closes #


Keavon:
!build

Keavon:
!build

Keavon:
It's looking like the height still goes negative when dragging in the upward direction from the start point of a shape. Could you fix that by actually normalizing the shapes? You can worry about the special case for odd-number-sided polygon/star shapes afterward (that's where the negative radius comes in).

PR #1968, opened 2024-09-09
Raw-rs: run tests in parallel

elbertronnie:
- Runs the tests in parallel by default in the CI which will be useful when more images are added.
- If the older sequential way is required (especially while print based debugging), enabling `RAW_RS_TEST_RUN_SEQUENTIALLY` env variable will make it sequential again.


TrueDoctor:
Are the tests even executed in the ci? Don't see them in the logs

TrueDoctor:
The code itself looks fine but I am somewhat doubtful this currently improve ci speeds but as a future proofing measure that should be fine

elbertronnie:
> Are the tests even executed in the ci? Don't see them in the logs
```
running 2 tests
test extract_data_from_dng_images ... ignored
test test_images_match_with_libraw ... ok

test result: ok. 1 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 25.29s
```

elbertronnie:
Here are the logs from #1954 when no optimization were present
```
running 2 tests
test extract_data_from_dng_images ... ignored
test test_images_match_with_libraw has been running for over 60 seconds
test test_images_match_with_libraw ... ok

test result: ok. 1 passed; 0 failed; 1 ignored; 0 measured; 0 filtered out; finished in 140.02s
```

So the time went from 140s to 25s in the CI.

elbertronnie:
@TrueDoctor I have modified the code in compile_demo_art.rs wherein I used `cfg_attr` instead of `attr`. It does compile correctly and there will be no need for the drop function too.

TrueDoctor (approved):
This should be good to merge once ci passes

PR #1972, opened 2024-09-12
Raw-rs: Refactor to run multiple steps in a single loop

elbertronnie:
- [x] Merge subtract black and scale colors steps
- [x] Merge demosiacing, convert to RGB and transform steps
- [x] Split scale colors into 2 steps


Keavon:
Approving with the assumption that we'll remove the `fortuples` dependency in a followup PR.

PR #1973, opened 2024-09-13
Make the primitive shape tools avoid setting a negative transform scale

moOsama76:
Fix negative height and unnecessary 180° rotation while drawing primitive tools, make odd-sided polygons and start flipped with radius instead of scale or rotation


moOsama76:
!build

0HyperCube (approved):
The code looks good; thank you @moOsama76. I extracted the star tool radius negation logic to a function.

Keavon:
!build

Keavon (approved):

PR #1975, opened 2024-09-14
Nudge only the shallowest selected layers to avoid amplified translation

0HyperCube:
Fixes the issue reported by @moOsama76 on discord whereby if a layer and its parent are selected and nudged, the nudge is applied twice.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -33,6 +33,7 @@
 	},
 	// Rust Analyzer config
 	"rust-analyzer.cargo.allTargets": false,
+	"rust-analyzer.check.command": "clippy",
 	// ESLint config
 	"eslint.format.enable": true,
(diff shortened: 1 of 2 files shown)


.vscode/settings.json
```diff
@@ -33,7 +33,6 @@
 	},
 	// Rust Analyzer config
 	"rust-analyzer.cargo.allTargets": false,
-	"rust-analyzer.check.command": "clippy",
```
Keavon, line 36: What was the reason for removing this? Feel free to mention it in #dev-tooling if this was intentionally committed and we can discuss its removal, but I'm currently not aware of any problems that it would cause. I'm going to remove it from this PR for the moment so it can be merged, as I'm assuming it was just inadvertently committed.

Keavon (approved):
💯

PR #1977, opened 2024-09-14
Correctly apply transforms to vector data and strokes

### LINKED ISSUE #1930: Make Fill and Stroke apply in-the-moment, not as part of the final rendered element
Currently, both Fill and Stroke apply a single choice of color to their respective fields in the VectorData. This means you can overwrite one with a later one. But we would rather make it apply the moment the data arrives at each Fill or Stroke node, pushing to a Vec of styles (fills and strokes). Additionally, the transform of the shape in that moment should become stored in the stroke or fill (when gradients are used, that's relevant to fills, too). This lets you have multiple fills (useful when transparency is involved, especially with gradients) and strokes (allowing you to have multiple strokes on a single shape drawn on top of each other). It should also be possible to put the fill on top of the stroke, or vice versa, depending on the ordering of each node in the graph (and thus the Vec of styles on the VectorData).

This change also lets us apply subsquent transforms to the shape that skew the earlier-set fill, without requiring a Merge layer to get involved. See #1931.

This will allow us to eventually make a user-facing Style node which acts as a subgraph encapsulating various orders of Fill and Stroke nodes letting the user only worry about styling from the Properties panel UI (setting one fill and one stroke, choosing which is on top of which, etc.) without cluttering the graph with multiple nodes.

See also #1834.

### LINKED ISSUE #1931: The Transform node sets its GraphicGroup's final transform, but should only affect the data-so-far in the graph flow
If you stick a Transform node underneath a layer stack, the whole stack gets moved even though, intuitively, it's applying to no data in the stack so far. If you stick the Transform node/layer in the middle of a stack, it affects both those layers above and below it. Definitely feels wrong.

That's because it is changing the final resulting GraphicGroup's `transform` field, instead of the transform on the individual layers up to that point which make up the partial stack (the part underneath the transform node).

The tricky part is preserving the correct behavior for rendering strokes (or gradient fills) which can get stretched and skewed.

This relates to #1930 and #1834.

0HyperCube:
As mentioned by @adamgerhant, adding a layer to a transformed group should not transform that layer.

Closes #1930 closes #1931


adamgerhant:
I don't think inverting the transform for graphic elements downstream of the transform node is the correct solution. It means that the downstream elements are rendered at a different size, then transformed to the size they should be. This means the strokes also get transformed. I think the best solution is to have the transform modify the transform of each graphic element, which I think should be the general implementation for any node on a group of data.

https://github.com/user-attachments/assets/6c8a6063-515a-406a-8dc9-44298ffadde8

0HyperCube:
@adamgerhant I have fixed this minor issue. I'm not sure why you believe your way to be more "correct" than what I've implemented here. However feel free to implement that anyway.

I also fixed some issues with the vello rendering of artboards etc so you might want to merge this anyway.

Keavon (approved):
Looks good! I just have to figure out how to update the Isometric Fountain demo art which has a weird gradient change-of-functionality that I'll just fix in the node graph and push to this before merging when I wake up tomorrow morning.

TrueDoctor (approved):
Code looks good

adamgerhant:
@0HyperCube My definition of correct is based on Keavons plan to have transforms directly modify the vector data. I personally disagree with this since it modifies referenced data

adamgerhant (approved):

Keavon:
This exhibits the correct behavior which is all that matters currently, and we can continue discussing the details of the data model changes that we'll do soon for spreadsheets/attributes, separately from this PR which is an excellent improvement for correctness.

Keavon (approved):
All good for gradients, merging now 😀

PR #1981, opened 2024-09-16
Use overlays to draw artboard names

### LINKED ISSUE #1863: Refactor artboard text label so it's an overlay
The artboard name drawn above the top left corner of every artboard is currently rendered as part of the artwork. But it isn't part of the artwork, it's an overlay, so the code should be refactored to make it an actual overlay.

0HyperCube:
Closes #1863


Keavon (approved):
Splendid, now the Vello renderer is identical to the browser SVG renderer! Thanks @0HyperCube :)

PR #1946, opened 2024-08-30
Render output clicktargets

adamgerhant:
Closes https://github.com/GraphiteEditor/Graphite/issues/1889

Get footprints from RenderOutput to allow caching for all layers. Caching currently causes click targets to be desynced since the transform from layer space to viewport space is not updated when panning.
 

Frame time for Rust execution in Painted Dreams currently (no caching): 78ms
Frame time with caching: 6.2ms

TODO:
Prevent rendering the entire network when a layer is selected
Fix using select tool on artboard

PR #1986, opened 2024-09-18
Fix many regressions introduced mostly in #1946

### LINKED ISSUE #1982: Text tool completly broken
The text tool is completely broken:
- The typing is always at the origin
- The bounding box doesn't disappear when finishing typing
- Text is only rendered properly after the document is panned
- etc.

### LINKED ISSUE #1983: Pen tool start position offset on offset artboard
When drawing a new path with the pen tool on an offset artboard (not at the origin), the start point is in the wrong place.

### LINKED ISSUE #1984: Incorrect pen tool overlay when placing a second anchor
When drawing a new path with the pen tool, the square overlay for the first anchor disappears when you are placing the second anchor. This should not happen.

Another regression introduced by #1946.

### LINKED ISSUE #1987: Position no longer relative to artboard
The position of a drawn shape should be relative to the parent artboard. Currently it is not. This means that moving an artboard doesn't move the child layers.

Another regression introduced by #1946.

### LINKED ISSUE #1988: Pen tool has overlay for closed path
The pen tool creates an overlay to close the path whilst it is being drawn.

Another regression by #1946.

### LINKED ISSUE #1989: Cannot make point co-linear
The co-linear checkbox in the path tool doesn't work on a point (e.g. from the freehand tool).

Another regression by #1946.

### LINKED ISSUE #1990: Cannot extend path with pen tool
If a new path is drawn at the start/end of an old path, then the old path should be extended.

A regression introduced by #1946

### LINKED ISSUE #1991: Cannot close path with the pen tool
When drawing a path with the pen tool, placing an anchor on the start point should close the path.

A regression introduced by #1946.

### LINKED ISSUE #1997: SVG import is broken
File > Import for an SVG doesn't work anymore.

adamgerhant:
Fixes #1982, fixes #1983, fixes #1987, fixes #1984, fixes #1988, fixes #1989, fixes #1990, fixes #1991, fixes #1997

Also fixes:
- Uncentered artboard when reloading and creating new document
- Colinear handles aren't working anymore (on ellipses)
- The Path tool is unable to move control point handles (on a shape created with the Path tool)
- Right clicking while drawing an ellipse is supposed to abort the drawing, but now it incorrectly accepts the shape. Also for dragging nodes in the node graph and right clicking which should abort to their initial location.
- Artboard tool broken (graph type error)

Adds a message buffering system so the tools can use the transform of the layer for their first interaction.

### DIFF
--- a/editor/src/dispatcher.rs
+++ b/editor/src/dispatcher.rs
@@ -119,10 +119,6 @@ impl Dispatcher {
 					let mut update_click_targets = VecDeque::new();
 					update_click_targets.push_back(DocumentMessage::UpdateClickTargets { click_targets }.into());
 					self.message_queues.push(update_click_targets);
-
-					// let mut update_vector_modify = VecDeque::new();
-					// update_vector_modify.push_back(DocumentMessage::UpdateVectorModify { vector_modify: vector_data }.into());
-					// self.message_queues.push(update_vector_modify);
 				}
 				Message::NoOp => {}
 				Message::Init => {
--- a/editor/src/dispatcher.rs
+++ b/editor/src/dispatcher.rs
@@ -110,11 +110,7 @@ impl Dispatcher {
 						self.message_queues.extend(buffered_queue);
 					};
 
-					let graphene_std::renderer::RenderMetadata {
-						footprints,
-						click_targets,
-						vector_data,
-					} = render_metadata;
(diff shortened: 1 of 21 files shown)


Keavon (approved):
Back to normal again, thank you!

PR #1992, opened 2024-09-19
Group layers with Ctrl+G into independent groups if they're spread across artboards

moOsama76:
Resolves the TODO comment [here](https://github.com/GraphiteEditor/Graphite/blob/77936c44b053a7a4cf5af677e024bb443738cdde/editor/src/messages/portfolio/document/document_message_handler.rs#L494C1-L494C68), here's a snippet:

```rs
DocumentMessage::GroupSelectedLayers => {
	responses.add(DocumentMessage::AddTransaction);

	let Some(parent) = self.network_interface.deepest_common_ancestor(&self.selection_network_path, false) else {
		// Cancel grouping layers across different artboards
		// TODO: Group each set of layers for each artboard separately
		return;
	};
```

A document can have multiple artboards, each containing its own layers. If layers are selected across multiple artboards, Ctrl+G currently yanks them out of their other artboards and puts them all in a group within a single artboard. This change makes the grouping happen on a per-artboard basis, so there ends up being multiple groups (one per artboard that had a selected layer).

### DIFF
--- a/editor/src/messages/input_mapper/input_mappings.rs
+++ b/editor/src/messages/input_mapper/input_mappings.rs
@@ -7,6 +7,7 @@ use crate::messages::input_mapper::utility_types::misc::MappingEntry;
 use crate::messages::input_mapper::utility_types::misc::{KeyMappingEntries, Mapping};
 use crate::messages::portfolio::document::node_graph::utility_types::Direction;
 use crate::messages::portfolio::document::utility_types::clipboards::Clipboard;
+use crate::messages::portfolio::document::utility_types::misc::GroupFolderType;
 use crate::messages::prelude::*;
 use crate::messages::tool::tool_messages::brush_tool::BrushToolMessageOptionsUpdate;
 use crate::messages::tool::tool_messages::select_tool::SelectToolPointerKeys;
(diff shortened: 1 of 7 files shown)


adamgerhant:
!deploy

adamgerhant:
!build

adamgerhant:
I think this would have several issues. If the artboard was not selected, then it would not be added to the hashmap and nothing would happen. The best way to approach this would be to collect all the layer paths, and check if there is an artboard in the path. If so, then the ROOT_PARENT gets popped from the path. If the path is just one item at this point, it can be filtered out (since it is an artboard). Once these paths are collected, a hash map can be used to group the paths based on the last element. Then each group of paths can be grouped. The parent is the deepest common ancestor, and the rest of the logic can be applied from there.

moOsama76:
Okay I'll rework it

adamgerhant:
!build

adamgerhant:
!build

adamgerhant:
I made a few changes in this commit: 4c4f326fdd228b6167b993b0ab624e457d4d66ce. Could you add it to your branch so this can get merged?

moOsama76:
@adamgerhant can you review?

adamgerhant:
I pushed my changes/fixes about 2 weeks ago, and it was ready to merge after that.

Keavon:
@adamgerhant in that situation, could you please remember to leave an approval on the PR? Thanks :)

adamgerhant:
I'm confused by what exactly happened with the commit history. It seems my commit got edited and then forced pushed? I reviewed/tested it though and its ready to merge

adamgerhant (approved):

moOsama76:
I checked this and I think it's okay

Keavon:
Sorry, what issue or `#code-todo-list` task is this solving? The PR description is blank and the title doesn't give me enough details to understand what this is meant to fix.

moOsama76:
> Sorry, what issue or `#code-todo-list` task is this solving? The PR description is blank and the title doesn't give me enough details to understand what this is meant to fix.

https://github.com/GraphiteEditor/Graphite/blob/77936c44b053a7a4cf5af677e024bb443738cdde/editor/src/messages/portfolio/document/document_message_handler.rs#L494C1-L494C68

Keavon:
Ah, cool, thanks. Yeah, that looks useful. I put that in the PR description. This will need to have its conflicts resolved against master but then I can code review it with that context in mind.

Keavon:
Superseded by #2239.

PR #1993, opened 2024-09-20
Fix Graphene CLI crash

0HyperCube:
Fixes the graphene-cli so it doesn't crash when used. It currently doesn't seem to do anything other than loop forever.

### DIFF
--- a/node-graph/graphene-cli/src/main.rs
+++ b/node-graph/graphene-cli/src/main.rs
@@ -78,9 +78,32 @@ fn init_logging() {
 		.unwrap();
 }
 
+// Migrations are done in the editor which is unfortunately not available here.
+// TODO: remove this and share migrations between the edtior and the CLI.
+fn fix_nodes(network: &mut NodeNetwork) {
+	for node in network.nodes.values_mut() {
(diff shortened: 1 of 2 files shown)


TrueDoctor:
The endless loop is intentional, when executed with a matching graaphite document, it should create a new window with the render result of the node graph

node-graph/graphene-cli/src/main.rs
```diff
@@ -78,33 +78,64 @@ fn init_logging() {
...
+		inputs: vec![],
 		..Default::default()
 	};
 
+	// TODO: Replace with "Output" definition?
+	// let render_node = resolve_document_node_type("Output")
+	// 	.expect("Output node type not found")
+	// 	.node_template_input_override(vec![Some(NodeInput::node(NodeId(1), 0)), Some(NodeInput::node(NodeId(0), 1))])
+	// 	.document_node;
+
```
TrueDoctor, line 129: That wouldn't be possible without adding the editor crate as a dependency. But we could move the wrap network in scope function into the interpreted executor crate

TrueDoctor (approved):

PR #1999, opened 2024-09-21
Improve node macro and add more diagnostics

TrueDoctor:


Keavon (approved):
This is getting really nice to work with now!

PR #2000, opened 2024-09-21
Path Bool library code cleanup

TrueDoctor:
- Fix endless loop when parsing paths with whitespace after the last Z command
- Add more tests


Keavon:
https://github.com/GraphiteEditor/Graphite/blob/261c79bf4d47a1aa74c0ee4649f9506aab2410b0/libraries/path-bool/src/parsing/path_data.rs#L115

Should this panic? Is that a code path reached by a user's path input? If so, that should return a Result, right?

Keavon (approved):

PR #2004, opened 2024-09-23
Fix Opacity slider in Layers panel adding Transactions on_update

singhutsav5502:
Make sure Opacity slider in Layers panel does not add **Transactions( History )** _on_update_ 
Only add **Transactions** _on_commit_.


Keavon:
!build

Keavon (approved):
Thanks!

PR #2011, opened 2024-09-28
Make Copy to Points and (Circular) Repeat and nodes output group data, and add flattening nodes

adamgerhant:
Output GraphicGroup from repeat nodes, add node to flatten vector elements to vector data


Keavon:
I think we should make sure that Copy to Points and the two Repeat nodes behave the same way with how they handle their data types and generics.

Keavon (approved):

PR #2015, opened 2024-09-30
Clear artboard

moOsama76:

### DIFF
--- a/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
@@ -236,6 +236,7 @@ impl MessageHandler<GraphOperationMessage, GraphOperationMessageData<'_>> for Gr
 								input: NodeInput::node(merge_node, 0),
 							});
 						} else {
+							// TODO: this will need to be rethought when there are nodes that can receive artboard's output
 							responses.add(NodeGraphMessage::SetInput {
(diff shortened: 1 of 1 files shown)


moOsama76:
Note that merge nodes are named "Artboard" as I reserve the name

Keavon:
!build

adamgerhant:
I would like to see the artboards name/position/downstream connnector/upstream connectors being saved, then the artboard is deleted, a layer node is created,  its name/position are set, then it is reconnecting. All using the same functions in the network interface as if the user was to do with.

Also, make sure to start and end a transaction at the start and end of the message. Its also important that no mutable network interface functions are called in a message that has a start/end transaction.

It would also be nice if the artboards position is then used to set the transform of the new layer, so that the contents stay at the same place. This would require using a monitor node to access the artboard data, where the translation is stored.

Keavon:
I'll mark this as a draft until those comments are addressed. Thanks for taking the time to review this, @adamgerhant.

adamgerhant:
You can use however many loops are necessary to achieve the correct behavior

moOsama76:
> Just checking in on the status of this @moOsama76.

I was taking a break, will do it this week

Keavon:
I'll close this PR for now, but please feel free to reopen it once you've addressed the code review feedback from Adam.

moOsama76:
> I'll close this PR for now, but please feel free to reopen it once you've addressed the code review feedback from Adam.

No problem I am getting back soon I promise :)

PR #2027, opened 2024-10-10
Update Wasm dependencies to fix WebGPU spec change crash in Firefox with Vello

0HyperCube:
Fixes running on firefox nightly (requires the latest wgpu to fix https://github.com/gfx-rs/wgpu/issues/6290#issuecomment-2383605551). This also required using vello with the git wgpu version.

We get a warning due to "use of deprecated method `web_sys::CanvasRenderingContext2d::set_stroke_style`". There is an alternative function but it is broken. I fixed it in https://github.com/rustwasm/wasm-bindgen/pull/4170 but it has not been released. There will apparently be a patch release "today".

### DIFF
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -38,10 +38,10 @@ impl OverlayContext {
 			self.render_context.line_to(quad.0[i].x.round() - 0.5, quad.0[i].y.round() - 0.5);
 		}
 		if let Some(color_fill) = color_fill {
-			self.render_context.set_fill_style(&wasm_bindgen::JsValue::from_str(color_fill));
+			self.render_context.set_fill_style_str(color_fill);
 			self.render_context.fill();
 		}
-		self.render_context.set_stroke_style(&wasm_bindgen::JsValue::from_str(COLOR_OVERLAY_BLUE));
+		self.render_context.set_stroke_style_str(COLOR_OVERLAY_BLUE);
 		self.render_context.stroke();
 	}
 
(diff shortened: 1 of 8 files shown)


Keavon:
It sounds like next week is when WGPU 0.23 is planned for release. I'll go ahead and merge this now but we should be able to update as soon as that comes out and Vello pushes an upgrade to it, letting us switch to the released WGPU and the official Git version of Vello instead of needing to use the Vello fork for now. But that'll all be an easy followup PR next week.

Keavon (approved):

editor/src/messages/portfolio/document/overlays/utility_types.rs
```diff
@@ -38,10 +38,10 @@ impl OverlayContext {
 			self.render_context.line_to(quad.0[i].x.round() - 0.5, quad.0[i].y.round() - 0.5);
 		}
 		if let Some(color_fill) = color_fill {
-			self.render_context.set_fill_style(&wasm_bindgen::JsValue::from_str(color_fill));
+			self.render_context.set_fill_style_str(color_fill);
```
Keavon, line 41: I just tested the build link and I'm noticing all our overlays, as set in this file, are no longer blue or whatever color— they're all the default black now. So I think some/all of these color setters aren't working, but I wouldn't be sure if that's due to the way we're calling them or an issue with the wasm-bindgen/web-sys/js-sys upstream code. Would you mind taking a look, please?

Keavon (changes requested):

editor/src/messages/portfolio/document/overlays/utility_types.rs
0HyperCube: You have to update your version of wasm-bindgen-cli (see the docs I updated here) @Keavon.

0HyperCube: `cargo install -f wasm-bindgen-cli@0.2.95`

Keavon (approved):
💯

PR #2028, opened 2024-10-10
Factor in artboard clipping to the click target x-ray function

0HyperCube:
As requested by @Keavon.

I chose to implement collecting the clipping artboards as part of collecting the metadata. If @adamgerhant wishes to redo this then they are welcome to.


adamgerhant:
I think using the render output is good for now until there is better monitor node insertion for the artboard data.

Keavon (approved):
Works beautifully! And that's some nice abstraction, I like it!

Issue #2026, opened 2024-10-09
Tracking Issue: Math expression parser/calculator


Keavon:
This is our library at [`/libraries/math-parser`](https://github.com/GraphiteEditor/Graphite/tree/master/libraries/math-parser) that uses a parsing framework that takes a string at runtime and calculates its result, including units and dimensional analysis.

## Roadmap

- [x] Scalars (real numbers)
- [x] Particle1 (complex numbers)
- [ ] Vector1/Vector2/Vector3 and Particle2/Particle3 (up to quaternions)
- [ ] Matrices (linear and affine maps)
- [ ] Static typing: sorts (values vs. matrices), the value ladder, the matrix diamond (Linear/Affine × 2D/3D), and refinements
- [ ] Units
- [ ] Documentation
- [ ] Error reporting *(partially done: parse errors carry byte spans and evaluation errors are typed, but there is no source-snippet rendering)*
- [ ] Compiling into Graphene AST

## Operators
- [x] Infix operators:
	- [x] Addition: `+`
	- [x] Subtraction: `-`, `−`
	- [x] Multiplication: `*`, `×`, `⋅`
	- [x] Division: `/`, `÷`
	- [ ] Modulo: `%` (floored, so the result takes the sign of the modulus and `(-3.2) % 2` is `0.8`, matching the Modulo node's default; `x % turn` wraps an angle into one revolution)
	- [x] Exponentiation: `^`
	- [x] Equals: `==`
	- [x] Not Equals: `!=`, `≠`
	- [x] Less Than or Equal To: `<=`, `≤`
	- [x] Greater Than or Equal To: `>=`, `≥`
	- [x] Less Than: `<`
	- [x] Greater Than: `>`
	- [x] Or: `||`, `∨`
	- [x] And: `&&`, `∧`
- [x] Prefix operators:
	- [x] Unary Plus: `+`
	- [x] Unary Minus: `-`, `−`
	- [x] Not: `!`, `¬`
- [x] Postfix operators:
	- [x] Factorial: `!`
- [x] Grouping operators:
	- [x] Parentheses: `( )`
	- [x] Magnitude bars: `|x|` (the Euclidean magnitude over all parts, the absolute value on a scalar; the per-part fold is `abs()`; `|a||b|` lexes as the magnitude of `a` OR `b`, so users must write `|a| |b|`, while a `||` with two magnitudes open closes both, so `|a*|b||` is valid but `|a|b||` is an error)

## Chained comparisons
[...]

PR #2033, opened 2024-10-12
Add math-parser library

urisinger:
Resolves #2026

For more info check out the refenced issue.


libraries/math-parser/src/constants.rs
```diff
@@ -0,0 +1,51 @@
...
+			"cot" => Some(Value::Number(Number::Complex(complex.tan().recip()))),
+
+			"invsin" => Some(Value::Number(Number::Complex(complex.asin()))),
+			"invcos" => Some(Value::Number(Number::Complex(complex.acos()))),
+			"invtan" => Some(Value::Number(Number::Complex(complex.atan()))),
+			"invcsc" => Some(Value::Number(Number::Complex(complex.recip().asin()))),
+			"invsec" => Some(Value::Number(Number::Complex(complex.recip().acos()))),
+			"invcot" => Some(Value::Number(Number::Complex((Complex::new(PI / 2.0, 0.0) - complex).atan()))),
+
+			_ => None, // Handle unknown function names
```
0HyperCube, line 46: A warning here (just to the console) might be useful.

libraries/math-parser/src/context.rs
```diff
@@ -0,0 +1,86 @@
...
+		self.0.insert(name, value);
+	}
+
+	pub fn contains(&self, name: &str) -> bool {
+		self.0.contains_key(name)
+	}
+
+	pub fn remove(&mut self, name: &str) -> Option<Value> {
+		self.0.remove(name)
+	}
```
0HyperCube, line 51: You can just implement deref mut so all of the HashMap functions are exposed.

0HyperCube (approved):
Looks good; I'm sorry that it has taken so long to merge anything related to this.

Keavon:
It looks like I don't have permission to push to your branch. Can you delete the test at the very bottom of `editor_api.rs` which is failing CI?

Keavon:
!build

Keavon:
From my tests, this seems to work well when replacing meval! Thank you. And sorry it took so long to get to.

libraries/math-parser/src/constants.rs
urisinger: when a function lookup fails we should check if an ident with that name exists, for now i changed it to be a hashmap lookup instad.

urisinger:
> From my tests, this seems to work well when replacing meval! Thank you. And sorry it took so long to get to.

no problem, this still needs quite a bit of work to work as its own node, mainly becuase it needs some sort of compile time node generation

Keavon:
That'll be a good step for the next PR. I'm looking forward to that! This one is good to merge now.

Keavon (approved):

PR #2055, opened 2024-10-19
Allow the Path tool to edit an upstream path even if there's a type conversion midway

adamgerhant:
I changed the upstream traversal for the tools to only iterate over the same type so that modifications to the output of a Boolean were applied to the vector data after the operation, not an upstream path node. This was a hacky solution, and we need better support for selection which path node to modify. This can be done with dots in the layer panel. More information in comment above `existing_node_id`

### DIFF
--- a/editor/src/messages/portfolio/document/graph_operation/utility_types.rs
+++ b/editor/src/messages/portfolio/document/graph_operation/utility_types.rs
@@ -5,6 +5,7 @@ use crate::messages::portfolio::document::utility_types::network_interface::{sel
 use crate::messages::prelude::*;
 
 use bezier_rs::Subpath;
+use graph_craft::concrete;
 use graph_craft::document::value::TaggedValue;
 use graph_craft::document::{NodeId, NodeInput};
 use graphene_core::raster::{BlendMode, ImageFrame};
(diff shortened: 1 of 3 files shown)


0HyperCube (approved):
The code looks good.

Automatically inserting the flatten elements before the path node foregrounds an issue with the ID collisions. If you have a group with two shapes with the same point ids (e.g. two rectangles) and flatten them, the ids for the second rectangle differ each time, meaning that it is not possible to drag them around.

https://github.com/GraphiteEditor/Graphite/blob/c3a3c4c907419e2ebb9a382546a1d7b66f57cdc0/node-graph/gcore/src/vector/vector_data/attributes.rs#L762C78-L763C1

adamgerhant:
Yes, that is why I created the generate_from_hash, which generates stable yet unique ids based on the NodeId of the merge node that the vector data flows through. It is implemented for the flatten vector elements in #2049, [here]<https://github.com/GraphiteEditor/Graphite/blob/49a5cd88311294f5a02de71f8bd7317de13291ba/node-graph/gcore/src/vector/vector_data/attributes.rs#L760> where I needed the functionality to add the joining of vector data with the pen tool (this should be split into a separate PR). 

It is also currently used by the repeat and copy to point nodes so that each new vector data has stable and unique ids.

This is still a work around to allowing the path tool to operate directly on vector data. A more robust method is necessary for the path node to input and output a GraphicGroup. I suppose each vector data will need a unique ID that the path node can reference, since the merge node id will change when it is copy and pasted.

Keavon:
Well, I couldn't find the crash either, so we'll have to live with it and fix it once we can reproduce it in the wild.

Keavon (approved):

PR #2058, opened 2024-10-20
Tab swaps to dragging the opposite handle

### LINKED ISSUE #1870: Tracking Issue: Pen and Path tool improvements
Much of this is explained/demonstrated in this detailed webcast: https://youtu.be/OmveqhzXJFg

Discord thread: https://discord.com/channels/731730685944922173/1319602515758682132

<details>
<summary>

## Path tool (completed)

</summary>

When dragging a handle:
- [x] <kbd>Alt</kbd> keeps the handles equidistant while pressed (instead of <kbd>Shift</kbd> in use now) (#2058)
- [x] <kbd>Alt</kbd>, as part of keeping the handles equidistant, also creates a matching handle if the other is zero-length or not a cubic Bézier (#2196)
- [x] <kbd>Shift</kbd> locks the handle angle to 15° increments (shouldn't interfere with <kbd>Shift</kbd>-clicking handles for multi-selecting points such as handles) (#2160)
- [x] <kbd>Ctrl</kbd> locks the current angle of the handle (#2160)
  - [x] If pressed while the non-colinear handle is over its anchor, it becomes temporarily colinear and is locked to the angle of its opposite handle, or if there is no opposite handle then it should not become colinear but it should become locked to the angle of its [shared adjacent segment angles](https://files.keavon.com/-/NovelGruesomePangolin/capture.png) (#2612)
  - [x] If pressed while over the an adjacent anchor, it locks to handle dragging along its tangent line [*(video)*](https://files.keavon.com/-/InferiorIllinformedSableantelope/capture_25_.mp4), except if that anchor has no handles then it has no tangent; if it has split (non-colinear) handles then it uses the one on the side adjacent to the handle being dragged. (#2663)
- [x] <kbd>Tab</kbd> swaps to dragging the opposite handle (native: teleports the cursor / web: enters pointer lock and draws a fake cursor at the new location) (#2058)
- [x] <kbd>Space</kbd> shifts the anchor (and handles) while pressed (#2065)
- [x] <kbd>C</kbd> toggles bent vs. colinear handles (instead of <kbd>Alt</kbd> in use now) (#2058)

When dragging an anchor:
- [x] <kbd>Alt</kbd> **(new since video was recorded)**, if held during the start of the click-drag, converts the anchor's handles to colinear and drags the handle pair instead of the anchor, keeping them equidistant unless <kbd>Alt</kbd> is released to temporarily return the opposite handle back to where it started before the <kbd>Alt</kbd>-drag began (which involves returning its colinear state back to what it was before). Should work with <kbd>Tab</kbd> to swap which is the dragged handle, so its opposite can be the one that's returned if <kbd>Alt</kbd> is temporarily released. Just <kbd>Alt</kbd> clicking but not dragging the anchor should remove both handles by setting them to zero length. (#2496)
- [x] <kbd>Shift</kbd>, locks the drag to the nearer of the X or Y axes from its drag starting point. (If pressed before the drag begins, it shouldn't break <kbd>Shift</kbd>-clicking points to add to the selection.) (#2449)
- [x] <kbd>Ctrl</kbd> **(new since video was recorded)**, if pressed before the drag begins, should select and drag a zero-length handle out of its anchor locked to the angle of the opposite non-zero-length handle, which it should be made colinear to. This applies only to anchors with two handles where exactly one is zero-length. Otherwise, the key is ignored and the anchor is dragged per usual. One more scenario is where a non-endpoint anchor has no handles (they're both zero-length), if <kbd>Ctrl</kbd> and <kbd>Alt</kbd> (see previous bullet point) are both pressed before dragging, it locks the angle to that of the line perpendicular to the average of the two tangents [*(diagram)*](https://files.keavon.com/-/NovelGruesomePangolin/capture.png) from that anchor to its adjacent curves. (#2620)

When double-clicking an anchor **(new since video was recorded)**:
- [x] Currently this swaps between smooth and sharp, but doesn't work if multiple handles are multi-selected to convert them all to smooth or sharp. If they're of mixed type, the specific one that's double clicked on should be the decider for the rest of them as to which they should all be swapped to. (#2498)
- [x] If a single anchor is double clicked, and it has only one handle (the other is zero-length or doesn't exist), currently this converts it to sharp by removing the handle. But we want that to make the new handle colinear and extend it to an equal length as the present handle. Only if both handles are present, should double-clicking remove both handles like it does currently in that situation. And if both handles are gone, it should also do what it does now by restoring them as colinear. (#2620)

When clicking a segment **(new since video was recorded)**:
- [x] Clicking without any modifier key should split the segment where clicked. Currently this has the extra complication of entering an insertion mode where the point of the segment split can be slid within the segment and left click confirms it while right click cancels. We want to remove that extra step and just insert directly. Important: clicking then immediately click-dragging to drag the point shouldn't become a double-click that converts from sharp to smooth. An overlay should be drawn in cases where the cursor is hovered over the segment such that clicking would insert a split, it should be a short line segment (20px in diameter) running perpendicular to the curve. (#2495)
- [x] <kbd>Alt</kbd> clicking should delete the segment that's clicked. Currently a user would have to split the segment then delete the newly created point. (#2495)
- [x] Clicking and dragging without any modifier key, as per #615, should [mold the segment](https://pomax.github.io/bezierinfo/#molding). (#2660)

When a single handle is selected:
- [x] <kbd>R</kbd> rotates about its anchor (#2180)
- [x] <kbd>S</kbd> scales about its anchor (#2180)

</details>
<details>
<summary>

## Pen tool (completed)

</summary>

When no drawing has begun yet (pointer is up):
- [x] **(New since video was recorded)** clicking a path's endpoint should begin drawing from it with an initial handle that's colinear and equidistant from the opposite handle, instead of the current behavior where clicking the endpoint begins drawing the segment with a zero-length handle. If the user wants the current behavior, they can simply click the endpoint anchor a second time to make the current segment's handle zero-length. This solves the pain point that switching away from the Pen tool, then back to it, causes a discontinuity at the previous endpoint where the handles can't easily be opposite each other like when drawing consecutive segments normally. (#2295)

When stretching out a segment to be placed (pointer is up):
- [x] <kbd>Backspace</kbd>/<kbd>Delete</kbd> removes the current segment's handle (equivalent to clicking the current anchor) (#2180)
- [x] <kbd>G</kbd> grabs (translates) the current segment's connected outgoing handle (#2211)
- [x] <kbd>R</kbd> rotates the current segment's connected outgoing handle about its anchor (#2211)
- [x] <kbd>S</kbd> scales the current segment's connected outgoing handle about its anchor (#2211)

When dragging the handles of the segment being placed (pointer is down):
- [x] <kbd>C</kbd> toggles bent vs. colinear handles (instead of the momentary <kbd>Alt</kbd> in use now) (#2242)
- [x] <kbd>Alt</kbd> stops keeping the handles equidistant while pressed, as long as they're currently colinear (otherwise it does nothing). Exception: the reverse is the case (it keeps the handles equidistant while pressed) when the click-drag began with a single handle (either by click-dragging the current segment's connected anchor, click-dragging an endpoint of an existing not-yet-being-edited layer, or click-dragging the start point to close the shape) but <kbd>Ctrl</kbd> was pressed to keep (#2242)
- [x] <kbd>Ctrl</kbd> locks the current angle of the handle (already implemented), but should also work to be colinear with the opposite handle if being dragged out of an existing anchor point, or if that anchor point has no handle then it should use the angle of the actual curve segment entering that handle, i.e. its tangent angle at its endpoint (#2284)
- [x] <kbd>Space</kbd> shifts the anchor (and handles) while pressed (#2416)
- [x] <kbd>Tab</kbd> swaps to dragging the opposite handle (native: teleports the cursor / web: enters pointer lock and draws a fake cursor at the new location) (#2452)
- [x] If the current segment is closing the shape by connecting to its start point, the handle that's being dragged should be the closing handle, not a fake handle on the opposite side, and all the above shortcuts should work on that anchor/handle group spanning the start/end point (<kbd>Tab</kbd> will modify the shape of the first segment while the current/final segment is still in-progress) (#2452)
- [x] All the above shortcuts should work with the "Bend Prev. Point" case where the current segment's connected anchor can be clicked to turn it sharp or click-dragged to bend its outgoing handle (turning its handles to not colinear), meaning <kbd>C</kbd> should be able to turn it back to colinear

</details>

DaraghD:
Unsure of naming of the function

Partially closes #1870


0HyperCube:
!build

0HyperCube (commented):
The code looks good so far and works well.

Currently if both the handles from an anchor are selected, <kbd>tab</kbd> switches to just having a random one selected. Is this the desired behaviour? I would have thought it would be more intuitive just to keep both handles selected (since they are both opposite to a formerly selected handle). However it probably doesn't matter too much.

Also please can you add a hint for this. These are the text at the bottom of the viewport:
screenshot with hint in

Thanks for your work on this contribution so far.

DaraghD:
> Currently if both the handles from an anchor are selected, tab switches to just having a random one selected. Is this the desired behaviour? I would have thought it would be more intuitive just to keep both handles selected (since they are both opposite to a formerly selected handle). However it probably doesn't matter too much.

Switched it to be that way, I think that makes more sense

> Also please can you add a hint for this. These are the text at the bottom of the viewport: 

Added one, again not sure of the naming

0HyperCube (approved):
Works well and the code looks good. Thanks again for working on this.

DaraghD:
Clippy failed on a ton of stuff that I didn't touch, is that normal?

0HyperCube:
@DaraghD the clippy lints are fixed by https://github.com/GraphiteEditor/Graphite/commit/a395fbf0637c23b8b05a7c451ff7b8421587b655.

Keavon:
I'll do a QA review on this before we merge it. It's in my queue that I'm working through presently.

Keavon:
@DaraghD upon testing this, I think the best near-term solution for the question of cursor alignment is to enable cursor lock so the user's cursor disappears, but don't yet attempt to render a fake cursor. Could you please add that part? You can reference how we do this in the NumberInput.svelte slider drag, for example. That's all frontend though, and this'll need to plumbed in from the backend. You could also look at the code paths which let the backend plumb to the frontend which cursor style to display, since this would be similar and could potentially be included in those code paths. Thanks!

Keavon:
!build

Keavon:
This will be ready to merge once GitHub fixes their CI outage and I can make it re-run to pass CI, which should be when I wake up.

PR #2064, opened 2024-10-21
Add switch node and fix log to console node

TrueDoctor:


Keavon:
It would be nice if the Switch node could take T for its two inputs and give T output. Since there are plenty of data types, like Option<Color>, this doesn't support. But merging it for now since it's useful.

Keavon (approved):

PR #2076, opened 2024-10-25
Allow the Pen tool to connect layers by their endpoints, merging into a single layer

### LINKED ISSUE #2077: Make the Pen tool connect separate layers by their endpoints to become one layer

adamgerhant:
Closes #2077

### DIFF
--- a/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
@@ -1,7 +1,7 @@
 use super::transform_utils;
 use super::utility_types::ModifyInputsContext;
 use crate::messages::portfolio::document::utility_types::document_metadata::LayerNodeIdentifier;
-use crate::messages::portfolio::document::utility_types::network_interface::{InputConnector, NodeNetworkInterface};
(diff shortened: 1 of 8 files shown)


Keavon:
We have a little bit of a bug [in this document](https://github.com/user-attachments/files/17528158/bug_scene.graphite.txt):

https://github.com/user-attachments/assets/c0f34752-8884-4842-a65a-14b0eb678f21

Keavon:
I'll temporarily mark this as a draft until then.

Keavon:
Because of the force pushes, it's hard for me to see exactly what changes you made. However, testing this again in the latest build link, the issue I showed above still exists 😢 I'll mark this as a draft again in the meantime to make it easier to get through my review backlog. As always, please un-mark as draft when it's ready (and perhaps ping me). Thanks.

Keavon:
Seems perfect now, except for one issue we can defer to another PR and basically call a new feature instead of a bug: we want the Pen tool to only add to an existing path (which seems to happen now for any selected layer that has a Path node) if a checkbox is selected in its tool options bar, rather than by default, since that could confuse users how new layers aren't being created and instead new subpaths are being added.

Keavon (approved):

PR #2081, opened 2024-10-26
Fix layer name text input

0HyperCube:
Fixes an issue created & reported by Keavon.

It appears that the text input value for the layer name should have been the layer name instead of the document name.


Keavon:
Oh wow, this was a copy-paste error all along! I was so focused on the issue with the units (#2080) that I was convinced it was the same problem causing both observations, not just a backend bug I'd made. Thank you for solving this!

Keavon (approved):

PR #2091, opened 2024-11-03
continuous across start/end points usig extra nodes

### LINKED ISSUE #2086: Fix the spline node algorithm to be continuous across start/end points
Currently, the Spline from Points node doesn't handle continuity across the start/end point portion of the edge flow. This is what a square looks like when the node is applied, which fails in the top left corner where the start/end points connect:

capture

Instead, this example should look like a symmetrical blob, with the same roundness around all corners.

This task involves updating the spline calculation algorithm, as pointed out by [this todo comment](https://github.com/GraphiteEditor/Graphite/commit/3f17e83e80d345a373f0df07ed642fc95de871e2#diff-e7a79f7e09ecb50a47584450e9b37dda9b437bbf0dbda93360c8ad1cdb8445c0R955).

kalle95data:
Fixes #2086

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -943,13 +943,16 @@ async fn splines_from_points<F: 'n + Send>(
 		}
 
 		let mut nr_output_segments = ordered_point_indices.len();
+		let nr_overshoot_segments = 9; // Seems good enough, at least > 6 should be enough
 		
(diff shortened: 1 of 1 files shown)


Keavon:
!build

kalle95data:
I haven't managed to get bezier with a loop / cyclic boundaries all the way yet. The best so far is where the handles were rotated 90 degrees, but unfortunately there is still something wrong. This solution commited now is probably good enough, visually almost perfect (haven't "proved" it mathematically, but feels good enough).

PR #2092, opened 2024-11-03
Fix the spline node algorithm to be continuous across start/end points

### LINKED ISSUE #2086: Fix the spline node algorithm to be continuous across start/end points

0HyperCube:
Closes #2086

### DIFF
--- a/libraries/bezier-rs/src/subpath/core.rs
+++ b/libraries/bezier-rs/src/subpath/core.rs
@@ -417,3 +417,107 @@ pub fn solve_spline_first_handle(points: &[DVec2]) -> Vec<DVec2> {
 
 	d
 }
+
+pub fn solve_spline_first_handle_closed(points: &[DVec2]) -> Vec<DVec2> {
+	let len_points = points.len();
+	if len_points < 3 {
+		return Vec::new();
+	}
+
+	// Matrix coefficients a, b and c (see https://mathworld.wolfram.com/CubicSpline.html).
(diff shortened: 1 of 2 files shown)


Keavon:
It doesn't look like this is producing the expected results.

Input:
capture

Dev:
capture

This PR:
capture

Keavon:
Understood, and yes, if you're offering to do that improvement here too, that's probably the most expedient option to move forward. Thank you!

0HyperCube:
@Keavon I have implemented the new algorithm here. I have also added a test to confirm that the second derivative is continuous across all adjacent pairs of segments.

Keavon (approved):
Super! Love the much cleaner code, and the solved continuous closed subpath solution. I can finally use this node in my Nodevember creations now that it doesn't have a random sharp point 😀

PR #2093, opened 2024-11-04
Improve the node graph with revamped top bar and disabling tools when graph is open

Keavon:
Implements these layer/node management and graph view features:

capture

Also disables tools while the graph is open and restores them upon closing the graph.

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -28,6 +28,7 @@ pub struct NodeGraphHandlerData<'a> {
 	pub collapsed: &'a mut CollapsedLayers,
 	pub ipp: &'a InputPreprocessorMessageHandler,
 	pub graph_view_overlay_open: bool,
+	pub graph_fade_artwork_percentage: f64,
 }
 
 #[derive(Debug, Clone)]
@@ -76,8 +77,9 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
 			breadcrumb_network_path,
 			document_id,
 			collapsed,
-			graph_view_overlay_open,
 			ipp,
+			graph_view_overlay_open,
+			graph_fade_artwork_percentage,
 		} = data;
 
 		match message {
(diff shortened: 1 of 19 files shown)


editor/src/messages/tool/tool_message_handler.rs
```diff
@@ -25,6 +25,7 @@ pub struct ToolMessageHandler {
 	pub tool_state: ToolFsmState,
 	pub transform_layer_handler: TransformLayerMessageHandler,
 	pub shape_editor: ShapeState,
+	pub tool_is_active: bool,
```
0HyperCube, line 28: Is this just the inverse of `document.is_graph_overlay_open()`? If so, could it be used instead to reduce duplicated state?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1457,100 +1475,226 @@ impl NodeGraphMessageHandler {
...
+				.on_update(|number_input: &NumberInput| {
+					NavigationMessage::CanvasZoomSet {
+						zoom_factor: number_input.value.unwrap() / 100.,
+					}
+					.into()
+				})
+				.increment_behavior(NumberInputIncrementBehavior::Callback)
+				.increment_callback_decrease(|_| NavigationMessage::CanvasZoomDecrease { center_on_mouse: false }.into())
+				.increment_callback_increase(|_| NavigationMessage::CanvasZoomIncrease { center_on_mouse: false }.into())
+				.widget_holder(),
```
0HyperCube, line 1686: Seeing as this is the same as the document navigation overlays (but with the text changed to node graph and using the node graph PTZ struct) and I don't really expect these to diverge (changes to the PTZ system would impact both the document and node graph), would it be preferable to use a shared function?

0HyperCube (approved):
Works well; these are some fantastic improvements to the usability and discoverability of the node graph UI.

editor/src/messages/tool/tool_message_handler.rs
Keavon: There might be other callers of the activate and deactivate methods in the future (or even now? I didn't look closely how it's currently used). So I don't want to necessarily couple those activate/deactivate methods to the concept of the graph being visible.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
Keavon: Good point, I'll see if that's workable.

PR #2094, opened 2024-11-04
Path Tool: Changed reference coordinates and added different behaviors for Mouse and Keyboard Inputs

### LINKED ISSUE #2083: Path tool point nudging should be in document coordinates, not viewport coordinates
If you select a point using the Path tool and nudge it to the right by 10 units with <kbd>Shift</kbd><kbd>Right Arrow</kbd>, starting at 100% viewport zoom but then zooming to 200%, you'll notice it takes two presses of <kbd>Shift</kbd><kbd>Left Arrow</kbd> to return back to its starting place. That's because it currently is moving relative to the viewport coordinates, but it should always work relative to the document coordinates so 10 units to the right means the same thing regardless of the current level of zoom. It also shouldn't be in the layer's local coordinate space, where a shape that has a 0.1 scale factor would take 10 times as many presses to travel the intended distance. The Select tool's nudging can be referenced for this, because it works correctly in document space (not layer-local or viewport space).

Kukkarmutta:
Closes #2083 

Hello, this is my first contribution. Hope the issue is resolved as intended.
If I've made any mistake or there is room for improvement, do let me know :)

Details:
>The move_selected_points function is used in dragging as well as for keyboard nudging, so modifying its default coordinate system initially resulted in undesirable dragging behaviour.
> Edited move_selected_points function. Added a bool parameter to check whether we're using keyboard (1) or dragging (0).
> If we use keyboard it will now use document coords as reference, but for dragging it uses default coordinate system


Keavon:
Hi, welcome! Please be sure to run `cargo fmt`, and set up your text editor to run it automatically when saving the file. (This should already be the default in VS Code if you have the Rust Analyzer extension installed, but you'll have to figure out how to do that in other editors yourself.) Once you commit the formatted code, I can begin reviewing this. Thanks!

Kukkarmutta:
Hello Again. Although I had rust-analyser installed. I have ran the formatting command. Kindly review and let me know if any further changes are required.

Keavon:
It looks like the code doesn't compile. Have you thoroughly tested everything, including both the intended functionality and other likely regressions? I'll mark this as a draft until you have it working. Thanks!

Kukkarmutta:
When I tried to fix it manually idk what happened and some lines got deleted :') (i noticed that much later)  
im fixing that issue and will submit it in a few minutes

Keavon:
#2083 was closed by #2095 last week.

PR #2095, opened 2024-11-06
Fix point nudging to work in document space

### LINKED ISSUE #2083: Path tool point nudging should be in document coordinates, not viewport coordinates

November-6:
Closes #2083

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -622,13 +622,14 @@ impl ShapeState {
 		in_viewport_space: bool,
 	) {
 		for (&layer, state) in &self.selected_shape_state {
-			let Some(vector_data) = document.network_interface.compute_modified_vector(layer) else {
-				continue;
-			};
+			let Some(vector_data) = document.network_interface.compute_modified_vector(layer) else { continue };
+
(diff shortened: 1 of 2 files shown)


Keavon:
!build

Keavon (approved):
Thank you!

November-6:
Thanks for the notes! 
I'll get working on it as soon as i get back from college today. In the mean time, can you suggest me some place where i can learn more about the specific maths used in this project?

Regards :)

Keavon:
Are you referring to the TODO comment I left in my latest commit? That's not something you need to take action on. This PR is ready to merge.

PR #2096, opened 2024-11-06
Fix Bevel node crash with zero-length segments

0HyperCube:
Fixes a crash with the bevel node on a zero length segment.

Bevelled ellipse with top point duplicated:
Bevelled ellipse with top point duplicated


Keavon (approved):
Thank you! Much less dangerous now to use this with booleans, which like to insert duplicated points :)

PR #2104, opened 2024-11-10
CSS fix for Preferences Dialog

### LINKED ISSUE #2103: Overlapping widgets in file->preferences menu
The file->preferences menu looks like:
preferences menu
It has some overlapping widgets which are most likely undesired. Introduced by https://github.com/GraphiteEditor/Graphite/commit/de366f951424fcdf4463a419db3fa659910fabfd.

RounakJoshi09:
This is added in order let the elements fill the space in the widget span row, this is indeed necessary here, because , the width is available in the dialog, but the elements are aligned to the start, and not utilizing the width properly, in order to fix this in a responsive way, I have added the `justify-content` css property, which uses the width available for the `WidgetSpan`.

Fixes #2103

Screen shots after the fix
image
image

### DIFF
--- a/frontend/src/components/widgets/WidgetSpan.svelte
+++ b/frontend/src/components/widgets/WidgetSpan.svelte
@@ -198,6 +198,7 @@
 		flex: 0 0 auto;
 		display: flex;
 		min-height: 32px;
+		justify-content: space-between;
 
 		> * {
 			--widget-height: 24px;


0HyperCube:
Resolved by https://github.com/GraphiteEditor/Graphite/commit/99cf8f0c4f91a051b59fc2c9e5cc6c7417bdd74b.

Keavon:
@RounakJoshi09 Just to chime in about why I didn't merge this PR, unfortunately your proposed CSS change would have a large impact on many parts of the UI beyond just the misaligned settings dialog content. But I'm also going to quickly check if it would have made desirable improvements or if it broke things (just by looking at the code, I'm having to assume the latter is more likely, but I'll check by triggering a build real quick).

Keavon:
!build

PR #2105, opened 2024-11-10
Polish user-created subgraph nodes: imports in the Properties panel; reorder/delete/rename imports/exports

### LINKED ISSUE #2043: Improved Node Graph support for user created nodes.
It should be possible to select a group of nodes, and then "group" them into a single node with a network implementation of those nodes. It should also be possible to change the number, order, and names of inputs/outputs. It could also be useful to "ungroup" a node. Once #2042 is complete, it should also be possible to change the annotated types for the inputs, as detailed in #2069

### LINKED ISSUE #2069: Add list of acceptable types when hovering over node input
As noted by [Lerc](https://discord.com/channels/731730685944922173/1134981447539761273/1298026394047217773), it would be helpful for the NodeInput tooltip to display all acceptable types that will not cause a node error. For protonodes, this would require accessing the list of implementations, which are defined in the node macro. The implementations would then be filtered based on the current type of all other inputs to return a list of all currently valid types for that input. Network nodes are a bit more complex, since the inputs will undergo automatic type casting ([2042](<https://github.com/GraphiteEditor/Graphite/issues/2042>)), and all usages need to be valid. This means the acceptable types will be the intersection of all types that implement From for the network input type, and all types that the import is connected to.

### LINKED ISSUE #2208: Full support for changing subgraph node signatures
We can now group nodes into subgraphs and add inputs and outputs. But we can't reorder them, rename them, delete them, or connect them to input rows in the Properties panel.

adamgerhant:
Closes #2208 and continues to resolve #2043. Also resolves #2069.

Allows removing, renaming, and reordering imports/exports.
Input based properties for network nodes to allow useable reordering
Adds Tooltips for acceptable types of a node input
Annotations for both node and input based overrides added to the node macro

Reworks how the properties panel is generated. A full node level override for the entire panel is still possible, which is stored as a lambda in the definition. Modifying the imports/exports means that the reference is lost, and the properties are instead generated individually for each input. By default, this will use the data type/type annotation to generate an appropriate set of widgets. This is done in `property_from_type`. For custom input overrides, each input has the PropertiesRow struct, which can store a reference to a widget override (a lambda which will generate a custom set of widgets). This lambda is stored in a hashmap, since it would not be possible to serialize it for every node. This also makes it easily extendible, just like how the definitions are stored. This lambda has access to the entire network interface, which it can use to access any information it needs to generate the widget. In addition to the widget override reference, each input also stores a hashmap of serializable data, which can store any information necessary. For the most part this will just contain the input name, but the author of the node could store additional data as they need.

TODO: 
- Rename inputs in properties panel (deferred)
- Automatic Into node insertion with drop down menu (deferred)
- [x] Protonode name at top of properties panel
- [x] Hiding input should automatically disconnect
- [x] Filtering out the primary inputs of autogenerated Properties panel node sections so only secondary inputs are shown (e.g. Ellipse and Merge nodes)
- [x] Valid input types for the identity node crashes
- [x]  Upgrading previous documents


Keavon:
Deferred code review comments/requests from Hypercube in the precursor PR to this one:

- https://github.com/GraphiteEditor/Graphite/pull/2097#discussion_r1835687634
- https://github.com/GraphiteEditor/Graphite/pull/2097#discussion_r1835689486
- https://github.com/GraphiteEditor/Graphite/pull/2097#discussion_r1835691147

Keavon:
- [ ] Upgrading previous documents
- [ ] Filtering out the primary inputs of autogenerated Properties panel node sections so only secondary inputs are shown (e.g. Ellipse and Merge nodes)

Keavon:
Please implement upgrading logic for the Noise Pattern node when you get the chance. Thanks!

Keavon:
I'll discuss this with you on Discord but here are my notes of unresolved questions/potential concerns from my code review:

- Weird capitalization in one of the first files that has all the spread out changes
- Graphite save documents have possibly unintentional fields in the JSON
- Can strings like `#[widget(ParsedWidgetOverride::Custom = "assign_colors_repeat_every_row")] repeat_every: u32,` be updated anytime without affecting saved documents? Some inconsistent naming like this example ending in `_row` which might mean we should do a full review on these before merging if that affects user documents.

Keavon (approved):

Issue #1895, opened 2024-08-05
Editor crash when loading large images


TrueDoctor:

TrueDoctor:
![image](https://github.com/user-attachments/assets/5d883a31-8dc4-45db-b849-ae1f83ce2af5)
After fixing the mutex errors, this seems to be related to the image serialization.

TrueDoctor:
This might actually be a memory issue as the in memory representation takes 384mb which is quite substantial

0HyperCube:
The crash shown is from the autosaving (see all the serde in the call stack). This will likely be solved as part of #1604.

Keavon:
See also discussion in #1231 which is a duplicate of this.

Issue #1231, opened 2023-05-21
Out of memory with large images


nat-rix:
When using a large image (about 4000x4000) and using any image adjustment node on it and play around for a small amount of time with its parameters, wasm dies of oom.
![image](https://github.com/GraphiteEditor/Graphite/assets/26610181/a7205c17-d784-4818-ac55-28ad62161525)
A small round of debugging showed me, the last allocation before the exception is always an `TaggedValue::ImageFrame`.

nat-rix:
I've just done some `Drop` and `static AtomicUsize` magic and counted all instances of `ImageFrame` that are instanciated. I hope, I've implemented it correctly. When creating a new document, this count is at about 31 (varying from times to times). When pasting my image, this number grows to 34. When creating a blend node, it gets 36. When inserting the node into the graph, it gets about 49. And when playing around with the parameters, it goes up and down (not under 49) and crashes when it gets >60.
Some quick maths: an 4000x4000 image with 4 f32 channels has `4000 * 4000 * 4 * 4 = ca 244MiB`. Assuming a maximum memory size of 4GiB, it needs only 17 copies of this image to overflow memory.

Keavon:
(Our wasm memory is only 2GB since wasm-bindgen has a bug that uses signed pointer values which wastes half the 4GB)

Keavon:
Consolidating this into #1895.

PR #2106, opened 2024-11-10
Import ARW files as Images

elbertronnie:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -351,7 +351,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.79",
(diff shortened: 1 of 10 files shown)


0HyperCube:
Just for context, this is currently stalled due to #1895 (duplicate of #1231).

The uncompressed RAW images use a lot of memory. We also appear to clone them a lot (e.g. when running the node graph and running autosave). This exceeds the WASM memory limits leading to a crash whenever you attempt to work with a RAW file.

Keavon:
And specifically #1604.

elbertronnie:
@0HyperCube I doubt that merging this will be useful. This PR only works on very small RAW images which do not exist and after the asset database, this entire PR will need to rewritten anyway.

Keavon:
This is stale enough that I'll close it for housekeeping reasons, but we can search for this PR in the future if we need to reference it.

PR #2107, opened 2024-11-11
Make Pen tool always snap to endpoint anchors, even when snapping is off

### LINKED ISSUE #1980: Pen tool clicking an anchor to turn it sharp or close a path shouldn't rely on snapping
In the Pen tool, clicking to turn a point sharp relies on snapping being enabled to work, which is bad: 

https://github.com/user-attachments/assets/70c74839-2519-42a5-8d40-af4a459da912

It should be able to do that irrespective of the snapping system. Essentially, the tool should have its own self-contained implementation of snapping just for the actively being-drawn-from segment endpoint, which is always active even when the snapping system is disabled, and the real snapping system shouldn't interfere with it.

james-ryans:
Closes #1980


Keavon:
!build

Keavon:
!build

PR #2109, opened 2024-11-14
Bezier-rs: Make rectangle constructor produce linear segments

### LINKED ISSUE #1696: Subpath::new_rect() creates cubic segments instead of linear segments
Hi,

The bezier-rs `Subpath::new_rect()` constructor creates cubic segments, but imho linear segments would be more fitting?

Thanks!

Sidharth-Singh10:
Closes #1696 

`new_rec` constructor now should create linear segments instead of cubic segments.


Keavon:
!build

Keavon (approved):
Looks good, thank you.

PR #2120, opened 2024-11-28
Disabling colinear state when both the handles are selected and moved

0SlowPoke0:
When using path tool ,when both the handles are selected and moved the anchor point and handles are no longer colinear but the colinear state is still on.


0HyperCube (approved):
Works well; thanks for this useful tooling improvement.

PR #2123, opened 2024-11-30
Simplify the implementation of the message buffering

0HyperCube:
This is an attempt to slightly simplify the logic of the message buffering, whilst also fixing the debug message tree during buffering such that the message parents are properly aligned.

Also improved was the functionality when attempting to run another message when a buffer is open. Instead of this new message being processed first, it is processed last.

sample message tree

### DIFF
--- a/editor/src/dispatcher.rs
+++ b/editor/src/dispatcher.rs
@@ -63,20 +63,33 @@ impl Dispatcher {
 		}
 	}
 
-	pub fn handle_message<T: Into<Message>>(&mut self, message: T) {
-		self.message_queues.push(VecDeque::from_iter([message.into()]));
-		while let Some(message) = self.message_queues.last_mut().and_then(VecDeque::pop_front) {
-			// Do not buffer the EndBuffer message
-			if !matches!(message, Message::EndBuffer(_)) {
-				if let Some(buffered_queue) = &mut self.buffered_queue {
-					// Store each message in a deque so that its children are added before future messages
-					let mut message_deque = VecDeque::new();
-					message_deque.push_back(message);
-					buffered_queue.push(message_deque);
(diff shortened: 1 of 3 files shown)


editor/src/dispatcher.rs
```diff
@@ -104,9 +117,9 @@ impl Dispatcher {
 					self.buffered_queue = Some(std::mem::take(&mut self.message_queues));
 				}
 				Message::EndBuffer(render_metadata) => {
-					// The buffered vec is added before the metadata messages, because the end of the vec is processed first
+					// Assign the message queue to the currently buffered queue
 					if let Some(buffered_queue) = self.buffered_queue.take() {
-						self.message_queues.extend(buffered_queue);
+						self.message_queues = buffered_queue;
```
adamgerhant, line 122: Is there a reason why the buffered queue replaces the current message queues? I think it could be possible for there to still be messages in self.message_queues, which would be lost here.

adamgerhant (approved):
This looks good, I tested it with offset transforms and everything still works.

In general I'm still a bit uncertain if the message buffering is the correct solution to "async" messages that can only be run once a "promise" is resolved. There essentially needs to be a way for the message to be queued and executed once some data is available after some number of renders. Currently this is necessary for layers to await until their transform has been computed and returned, but in the future a more flexible system might be necessary where for instance a message can be queued for a certain number of milliseconds or until some state changes in the network, which may take multiple renders

editor/src/dispatcher.rs
0HyperCube: The current message queues will always be empty (except for the end buffer event), since they are cleared by the start buffer event and all other messages are added directly to the buffer.

PR #2131, opened 2024-12-06
Fix SVG `viewBox` capitalization in renderer

### LINKED ISSUE #2130: SVG export `viewbox` typo.
When exporting an SVG, the `viewBox` element is incorrectly spelled in lowercase. This breaks the SVG when rendering in the browser if you try to scale it.

reference:
https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox

paulwrath1223:
Check issue #2130 for description.

The `viewBox` attribute was erroneously lowercase, but the spec and browser rendering need it to be _"viewBox"_.

Closes #2130


Keavon:
TIL that would be case sensitive in some contexts! Thanks for this.

0HyperCube (approved):

PR #2136, opened 2024-12-12
Selection history navigation with Alt+[ and Alt+] keys should filter out empty selections

0SlowPoke0:
filters empty selections made by select tool  before adding in the selection history


0HyperCube:
!build

0HyperCube (approved):
Thanks for this fix; it appears to work as expected (although perhaps @Keavon will want to check as well).

Keavon:
@0SlowPoke0 @0HyperCube can you both please confirm that my simplification commit c362e26f97b4d21f40bbcf5e05b47dd126d2a81d is indeed equivalent? I don't understand why it was so complicated before, and I'm pretty sure the changes I made are correct, but I'm asking to make sure I didn't miss any subtleties that explained why the previous code was so complicated.

Keavon:
!build

Keavon:
This PR breaks the editor (unable to draw more shapes, unable to select layers, etc.) after you've drawn the first shape.

Keavon:
!build

0SlowPoke0:
@Keavon  I think this solves the issue can you please review the build, thanks.

Keavon:
Superseded by #2138

PR #2138, opened 2024-12-14
Improve issues with selection history

### LINKED ISSUE #2137: Drawing new shapes creates multiple selection history states
- Create a layer (e.g. with the rectangle or polygon tool)
- Go back in the selection history <kbd>Alt</kbd> + <kbd>[</kbd>
- Observe how the rectangle or polygon node is selected.

This is not the desired behaviour. We would expect the selection that was made before drawing the layer to be restored.

bakayu:
By merging this PR the following fixes/improvements will be implemented:
- Empty selections are no longer added to the selection history (fix for [discord's code-todo-list](https://discord.com/channels/731730685944922173/881073965047636018/1299601627006631966)).
- This closes #2137.
- Previously, If a new layer was drawn, we had to press `Alt` + `[` / `Alt` + `]` twice in order to move between the selections. This has been fixed, ie only one call is required now.
- Previously, If we pressed `Alt` + `[` many times after reaching the end of selection history, we had to press `Alt` + `]` that many times to access the selections again. This is fixed now (ie, it takes only one call of `Alt` + `]` to access the selections again).

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -1592,13 +1592,17 @@ impl NodeNetworkInterface {
 			let empty = current.selected_layers_except_artboards(self).next().is_none();
 			(current, previous, empty)
 		};
-
 		self.unload_stack_dependents(network_path);
 
 		let Some(network_metadata) = self.network_metadata_mut(network_path) else {
 			log::error!("Could not get nested network_metadata in selected_nodes");
 			return None;
 		};
+
+		// Initialize default value if selection_undo_history is empty
+		if network_metadata.persistent_metadata.selection_undo_history.is_empty() {
+			network_metadata.persistent_metadata.selection_undo_history.push_back(SelectedNodes::default());
+		}
 
 		// Update history only if selection is non-empty/does not contain only artboards
 		if !is_selection_empty && prev_state.as_ref() != Some(&last_selection_state) {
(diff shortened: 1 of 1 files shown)


0HyperCube:
I don't think this works as expected.

If I draw two rectangles and then press the <kbd>alt</kbd> + <kbd>[</kbd>, then the first rectangle should be selected again. In this implementation, nothing is selected.

Keavon:
!build

bakayu:
> I don't think this works as expected.
> 
> If I draw two rectangles and then press the alt + [, then the first rectangle should be selected again. In this implementation, nothing is selected.

ah I see, I missed that while testing, this seems to be happening because I implemented a transaction check before updating the history.
Moroever I noticed that #2136 tries to fix a similar, that fix seems to have fixed this issue as well

bakayu:
As I thought, #2136 has fixed this issue as well, but while testing I found that, when we add 2 rectangles, we need to press `alt` + `[` twice to get the selection back to the previous rectangle, same for going ahead in selection history.
But when selection tool is used to select them this issue does not occur.
I can propose a fix to this issue after #2136 is merged if you'd like.

bakayu:
@Keavon @0HyperCube I think the fix is complete, I did some testing and it now works as intended, please review the changes.

Keavon:
!build

Keavon:
I am unable to even select a layer.

bakayu:
It works locally but seems to be breaking during build.. I will look into this and report back.

bakayu:
@Keavon would you mind producing a build to test the latest changes ^^

0HyperCube:
!build

Keavon:
!build

Keavon (approved):

PR #2139, opened 2024-12-15
Display acceptable input types  for protonodes

### LINKED ISSUE #2069: Add list of acceptable types when hovering over node input

prikshitsingh24:
Closes #2069

It shows the acceptable input types of the protonodes which looks something like this:

brave_8ZQtQMwnrm

brave_nWHsuolAfX

brave_qTkFsElQYy

### DIFF
--- a/frontend/src/components/views/Graph.svelte
+++ b/frontend/src/components/views/Graph.svelte
@@ -696,7 +696,7 @@
 							style:--data-color-dim={`var(--color-data-${node.primaryInput.dataType.toLowerCase()}-dim)`}
 							bind:this={inputs[nodeIndex + 1][0]}
 						>
-							<title>{`${dataTypeTooltip(node.primaryInput)}\n${inputConnectedToText(node.primaryInput)}`}</title>
+							<title>{`Supported types:\n${node.primaryInput.dataType}\n${dataTypeTooltip(node.primaryInput)}\n${inputConnectedToText(node.primaryInput)}`}</title>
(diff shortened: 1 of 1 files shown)


0HyperCube:
@prikshitsingh24 the idea is to list all the possible types that are possible; not just the currently chosen one. For example the transform node supports `VectorData`, `GraphicGroup`, `ImageFrame<Color>`, and `TextureFrame`.

prikshitsingh24:
Yep understood also I have to change how frontend_import sends data right ?

Keavon:
@0HyperCube @adamgerhant are either of you familiar with the status of where this PR resides in relation to the latest in our code base?

adamgerhant:
This feature is completed and active in dev

Keavon:
Thanks, I'll close this then as superseded.

Issue #1912, opened 2024-08-08
Make all dragging actions abortable with right click


Keavon:
This is an ongoing list of all actions we find which need to be made abortable by right-clicking. Right clicking should be possible anywhere in or outside the app window, not just over the widget that's being dragged.

## Header bar and widgets
- Menu bar entries should—first of all, open on mousedown and let you drag down and release on an entry—then right click in the menu to abort and close it
- Same as above for dropdown menus

## Layers panel
- Dragging to rearrange layers (in Firefox, this actually confirms the action instead of aborting it, in Chrome it just does nothing)

## Color picker
- Saturation/Value square
- Hue track
- Alpha track
- Gradient spectrum input knobs

## Viewport
- Scrollbars

## Tools
- Select & Path tools: drag-selecting, upon right clicking, should restore the selection before drag-selecting

## Node graph
- Box selection should, upon right clicking, restore the selection before drag-selecting
- Dragging nodes should restore their initial positions
- Dragging a wire should restore its initial connection

a-khushal:
@Keavon is this the expected behaviour??
https://github.com/user-attachments/assets/19489765-4ab1-4f29-a9f7-9b14bae62589

Keavon:
@a-khushal did you upload the wrong video?

a-khushal:
The video is the solution for this issue: "Menu bar entries should—first of all, open on mousedown and let you drag down and release on an entry—then right click in the menu to abort and close it", whenever mousedown happens on a menu, it opens the dropdown, and when dragged down and released on an entry, it selects that entry

Keavon:
@a-khushal I see, although I won't be able to tell much from the video. Please open a PR and I'll try it out.

a-khushal:
@Keavon please refer #2164

PR #2143, opened 2024-12-17
Make all dragging actions abortable with right click(Select-Tool) 

0SlowPoke0:
fixes only for the select-tool-part of #1912


Keavon:
!build

PR #2144, opened 2024-12-17
Fix text layer getting deselected after clicking out of Text tool interactive editing

### LINKED ISSUE #2127: Keep selection when clicking in properties after creating text
- Create a new text box
- Type something
- Attempt to modify something in the properties of the newly added text layer (e.g. change the fill colour)
- The layer is deselected so the properties disappear.

It would be nice to keep the properties around when the user is trying to edit them.

Sidharth-Singh10:
Closes #2127

- Introduced the `active_panel` field in `DocumentMessageHandle`r to track the currently active interface panel.
- Added a match expression in `set_editing` fn to verify if the active panel is the Properties Panel, preventing its deselection during edits.

https://github.com/user-attachments/assets/dd34f593-b3e0-4f54-b65a-f8cde3af87ff

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -144,7 +144,6 @@ impl Default for DocumentMessageHandler {
 			graph_view_overlay_open: false,
 			snapping_state: SnappingState::default(),
 			graph_fade_artwork_percentage: 80.,
-
 			// =============================================
 			// Fields omitted from the saved document format
 			// =============================================
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -17,7 +17,6 @@ use crate::messages::portfolio::document::properties_panel::utility_types::Prope
 use crate::messages::portfolio::document::utility_types::document_metadata::{DocumentMetadata, LayerNodeIdentifier};
 use crate::messages::portfolio::document::utility_types::misc::{AlignAggregate, AlignAxis, DocumentMode, FlipAxis, PTZ};
 use crate::messages::portfolio::document::utility_types::nodes::RawBuffer;
-use crate::messages::portfolio::utility_types::PanelType;
 use crate::messages::portfolio::utility_types::PersistentData;
 use crate::messages::prelude::*;
 use crate::messages::tool::common_functionality::graph_modification_utils::{self, get_blend_mode, get_opacity};
@@ -91,8 +90,6 @@ pub struct DocumentMessageHandler {
 	pub graph_view_overlay_open: bool,
 	/// The current opacity of the faded node graph background that covers up the artwork.
 	pub graph_fade_artwork_percentage: f64,
-	/// Tracks the currently active panel in the application interface.
-	pub active_panel: PanelType,
 
 	// =============================================
 	// Fields omitted from the saved document format
@@ -147,7 +144,7 @@ impl Default for DocumentMessageHandler {
 			graph_view_overlay_open: false,
 			snapping_state: SnappingState::default(),
 			graph_fade_artwork_percentage: 80.,
-			active_panel: PanelType::Document,
+
 			// =============================================
 			// Fields omitted from the saved document format
 			// =============================================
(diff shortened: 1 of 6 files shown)


0HyperCube (approved):
Works well; thanks for this contribution.

Keavon:
@adamgerhant could you please review this in regards to the introduced ActivePanel code, to make sure it doesn't conflict with your existing implementation of the related concept of the active panel used for showing whether the graph's subnetwork or Layers panel is active?

adamgerhant (changes requested):
The active panel is currently being stored in the portfolio message handler. There is no need to store it in the document message handler.

0HyperCube:
@adamgerhant the text tool code can't access the portfolio message handler.

adamgerhant:
It could be passed as a parameter, but I think it makes sense for the active panel to be stored on a per document basis instead of the entire portfolio/application. This means switching tabs will store independent active panel states. In this case the active panel should be removed from the portfolio message handler.

adamgerhant:
Why not just leave the text layer as selected when editing ends?

Sidharth-Singh10:
I haven't worked much with design tools much, but doesn't keeping text layers selected after editing deviates from standard design tool patterns where edits typically end in deselection.
Persistent selection might create risk for accidental modifications - users may unintentionally resize/move/style the text layer when they meant to work on something else.

In this implementation the layer while in editing is deselected when clicked anywhere else expect properties panel

adamgerhant (approved):

Keavon:
@Sidharth-Singh10 I like where you're going with that thought process, although I believe for consistency with other drawing tools in Graphite, it would be best to keep it selected after clicking out. If you draw with any other tool like Rectangle or Ellipse, you'll notice if you look in the Layers panel, that layer remains selected after finishing drawing it. So it'd be best if the layer remains selected if you click anywhere in the viewport, or anywhere else in the UI, after interactively editing the text. Thanks :)

Keavon (changes requested):
See my previous comment for the updated approach, thanks!

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -144,6 +144,7 @@ impl Default for DocumentMessageHandler {
 			graph_view_overlay_open: false,
 			snapping_state: SnappingState::default(),
 			graph_fade_artwork_percentage: 80.,
+
```
Keavon, line 147:
```suggestion
```

Keavon:
!build

Keavon:
Can I have you add another little improvement to this PR while I'm thinking about it? If the user is editing the interactive text in the viewport, and hits escape, it should delete the entire text layer if escape was pressed when the text was empty. So if a user creates text in the wrong spot but doesn't type anything (or types then deletes it all), escape will delete the layer so it doesn't get invisibly stuck behind.

Keavon:
!build

Keavon:
!build

Keavon:
It looks like <kbd>Escape</kbd> is working but right-click isn't for canceling an empty text layer for it to get deleted. Can you please look into making it work when it's empty and the user right-clicks?

(Also note that I rebased, so you'll need to pull before continuing to code on it.)

Sidharth-Singh10:
Hey @Keavon, I added an entry for `MouseRight` with `action_dispatch` set to `TextToolMessage::CommitText` in `input_mapping.rs`, but it doesn't seem to have any effect.

To confirm I was editing the correct location, I commented out the `MouseLeft` entry with `action_dispatch` set to `TextToolMessage::Interact`. This prevented placing or editing text, which suggests my changes are indeed in the right place.

Could you nudge me in the right direction?

Keavon:
I'll ping @0HyperCube to see if he can answer your question about that.

0HyperCube:
@Sidharth-Singh10 

The actions that are advertised to the input mapper system don't include the commit text when the text is not being edited:
```rs
fn actions(&self) -> ActionList {
	match self.fsm_state {
		TextToolFsmState::Ready => actions!(TextToolMessageDiscriminant;
			Interact,
		),
		TextToolFsmState::Editing => actions!(TextToolMessageDiscriminant;
			Interact,
			Abort,
			CommitText,
		),
	}
}
```

Any time any mouse button is pressed that is not inside the text input, `input.ts:175` sends `editor.handle.onChangeText` which sets editing to false.

Sidharth-Singh10:
Thanks for pointing this out!

Keavon:
!build

Keavon:
@Sidharth-Singh10 your solution to emulate an escape keypress from the frontend TS code is a hack that works but isn't very clean. Can you please take into account what Hypercube wrote and implement a solution in the backend Rust code? In other words, you need to have something run before the tool state leaves `Editing` and returns to `Ready` to check if that TS code he mentioned is causing it to return to `Ready`.

Keavon:
!build

Keavon (approved):

Issue #1896, opened 2024-08-05
Tracking Issue: Quick measurement overlays while holding the Alt key


Keavon:
- [x] First basic implementation (#1894)
- [x] Improvements to match the spec described below (#2147)
- [x] Tidy up the number text alignment and decimal precision (#2155)
- [ ] Make it work when the layers are rotated and skewed (#2396)

---

When holding the <kbd>Alt</kbd> key with a layer (or multiple) selected, while hovering over another layer (or group), an overlay should appear visualizing the AABB of both the selected and hovered layers, with lines and distance numbers (in document coordinates) between them as shown below.

When the document is viewed with a tilt in the viewport PTZ navigation, the lines become angled but the text labels stay upright for readability.

Here are, I believe, all the scenarios. The three diagrams with text do care about which is the selected layer and which is the hovered layer; all the others are drawn the same in either order, so it doesn't matter which is the selected or hovered one. The two colored boxes represent AABBs, not actual shapes in the artwork— those should be drawn with a blue outline around both their AABBs.

| Scenarios |
| - |
| <strong>Terminology definitions:</strong><br /><br />- "Boxes overlap along 0/1/2 axes" means extruding the box's edge infinitely left/right, or up/down, would hit the other box<br />- "0/1/2 of their centerlines cross the other boxes" means extending both box's center points infinitely left/right, or up/down, would hit the other box<br />- "Along axis A/B" means either the X and Y, or the Y and X, axes<br />- "box A/B" means either one particular box and its opposite, or the reverse<br /><br /> |
| Boxes overlap along 0 axes:<br /><br />![image](https://github.com/user-attachments/assets/4f1e6752-f4c4-4494-8b76-c34b3903f20b) |
| Boxes overlap along 1 axis and 0 of their centerlines cross the other boxes:<br /><br />![image](https://github.com/user-attachments/assets/3dd9bbc2-a55a-4465-b5bb-7c9bd1511f15) |
| Boxes overlap along 1 axis and 2 of their centerlines cross the other boxes:<br /><br />![image](https://github.com/user-attachments/assets/79bb251b-31bd-49d6-90a4-cdc4518715fb) |
| Boxes overlap along 1 axis and 1 of their centerlines cross the other boxes:<br /><br />![image](https://github.com/user-attachments/assets/2fdd2181-93f8-4780-b7d5-afdee3f9ab09) |
[...]

PR #2147, opened 2024-12-21
Improve quick measurement overlays across all possible arrangement scenarios

singhutsav5502:
Adds further code implementation for #1896. This PR adds code that handles all 9 cases as described within the issue, support for individually rotated objects still needs to be added.

### DIFF
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -72,6 +72,36 @@ impl OverlayContext {
 		self.render_context.line_to(end.x, end.y);
 		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
 		self.render_context.stroke();
+		self.render_context
+			.set_line_dash(&JsValue::from(js_sys::Array::new()))
+			.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
+			.ok();
+	}
+
+	pub fn dashed_line_with_pattern(&mut self, start: DVec2, end: DVec2, color: Option<&str>, dash_width: f64, gap_width: f64) {
+		let start = start.round() - DVec2::splat(0.5);
+		let end = end.round() - DVec2::splat(0.5);
+
+		// Set the dash and gap pattern
+		let array = js_sys::Array::new();
+		array.push(&JsValue::from(dash_width));
+		array.push(&JsValue::from(gap_width));
+		self.render_context
+			.set_line_dash(&JsValue::from(array))
+			.map_err(|error| log::warn!("Error setting line dash: {:?}", error))
+			.ok();
+
+		self.render_context.begin_path();
+		self.render_context.move_to(start.x, start.y);
+		self.render_context.line_to(end.x, end.y);
+		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
+		self.render_context.stroke();
+
+		// Reset the dash pattern after drawing
+		self.render_context
+			.set_line_dash(&JsValue::from(js_sys::Array::new()))
+			.map_err(|error| log::warn!("Error resetting line dash: {:?}", error))
+			.ok();
 	}
 
 	pub fn manipulator_handle(&mut self, position: DVec2, selected: bool) {
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -72,6 +72,8 @@ impl OverlayContext {
 		self.render_context.line_to(end.x, end.y);
 		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
 		self.render_context.stroke();
+
+		// Reset the dash pattern after drawing
 		self.render_context
 			.set_line_dash(&JsValue::from(js_sys::Array::new()))
 			.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -53,7 +53,7 @@ impl OverlayContext {
 		let start = start.round() - DVec2::splat(0.5);
 		let end = end.round() - DVec2::splat(0.5);
 		if let Some(dash_width) = dash_width {
-			let gap_width = gap_width.unwrap_or(1.0);
+			let gap_width = gap_width.unwrap_or(1.);
 			let array = js_sys::Array::new();
 			array.push(&JsValue::from(dash_width - 1.));
 			array.push(&JsValue::from(gap_width));
@@ -74,7 +74,7 @@ impl OverlayContext {
 		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
 		self.render_context.stroke();
 
-		// Reset the dash pattern after drawing
+		// Reset the dash pattern to solid after drawing
 		self.render_context
 			.set_line_dash(&JsValue::from(js_sys::Array::new()))
 			.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -46,16 +46,17 @@ impl OverlayContext {
 	}
 
 	pub fn line(&mut self, start: DVec2, end: DVec2, color: Option<&str>) {
-		self.dashed_line(start, end, color, None)
+		self.dashed_line(start, end, color, None, None)
 	}
 
-	pub fn dashed_line(&mut self, start: DVec2, end: DVec2, color: Option<&str>, dash_width: Option<f64>) {
+	pub fn dashed_line(&mut self, start: DVec2, end: DVec2, color: Option<&str>, dash_width: Option<f64>, gap_width: Option<f64>) {
 		let start = start.round() - DVec2::splat(0.5);
 		let end = end.round() - DVec2::splat(0.5);
 		if let Some(dash_width) = dash_width {
+			let gap_width = gap_width.unwrap_or(1.0);
 			let array = js_sys::Array::new();
-			array.push(&JsValue::from(1));
 			array.push(&JsValue::from(dash_width - 1.));
+			array.push(&JsValue::from(gap_width));
 			self.render_context
 				.set_line_dash(&JsValue::from(array))
 				.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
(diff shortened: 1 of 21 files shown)


Keavon:
!build

Keavon:
!build

singhutsav5502:
The Clippy build fail seems unrelated to the code i have changed.

Keavon:
Yeah, that randomly began failing recently. It's because PR from forks can't access our repo's secret key used by the bot to write the results comment. But it wasn't an issue before. I'm still working on figuring out a strategy to deal with that, such as maybe getting it to send the comment text to a separate workflow that is allowed to access the secret key in order to make the API call to make the comment. (If you happen to have any experience in these matters, feel free to give suggestions, but that's the context here.)

Keavon:
Nice, comprehensive work on this! Some quick QA feedback:

All the cases presented in the issue are working as expected, except for the first two scenarios. Could you please take a closer look at those two? Please pay attention, specifically, to the dashed lines, where the number labels go, and for the first scenario also how the lines connect to the corners of the opposing corner.

Thank you!

singhutsav5502:
Thanks! unfortunately i have no experience with clippy, Will fix the first 2 cases ASAP.

singhutsav5502:
I have pushed the changes for the first 2 cases @Keavon

Keavon:
!build

Keavon:
Ok, nice! A few notes:
- The dashed line should please be 2px of blue then 2px of blank
- It seems like, once activated, the dash begins infecting a bunch of the other overlays like the transform cage itself; whatever engineering choice was made to cause this to happen seems to be stateful? Assuming it's stateful, please try to pick an alternative state-free approach so this doesn't have any possibility of happening.

singhutsav5502:
I see, i probably overlooked the issue. I have been using the `dashed_line` method of the `overlay_context` to render the dashed lines. I will try to come up with a solution that uses the normal line method instead so that the transform cage doesn't get affected.

Keavon:
Oh I see, it's the JS drawing API that's keeping the state. I believe you just need to add this to the very end of the `pub fn dashed_line()` function to reset it:

```rs
self.render_context
	.set_line_dash(&JsValue::from(js_sys::Array::new()))
	.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
	.ok();
```

Since that, I believe, should satisfy [MDN's note](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/setLineDash):

> Note: To return to using solid lines, set the line dash list to an empty array.

singhutsav5502:
hi @Keavon i have pushed the needed changes.

Keavon:
!build

editor/src/messages/portfolio/document/overlays/utility_types.rs
```diff
@@ -72,6 +72,36 @@ impl OverlayContext {
...
+			.map_err(|error| log::warn!("Error setting line dash: {:?}", error))
+			.ok();
+
+		self.render_context.begin_path();
+		self.render_context.move_to(start.x, start.y);
+		self.render_context.line_to(end.x, end.y);
+		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
+		self.render_context.stroke();
+
+		// Reset the dash pattern after drawing
```
Keavon, line 100: Please replicate this comment in the function above.

editor/src/messages/portfolio/document/overlays/utility_types.rs
```diff
@@ -72,6 +72,36 @@ impl OverlayContext {
 		self.render_context.line_to(end.x, end.y);
 		self.render_context.set_stroke_style_str(color.unwrap_or(COLOR_OVERLAY_BLUE));
 		self.render_context.stroke();
+		self.render_context
+			.set_line_dash(&JsValue::from(js_sys::Array::new()))
+			.map_err(|error| log::warn!("Error drawing dashed line: {:?}", error))
+			.ok();
+	}
+
+	pub fn dashed_line_with_pattern(&mut self, start: DVec2, end: DVec2, color: Option<&str>, dash_width: f64, gap_width: f64) {
```
Keavon, line 81: What's your reasoning for making two similar functions? I think we could reduce the lines of code by just making the one function more customizable to encompass all the requirements.

Keavon:
Please make the dash pattern 2 pixels of blue, 2 pixels of blank. I think right now it's 3 pixels of blue, 1 pixel of blank?

editor/src/messages/portfolio/document/overlays/utility_types.rs
singhutsav5502: The pre-existing function seems to be incorrectly defined the `dash_width` is being used as the 2nd parameter of the array which effectively makes it the length of the `gap` between two dashes of the line. and the actual dash width (the 1st parameter) seems to be hard set at `1` . 
And this function seems to be used at multiple other places, so i decided to make an extra function that exposes both gap_width and dash_width as individual parameters. ( since a 2px dash was needed to be followed with  a 2px blank space ).

Keavon: I see, thanks for the explanation. Can you please refactor it to use one function, preserving the appearance of existing usages (where the width is `gap` and `1` functionality you described)?

editor/src/messages/tool/common_functionality/measure.rs
```diff
@@ -39,7 +39,7 @@ fn draw_zero_axis_crossings(selected_bounds: Rect, hovered_bounds: Rect, transfo
 	let min_viewport = transform.transform_point2(line_start);
 	let max_viewport = transform.transform_point2(line_end);
 
-	overlay_context.dashed_line(min_viewport, max_viewport, None, Some(8.0));
+	overlay_context.dashed_line_with_pattern(min_viewport, max_viewport, None, 2.0, 2.0);
```
singhutsav5502, line 42: based on the function description and the input parameters the dash length and the gap length should both be set to 2px. I will look into this.

singhutsav5502: on actual render it does seem to be 3pixels of blue and 1 pixel of blank. I have probably overlooked something in the function definition. will fix ASAP.

Keavon: Yeah, basically, I'd say it's best if you try to combine these functions into one, and then figure out how to make all existing and new usages display their intended dash spacing so the end result is a function that has parameters that behave expectedly but without changing the actual behavior of other usages. Please make note of each existing dash usage and test to ensure it looks identical afterwards compared to https://dev.graphite.rs (the latest master build). It would also be helpful if you could list those locations in a comment here so I can double check them for you.

editor/src/messages/portfolio/document/overlays/utility_types.rs
singhutsav5502: sure!

editor/src/messages/tool/common_functionality/measure.rs
singhutsav5502: -  the `dashed_line` method was being used in the `grid_overlays.rs` file.

-  the `measure.rs` file ( the one currently being edited ). 

Other than that there doesn't seem to be a direct reference to the method in the entire code base. The `line` method leverages the `dashed_line` method indirectly. 
the `line` method is being directly used by the following: ( i am not yet sure if they indirectly leverage dashed_line in some manner, testing would need to be done. )

- `snapping.rs`
- `pen_tool.rs`
- `utility_functions.rs` for bezier functions.

I have refactored the original function to work properly ( as it did before ) for cases where `gap_width` needs to be set to 1.
I have also set the `gap_width` parameter to None in the appropriate places (`grid_overlays.rs` and the `line` method implementation).

Keavon: Thanks! Does this also solve the 2x2 dashes appearing as 3x1 dashes? I can do another build once you've resolved that too.

singhutsav5502: From the things i have tested on my device so far the following seem to be working as expected: Grids, Spline Tool (dash parameters ), Line Tool( dash parameters )  and the Pen Tool ( dash parameters ). Not sure if i missed something but do look into it.

Keavon:
!build

editor/src/messages/tool/common_functionality/measure.rs
singhutsav5502: > Thanks! Does this also solve the 2x2 dashes appearing as 3x1 dashes? I can do another build once you've resolved that too.

It seems to be an issue that was already resolved in the pre existing function with how `1` was being subtracted from the dash_width, The dash width and gap width should work properly they looked to be equidistant on my local build.

Keavon:
With this build, it's 2px blue 1px blank. Can you please adjust accordingly?

singhutsav5502:
> With this build, it's 2px blue 1px blank. Can you please adjust accordingly?
Hi, does the image below seem okay? if so, i will push the code.
image

Keavon:
Hard to tell from the screenshot but probably!

singhutsav5502:
pushed the new code.

Keavon:
!build

Keavon:
The dash spacing is perfect now, thanks!

Keavon:
So my only remaining comment about this is the text alignment, however I think that's best to save that for a later PR. Are you interested in working on that after we merge this?

singhutsav5502:
> So my only remaining comment about this is the text alignment, however I think that's best to save that for a later PR. Are you interested in working on that after we merge this?

Sure! Though I think I would like to solve a few more smaller issues from the task list before i come back to this code again.

Keavon:
Sure, sounds good. Also the TODO you left about the rotation will also need to be solved eventually too, but it's also best to wait for a later PR. Feel free to work on these or other issues in any order you prefer! Ping me on Discord and we can also help you pick out good ones, but I think you're at a skill level where you can handle most of the issues without too much challenge (as long as it's not blocked by anything else at the moment which would be context only we may have in our knowledge).

singhutsav5502:
Absolutely. Thanks!
I will try and be a lot more active going forward since I am currently a uni student and plan to come up with a gsoc proposal as well. 👍

Keavon:
Sweet, we'll be glad to have your work over the coming months and working together on your GSoC proposal when the applications open (of course, contingent on our org getting accepted which technically hasn't happened yet because we org applications haven't opened yet).

singhutsav5502:
Hi, is there any actionable left from my side which might be an hinderance for merging this?

Keavon:
I just have to do a code review. I pushed a quick commit with some nit picks (feel free to read over the diff for learning purposes about our stylistic preferences). I need to do a detailed code read-through tomorrow. If you're able to improve the code by means of reducing code duplication, that's my current concern for areas that are most likely to require change requests from me, since I feel like there's a sizable amount of similarity in `measure.rs` within all those similar functions.

singhutsav5502:
> I just have to do a code review. I pushed a quick commit with some nit picks (feel free to read over the diff for learning purposes about our stylistic preferences). I need to do a detailed code read-through tomorrow. If you're able to improve the code by means of reducing code duplication, that's my current concern for areas that are most likely to require change requests from me, since I feel like there's a sizable amount of similarity in `measure.rs` within all those similar functions.


Sure, the biggest part of duplicated code seems to be the call for the `line` function. I can probably wrap it into a smaller function that exposes the only 2 changing variables in each call ( rest of the attributes remain same across all function calls in the `measure.rs` file) , for this specific use case but i am not sure if that would be considered redundant or not.

Keavon:
Sounds like a good change. Any code that does the same thing in multiple places should be consolidated, because that helps reduce the line count of your PR (less code for maintainers to read in that file and thus quicker to understand) and more maintainable by avoiding the possibility of changing one place but forgetting to replicate identical changes in related places. Apply that philosophy everywhere, since we aim for minimal lines of code additions and avoiding duplication wherever it can sensibly be achieved.

singhutsav5502:
> Sounds like a good change. Any code that does the same thing in multiple places should be consolidated, because that helps reduce the line count of your PR (less code for maintainers to read in that file and thus quicker to understand) and more maintainable by avoiding the possibility of changing one place but forgetting to replicate identical changes in related places. Apply that philosophy everywhere, since we aim for minimal lines of code additions and avoiding duplication wherever it can sensibly be achieved.

Hi @Keavon, just pushed a code refactor it reduces the code length quite a bit.

Keavon:
!build

Keavon (approved):

PR #2149, opened 2024-12-22
Fix transform cage rotation abort causing broken state upon next transformation

singhutsav5502:
Ensures Abort and Redraw events are added to the responses when an Abort event occurs.


Keavon:
!build

Keavon (approved):

PR #2164, opened 2024-12-25
Make menu bar entries clickable by dragging from the header button and releasing on the entry

### LINKED ISSUE #1912: Make all dragging actions abortable with right click

a-khushal:
Closes #1912 (header bar and widgets sub-issue).

### DIFF
--- a/frontend/src/components/floating-menus/MenuList.svelte
+++ b/frontend/src/components/floating-menus/MenuList.svelte
@@ -427,6 +427,7 @@
 					on:click={() => !entry.disabled && onEntryClick(entry)}
 					on:pointerenter={() => !entry.disabled && onEntryPointerEnter(entry)}
 					on:pointerleave={() => !entry.disabled && onEntryPointerLeave(entry)}
+					on:mouseup={() => !entry.disabled && onEntryClick(entry)}
 				>
 					{#if entry.icon && drawIcon}
 						<IconLabel icon={entry.icon} class="entry-icon" />
--- a/frontend/src/components/widgets/buttons/TextButton.svelte
+++ b/frontend/src/components/widgets/buttons/TextButton.svelte
@@ -8,6 +8,7 @@
 	import TextLabel from "@graphite/components/widgets/labels/TextLabel.svelte";
 
 	let self: MenuList;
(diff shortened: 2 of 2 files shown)


Keavon:
!build

Keavon:
Your implementation removes the ability to click on the menus without them disappearing again immediately.

a-khushal:
Can you please explain about the expected behaviour

Keavon:
The expected behavior should be the same as with any other software application that has a menu bar, e.g. VS Code. Give that a try and experiment a bit.

a-khushal:
@Keavon, can you please check now, I have pushed changes to the same branch. 
I have implemented features similar to VS code's menubar:
- select a menu entry by dragging from the header and releasing on that entry 
- right click on menu header to close the menu

Keavon:
!build

Keavon:
This breaks when holding down the mouse and moving left/right from one menu bar button to another, like from "File" to "Edit". It then requires several more clicks to regain a non-broken state so the menus work correct. Reference VS Code again for how you should be able to switch left and right between menus while click-and-release or click-and-hold ing.

Keavon:
I'm going to close this since it seems to be stalled, but feel free to reopen it if you make progress.

PR #2175, opened 2025-01-03
Bezier-rs: Add function to calculate if a subpath is inside polygon

### LINKED ISSUE #2174: Add bezier-rs api to check if Subpath is inside the Polygon

indierusty:
Closes #2174

### DIFF
--- a/libraries/bezier-rs/src/subpath/solvers.rs
+++ b/libraries/bezier-rs/src/subpath/solvers.rs
@@ -1,6 +1,6 @@
 use super::*;
 use crate::consts::MAX_ABSOLUTE_DIFFERENCE;
-use crate::utils::{compute_circular_subpath_details, line_intersection, SubpathTValue};
+use crate::utils::{compute_circular_subpath_details, is_rectangle_inside_other, line_intersection, SubpathTValue};
 use crate::TValue;
 
 use glam::{DAffine2, DMat2, DVec2};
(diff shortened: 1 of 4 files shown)


libraries/bezier-rs/src/utils.rs
```diff
@@ -179,6 +179,19 @@ pub fn do_rectangles_overlap(rectangle1: [DVec2; 2], rectangle2: [DVec2; 2]) ->
 	top_right1.x >= bottom_left2.x && top_right2.x >= bottom_left1.x && top_right2.y >= bottom_left1.y && top_right1.y >= bottom_left2.y
 }
 
+pub fn is_subpath_inside_polygon<PointId: crate::Identifier>(polygon: &[DVec2], subpath: &Subpath<PointId>) -> bool {
+	let polypath = Subpath::from_anchors_linear(polygon.to_vec(), true);
```
Keavon, line 183: Please make the caller do this rather than taking a `&[DVec2]` as a function parameter.

Keavon (changes requested):

libraries/bezier-rs/src/utils.rs
```diff
@@ -179,6 +179,19 @@ pub fn do_rectangles_overlap(rectangle1: [DVec2; 2], rectangle2: [DVec2; 2]) ->
...
+	if !subpath.subpath_intersections(&polypath, Some(0.02), Some(0.05)).is_empty() {
```
Keavon, line 184: Please leave a comment about the reasoning for needing to replace the defaults with 0.02 and 0.05.

libraries/bezier-rs/src/utils.rs
indierusty: Got it. Correcting that.

libraries/bezier-rs/src/utils.rs
indierusty: Is it better to let the caller decide error and minimum separation parameters

Keavon: Potentially, if needed. Did you encounter problems with the defaults that made you specify something other than the defaults?

libraries/bezier-rs/src/utils.rs
```diff
@@ -179,6 +179,19 @@ pub fn do_rectangles_overlap(rectangle1: [DVec2; 2], rectangle2: [DVec2; 2]) ->
...
+		return false;
+	}
+	for anchors in subpath.anchors() {
+		if !polypath.contains_point(anchors) {
+			return false;
+		}
+	}
```
Keavon, line 191: We should do benchmarks to compare, but I have a feeling this is faster and should be put first. I also suggest doing the cheap early-exit bounding box check that you had suggested, and coming up with other optimizations.

Please also put a comment describing the logical justification for each of these checks and how it narrows the possibility space to result in a correct answer.

libraries/bezier-rs/src/utils.rs
indierusty: I dicided the parameter to be less than default for precise lasso selection but on giving second though it could be waste efficiency as user may not need that precision. we can always change it later.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -237,6 +237,37 @@ impl<PointId: crate::Identifier> Subpath<PointId> {
 		false
 	}
 
+	/// Returns `true` if this subpath is completely inside the other subpath.
```
Keavon, line 240: Please add the iframe demo like in other API functions.

libraries/bezier-rs/src/subpath/solvers.rs
```diff
@@ -237,6 +237,37 @@ impl<PointId: crate::Identifier> Subpath<PointId> {
...
+			}
+		}
+
+		// Eliminate this if its subpath is intersecting wih the other's subpath
+		if !self.subpath_intersections(&other, error, minimum_separation).is_empty() {
+			return false;
+		}
+
+		// here (1) this subpath bbox is inside other bbox, (2) its anchors are inside other subpath and (3) it is not intersecting with other subpath.
+		// hence this this subpath is completely inside given other subpath.
```
Keavon, line 267: Nit: capitalize the starts of sentences, please.

Keavon:
Because this branch was named `master` and I had to do a bunch of workarounds to even be able to pull this down locally for my code review, I ended up accidentally pushing my commit to the `origin` remote instead of your fork's remote, which merge all these commits into the mainline master and I had to pop those off and revert it. But unfortunately GitHub thinks this PR was merged, and that can't be undone to reopen it, so you'll need to open a new one. Before you do that, please rename your branch something other than `master`.

PR #2177, opened 2025-01-04
Fix Document > Clear Artboards so it doesn't also clear everything else

moOsama76:
Continuation of first attempt with unmerged PR #2015


Keavon:
!build

Keavon:
Problem: this introduces an intermediate history step which you'll see if you open the graph and undo three times to get back when it should have taken just one step.

Other minor problem: it seems that the Artboard layer gets replaced by a Merge layer but it ends up in a different position in the graph. Ideally this wouldn't cause the layers to shift. In a case with multiple artboards each containing layers, I'm finding this even results in things overlapping in the graph after running this which isn't a suitable outcome.

Thanks, and welcome back!

Keavon:
!build

Keavon:
Can you please make it add a Transform node, feeding into the replacement Merge node, if the artboard has a nonzero *Location*? Currently, all former artboards teleport to having a top-left location at the canvas origin since the artboard offset is lost.

Keavon:
Marking this as a draft in the meantime. And copying what I sent you in Discord for convenient reference:

> Hi, I got a chance to review your PR again (sorry I've been so so so busy lately). It looks like the way you apply transformation of the artboard to the constituent layers is by deeply traversing the layer hierarchy and applying it to the roots. But this seems to cause bugs (for example, try setting the artboard for Isometric Fountain to a non-default X, Y position then clearing artboards and watch how the plants and pond ripples move). It's also not necessarily desirable how it ends up adding Transform nodes to all the individual layers rather than setting it on just one upstream Transform node that already exists. Can you make it set the Transform node on the least-nested layer of each artboard-that-becomes-a-layer rather than deeply setting it on all deepest-nested layers?

Thanks and great work so far, this is close!

moOsama76:
> Marking this as a draft in the meantime. And copying what I sent you in Discord for convenient reference:
> 
> > Hi, I got a chance to review your PR again (sorry I've been so so so busy lately). It looks like the way you apply transformation of the artboard to the constituent layers is by deeply traversing the layer hierarchy and applying it to the roots. But this seems to cause bugs (for example, try setting the artboard for Isometric Fountain to a non-default X, Y position then clearing artboards and watch how the plants and pond ripples move). It's also not necessarily desirable how it ends up adding Transform nodes to all the individual layers rather than setting it on just one upstream Transform node that already exists. Can you make it set the Transform node on the least-nested layer of each artboard-that-becomes-a-layer rather than deeply setting it on all deepest-nested layers?
> 
> Thanks and great work so far, this is close!

It's okay I am not like waiting for it to be merged, I appreciate you reviews :)

Keavon:
!build

Keavon (approved):

Keavon:
Nice work on this, thanks.

moOsama76:
> Nice work on this, thanks.

thanks for code review I read it line by line

moOsama76:
@Keavon actually there are 2 problems with the current code
1 - adding a transaction actually makes it take 2 undos to go back to original state, I didn't know why but I just didn't add because it's added somehow

https://github.com/user-attachments/assets/e9f20265-3eb4-45fd-be6b-d433ce444d2d


2- returning from the scope if there are no artboards doesn't prevent the need for N undos, that's why I undo before returning

the reason for both behaviors is likely the same but I couldn't figure it out, it's probably frontend related

PR #2178, opened 2025-01-05
Line tool drawing with start and end points

### LINKED ISSUE #2090: The shape tools should set shape node parameters, not transformations of unit shapes
Should go hand-in-hand with #1715.

Presently, if I use the Line tool to draw a line:

capture

It doesn't produce just a Line node with *Start* and *End* set to the appropriate coordinates. Instead, it uses the unit line `(0, 0)` to `(1, 0)` and applies some highly scaled transform with a Transform node to actually position it in space:

capture

The desired result of drawing with the Line tool shouldn't include a Transform node.

`line_tool.rs` is a good place to start looking for this code (although it could potentially be in a helper function in another file).

---

This also happens with the Rectangle tool:

capture
capture

Here, the rectangle is a unit square (`1`x`1` units). The desired result of drawing with the Rectangle tool should apply the scale to the Rectangle node and the translation (and if the canvas is tilted, also rotation) to the Transform node, leaving it at the default scale factor of `1`x`1`.

`rectangle_tool.rs` is a good place to start looking for this code.

---

This also happens with the Ellipse tool:

capture
capture

The desired behavior is the same as the Rectangle tool.

`ellipse_tool.rs` is a good place to start looking for this code.

---

This also happens with the Polygon tool, in "convex" mode:

capture
capture

And in "star" mode:

capture
capture

In this case, like with Ellipse and Rectangle, we want to represent the scale with the shape generator nodes. However, because the Star and Regular Polygon nodes don't have parameters for non-uniform aspect ratios, the Transform node will still need to represent the stretch factor. We just don't want to use the unit (`1`x`1`) shapes anymore. It would probably be preferable to represent the smaller dimension as a scale factor of `1` in the Transform node so that the larger dimension can be represented as a value larger than `1` in the Transform node in order to give it the desired stretch.

`polygon_tool.rs` is a good place to start looking for this code.

November-6:
Closes  #2090

This isnt a proper fix for the issue but a starting point for it.
> I have moved the layer and node creation to when the dragstop message fires. This way we have the starting and end points with us when we want to draw the line. 
> But, with this approach the line preview is not being shown i.e i cant see the line being formed as im drawing it. I only see it once i release click.
> Please provide help with how to get started with implementing a preview of line as well as the snapping and angle functionality.


Keavon:
Welcome back! In regards to your quoted text, how does that relate to the current situation after your rollback? What is the status of the code as you've committed it presently?

November-6:
The current code implements the required functionality in its most basic form, it draws a line based on starting and end points. Although preview drawing of line and shift ctrl functionalities are lost which i need help introducing again. You can see this in the issues tab where I presented a screen recording of it. The code is (mostlyyyyy) in the same state as it was when I uploaded that video. 

The rollback I referred to was from a solution to a node problem I was having where I tried to use Rc/Arc to reference the node instead of passing it directly into nodes vector. It had its own issues so I came back to my first draft.

Also regarding the review I added is actually a bug for other tools which I recently introduced while trying to implement a preview myself. Its fixed now.

Keavon:
Since there isn't much recent progress and this PR accounts for only a small portion of the main issue, I am going to close this. Feel free to keep experimenting in your fork's branch and reopen this same PR if you make progress and fully implement the scope of the original issue in some state that could be merged. Thanks!

PR #2179, opened 2025-01-06
Fix Alt-drag duplication not preserving the original layer order

pratikagr:
- in the function `start_duplicates`, `get_calculated_insert_index` was taking all the layers in the parameter but we needed to only pass the layer that is currrently in the loop so that it will return the correct index to push the layer.

- I passed the current layer `SelectedNodes(vec![layer.to_node()])` under the loop to that function.

https://discord.com/channels/731730685944922173/881073965047636018/1294525859142569994


Keavon:
!build

Keavon:
I'm cleaning up old incomplete issues, so this is getting closed since it doesn't work and it's been a couple weeks without progress. Feel free to reopen this same issue when you have it working and fully tested.

Issue #1490, opened 2023-12-07
Avoid using magic number indexes in the node system


Keavon:
There are tons of fragile index values like `0`, `1`, `2`, etc. used to reference inputs and outputs (amongst other things?) between nodes and node internals. This is extremely fragile, makes it hard to refactor, provides a difficult developer experience, lacks self-documenting clarity, and is hard to debug.

One potential solution could be using enums. But thinking through some alternative ways to achieve this could also be good.

Kakapio:
Hey Keavon,

I'd be interested in contributing to this. Any chance we can chat about this issue?

Thanks

Keavon:
Hi @Kakapio, I don't think this issue is likely to be an easily-solved one that's easy for newcomers to jump on. But I'd encourage you to ask me on Discord for some advice about which issues you can get started with based on your background and I can direct you to some issues that are easier to begin with. Thanks!

Tyooughtul:
I'll work on this issue, please assign it to me.😊

Keavon:
We don't do assignments until your PR is merged, but go for it! If you don't open a PR within a few days, this is fair game for someone else to work on.

Keavon:
Also, see the note about AI usage I just added to the top of this issue description.

Tyooughtul:
<img width="877" height="219" alt="Image" src="https://github.com/user-attachments/assets/93530214-7f0c-4df8-8f24-86d8c0fd3139" />

 Hi@Keavon!  I noticed that the test is failing, but I think it's unrelated to my PR for issue #1490. From what I can tell, it's a precision issue with the LUT in discrete_srgb.rs, the precomputed values don't exactly match the reference implementation due to floating-point rounding errors. I'd be happy to open a separate PR to fix this LUT issue if you'd like. Let me know what you think. Thanks!😊

Keavon:
@Tyooughtul does the failure occur on the master branch? If so, that would indicate it is happening irrespective of your work in this branch. I'll take a look when you open a PR.

Tyooughtul:
<img width="748" height="295" alt="Image" src="https://github.com/user-attachments/assets/b21c3363-2763-451e-b4da-b0c6d5cb0980" />  
it's true the error happens in the master branch

PR #2181, opened 2025-01-06
Add tests to the Ellipse, Artboard, and Fill tools

### LINKED ISSUE #2146: Artboard constrain to square broken on tilted document
When the document is tilted (<kbd>Alt</kbd>+middle mouse drag), constrain to square with <kbd>Shift</kbd> on the artboard doesn't work and instead constrains to some other aspect ratio.

Found whilst trying to write my first test.

0HyperCube:
A not very good attempt at #1490

Closes #2146 because it broke the test.


Keavon (approved):
Fantastic, thank you for pushing for having more tests!

Keavon:
> A not very good attempt at https://github.com/GraphiteEditor/Graphite/issues/1490

Is that accurate or did that get referenced in the wrong ticket? I didn't see changes relating to that when I read the diff.

0HyperCube:
@Keavon it does provide a system where an identifier can be used instead of an index to access an input (see changes in this PR to the codegen.rs). However I have only used these as part of the new tests and haven't yet modified other instances of the indices. An example of getting the input from an instrumented graph:
```rs
instrumented.grab_protonode_input::<ellipse::RadiusXInput>(&vec![ellipse_node], &editor.runtime).unwrap()
```

Keavon:
I see, that's cool! Looking forward to rolling that out throughout the rest of the code. I like that syntax.

PR #2185, opened 2025-01-10
Update the bounding box snapping modes to use Align with Edges for edges and alignment

mTvare6:
Fixes snapping behaviour which defaulted to X and Y axes.

Before and after:

Before After


Keavon:
!build

PR #2210, opened 2025-01-20
feat: add selection tolerance

### LINKED ISSUE #2141: Small click target when dragging/moving a straight line or small objects
Make a straight line with e.g. the pen tool. Using the select tool, try to move the line. Notice how it is not possible -- either the line is rotated, or the point of reference for rotation is moved -- the line cannot be dragged. Moving with the "G" button (grab) works, but that mode does not seem to support snapping.

A (similar?) issue occurs for 2D objects when they are (too) small: they cannot be dragged anymore, only resized or rotated.

Sidharth-Singh10:
Closes #2141
Modified the `check_layer` function in the document_message_handler to add tolerance-based selection

https://github.com/user-attachments/assets/e266ca3d-7585-4d43-9498-a2fcc5b78e3b

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -29,7 +29,7 @@ use graph_craft::document::value::TaggedValue;
 use graph_craft::document::{NodeId, NodeNetwork, OldNodeNetwork};
 use graphene_core::raster::{BlendMode, ImageFrame};
 use graphene_core::vector::style::ViewMode;
-use graphene_std::renderer::{ClickTarget, Quad};
+use graphene_std::renderer::{ClickTarget, Quad, Rect};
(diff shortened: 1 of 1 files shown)


Keavon:
!build

Keavon:
It looks like you're using the bounding box rather than actual proximity to the path. That causes paths like this (which have no Fill node) to be selectable from far too generous of a distance, because anywhere within its bounding box makes it accept the selection.

capture

Please note, we already have a logic for extending the selection to some slightler-larger-than-1.0 scale factor of the shape's stroke thickness. Notice how you can inflate the stroke and it becomes easy to select, even a bit beyond its stroke. I think the proper solution here is to discard your current approach and actually just modify that scale factor: use exactly 1.0 for large values (as measured in viewport space) and cap it to a minimum thickness for smaller values (as measured in viewport space) to ensure no line's click target becomes less thick than some minimum.

PR #2217, opened 2025-01-23
Make the layer tree show duplicate branches where they are used multiple times in the graph

### LINKED ISSUE #2153: The Layers panel doesn't show expandable children for the lower layer, only the top layer
capture

In the Layers panel on the right, *both* parent layers should show an expand arrow with child layers. Not only the top one.

mTvare6:
Fixes #2153


Keavon:
!build

Keavon:
With the following graph setup, upon closing the node graph or switching to the Select tool, we enter an infinite loop.

capture

Here's what the profiler shows is running in the infinite loop.

capture

Keavon:
If, during testing, you end up with the document autosaved and the browser freezing, you may have trouble opening localhost again after closing then revisiting the `localhost:8080` tab. If that occurs, you may need to quickly use F12 to open dev tools and expand IndedexDB storage to clear it, potentially fast enough before the infinite loop begins rendering dev tools unresponsive. Here's how to clear it:

Chrome:

capture

Firefox:

capture

mTvare6:
Seems to be the case that changing order works fine:

Keavon:
Closing as stale/stalled.

PR #2224, opened 2025-01-25
Give the current snapping target layer(s) an outline

### LINKED ISSUE #2220: Outline entire layer for snap to intersections/paths
When snapping to an intersection or path, currently just the relevant bézier curve(s) are displayed. @Keavon wishes for the entire layer to be displayed instead.

Keavon notes:

> Lighting up just the relevant subpath involved in the snapping would be best, but all the layer's paths would be fine too if that's much easier to implement.

Hypercube notes:
> The issue with subpaths is that we don't have any way to identify them due to the mesh vector data format.

0HyperCube:
Closes #2220


Keavon:
Could you please make this work with `[Path: Anchor Point] to [Path: Anchor Point]`?

https://github.com/user-attachments/assets/63c7ab60-eec7-4b27-8a19-c91a49087f16
https://github.com/user-attachments/assets/0cf424be-b76f-4f2f-a1d4-74722704fb4c

0HyperCube:
@Keavon I have added this additional feature.

Keavon (approved):
Thanks, this is a nice QoL improvement to make it feel just a bit more polished in ways users often don't even notice but do feel.

PR #2226, opened 2025-01-26
Make the transform cage resize about the pivot when Alt is pressed

0HyperCube:
As described in the voice meeting. Solves https://discord.com/channels/731730685944922173/881073965047636018/1332463810841411594

The resizing bounds now acts around the old pivot rather than centring on the old pivot.


Keavon:
Can you please add the division by zero check to ignore the <kbd>Alt</kbd> key when the pivot is on the edge being dragged? (Or perhaps if it's, say, within 1% the bounding box width which would mean holding <kbd>Alt</kbd> would result in a 100x growth ratio between the side being dragged and the fast-growing opposite side in the most extreme case before we ignored <kbd>Alt</kbd>.)

Keavon:
(This worked in this prior point in time https://65b3a12e.graphite.pages.dev/ however it suffered from the same division-by-zero edge case.)

Keavon (approved):

Issue #2184, opened 2025-01-10
Make the Spline tool draw a polylines with a `Path -> Spline from Points` node setup


Keavon:
- [x] **Part 1** (#2200)

    Listen to the task description [in this video](https://files.keavon.com/-/ChillyTenseAdamsstaghornedbeetle/capture_16_.mp4).

- [x] **Part 2** (#2203)

1. Make the Spline tool visualize the endpoints of each spline path just like the Freehand tool does (even if there are multiple spline paths because the Path node has multiple subpaths)
2. Allow clicking on the endpoints of a spline (on either end), when not drawing, to begin drawing a continuation of the spline
3. Allow clicking the start point while drawing spline segments to add a segment which closes the path given to the Spline from Points node
4. Allow holding <kbd>Shift</kbd> to begin drawing a new spline subpath in the same layer, instead of making it add a new layer (just like the equivalent functionality with the Pen tool)

- [x] **Part 3** (#2274)

1. Create an upgrade script to transform a single existing "Spline" node, in the graph of any existing documents people have today, into a `Path -> Spline from Points` node pair with the point data transferred over to the Path node's geometry
2. Delete the "Spline" node
5. Rename the "Spline from Points" node to just "Spline"
6. Create an upgrade script to rename the "Spline from Points" node in existing documents to "Spline"

    Search the project for "document upgrade code" and you'll find all the comments relating to the upgrading scripts. The ones in `portfolio_message_handler.rs`, within the handler for the `PortfolioMessage::OpenDocumentFileWithId` message, are most relevant to read and that's likely where you'll be adding most or all the upgrading code to.

- [ ] **Part 4** (PR 1 of ?: #2269, PR 2 of ?: coming soon)
  
  The advanced parts in https://files.keavon.com/-/PunyUprightStagbeetle/capture_35_.mp4

kliwongan:
Can I take this issue as my first issue?

Keavon:
@kliwongan this is currently being worked on by @indierusty (here's the [Discord thread](https://discord.com/channels/731730685944922173/1327206573071667202). If there isn't a viable PR opened within a week, feel free to ask again if it's available then.

Keavon:
@kliwongan I ended up adding a new part to this issue, so now it's a three-parter. Feel free to work on part 3, which isn't blocked by the previous two parts, if you would like. If you plan to work on it, please let me know when you'll be beginning.

indierusty:
[...]

PR #2227, opened 2025-01-26
Add joining of path endpoints with Ctrl+J in the Path tool

Sidharth-Singh10:
fixes: https://discord.com/channels/731730685944922173/881073965047636018/1289168449117097994

Implements a keyboard shortcut (Ctrl+J) to close open paths, with support for multiple scenarios:

1. With selected points: Connects two selected endpoints
2. Without selection: Automatically closes single continuous subpaths by connecting their endpoints

The implementation:
- Checks that selected points are endpoints before connecting
- Handles points in different layers by copying to topmost layer 
- Only connects valid endpoints (points connected to single segment)
- Creates straight line segments without handles


Keavon:
Please see the Pen tool for the behavior and code paths that occur when the endpoints of two layers are clicked on to draw a segment between the two. You'll notice how the two layers get joined into one. Check the graph to see the before and after of the node setup. Try reusing that code for when joining points across layers, which currently doesn't work (it just draws a line that happens to go to the location of the other one, on one of the pair of layers).

Keavon (approved):

Sidharth-Singh10:
Got it! I’ll handle this in a follow-up PR

Keavon:
This also relates to some of the same concepts in https://github.com/GraphiteEditor/Graphite/issues/2184, particularly the video mentioned in part 4 at the very end of the issue.

Sidharth-Singh10:
Will look into it!!

PR #2229, opened 2025-01-27
Polish the G/R/S feature behavior, visualizations, and hints

mTvare6:
- [x] Another change: when typing values, we shouldn't round the number to 2 decimal places, instead it should show the exact number the user types, including trailing zeros if the user types those.
- [x] For Scale, the solid vs. dashed line that follows the cursor is following the projected location of the cursor onto that line. It should not be doing that, instead it should be displaying the actual scale ratio. 0 means at the center. 1 means at the starting point of the Sscale. 0.5 means half is solid, half is dashed. Notice in my video below, I'm able to make the solid line's end go all the way to the center, which should occur at a scale factor of 0, without ever having the scale factor even go below 0.5. Because currently the two values aren't related. Furthermore, this is necessary to visualize snapping when Ctrl is pressed and to properly handle slowing when Shift is pressed. [Video](https://files.keavon.com/-/QuirkyYummyGrayfox/capture_38_.mp4)
- [x] Implement integer-value snapping for Grab when Ctrl is pressed? It should work whether we're freely moving or constrained to X or Y (but still mouse controlled; Ctrl doesn't affect typed numbers). So when Ctrl is held, we shouldn't see decimal values. You'll need to zoom in past 100% to see decimal values.
- [x] We don't enter G, R, or S mode until after pressing one of those keys and then moving the mouse cursor. It should happen immediately upon pressing the key, before needing to move the cursor.
- [x] The hints need to be updated to work dynamically with the state of global vs. local X and Y constraint, and include `-` negation, numbers, and Backspace if numbers are typed.
- [x] In the Path tool, we actually do want the Rotation to start from the direction of the cursor rather than the local rotation of the layer. So if you can undo that change specifically for the Path tool, but keep it for the Select tool, that would be ideal.
- [x] When G is constrained to Y in both Path and Select tools, the number label shouldn't swap sides based on moving the mouse left and right offset by the pivot-to-start-point X offset distance. Alternatively, you could keep this behavior but make it swap to the side that the mouse is currently on (removing the pivot-to-start-point X offset from the current mouse position). When solving this, just be sure you don't break the correct behavior for G when unconstrained to an axis, since we like how the Y component label swaps sides. Please also double-check that you're not drawing any extra (overlapping) overlay lines than the necessary 1 when in X or Y constraint mode, or when X or Y happen to be precisely 0. [Video](https://files.keavon.com/-/EachWeeArcticseal/capture_39_.mp4)
- [x] Remove the bottom-left viewport overlay labels for the G/R/S status
- [x] If you type `-` it negates the mouse-controlled GRS. Then if you type a number, it becomes that negative number. If you then type `-` it negates that number, etc.

### DIFF
--- a/editor/src/messages/input_mapper/input_mappings.rs
+++ b/editor/src/messages/input_mapper/input_mappings.rs
@@ -93,7 +93,7 @@ pub fn input_mappings() -> Mapping {
 		entry!(KeyDown(Minus); action_dispatch=TransformLayerMessage::TypeNegate),
 		entry!(KeyDown(Comma); action_dispatch=TransformLayerMessage::TypeDecimalPoint),
 		entry!(KeyDown(Period); action_dispatch=TransformLayerMessage::TypeDecimalPoint),
-		entry!(PointerMove; refresh_keys=[Control, Shift], action_dispatch=TransformLayerMessage::PointerMove { slow_key: Shift, snap_key: Control }),
+		entry!(PointerMove; refresh_keys=[Control, Shift], action_dispatch=TransformLayerMessage::PointerMove { slow_key: Shift, increments_key: Control }),
 		//
 		// SelectToolMessage
 		entry!(PointerMove; refresh_keys=[Control, Alt, Shift], action_dispatch=SelectToolMessage::PointerMove(SelectToolPointerKeys { axis_align: Shift, snap_angle: Control, center: Alt, duplicate: Alt })),
--- a/editor/src/messages/input_mapper/utility_types/input_keyboard.rs
+++ b/editor/src/messages/input_mapper/utility_types/input_keyboard.rs
@@ -295,8 +295,7 @@ impl fmt::Display for Key {
 			Self::MouseMiddle => "MMB",
 			Self::MouseBack => "Mouse Back",
 			Self::MouseForward => "Mouse Fwd",
-
-			Self::NumKeys => "0-9",
+			Self::NumKeys => "0–9",
 
 			_ => key_name.as_str(),
 		};
@@ -319,20 +318,6 @@ struct LayoutKey {
 	key: String,
 	label: String,
 }
-/*
-impl Serialize for Key {
-	fn serialize<S: serde::Serializer>(&self, serializer: S) -> Result<S::Ok, S::Error> {
-		let key = format!("{:?}", self.0);
-		let label = self.0.to_string();
-
-		assert_eq!(serde_json::to_string(Key::KeyEscape), {"key": KeyEscape, "label": "Esc"});
-
-		let mut state = serializer.serialize_struct("KeyWithLabel", 2)?;
-		state.serialize_field("key", &key)?;
-		state.serialize_field("label", &label)?;
-		state.end()
-	}
-}*/
 
 pub const NUMBER_OF_KEYS: usize = Key::NumKeys as usize;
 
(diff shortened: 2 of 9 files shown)


Keavon:
!build

Keavon:
- `S`cale when controlled by mouse (not keyboard numbers):
  - Going to the opposite side should display a `-` sign in front of the number label.
  - When typing `-` to swap sides, the current mouse mapping is reversed from what's desired. Getting closer to the center should make it shrink, not grow.
  - I've decided that we want to control scaling by its radial value from the center to the mouse position (and negated across the reflection line from the cursor start point) like we have it implemented now for unconstrained scaling, but when `X` or `Y` constrained scaling is used, it should instead use linear instead of radial along that axis. The result is that the scaling visualization, where the dashed line meets the solid line, should be at the point projected from the mouse position. After testing, I think this produces the more intuitive behavior when axis-constrained.
- `G`rab when `Y`-axis constrained (or not constrained but with zero X offset): we want the number label to be on the right of the vertical visualization line, not centered within it.
- Let's actually add `-` inverting support for `G`rab as well. Doing this results in the X and Y offsets getting negated, but the mouse position stays correctly mapped but just with an offset (the same kind of offset that can occur from `Shift` slowing). This way using `-` to negate the mouse-controlled motion is consistent for all three modes. Add a hint for `-` during each mouse-controlled state
- When grabbing, rotating, or scaling: hints should include the other two of G, R, and S, so two of these three (both the keys and slash-delimited text labels):
  capture
- When `Y` constrained, the hints should still show `X` before `Y`. Also, "Unconstraint" should say "Clear Constraint". When in non-local mode (a single `X` or `Y` tap), a hint should also say "`X` `X` Clear Constraint" or "`Y` `Y` Clear Constraint." That's two `X` keys similar to the screenshot above that shows "`G` `R` `S`" as three keys in one hint.
- Make sure your hint labels are `Title Case`, not `Sentence case`, please. But I'll probably tweak the exact wording after the final changes and before I merge this PR.
- The `NumKeys` hint should be labeled `0-9` instead, and those should show up whenever you can *start* typing numbers.
- Backspace still doesn't correctly remove the typed `-` sign, either for the reversed mouse-controlled `S`/`R` mode or for ensuring the next-typed number for `G`/R`/`S` ends up not being negative.
- The visualization overlays don't correctly handle panning (e.g. with the mouse scroll wheel) or zooming.

That's everything I've noticed so far, thank you!

Keavon:
!build

Keavon:
- In `G`, we can type a number and it applies to the X axis. That shouldn't happen unless an axis was first chosen.
- We need the `0-9` hint showing at any time you can begin typing a number (so that's always except in unconstrained `G`). Likewise for a `-` hint to show you can negate the value.
- I'd like to try, once more, seeing what it looks like for the uniform (non-axis-constrained) scaling to follow the cursor rather than sticking to the original angle when the scaling began. This should be a separate commit so it's easy to revert if the current behavior remains my preference.
- A one-frame visual bug: if you `S`, `Y`, move the mouse so the scale becomes negative, then `Y` to turn it to a local scale; for one frame it becomes un-inverted:  
  
  https://github.com/user-attachments/assets/d23d0cdf-9f2d-4425-8df9-aea882c830b7

- In `R` mode, there should be no separator line at the very end of the hints bar, when nothing comes after it.
- All G/R/S modes and states need to have a separate section at the start for left click or <kbd>Enter</kbd> to confirm and right click or <kbd>Esc</kbd> to cancel (copy latter from the hints when drawing a line, and adapt it for the former case).
- When resizing or rotating the transform cage, we'd like hints for cancel, <kbd>Shift</kbd> constrain aspect ratio (scaling the layer) or along axis (dragging the layer), <kbd>Alt</kbd> around the pivot (scaling the layer), and <kbd>Ctrl</kbd> snapping 15° increments (rotating the layer).

Keavon (approved):

Issue #1834, opened 2024-07-15
Tracking Issue: Instances tables


Keavon:
- [x] 1. Write the `Instances<T>` struct. Wrap `VectorData` to become `Instances<VectorData>`, `ImageFrame` to become `Instances<ImageFrame>`, `GraphicGroup` to become `Instances<GraphicGroup>`, which lets us leave `transform`/`alpha_blending` alone for now. Places that read the single piece of data now read from a "getter" function that calls `.next().expect("...")` and write to a "setter" function that loops through each row (but there's just 1 presently).
  (#2230)
- [x] 2. Move the `transform`/`alpha_blending` up a level, which removes those fields from `VectorData`, `ImageFrame`, and `GraphicGroup`.
  (#2249)
- [x] 3. Since the `ImageFrame<P>` struct now has only one field left (`pub image: Image<P>`), delete this redundant struct and replace its usage in `Instances<ImageFrame<P>>` with `Instances<Image<P>>`.
  (#2256)
- [x] 4. Wrap `ArtboardGroup` to become `Instances<ArtboardGroup>`. Since the `ArtboardGroup` struct now has only one field (`artboards: Vec<(Artboard, Option<NodeId>)>`), delete this redundant struct and replace its usage with `Instances<Artboard>`. Move the `Option<NodeId>` to be an attribute called `source_node_id` in `Instances<T>`.
(#2265)
- [x] 5. Since the `GraphicGroup` struct now has only one field left (`elements: Vec<(GraphicElement, Option<NodeId>)>`), delete this redundant struct and replace its usage in `Instances<GraphicGroup>` with `Instances<GraphicElement>`. Move the `Option<NodeId>` to the `source_node_id` attribute in `Instances<T>`. At this point, we have multi-row tables for group data but single-row image and vector data tables.
  (#2363)
[...]

PR #2230, opened 2025-01-27
Instance tables refactor part 1: wrap graphical data in the new Instances<T> struct

Keavon:
Part 1 of #1834


node-graph/gcore/src/graphic_element.rs
```diff
@@ -38,6 +41,67 @@ impl AlphaBlending {
...
+		for instance in &self.instances {
+			let instance = instance.lock().unwrap();
+			instance.hash(state);
+		}
+	}
+}
+
+impl<T: Into<GraphicElement> + PartialEq + StaticType + 'static> PartialEq for Instances<T> {
+	fn eq(&self, other: &Self) -> bool {
+		self.id == other.id && self.instances.len() == other.instances.len() && { self.instances.iter().zip(other.instances.iter()).all(|(a, b)| *a.lock().unwrap() == *b.lock().unwrap()) }
```
0HyperCube, line 97: I'm pretty sure it will deadlock if you compare an array to a clone of itself. This is because the same mutexes will be in self and other. Therefore when we try and get a lock on both a and b at the same time which can never happen.

Keavon:
All working now, without any known regressions.

PR #2235, opened 2025-01-29
Add lasso selection mode

### LINKED ISSUE #2161: Lasso selection with the Select tool
The Select tool currently lets you click-and-drag to box select a region of shapes within the rectangle. If <kbd>Shift</kbd> is held down before dragging, this lets the box selection *add* to the selection instead of *replacing* the existing selection.

Sometimes, users want more precise control over the region being selected than a rectangle can accommodate. For that, we'd like to have lasso selection. It should be triggered if <kbd>Ctrl</kbd> is held down before dragging. Unlike box selection described above where <kbd>Shift</kbd> has to be held down to *add* instead of *replacing* the selection, lasso selection will always *add* to the selection. Another difference from box selection is that the lasso must fully encompass the layer, rather than just barely touching it (as with box selection), to be included in the selection. Here's how it should work (a video mockup made using Blender):

https://github.com/user-attachments/assets/a5596190-1711-4e18-857e-00da14195419

The overlay system for drawing the blue lines, which is used for drawing the box selection rectangle, will also need to be used to draw the polyline of each mouse coordinate visited during the lasso selection drag. I don't think we currently have an existing function in the overlays system for drawing a polyline but it should be easy to adapt an existing function for that purpose.

indierusty:
Closes #2161


Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon (approved):

Issue #1128, opened 2023-04-15
G/R/S visualization overlays


Keavon:
G/R/S, used with both the Select and Path tools, should display an X or Y line when grabbing or scaling in the X or Y axis. (The line should be aligned with the common center, i.e. the pivot used for scaling or rotating.) It should go from one end of the viewport to the other. It should also display a dot around the pivot while rotating or scaling. Reference how other overlays are drawn (like those visualizing snapping relationships) for how to draw overlay shapes like lines and dots.

Also, text should be drawn next to those lines visualizing the numerical offsets. Currently these values are shown in the hints bar, which is a temporary workaround until this is implemented.

Also: a line should be drawn when dragging with the Select tool while holding <kbd>Shift</kbd> to constrain it to the X or Y axis. (That behavior already exists, but the line should be added to visualize it.)

- [x] #2195
- [x] #2229
- [x] #2243
- [x] #2285

mTvare6:
Commenting so you can assign.

PR #2243, opened 2025-01-31
Further polishing of G/R/S visualization and features

mTvare6:
Followup to [this feedback](https://github.com/GraphiteEditor/Graphite/pull/2229#issuecomment-2625546292) in #2229, which gets closer to closee #1128.

- [x] In `G`, we can type a number and it applies to the X axis. That shouldn't happen unless an axis was first chosen.
- [x] We need the `0-9` hint showing at any time you can begin typing a number (so that's always except in unconstrained `G`). Likewise for a `-` hint to show you can negate the value.
- [x] A one-frame visual bug: if you `S`, `Y`, move the mouse so the scale becomes negative, then `Y` to turn it to a local scale; for one frame it becomes un-inverted:  
- [x] In `R` mode, there should be no separator line at the very end of the hints bar, when nothing comes after it.
- [x] All G/R/S modes and states need to have a separate section at the start for left click or <kbd>Enter</kbd> to confirm and right click or <kbd>Esc</kbd> to cancel (copy latter from the hints when drawing a line, and adapt it for the former case).
- Removed: I'd like to try, once more, seeing what it looks like for the uniform (non-axis-constrained) scaling to follow the cursor rather than sticking to the original angle when the scaling began. This should be a separate commit so it's easy to revert if the current behavior remains my preference.
- [x] When resizing or rotating the transform cage, we'd like hints for cancel, <kbd>Shift</kbd> constrain aspect ratio (scaling the layer) or along axis (dragging the layer), <kbd>Alt</kbd> around the pivot (scaling the layer), and <kbd>Ctrl</kbd> snapping 15° increments (rotating the layer). *(This also fixes <https://discord.com/channels/731730685944922173/881073965047636018/939265895509925898>.)*


Keavon:
- Let's actually get rid of `S`cale's radial behavior (which Blender uses) and make it instead work like `X` and `Y` do where they are linear instead of radial, but using the starting-point-to-pivot line (which is visualized by the overlays) as that line to travel along to adjust the scale. This should help make it clearer, and it removes the discontinuity where the number flips, and also ensures that the solid-to-dashed line transition point is a projection of the current cursor location (unless an offset occurred via Shift, panning, zooming, etc.) just like it is for `X` and `Y`.
- For `G` (but not `S`, which currently works as intended), if the user is in typing mode with `X` or `Y`, then hits `X` or `Y` again to go into local mode, then hits it again to exit the typed number constraint, it should return to being movable with the mouse rather than an `X` constraint (but where the hints don't indicate that). But when returning to this mouse mode, the typed `-` state should be taken into account, and if the user goes back into `X` or `Y` mode, it should return to the previously typed value rather than forcing the user to type those numbers and `-` state again.
- If a shape has been scaled along one axis by -1 to form its mirror image, `G` with a local constraint acts in reverse which is unexpected.

Keavon (approved):

PR #2248, opened 2025-02-01
Make the Select tool avoid updating hints just when clicking but not dragging

singhutsav5502:
Solves https://discord.com/channels/731730685944922173/881073965047636018/1334123140842000386

Make Fsm `update_hints` function borrow `tool_data` and update individual tool implementations accordingly, use `tool_data` in select_tool `update_hints` function as well as add a `UpdateHints` response when the pointer moves after the change from  DragStart to Dragging state.


singhutsav5502:
Hi, @Keavon  please review.

Keavon:
Will try to do within the next few days. Working to catch up on reviews.

Keavon (approved):

PR #2250, opened 2025-02-02
Fix some number input widgets becoming selected after dragging left/right in Firefox

### LINKED ISSUE #2215: In Firefox, tool control bar number input field dragging causes text entry afterward
This only happens in Firefox which likely means it is a result of event ordering. Events that happen at the same time (such as `click` and `mouseup` and `pointerup`) don't get fired in the same order in FF as in Chrome. This has caused problems in the past. The solution is tracing the logic and then updating it to be more robust to either ordering. This is a sizable chance that some of the debugging will take place in the Rust code, not just `NumberInput.svelte` and other frontend code.

For only the tool control bar (the left side of the control bar, above the viewport), if you drag left/right on number input boxes (such as "Diameter" for the Brush tool, "Weight" for any of the vector drawing tools, also the Polygon and Text tool's various number inputs, but apparently **not** the "X"/"Y" point position of the Path tool). It doesn't seem to happen for any other number inputs throughout the editor from what I could find, even those on the right side of the control bar.

Image

prathmesh703:
Closes #2215 

https://github.com/user-attachments/assets/828c905a-23ce-466c-9ee8-4e47ae9e6dbf

### DIFF
--- a/frontend/src/components/widgets/inputs/NumberInput.svelte
+++ b/frontend/src/components/widgets/inputs/NumberInput.svelte
@@ -57,10 +57,9 @@
 	export let incrementCallbackDecrease: (() => void) | undefined = undefined;
 
 	let self: FieldInput | undefined;
-	let isdragging: boolean;
-	isdragging = false;
 	let inputRangeElement: HTMLInputElement | undefined;
 	let text = displayText(value, unit);
+	let isDragging = false;
 	let editing = false;
 	// Stays in sync with a binding to the actual input range slider element.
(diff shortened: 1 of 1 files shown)


Keavon:
Within the next few days. I'm working to get caught up on reviews.

Keavon (approved):

Keavon:
Can you please leave a comment in #2215 that you fixed it in this PR number? I need it for tracking purposes so I can make you the assignee.

Keavon:
Sorry, disregard, you did already leave a comment in that issue.

PR #2251, opened 2025-02-02
Add support for skewing the transform cage by Ctrl-dragging its edges

0HyperCube:
As suggested by Keavon [on discord](https://discord.com/channels/731730685944922173/1134981447539761273/1335542641500360734).

I haven't added a hint for this since I'm not sure where it should go.


Keavon:
Looks great. Could you please also make it so holding Ctrl when hovering over the edge, or when actively skew dragging, it changes the cursor from horizontal to vertical; and vertical to horizontal? That would hint to the user that, because Ctrl is currently pressed, skew will occur instead of regular scaling.

Keavon:
I'll merge without that requested improvement and include it as an improvement for new contributors, in the interest of getting this PR merged sooner rather than later.

Keavon (approved):

PR #2249, opened 2025-02-02
Instance tables refactor part 2: move the transform and alpha_blending fields up a level

Keavon:
Part 2 of #1834

Currently based on commit 252de1b33cc897ca4bd6aa5a69a88e5aef6993ff from #2232. The first commit in the chain of this PR's changes is 2fc5fb34dc1c349d842c9d9636038c6774e9e038. This PR is blocked on that PR being merged first.

PR #2256, opened 2025-02-03
Instance tables refactor part 3: flatten ImageFrame<P> in lieu of Image<P>

Keavon:
Part 3 of #1834.

Flattens the one-field `ImageFrame<P>` struct, removing it to put `Image<P>` in its place.

This PR is based on top of #2249. The first commit in the chain of this PR's changes is b36f57fe8a4b16f7ccd182ef59a0601c9e6c297b. This PR is blocked on that PR being merged first.


Keavon:
Merging with one known issue:

capture

This occurs while upgrading a previous document that contains Brush tool strokes. It doesn't seem to actually affect things, except for the fact that it alters the bounding box of the brush strokes to extend to the origin. After saving and reloading the upgraded document, that remains true but the console error doesn't return. Since brushes are not fully supported and usage is likely to be rare, I'm not going to investigate and try to fix this.

PR #2264, opened 2025-02-04
Add Pen and Path tool modes to avoid showing all handles

### LINKED ISSUE #2261: Reduce path editing clutter with options to hide non-selected anchors' handles
There should be three modes, explained below. But in summary:
- The first mode is the current behavior where all handles are shown.
- The second mode is where all selected anchors' segments' handles are shown.
- The third mode is where only the frontier handles are shown.

## Pen tool

In the Pen tool, only these circled handles should appear:
 
- Only the last-drawn anchor and its handles are shown, plus the handle on the other end of its connected segment. The Pen tool only has modes 1 and 2, but not 3.
  
  Image

## Path tool

In the Path tool, which allows multi-selection of anchors, these are the cases where it should appear:
  
- When one anchor is selected, its handles should be shown, as well as the handles of its connected segment(s). When one anchor is selected, this same situation applies to both modes 2 and 3, because mode 3 only becomes active with multi-selection (otherwise mode 2 is used).
  
  Image

- **Mode 2:** When two or more anchors are selected, all handles (circled in red) on either end of the adjacent segments (purple) are shown.
  
  Image

- **Mode 3:** When two or more anchors are selected, only the handles on the **frontier** are shown. The green shows the adjacent segments to the selected anchors, and the ends of the green are the frontiers. Selected in red are the only handles to be shown.
  
  Image

Remember: only the handles in the diagrams above which are circled in red should appear, all other handles should be hidden.

The three modes (two for Pen tool) should be in the respective tools' control bars as a radio button. just use "1", "2", and "3" text for now until I come up with names or icons. The control bar is the area on the left above the drawing viewport where the tool controls are located.

4adex:
The objective was to provide 3 (2 in case of pen tool) modes for the viewer that changes how the handles are shown of the selected anchors. The specific requirements and intended behaviour is well written in the issue which this closes #2261.
The approach is summarised below - 

- Added `overlay_mode_type` in options of both tools which can be modified by the radio buttons in the tool options layout.
- For each tool they are enum which control the behaviour in `Overlays` tool message.
- `path_overlays` is made more modular by giving another argument, which specifies whether all handles, specific handles, frontier handles, or no handles to be drawn.
- `get_selected_segments` is for getting the vector of segmentIds of all segments which are associated with selected elements, which is used specifically for implementing mode 2 and 3.
- For mode 3, the core logic is finding those points which are only attached to one segment in all of the relevant segments, this approach also takes care of base cases like when there is a loop, etc.
- `overlay_bezier_handles` utility function is created to avoid repetition of code.


Keavon:
!build

4adex:
Changes after first review - 
- While in mode 2 or mode 3, when the new selection (on mouse down) has only anchors, then previous selection of points is stored in `saved_points_before_handle_drag` in `tool_data`
- In mode 3, if only one point is selected then behaviour is of mode 2.
- While dragging a handle you can also see the opposite handle.

Keavon:
!build

Keavon:
!build

Keavon:
Here are four separate frontiers:

capture

But the current implementation only treats 1 and 4 as frontiers, which means we don't see handles on either end of frontiers 2 and 3. We should see both handles around that unselected point, as a result.

Keavon (approved):

Keavon:
Another issue: in the Pen tool, when not drawing, all handles are shown regardless of mode.

Keavon:
We made it so when dragging a handle, once done, it returns the selection to the previous state— that works when you release the drag, but not when you abort the drag by right clicking or hitting escape.

PR #2269, opened 2025-02-06
Implement initial infrastructure for the Spline tool to join itself with other splines and paths

Sidharth-Singh10:
Closes part_4 of #2184


Keavon:
!build

Keavon:
Seems like this breaks if I try to connect a spline to a line (drawn with the Line tool). There are two modes for failure:

1. Start with a line, then draw the spline and end with connecting to an endpoint of the line. They don't connect together.
2. Start with a line and a spline, then use the Spline tool to extend the spline into the line. The line disappears to the top left of the artboard with a straight line connector going to that origin corner.

Keavon:
!build

Keavon:
Merging with known issue in my case `#2` described above being more broken than before (invalidating the graph).

Keavon (approved):

PR #2294, opened 2025-02-15
Add drag and release to menu lists

### LINKED ISSUE #2280: Menu lists should allow dragging from their spawner
Currently, menu bar entries require clicking the menu bar button to open it, then clicking to pick a row inside. Menu bar entries should also support opening on mousedown and letting users drag down and then release on an entry. Additionally, right click and <kbd>Esc</kbd> should allow aborting when the left mouse is held down (either for a drag or just a click), such that the menu closes without being selected. This applies for all menu lists, such as dropdown menus as well.

Stargazer10101:
Closes #2280

Implemented events: mousedown(Drag), mouseup(select), right click and esc button(abort).

### DIFF
--- a/frontend/src/components/floating-menus/MenuList.svelte
+++ b/frontend/src/components/floating-menus/MenuList.svelte
@@ -297,6 +297,9 @@
 			// Keep the menu stack open
 			return false;
 		}
+		//Close menu with right click
+		
+		
 
 		// Click on a highlighted entry with the enter key
 		if (menuOpen && highlighted && e.key === "Enter") {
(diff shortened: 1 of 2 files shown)


Keavon:
!build

Keavon:
This breaks clicking to close. It also doesn't work on all menu lists, just the menu bar items.

Keavon:
> Can you please elaborate what you meant by "clicking to close". You may mark this as draft while I make changes to make it work on all menu lists.

Unfortunately I can't spend the time elaborating such precise details in places where contributor discretion is required. It's not too much trouble in one case, but it becomes excessively burdensome in the aggregate. I trust that if you're able to solve the coding problem, you can also figure out what is intended based on how regular software works. To some extent, this is a way to filter out inexperience. Thanks for your understanding.

PR #2295, opened 2025-02-16
Make the Pen tool extend an endpoint by starting with a colinear, equidistant handle

4adex:
Part of #1870 

- clicking a path's endpoint should begin drawing from it with an initial handle that's colinear and equidistant from the opposite handle, instead of the current behavior where clicking the endpoint begins drawing the segment with a zero-length handle. If the user wants the current behavior, they can simply click the endpoint anchor a second time to make the current segment's handle zero-length. This solves the pain point that switching away from the Pen tool, then back to it, causes a discontinuity at the previous endpoint where the handles can't easily be opposite each other like when drawing consecutive segments normally.

### DIFF
--- a/editor/src/messages/tool/tool_messages/pen_tool.rs
+++ b/editor/src/messages/tool/tool_messages/pen_tool.rs
@@ -624,52 +624,9 @@ impl PenToolData {
 		self.handle_end = None;
 
 		let tolerance = crate::consts::SNAP_POINT_TOLERANCE;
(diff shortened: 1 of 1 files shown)


Keavon (approved):

Keavon:
One change: it should make the handle colinear, which means hitting G or R should also move the opposite handle.

Keavon:
!build

PR #2300, opened 2025-02-18
Add draggable skew triangles to the transform cage

### LINKED ISSUE #2299: Transform cage skew sliding triangle gizmos
Image

As of #2251, you can currently skew by holding Ctrl when dragging an edge of the transform cage. I'd like to add these triangles that appear when hovering over an edge's resize click target. Only the edge for which you're hovering should show the pair of triangles. Triangles are the same dimensions as the squares (7 x 7 px), but with a vertex at `7 / 2 = 3.5` instead of the square's corners. They are located 4 px away from the square. After adding this, we can remove the `Ctrl` shortcut which isn't needed anymore since this is easier and more discoverable. However, when dragging a triangle, `Ctrl` can be assigned to allow the dragging of the entire edge instead of just sliding it. Meaning it's allowed to follow the cursor along both the local X and local Y axis, not just sliding along the local X or Y axis (whichever edge is being dragged). If `Ctrl` is released while dragging, it should return to its previous sliding-only position, as if the user had never held `Ctrl` but merely moved the cursor around, only sliding it along the local axis. The hints, during the skew sliding, should indicate that `Ctrl` unlocks the slide.

mTvare6:
Fixes #2299


Keavon:
!build

Keavon:
!build

Keavon:
!build

0HyperCube:
Just to note that I find it very hard to trigger the skew with the new regime due to the very low tolerance. This makes the feature inaccessible to those with poor motor skills. Since it is not possible to skew in any other way, editor functionality becomes inaccessible.

Furthermore the arrows are rather subtle and probably quite easy to miss, especially considering that the cursor will be over them.

Keavon:
!build

Keavon:
!build

Keavon (approved):

PR #2306, opened 2025-02-18
Transfer pasted layers to the viewport center when outside the current view area

### LINKED ISSUE #2301: Pasting should be transfered in view if its placement wouldn't be in view
Pasting a layer makes it go where it was copied from. If the viewport isn't looking at that location, and the pasted object(s) would be **fully** outside the view, we want to transform it/them (i.e. its/their bounding box) to the center of the viewport, by an integer number of pixels offset.

PRs with useful diffs to reference: #2212, #2177

mTvare6:
Closes #2301


Keavon:
This is a draft, what is it waiting on?

Keavon:
!build

Keavon:
When this gets pasted, it ends up not being centered, instead it goes to the bottom right.

capture

(The group has no transform applied to it.)

Keavon (approved):

PR #2307, opened 2025-02-19
New node: Merge by Distance

richard-uk1:
Here's a patch that adds a new node to collapse segments with chord shorter than a distance (parameter to the node). It's not finished but this version works - I tried it out by drawing freehand then watching the segments collapse as the distance increases. 

Things still to do:
 - [x] Also move bezier control points (I left this until after the first version because it is fairly orthogonal to finding and collapsing the short segments)
 - [x] Updating faces if the end segments were removed
 - [x] General tidy up (I've used `petgraph` for the clustering algo but unless you want to take the dep, I assume I'll need to figure it out myself and replace it with hand-written code.)
   - [x] If you did want to take the dep, I'd need to look at minimal features to get what is needed, to avoid bloating build time
 
I do think you should consider using a graph library though, because the relationship between points, segments and faces is inherantly graph-like, and one might as well take advantage of all the experties that's been developed for graphs.

**EDIT** I measured binary size before/after this PR:
 - with `master` size is 15.93M
 - with PR, size is 15.98M


node-graph/gcore/src/vector/vector_data/attributes.rs
```diff
@@ -149,6 +152,15 @@ impl PointDomain {
 		self.positions[index] = position;
 	}
 
+	pub fn set_position_by_id(&mut self, id: PointId, position: DVec2) {
```
richard-uk1, line 155: TODO this is dead and will be removed

node-graph/gcore/src/vector/vector_data/attributes.rs
```diff
@@ -504,6 +542,12 @@ impl RegionDomain {
 	}
 }
 
+pub struct Region {
```
richard-uk1, line 545: TODO also remove this dead code

node-graph/gcore/src/vector/vector_data/indexed.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use bezier_rs::BezierHandles;
+use glam::DVec2;
+use petgraph::graph::{EdgeIndex, NodeIndex, UnGraph};
+use rustc_hash::FxHashMap;
+
+use super::{PointId, SegmentId, VectorData};
+
+pub struct IndexedVectorData<'a> {
+	/// This structure owns the data to allow convenient modification + storage
```
richard-uk1, line 11: Comment out of date

node-graph/gcore/src/vector/vector_data/indexed.rs
```diff
@@ -0,0 +1,146 @@
...
+	/// Can work on owned data or a reference
+	pub(crate) data: Cow<'a, VectorData>,
+	/// Points and segments form a graph. Store it here in a form amenable to graph algos
+	pub(crate) point_graph: UnGraph<Point, Segment>,
```
richard-uk1, line 15: Not all fields are used here. They would be useful in other contexts - TODO consider whether to delete or not.

node-graph/gcore/src/vector/vector_nodes.rs
```diff
@@ -1220,6 +1220,29 @@ async fn bevel<F: 'n + Send + Copy>(
...
+		(),
+		Footprint,
+	)]
+	footprint: F,
+	#[implementations(
+		() -> VectorDataTable,
+		Footprint -> VectorDataTable,
+	)]
+	source: impl Node<F, Output = VectorDataTable>,
+	#[default(10.)] distance: Length,
```
richard-uk1, line 1235: Perhaps a different default would be better.

Keavon:
For the 0.05 MB I think we can accept the dependency. There's a chance we use it in the future for other graph data structures, too.

richard-uk1:
It's probably overkill for just this 1 problem, but I can see it being useful in a number of situations. It might even be the most efficient way to store path info.

Keavon:
!build

richard-uk1:
I'll get back on this.

Keavon:
Just a reminder since it's the weekend and you might have some time to fix this up for the merge.

node-graph/gcore/src/vector/merge_by_distance.rs
```diff
@@ -0,0 +1,109 @@
...
+
+				short_edges.add_node(start);
+				short_edges.add_node(end);
+				short_edges.add_edge(start, end, seg_id);
+			}
+		}
+
+		// Now group connected segments - all will be collapsed to a single point.
+		// Note: there are a few algos for this - perhaps test empirically to find fastest
+		let collapse: Vec<FxHashSet<PointId>> = petgraph::algo::tarjan_scc(&short_edges).into_iter().map(|connected| connected.into_iter().collect()).collect();
```
TrueDoctor, line 35: Tarjan will almost certainly be faster than Kosaraju because it only requires a single graph traversal

Keavon:
Hi, just one more friendly reminder, it'd be awesome to get this landed. Thanks for your work so far on it!

richard-uk1:
OK @Keavon ready for review.

Keavon:
!build

Keavon:
Thank you for making this! I'm merging this now, although I have found one behavior issue that I hope you'll be able to fix in a followup PR. We want the algorithm to only allow a collapse to a set of points if all of them are within `distance` of each other. Currently, it's possible to extend an unlimited run of points so long as each one is within `distance` of its adjacent point, even though all of them collectively extend very far.

https://files.keavon.com/-/TrueSuperbBlackfootedferret/capture_31_.mp4

node-graph/gcore/src/vector/vector_data/indexed.rs
```diff
@@ -0,0 +1,90 @@
...
+/// Important: It is the user's responsibility to ensure the indexes remain valid after mutations to the data.
+pub struct VectorDataIndex {
+	/// Points and segments form a graph. Store it here in a form amenable to graph algorithms.
+	///
+	/// Currently, segment data is not stored as it is not used, but it could easily be added.
+	pub(crate) point_graph: UnGraph<Point, ()>,
+	pub(crate) segment_to_edge: FxHashMap<SegmentId, EdgeIndex>,
+	/// Get the offset from the point ID.
+	pub(crate) point_to_offset: FxHashMap<PointId, usize>,
+	// TODO: faces
```
Keavon, line 23: Is this still needed or is this comment a leftover?

Keavon (approved):

node-graph/gcore/src/vector/vector_data/indexed.rs
richard-uk1: This is leftover.

Keavon:
Hi @richard-uk1, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

PR #2329, opened 2025-02-28
Add the Select Parent command

### LINKED ISSUE #2326: Add a "Select Parent" command
It should go under *Layer* > *Deselect All* (it can skip having an icon for now). For every selected layer, it should move the selection up a level in the hierarchy. Below are a couple before/after examples.

| Before | After |
|--------|--------|
| Image | Image |
| Image | Image |

prathmesh703:
Closes #2326 

https://github.com/user-attachments/assets/d10efad0-603b-48e8-b509-12cd15f9cbdf


Keavon (approved):

Keavon (approved):

Keavon:
Thanks! Would you like to move on to a "Select Siblings" command next? We'll also want an "Invert" command but it should probably have several variants that I need to think more about and decide on a plan for.

prathmesh703:
> Thanks! Would you like to move on to a "Select Siblings" command next? We'll also want an "Invert" command but it should probably have several variants that I need to think more about and decide on a plan for.

@Keavon yes

PR #2331, opened 2025-03-01
Add a workaround to prevent nudge resizing from giving lines a NaN scale

indierusty:
Fixes https://discord.com/channels/731730685944922173/881073965047636018/1344535507635404902

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -702,6 +702,8 @@ impl MessageHandler<DocumentMessage, DocumentMessageData<'_>> for DocumentMessag
 				};
 
 				let size = existing_bottom_right - existing_top_left;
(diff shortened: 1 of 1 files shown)


Keavon:
!build

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -702,6 +702,8 @@ impl MessageHandler<DocumentMessage, DocumentMessageData<'_>> for DocumentMessag
 				};
 
 				let size = existing_bottom_right - existing_top_left;
+				// If size is zero we clamp it to minimun value to avoid dividing by zero vector to calculate enlargement.
```
Keavon, line 705:
```suggestion
				// TODO: This is a hacky band-aid. It still results in the shape becoming zero-sized. Properly fix this using the correct math.
				// If size is zero we clamp it to minimun value to avoid dividing by zero vector to calculate enlargement.
```

Keavon (approved):

PR #2339, opened 2025-03-02
Disable the Path tool's "Colinear Handles" checkbox when no interior anchors are selected

indierusty:
Resolves https://discord.com/channels/731730685944922173/881073965047636018/1345303020136828958


Keavon:
!build

Keavon:
Can you please make it so, when multi-selecting both inner points and end point(s), it filters out the end points and displays the checkmark status of just the inner points? And then if clicking the checkbox to toggle those inner points, it ignores the endpoints but still correctly toggles it for the rest, updating the checkbox status as well.

Keavon:
!build

Keavon (approved):

Keavon (approved):

Keavon:
@indierusty would you like to work on the related feature to this one, from #1870?

> When double-clicking an anchor:
> - Currently this swaps between smooth and sharp, but doesn't work if multiple handles are multi-selected to convert them all to smooth or sharp. If they're of mixed type, the specific one that's double clicked on should be the decider for the rest of them as to which they should all be swapped to.

PR #2363, opened 2025-03-05
Instance tables refactor part 5: unwrap GraphicGroup as multi-row Instance<GraphicElement> tables and move up transforms

Keavon:
Part 5 of #1834.

Discussion about squashing some bugs (for the record): https://discord.com/channels/731730685944922173/860830913390575646/1348437590914895893

# Issues

- [ ] All leaf nodes are incorrectly being grouped, with their transform and alpha blending being applied to that group instead of the direct leaf SVG element (deferred for after part 6 or 7)
  - [ ] Fix for Vello as well (if necessary), and test Vello in general (deferred for after part 6 or 7)
- [x] Transform nodes mess up stroke transforms
- [x] Empty vectordata: there was a default empty vectordata in all groups.
- [x] Transforming a group with the tools resets the transform to identity.
  We need to have an original transform when using tooling e.g. grs. We use `GraphOperationMessage::TransformSet` with the values stored in `OriginalTransforms`. This allows us to stop the transform being stuck at zero when the user drags the bounds through zero.
  We also need a local transform for orientating the bounding box.
- [x] What is the original transform for a group?
  If you apply a transform node to the left of a group, it actually just transforms all of the elements in the group.
  If you apply a transform node above the group, it will also transform other layers which we don't want.
  
  image


TrueDoctor:
Only very few because we can use the automated edition upgrade on this pr as well. that should resolve most of the conficts

Issue #1832, opened 2024-07-15
Tables and attributes for graphical data


Keavon:
This involves moving towards a unified conceptual model for procedural graphical data based around tables (a data type represented by a spreadsheet) and attributes (the columns of a table).

<details><summary>Design notes (contains outdated terminology and concepts)</summary>

## New instances-based concept design notes (as we designed during the Maker Faire)

> Here are the notes Adam took while he, True Doctor, and Keavon discussed the design while driving to the Maker Faire:
> 
> All data is
> ```rs
> struct Instances<T> {
>     // Datastore: HashSet<Reference, T> (this is an older idea, before choosing to use Rc<Mutex<T>>)
>     instances: Vec<(String, Vec<TaggedValue)>>
>     // First column is Rc<Mutex<T>> second column is transform, footprint, click targets, vector modify,
>     // overlays, other columns that the user adds. This forms the spreadsheet.
> ```
> 
> Vector data transforms stored in `Vec<(NodeId, Transform)>`, each monitor node adds an entry, the transform node multiples the rightmost.
> 
> No appearance data, color/thickness/properties used by renderer are stored in table.
> 
> Rectangle node creates `Instances<VectorData>`, which can be added to a group of `Instances<GraphicElementInstances>` when it enters a layer stack
> 
> ```rs
> enum GraphicElement {
>     Vector(Instances<VectorData>),
>     ImageFrame(Instances<ImageFrame>)
>     // Etc.
> }
> ```

## Old pre-instances concept design notes (prior to the Maker Faire)

**Design for the new table-based graphical data structure**

Using this syntax:
| Definition | Description|
|:-|:-|
| `Struct = { Field1, Field2, ... }` | Rust struct |
| `[Table] = [Attribute1[], Attribute2[], ...]` | Either a HashMap, or Rust struct,<br />of parallel `Vec`s shown as a spreadsheet |
| `Union = Type1 \| Type2 \| ...` | Rust enum of 1-tuple type<br />variants each named after its type | 

Elements, which all implement `trait Graphical` (renamed from the currently named `GraphicElementRendered`):
- `[Element] = [ElementId[], (dyn Graphical)[], Transform[], AlphaBlending[]]`
- `VectorData = { [Point], [Segment], [Region], ColinearManipulators, FillRule, SvgFilter, FutureMeshVectorInfo }`
- `Raster = // to be designed`
- `SDF = // to be designed`
- `Color = // to be designed`
- `Gradient = // to be designed`

[...]

PR #2368, opened 2025-03-06
Move the Spline tool to a spline/path mode option in the Pen tool

indierusty:


Keavon:
Can you please resolve these merge conflicts?

Keavon:
I'll mark this as draft for the moment. Please make it ready for review when you've finished the conflict and making it compile again. Thank you!

Keavon:
!build

Keavon:
!build

Keavon:
If you get a chance soon to resolve the conflicts, I can to a code review and prepare to merge it now that I finally have the time to put my attention towards this. Thanks and sorry for the delay on my part causing you the extra work.

indierusty:
> If you get a chance soon to resolve the conflicts, I can to a code review and prepare to merge it now that I finally have the time to put my attention towards this. Thanks and sorry for the delay on my part causing you the extra work.

No problem about the delay. I was actually sick for a few days but I'm back today and ready to work to task I'm assigned to. I'll get those conflicts resolved.

Keavon:
Right clicking to end drawing the spline results in the whole new spline layer being deleted. It looks like this was intentional because of how the hints are shown. But we'll want to make it reflect the regular (path mode) Pen tool behavior of confirming all previously drawn segments but canceling the currently-being-extended segment; and if the user is clicking and dragging to slide around the new proposed segment, in that case <kbd>Enter</kbd> should accept it where it is while <kbd>Esc</kbd>/RMB should cancel that currently-being-placed-and-sliding segment. So in summary, if the mouse is up, <kbd>Enter</kbd>/<kbd>Esc</kbd>/RMB behave equivalently, but if the mouse is down and sliding, <kbd>Esc</kbd>/RMB behave differently from <kbd>Enter</kbd>. Check the Pen tool (path mode) for how that works and look at the input hints shown for both cases.

indierusty:
I tried to read all the changes made to spline tool but I probably missed it.

indierusty:
> It looks like we lose the functionality that we currently have in master where the Pen tool can extend a Spline tool spline from its endpoint, creating the necessary node changes to support that. But now that I think about it, I think it might be better to just keep it this way for simplicity at this time. We're likely to use a more spreadsheet/attribute-oriented approach in the future for marking which anchors should be treated as a spline, that way the node graph can be considerably simplified. (Pending further progress on #1832 and #2522.)

I think we can still extend the spline created in spline mode using path mode as path tool modifies the path node that feeds the spline node.

Keavon:
Consensus on our call together: replace this PR with one that reimplements the desired functionality: a "Polyline" mode that opts out of handle-dragging behavior in the Pen tool, and a "Spline" mode that is just "Polyline" mode but with the inclusion of a Spline node in the layer that's created.

Keavon:
Closing this per the above change in approach. We can use a separate PR in the future.

PR #2388, opened 2025-03-07
Allow printing proto graph in graphite-cli

bradrn:
As in the title. I see this change as being useful for debugging purposes. Additionally, since executing the node graph isn’t always needed, this adds another option `--no-run` to prevent this.

(This was previously discussed on the Discord server.)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -935,21 +935,23 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.28"
+version = "4.5.31"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3e77c3243bd94243c03672cb5154667347c457ca271254724f9f393aee1c05ff"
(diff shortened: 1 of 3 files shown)


TrueDoctor:
Could you make this a subcommand instead of a flag?
similar to how there is `cargo build` and `cargo run`
we could have
`gstd compile` (--print-proto)
`gstd upgrade` 
`gstd run` (--loop)
etc.

bradrn:
That makes sense, sure.

> `gstd upgrade`

What would this one do though? (Also, why `gstd` as the command name?)

TrueDoctor:
> What would this one do though? (Also, why gstd as the command name?)
Ah, sorry, meant to write `gcli`  there. 
We currently have some custom logic to upgrade documents between graphite versions but this does require more work. This was more of a "we can do this in the future" thing

TrueDoctor:
Merged manually as 85fac63bb264f281e63017df677c8e212e8cb63a

PR #2396, opened 2025-03-07
Fix quick measuring of skewed and rotated layers by using the viewport space AABB

### LINKED ISSUE #1896: Tracking Issue: Quick measurement overlays while holding the Alt key

singhutsav5502:
Make sure that quick measure overlays are based on AABBs drawn in the viewport local space.

Closes #1896


Keavon:
!build

Keavon:
Please make it draw a dashed line around the (non-sheared, non-rotated) AABB for the objects on both end of the measurement. That'll make it easier to understand what's visualized. Thanks.

singhutsav5502:
okay!

singhutsav5502:
@Keavon please review.

singhutsav5502:
Hi, @Keavon just a quick ping in case you missed the previous one.

Keavon:
!build

Keavon:
Can you please update it so the new bounding box spacing uses the same dashed stroke spacing as the dashed lines used by the quick transforms?

singhutsav5502:
Done! @Keavon

Keavon:
!build

Keavon:
Can you also please make it draw the dashed line *behind* everything else? Currently it's drawn on top, which causes it to show above the inner white parts of the transformation square handles at the corners and midpoints of the transform cage. It'd be nice if it was drawn behind the whole transform cage so that doesn't become visible when the standard rectangular transform cage is also shown.

singhutsav5502:
oh okay, i will give it a look.

singhutsav5502:
Hi, @Keavon give it a look. I couldn't really figure out how to draw *behind* the existing canvas content ( since the overlay happened after the selection ) , instead i simply redrew the selected bounds on top of the dashed overlay lines.

Keavon:
You just have to draw it before the transform cage.

singhutsav5502:
The transform cage is being drawn on selection , Before we can use the quick measure functionality ( since a selection needs to be made ). I am not sure how we can draw the dashed bounds before the general draw of the transform cage. ( since the dashed bounds are being drawn only for quick measure, drawing before it would mean that it is being drawn for every selection ).

Keavon:
Thanks! I was concerned about the redundant draw call you'd mentioned. Glad you've removed that.

Keavon:
!build

Keavon:
Regression: these triangles are now drawn behind the bounding box.

capture

singhutsav5502:
@Keavon it probably occurred because i brought down the render call of the transform cage instead of simply brining up the render call of the overlay ( fixed this now though ). Should be working as expected now.

Keavon:
!build

Keavon:
I just found another case we originally didn't account for:

capture

When they share an aligned edge, but without any overlap, we don't draw any measurement between them for their distance.

Can you please add that as well? This PR would probably be the best one to add it to, since I won't be able to do a code review tonight before merging this anyways, meaning this won't be merged for at least 12 hours but probably a couple of days since I'll be busy between a number of other tasks.

Thanks!

Keavon:
Yes.

singhutsav5502:
Hi, @Keavon I have pushed some new code, do note that i have added a `Tolerance` of 0.01 for line length ( any solid line smaller than that won't be rendered since it won't be visible and also because we were rounding off length to 2 floating points ) any line smaller than 0.01 in length was being displayed to have a length of 0 until now ( and the line wasn't visible since the width of the transform cage lines is bigger than that ) .

Keavon:
!build

Keavon (approved):

Issue #2397, opened 2025-03-07
Noise Pattern node is broken


Human43215:
i cant get the noise pattern node to work. it looks like it expects more inputs than there are in the properties panel.
(there is only one boolean input in the properties panel )

this is a bug, right? 
or am i using it wrong?

error message:

This node isn't compatible with the com-
bination of types for the data it is given:
• Input 1: () → bool          
• Input 2: () → bool
• Input 3: () → u32
• Input 4: () → f64
• Input 5: () → NoiseType
• Input 6: () → DomainWarpType
• Input 7: () → f64
• Input 8: () → FractalType
• Input 9: () → u32
• Input 10: () → f64
• Input 11: () → f64
• Input 12: () → f64
• Input 13: () → f64
• Input 14: () → CellularDistanceFunction
• Input 15: () → CellularReturnType
• Input 16: () → f64

Each invalid input should be replaced by
data with one of these supported types:
• Input 1: () → ()

Keavon:
That's a recently introduced bug, thank you for pointing it out. We'll try and get that fixed soon.

EllenGYY:
fix in #2412

Keavon:
@Human43215 you can use this in https://dev.graphite.rs but it won't be deployed to https://editor.graphite.rs until the next release at some point in the future (likely a few weeks).

Issue #2332, opened 2025-03-01
Image trace vectorization


Keavon:
One possibility of library is https://github.com/visioncortex/vtracer but there may be others worth researching and considering.

ShopaFree:
Any ETA?  Almost filed a duplicate question just now, as I was SOOOoooo hoping for this to already exist.  I mean, I _DID_ spend almost 42 seconds looking for it before I gave up and ~~posted~~, searched for this and commented, so, I think I'm doing my part here.  You're welcome :)  Great stuff Keavon & Dennis! 🤘

FWIW, best online vector conversion tracing I've seen so far, by far, is the vectorizer in Creative Fabrica studio, if that helps provide any insights.  https://studio.creativefabrica.com/vectorizer

Keavon:
This is near the top of the present priorities, so probably not much longer (perhaps weeks to months away). Still waiting on a contributor to tackle it, but if that doesn't occur, I may do it myself within the next month or two.

itsgauravjain22:
An automatic powerful image tracing feature is a highly used feature for anyone using graphic design software. I hope it will be considered as an important feature

peteroupc:
An algorithm I would like to see is one that "vectorizes" traditional user interface icons from the early to mid-1990s (generally 32 &times; 32 and in 16 colors), as follows:

- When rendered without antialiasing or scaling, the vector graphic should closely match the pixel image.
- Single-pixel, single-color line segments or curves should be rendered as line or curve commands.
- The scaling factors should range from 50% to 300%.
- Simple polygons of a single color in the input image should be rendered as such in the vector graphic.

For details and an example, see my document "[Converting Pixel Images to Vector Graphics](https://github.com/peteroupc/classic-wallpaper/blob/main/pixeltovector.md)".

PR #2398, opened 2025-03-07
Fix fill tool on raster image temporarily breaks the graph

### LINKED ISSUE #2270: Using fill tool on a raster layer temporarily breaks the graph
To reproduce:
- Paste a raster image
- Select the fill tool
- Click on the image
- Observe that the node graph has an error since the fill node does not support the raster input.

Since the tooling should avoid breaking the graph as part of #2214.

EllenGYY:
Closes #2270 

Detect whether the current layer is backed by a raster image. If it is, avoid calling the vector fill function to prevent temporarily breaking the graph.


editor/src/messages/tool/common_functionality/graph_modification_utils.rs
```diff
@@ -417,4 +417,19 @@ impl<'a> NodeGraphLayer<'a> {
...
 	}
+
+	// Check if the current layer is backed by a raster image
+	pub fn is_raster_image(layer: LayerNodeIdentifier, network_interface: &NodeNetworkInterface) -> bool {
+		// Create a NodeGraphLayer instance with the layer node identifier
+		let layer = NodeGraphLayer::new(layer, network_interface);
+		// Get the last node in the layer's flow and check if it is a raster image
+		if let Some(node_id) = layer.horizontal_layer_flow().last() {
+			let node_name = network_interface.frontend_display_name(&node_id, &[]);
+			if let Some(TaggedValue::ImageFrame(_)) = layer.find_input(&node_name, 1) {
```
0HyperCube, line 428: This isn't a particularly robust approach as nodes can produce raster data without an image frame input. For example the noise pattern node (if it wasn't broken).

EllenGYY: I change it to checking the document node definition catogory for now. As long as the catogory is correctly maintained when adding new nodes it should work fine. The noise pattern node is broken as mentioned in #2397.

0HyperCube: @EllenGYY Would it perhaps be better to check the actual input type?
```rs
let layer_input_type = self.network_interface.input_type(&InputConnector::node(layer.to_node(), 1), &[]).0.nested_type();
layer_input_type == concrete!(graphene_std::RasterFrame)
```

EllenGYY: Done. Currently I use soemthing similar to this `Some(x) if x == TypeId::of::<RasterFrame>() || x == TypeId::of::<ImageFrameTable<Color>>() || x == TypeId::of::<TextureFrameTable>() ` to check for the three raster types, since the input type is not wrapped in a RasterFrame.

0HyperCube (approved):
Works well; thanks for the contribution.

I think that checking the types is the best approach as it deals with cases such as the conversion of raster->vector with #2332. Thanks for your patience working through that.

PR #2412, opened 2025-03-09
Fix noise pattern parameter issue

### LINKED ISSUE #2397: Noise Pattern node is broken

EllenGYY:
Closes #2397 

A small bug in the node definition parameter causes the issue. Now fixed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ target/
 perf.data*
 profile.json
 flamegraph.svg
+
+.cursor/
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,3 @@ target/
 perf.data*
 profile.json
 flamegraph.svg
-
-.cursor/
-.vscode/
--- a/editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
+++ b/editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
@@ -727,7 +727,8 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 					manual_composition: Some(concrete!(Context)),
 					implementation: DocumentNodeImplementation::ProtoNode(ProtoNodeIdentifier::new("graphene_std::raster::NoisePatternNode")),
 					inputs: vec![
-						NodeInput::value(TaggedValue::Bool(false), false),
+						NodeInput::value(TaggedValue::None, false),
+						// NodeInput::value(TaggedValue::Bool(false), false),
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,6 @@ target/
 perf.data*
 profile.json
 flamegraph.svg
+
+.cursor/
+.vscode/
```
Keavon, line 9: Please revert this file change.

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -727,7 +727,8 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
 					manual_composition: Some(concrete!(Context)),
 					implementation: DocumentNodeImplementation::ProtoNode(ProtoNodeIdentifier::new("graphene_std::raster::NoisePatternNode")),
 					inputs: vec![
-						NodeInput::value(TaggedValue::Bool(false), false),
+						NodeInput::value(TaggedValue::None, false),
+						// NodeInput::value(TaggedValue::Bool(false), false),
```
Keavon, line 731: We don't like to commit commented out lines unless there's a specific reason they will be needed in the future, usually with a TODO comment note about what it's for.

.gitignore
EllenGYY: fixed

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
EllenGYY: fixed

0HyperCube (approved):
Thanks for this fix @EllenGYY.

PR #2414, opened 2025-03-09
fix: Box/lasso selection respect Deep Select mode

### LINKED ISSUE #2392: Shallow Select mode fails to transfer selection to selected sibling
In the Select tool with *Shallow Select* mode set, I should be able to single-click the green shape in the viewport and transfer the selection to it, because it is the sibling of the already-selected layer. Currently it incorrectly selects the shared parent folder instead of the green shape's "Boolean Operation" layer if it's clicked. Clicking a layer within the same shared ancestor shouldn't return selection to being outside the scope of that shared ancestor.

Image

Dhanraj30:
Closes #2392
These changes ensure that:
When in Deep Select mode, clicking or box/lasso selecting will select the deepest layers
When in Shallow Select mode, clicking or box/lasso selecting will select the shallowest layers
The selection behavior is consistent between single-click and box/lasso selection
The fix should now properly respect the Deep Select mode setting when using box/lasso selection, making the behavior consistent with the user's expectations.

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -1355,71 +1355,27 @@ impl Fsm for SelectToolFsmState {
 				SelectToolFsmState::Ready { selection }
 			}
 			(SelectToolFsmState::Drawing { selection_shape, .. }, SelectToolMessage::DragStop { remove_from_selection }) => {
-				let quad = tool_data.selection_quad();
-
-				let selection_mode = match tool_action_data.preferences.get_selection_mode() {
-					SelectionMode::Directional => tool_data.calculate_selection_mode_from_direction(),
-					selection_mode => selection_mode,
+				let selected = match selection_shape {
+					SelectionShapeType::Box => {
+						let selection_box = tool_data.selection_box();
+						tool_action_data.document.get_layers_in_box(selection_box, tool_action_data.input)
+					}
+					SelectionShapeType::Lasso => tool_data.intersect_lasso_no_artboards(tool_action_data.document, tool_action_data.input),
 				};
 
-				let intersection: Vec<LayerNodeIdentifier> = match selection_shape {
-					SelectionShapeType::Box => document.intersect_quad_no_artboards(quad, input).collect(),
-					SelectionShapeType::Lasso => tool_data.intersect_lasso_no_artboards(document, input),
-				};
-				let new_selected: HashSet<_> = if selection_mode == SelectionMode::Enclosed {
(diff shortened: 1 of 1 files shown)


editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -1588,31 +1544,16 @@ fn not_artboard(document: &DocumentMessageHandler) -> impl Fn(&LayerNodeIdentifi
...
-		tool_data.layers_dragging.push(new_selected);
+	// Only apply shallowest manipulation if we're in Shallow Select mode
+	if tool_data.nested_selection_behavior == NestedSelectionBehavior::Shallowest {
+		let selected: Vec<_> = selected.into_iter().filter(not_artboard(document)).collect();
+		if !selected.is_empty() {
+			responses.add(DocumentMessage::SetSelectedLayers { selected });
+		}
+	} else {
+		// In Deep Select mode, use the deepest manipulation logic
+		drag_deepest_manipulation(responses, selected, tool_data, document);
```
0HyperCube, line 1555: It is rather confusing to have the `drag_shallowest_manipulation` call the `drag_deepest_manipulation`.

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -1355,71 +1355,27 @@ impl Fsm for SelectToolFsmState {
...
-								if *layer != LayerNodeIdentifier::ROOT_PARENT {
-									Some(layer.to_node())
-								} else {
-									log::error!("ROOT_PARENT cannot be part of tool_data.layers_dragging");
-									None
-								}
-							})
-							.collect(),
-					});
+				if has_drawn {
```
0HyperCube, line 1366: This doesn't compile:
```
error[E0425]: cannot find value `has_drawn` in this scope
    --> editor/src/messages/tool/tool_messages/select_tool.rs:1366:8
     |
1357 |             (SelectToolFsmState::Drawing { selection_shape, .. }, SelectToolMessage::DragStop { remove_from_selection }) => {
     |              --------------------------------------------------- this pattern doesn't include `has_drawn`, which is available in `Drawing`
...
1366 |                 if has_drawn {
     |                    ^^^^^^^^^ not found in this scope
```

0HyperCube (changes requested):
Thanks for the contribution.

However please ensure that it compiles and works properly before creating a PR in the «ready to review» state.

Keavon:
Thanks for showing an interest in contributing to Graphite. Unfortunately, your PRs do not meet the minimum bar for quality as they are untested and do not compile (and chances are high that they are not human-authored) so we won't be considering them further. Best of luck in your future programming journey.

Issue #2007, opened 2024-09-26
Outline render mode support in Vello


Keavon:
Enable Vello from File > Preferences > Vello Renderer (requires a Chromium based browser and not on Linux).

EllenGYY:
fix in #2455

PR #2417, opened 2025-03-10
Fix inconsistent stroke width in 'Outline' view mode

### LINKED ISSUE #2387: Outline view mode doesn't always render with a consistent thickness
Outline (aka wireframe) mode has the wrong stroke widths (see the Procedural String Lights demo art), causing many layers to be drawn much thicker than the 1px desired.

Image

Image

EllenGYY:
Closes #2387

This issue occurs in vectors that do not have a stroke node and have a transform scale other than 1. The problem arises because a "stroke" always exists, even if it is not explicitly present in the node graph—it simply defaults to a stroke with a weight of 0. However, in these cases, stroke.transform is not correctly set.

Since the stroke is considered valid regardless, the original check prevented it from falling back to instance.transform() when necessary. To fix this, I added an additional check to ensure that the stroke is only treated as valid if its weight is greater than 0. Otherwise, we should use the fallback instead of relying on the default stroke.transform.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ target/
 perf.data*
 profile.json
 flamegraph.svg
+
+.cursor/
+.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ target/
 perf.data*
 profile.json
 flamegraph.svg
+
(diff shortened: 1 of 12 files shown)


Keavon:
!build

Keavon:
Thanks. You might want to move on to https://github.com/GraphiteEditor/Graphite/issues/2007 next which is similar.

PR #2418, opened 2025-03-10
Remove subtyping for () from node graph type system

bradrn:
The current type system for the node graph has a special case where all types are considered subtypes of the unit type `()`. However, it seems that this case never occurs in practice — I can’t find anything which breaks when this case is removed. Furthermore, it makes the type system (which is already somewhat confusing) even more difficult to reason about.


TrueDoctor (approved):

TrueDoctor:
@bradrn can you remove the unnecessary import? (basically run cargo check)

TrueDoctor:
Other than that everything seems to be fine

TrueDoctor (approved):

PR #2421, opened 2025-03-10
Fix color hex output inconsistency caused by lack of gamma conversion

### LINKED ISSUE #1710: Color::from_rgb_str() and Color::rgb_hex() inconsistency
When creating a color from a hex string with `Color::from_rgb_str()`, the hex string returned with `Color::rgb_hex()` is different.

Example:
```
let hex_color_before = "cccccc";
use graphene_core::raster::color::Color;
let color = Color::from_rgb_str(hex_color_before).unwrap();
let hex_color_after = color.rgb_hex();
console_log::init_with_level(log::Level::Info);
log::info!("rgb_hex: {:?}", hex_color_after); //logs "999999"
```

I'm pretty sure this is not intended behavior

mrodiduger:
Closes #1710 

The fix addresses an asymmetry in color representation when converting between hex strings and internal color values. 

Hex color values are read as sRGB (gamma-encoded) and converted to Alpha-premultiplied linear RGB for internal representation. Converting from internal color representation to a hex value was missing the alpha-unmultiplication and gamma encoding.  

Here is before (wrong):
Screenshot from 2025-03-11 00-25-39

Here is after (correct):
Screenshot from 2025-03-11 00-24-03


Keavon:
Thanks for this attempt. I ended up getting sucked down a multi-day rabbit hole when I started reviewing this. In short, this causes major rendering regressions so the approach isn't suitable. But I did spend the past couple days trying to find the root of the issue, getting led down towards many red herrings, and realizing it's a sizable refactor that I ended up carrying out. I'm close to done now, but I just wanted to update you that this was quite a lot less trivial than you likely had reason to expect. In the effort of tidying up the PR queue, I'm going to close this, but thank you for attempting this. If you'd like to start a new task, just ask in Discord and I'd be happy to point you to something else.

PR #2430, opened 2025-03-13
trying to fix shallow select

seif-hatem:
Closes #


Keavon:
This breaks Shallow Select entirely.

PR #2436, opened 2025-03-14
Add Convex Hull node

### LINKED ISSUE #1673: Convex Hull Node
@Keavon Just wanted a run-down of how you want this to work. I'm assuming I should just take in a VectorData, grab the points, and then generate line segments between them?

MixedMatched:
Closes #1673

Creates a [convex hull](https://en.wikipedia.org/wiki/Convex_hull) node using a [gift wrapping algorithm](https://en.wikipedia.org/wiki/Gift_wrapping_algorithm).

This PR still needs a couple "fixes" for Bézier curves:
- [ ] Gift wrap needs to check whether a given point is left of a given segment. That check needs to be updated to be for Bézier curves instead of lines.
- [ ] Instead of Bézier curves going beyond the convex hull, they need to be cut and added to the hull (issue shown below).

Screenshot from Graphite showing a black-stroked ellipse, rectangle, and pentagon surrounded by a red hull. Part of the ellipse is outside the hull because it's overlapping with the rectangle.

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -1075,6 +1075,87 @@ async fn centroid(ctx: impl Ctx + CloneVarArgs + ExtractAll, vector_data: impl N
 	DVec2::ZERO
 }
 
+#[node_macro::node(category("Vector"), path(graphene_core::vector))]
+async fn convex_hull(_: impl Ctx, mut vector_data: VectorDataTable) -> VectorDataTable {
+	// This is an implementation of convex hulls using gift wrapping, but a more efficient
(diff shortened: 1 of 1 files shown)


Keavon:
Welcome! Glad to see the beginnings of progress towards this :)

0HyperCube:
Hi @MixedMatched. Finding the time to cut the bézier can be done with [`Bezier::tangents_to_point`](https://graphite.rs/libraries/bezier-rs/#bezier/tangents-to-point), the cutting with `Bezier::split`.

I'm not sure how you'd translate the checking on the left to bézier curves. I assume this is the same as the `Bezier::winding`. This casts a ray to the left and count intersections.

I hope that helps.

PR #2437, opened 2025-03-14
Create outline with "non-scaling-stroke" SVG tag

### LINKED ISSUE #2435: Outline mode stroke width inconsistent related to parent layer scale value
If the parent layer of a vector shape has a non-1 transform scale, it will cause the outline mode rendering to have an inconsistent width. 

Image

EllenGYY:
Uses `vector-effect="non-scaling-stroke"` for outline rendering to ensure the stroke is unaffected by the parent’s transform. The normal rendering behavior remains unchanged. 

Closes #2435


Keavon:
Superseded by #2455.

PR #2457, opened 2025-03-18
Refactor many usages of Color to natively store linear not gamma

### LINKED ISSUE #1710: Color::from_rgb_str() and Color::rgb_hex() inconsistency

Keavon:
Closes #1710


editor/src/messages/portfolio/document/overlays/utility_types.rs
```diff
@@ -357,7 +357,10 @@ impl OverlayContext {
 
 		// Hover ring
 		if show_hover_ring {
-			let mut fill_color = graphene_std::Color::from_rgb_str(COLOR_OVERLAY_BLUE.strip_prefix('#').unwrap()).unwrap().with_alpha(0.5).rgba_hex();
+			let mut fill_color = graphene_std::Color::from_rgb_str(COLOR_OVERLAY_BLUE.strip_prefix('#').unwrap())
```
TrueDoctor, line 360: should this method be renamed to from_rgb_str_srgb?

Keavon: Probably but there are likely dozens that need that kind of renaming and I plan to revisit this code soon. It needs a *lot* more changes since it's extremely loosy-goosy with gamma and premult at the moment.

node-graph/gcore/src/application_io.rs
```diff
@@ -91,7 +91,7 @@ impl PartialEq for ImageTexture {
 		}
 		#[cfg(not(feature = "wgpu"))]
 		{
-			true // Unit values are always equal
```
TrueDoctor, line 94: Why did you make this change? Seems nonsensical

Keavon: Clippy

TrueDoctor: Then just remove the compile time distinctinons, we now the same code twice which is confusing

PR #2460, opened 2025-03-18
Experimental integration of the Rhai scripting language

TrueDoctor:


Keavon:
While this is cool and was worth experimenting with, I have sort of decided (though it was a hard choice) that the extra megabyte of gzipped binary size and uncommon syntax of Rhai isn't worth the tradeoff of having what will likely be a relatively obscure, undocumented feature. We're already heading in a direction where the data types (with tables/attributes) and expressiveness of the node system (with fields and granular control over the data) will reduce the need for scripting, and then we can hopefully aim to integrate a language like JS or Rust in the not-too-distant future, without having to support Rhai for backwards compatibility. Still, this was a cool experiment that showed what scripting can offer, and that's worth taking those learnings into future ideas for JS or Rust scripting.

PR #2464, opened 2025-03-18
Update #[min/max] node macro attributes to #[soft/hard]_[min/max] and make them clamp their input data

mTvare6:


Keavon:
Requires hard/soft range rework discussed in Discord.

mTvare6:
@Keavon I've changed them into `hard_min` and `min`(considering it to be the default). Do let me know if it should instead be called `soft_min`.

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -72,11 +73,11 @@ fn rectangle<T: CornerRadius>(
 }
 
 #[node_macro::node(category("Vector: Shape"))]
-fn regular_polygon<T: AsU64>(
+fn regular_polygon<T: AsU64 + std::cmp::PartialOrd + FromPrimitive>(
```
Keavon, line 76: @TrueDoctor can we avoid needing to include these trait bounds for `T` here?

TrueDoctor: Yes I described the solution when this was discussed originally, I still might have the partial code for that somewhere. These bounds should basically automatically be added by the macro

TrueDoctor:
Why do we have hard min and max but no hard/ soft ranges? And I'm not a fan of how the attributes are stored after parsing. Please choose a data representation which more closely aligns with what is a valid state. ( e.g. it shouldn't be possible to specify both hard and soft min and there should be a good error message in that case

Keavon:
> Why do we have hard min and max but no hard/ soft ranges?

Range is always soft, it refers to how far left and right the draggable range slider goes. Specifying a range (to get the dragging slider kind of number input) and a hard min/max would be the expected way to set a range slider and prevent the user from typing values beyond the intended allowable bounds.

> it shouldn't be possible to specify both hard and soft min and there should be a good error message in that case

If the soft min is larger than the hard min, that's allowed. But you're right that an error should be shown if they are equal (if so, soft should be removed) or reversed (this would indicate a mistake).

node-graph/gcore/src/vector/generator_nodes.rs
TrueDoctor: But we can also add that as a todo and implement that later if there are enough users that not having it becomes annoying

Keavon: Fixed in 1d361c030b7058bc2d956547778d92d0b26bfc23, if the code is correct. (@TrueDoctor)

Keavon (approved):

node-graph/node-macro/src/codegen.rs
```diff
@@ -134,14 +134,22 @@ pub(crate) fn generate_node_code(parsed: &ParsedNodeFn) -> syn::Result<TokenStre
...
 		.collect();
 	let number_max_values: Vec<_> = fields
 		.iter()
 		.map(|field| match field {
-			ParsedField::Regular { number_max: Some(number_max), .. } => quote!(Some(#number_max)),
+			ParsedField::Regular { number_soft_max, number_hard_max, .. } => match (number_soft_max, number_hard_max) {
+				(Some(soft_max), _) => quote!(Some(#soft_max)),
+				(None, Some(hard_max)) => quote!(Some(#hard_max)),
+				(None, None) => quote!(None),
+			},
```
Keavon, line 152: We have the hard and soft cases of these match arms both producing the same result. Is that intentional and sensible here? Assuming so, they should also perhaps be combined, replacing `None` with `_` and using ` | ` to put them on the same line.

Keavon:
Followup request: make the `#[range(...)]` min/max control the soft min/max of a range slider number input widget but still allow typing values above/below. This works successfully on some nodes like 'Copy to Points' because it's using the `f64` typedef variant, `Multiplier`, where `Some("Multiplier") => number_widget(default_info, number_input.unit("x")).into(),` doesn't constrain a min/max, but others like the 'Generate Handles' node which uses a regular `f64` with `#[range((0., 1.))]` does end up setting the `.min(...)` and `.max(...)` parts of the number input builder pattern construction. Specifically see [this code](https://github.com/GraphiteEditor/Graphite/blob/fix-min-max-macro/editor/src/messages/portfolio/document/node_graph/node_properties.rs#L132-L136).

Issue #2465, opened 2025-03-18
Testing for GRS


0HyperCube:
Things to test:
- [x] Should work after the scale is set to 0, or at least not crash (#2491)
- [x] Should work with only one anchor (see #2502)
- [x] Should work when the document is panned, zoomed, and rotated (#2492)
- [x] Should work when only one anchor point is selected (#2502)
- [x] Cancelling works properly, using right click and escape (#2467)
- [x] If the commands are chained (e.g. pressing G and then R without committing / cancelling) then this should work nicely (#2475)
- [ ] Should work with different selections and groups (#2518)

rahat2134:
@0HyperCube Please assign me the issue.

Keavon:
@rahat2134 we don't do issue assignments until PRs are merged, but you can post your PR number as a comment here once you have a reasonably complete PR ready, that way people know you're working on it.

rahat2134:
@Keavon Sounds good!

PR #2467, opened 2025-03-19
Add tests for GRS transform cancellation

rahat2134:
Fix part of #2465 

This PR adds tests to verify that cancellation works properly for the GRS (Grab, Rotate, Scale) transformation operations.

I implemented tests that:
Create a rectangle
Start a transform operation (G/R/S)
Move the cursor to change the transform
Cancel the operation
Verify the transform is unchanged after cancellation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ perf.data*
 profile.json
 flamegraph.svg
 
-.qodo
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ perf.data*
 profile.json
 flamegraph.svg
 
+.qodo
--- a/editor/src/dispatcher.rs
+++ b/editor/src/dispatcher.rs
@@ -13,6 +13,7 @@ pub struct Dispatcher {
 
 #[derive(Debug, Default)]
 pub struct DispatcherMessageHandlers {
+	animation_message_handler: AnimationMessageHandler,
 	broadcast_message_handler: BroadcastMessageHandler,
 	debug_message_handler: DebugMessageHandler,
 	dialog_message_handler: DialogMessageHandler,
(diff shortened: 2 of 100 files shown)


0HyperCube (commented):
Thanks for the contribution; the code and approach look good.

Please can you also add a testcase that doesn't cancel to ensure that we don't get the original transform in that case? This ensures that we are actually testing that the cancelling works.

rahat2134:
@0HyperCube @Keavon PTAL!

.gitignore
```diff
@@ -5,3 +5,4 @@ perf.data*
 profile.json
 flamegraph.svg
 
+.qodo
```
0HyperCube, line 8: Please mention when you have used the AI to write your PRs @rahat2134. This is due to the fact that your AI model may have been trained on GPLd software which we cannot include here due to 5c:
> 5. Conveying Modified Source Versions.
>
>  You may convey a work based on the Program, or the modifications to produce it from the Program, in the form of source code under the terms of section 4, provided that you also meet all of these conditions:
>
>    c) You must license the entire work, as a whole, under this License to anyone who comes into possession of a copy. This License will therefore apply, along with any applicable section 7 additional terms, to the whole of the work, and all its parts, regardless of how they are packaged. This License gives no permission to license the work in any other way, but it does not invalidate such permission if you have separately received it.

rahat2134: @0HyperCube The tests are my own work - I just used AI to better understand the codebase (which is a part of learning). All implementation and debugging was done by me directly.
Should I still mention AI? I think here it is not connected with the tests, I wrote them on my own.
Thanks for prompting!

Keavon: You should be good then, thanks

0HyperCube (approved):
Look good; thanks for the contribution.

Issue #1715, opened 2024-03-29
Shape tools should allow modifying their drawn shapes


0HyperCube:
Should go hand-in-hand with #2090.

Tools such as the rectangle and shape tool should allow for editing of the created shape.

Tools:
- [x] Line tool adjusting start and end
- [x] Rectangle allowing for moving top left and bottom right corners and ideally for rounding
- [x] Ellipse should allow dragging a handle for the X extent and Y extent and ideally creating an arc/chord
- [x] Polygon tool handle for the radius and offset of the first vertex. For a star the inner and outer radius should be controlled.

These can be done as separate MRs.

PR #2470, opened 2025-03-19
New node: Arc

### LINKED ISSUE #2434: Add arc/unclosed shape options to circle node
I'm trying to use Graphite to generate sprite sheets for a circular progress indicator. Inkscape has the ability to specify a start and end angle for the stroke, creating an unclosed cirlce:

Image

Having node editor properties to set these in Graphite would allow using nodes to generate the different images, with the arc filling up to close the circle as the progress indicator gets to 100%.

optozorax:
Closes #2434


optozorax:
Most of the code copied from https://github.com/GraphiteEditor/Graphite/blob/08a4b69948d9f2b8cdaf1a47caa30db7f3029aab/editor/src/messages/portfolio/document/overlays/utility_types.rs#L278

Tested it against circle with stroke +1px to see differences between them. Found none, seems pretty accurate.

Works for any angle from 0 to 360.

image

Keavon:
!build

mTvare6:
Doesn't the ellipse part of #1715 also cover the use case of an arc. I believe implementing that would make an arc node redundant. 
Inkscape for reference.

pie
segment

Keavon:
!build

Keavon:
I made some improvements, now it can animate with values of any range without discontinuities:

https://github.com/user-attachments/assets/c28bf5ea-dcee-48af-92fe-2c6489e78ad0

PR #2475, opened 2025-03-23
Add test for chained GRS transformations

rahat2134:
Fixes a part of #2465 
What is being tested:
Sequential transformation operations (G→R→S) without committing between operations
Verifies each transformation accumulates correctly during chaining
Confirms final transform matches expected combined transform

Implemented tests:
test_grab_rotate_scale_chained: Verifies proper sequential application of grab, rotate, and scale operations, ensuring each transformation is preserved through the chain
This functionality allows users to smoothly transition between different transformation types while maintaining previously applied transforms.

### DIFF
--- a/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
+++ b/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
@@ -914,15 +914,13 @@ mod test_transform_layer {
 		let mut editor = EditorTestUtils::create();
 		editor.new_document().await;
 		editor.drag_tool(ToolType::Rectangle, 0., 0., 100., 100., ModifierKeys::empty()).await;
-
 		let document = editor.active_document();
 		let layer = document.metadata().all_layers().next().unwrap();
 		editor.handle_message(NodeGraphMessage::SelectedNodesSet { nodes: vec![layer.to_node()] }).await;
-
 		let original_transform = get_layer_transform(&mut editor, layer).await.unwrap();
 
 		editor.handle_message(TransformLayerMessage::BeginGrab).await;
-		editor.move_mouse(50.0, 50.0, ModifierKeys::empty(), MouseKeys::NONE).await;
+		editor.move_mouse(150.0, 130.0, ModifierKeys::empty(), MouseKeys::NONE).await;
 		editor
 			.handle_message(TransformLayerMessage::PointerMove {
 				slow_key: Key::Shift,
@@ -931,27 +929,36 @@ mod test_transform_layer {
 			.await;
 
 		let after_grab_transform = get_layer_transform(&mut editor, layer).await.unwrap();
-		assert!(!after_grab_transform.abs_diff_eq(original_transform, 1e-5), "Grab should change the transform");
+		let expected_translation = DVec2::new(50.0, 30.0);
+		let actual_translation = after_grab_transform.translation - original_transform.translation;
+		assert!(
+			(actual_translation - expected_translation).length() < 1e-5,
+			"Expected translation of {:?}, got {:?}",
+			expected_translation,
+			actual_translation
+		);
 
+		// 2. Chain to rotation - from current position to create ~45 degree rotation
 		editor.handle_message(TransformLayerMessage::BeginRotate).await;
-		editor.move_mouse(150.0, 50.0, ModifierKeys::empty(), MouseKeys::NONE).await;
+		editor.move_mouse(190.0, 90.0, ModifierKeys::empty(), MouseKeys::NONE).await;
 		editor
 			.handle_message(TransformLayerMessage::PointerMove {
 				slow_key: Key::Shift,
 				increments_key: Key::Control,
 			})
 			.await;
-
 		let after_rotate_transform = get_layer_transform(&mut editor, layer).await.unwrap();
-		assert!(!after_rotate_transform.abs_diff_eq(after_grab_transform, 1e-5), "Rotation should change the transform after grab");
-		// Verifying rotation matrix components have changed for more robustness
+		// Checking for off-diagonal elements close to 0.707, which corresponds to cos(45°) and sin(45°)
 		assert!(
-			!after_rotate_transform.matrix2.abs_diff_eq(after_grab_transform.matrix2, 1e-5),
-			"Rotation should change matrix components"
+			!after_rotate_transform.matrix2.abs_diff_eq(after_grab_transform.matrix2, 1e-5) &&
+			(after_rotate_transform.matrix2.x_axis.y.abs() - 0.707).abs() < 0.1 &&  // Check for off-diagonal elements close to 0.707
+			(after_rotate_transform.matrix2.y_axis.x.abs() - 0.707).abs() < 0.1, // that would indicate ~45° rotation
+			"Rotation should change matrix components with approximately 45° rotation"
 		);
 
+		// 3. Chain to scaling - scale(area) up by 2x
 		editor.handle_message(TransformLayerMessage::BeginScale).await;
-		editor.move_mouse(200.0, 200.0, ModifierKeys::empty(), MouseKeys::NONE).await;
+		editor.move_mouse(250.0, 200.0, ModifierKeys::empty(), MouseKeys::NONE).await;
 		editor
 			.handle_message(TransformLayerMessage::PointerMove {
 				slow_key: Key::Shift,
@@ -960,17 +967,19 @@ mod test_transform_layer {
 			.await;
 
 		let after_scale_transform = get_layer_transform(&mut editor, layer).await.unwrap();
-		assert!(!after_scale_transform.abs_diff_eq(after_rotate_transform, 1e-5), "Scaling should change the transform after rotation");
-		// Verifying scale changed the determinant (area) for more robustness
 		let before_scale_det = after_rotate_transform.matrix2.determinant();
 		let after_scale_det = after_scale_transform.matrix2.determinant();
-		assert!((after_scale_det - before_scale_det).abs() > 1e-5, "Scaling should change the determinant of the matrix");
+		assert!(
+			after_scale_det >= 2.0 * before_scale_det,
+			"Scale should increase the determinant of the matrix (before: {}, after: {})",
+			before_scale_det,
+			after_scale_det
+		);
 
 		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
 		let final_transform = get_layer_transform(&mut editor, layer).await.unwrap();
-		// Applying the transform keeps the same values
+
 		assert!(final_transform.abs_diff_eq(after_scale_transform, 1e-5), "Final transform should match the transform before committing");
-		// Verifying the whole sequence produced a different transform from original
 		assert!(!final_transform.abs_diff_eq(original_transform, 1e-5), "Final transform should be different from original transform");
 	}
 }
--- a/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
+++ b/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
@@ -908,4 +908,71 @@ mod test_transform_layer {
 		let translation_diff = (after_cancel.translation - original_transform.translation).length();
 		assert!(translation_diff < 1.0, "Translation component changed too much: {}", translation_diff);
 	}
+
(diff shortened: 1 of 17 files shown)


rahat2134:
@0HyperCube Please take a look.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -908,4 +908,71 @@ mod test_transform_layer {
...
+
+		// Verifying scaling was applied
+		let scale_diff_x = (after_scale_transform.matrix2.x_axis.length() - after_rotate_transform.matrix2.x_axis.length()).abs();
+		let scale_diff_y = (after_scale_transform.matrix2.y_axis.length() - after_rotate_transform.matrix2.y_axis.length()).abs();
+		assert!(scale_diff_x > 0.1 || scale_diff_y > 0.1, "Scaling should be applied in the chain");
+
+		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
+
+		let final_transform = get_layer_transform(&mut editor, layer).await.unwrap();
+		assert_eq!(after_scale_transform, final_transform, "Final transform should match the chained operations");
```
0HyperCube, line 976: Why are you comparing the `get_layer_transform` to the `get_layer_transform` from before a few matrix multiplications and asserts (which won't change the transform at all)?

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -908,4 +908,71 @@ mod test_transform_layer {
...
+				slow_key: Key::Shift,
+				increments_key: Key::Control,
+			})
+			.await;
+
+		let after_scale_transform = get_layer_transform(&mut editor, layer).await.unwrap();
+
+		// Verifying translation is preserved
+		let translation_diff = (after_scale_transform.translation - original_transform.translation).length();
+		assert!(translation_diff > 1.0, "Translation should be preserved through the chain");
```
0HyperCube, line 962: A tolerance of 1 is really quite high considering that you are working with doubles. Perhaps use `1e-5` or something?

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -908,4 +908,71 @@ mod test_transform_layer {
...
+
+		// Verifying rotation is preserved
+		let rotation_matrix_diff = (after_scale_transform.matrix2.x_axis - after_grab_transform.matrix2.x_axis).length();
+		assert!(rotation_matrix_diff > 0.1, "Rotation should be preserved through the chain");
+
+		// Verifying scaling was applied
+		let scale_diff_x = (after_scale_transform.matrix2.x_axis.length() - after_rotate_transform.matrix2.x_axis.length()).abs();
+		let scale_diff_y = (after_scale_transform.matrix2.y_axis.length() - after_rotate_transform.matrix2.y_axis.length()).abs();
+		assert!(scale_diff_x > 0.1 || scale_diff_y > 0.1, "Scaling should be applied in the chain");
```
0HyperCube, line 971: I don't really understand why you are doing this. I'd suggest that you utilise the [`DAffine2::abs_diff_eq`](https://docs.rs/glam/latest/glam/f64/struct.DAffine2.html#method.abs_diff_eq) to compare the transformation to an expected matrix (which you can build using [`DAffine2::from_scale_angle_translation`](https://docs.rs/glam/latest/glam/f64/struct.DAffine2.html#method.from_scale_angle_translation)).

0HyperCube (changes requested):
Thanks for your work on this. I'm not convinced that the asserts really accomplish what we want at the moment though.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: Done, Thanks for the feedback!

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: Ya, I misconfused it somewhere; I just tried correcting it. Maybe you can have a look now.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: @0HyperCube instead of this, I am trying the determinant method as when we will do scaling, the area will change.

Does this make sense? Although i have implemented the changes you can have a look.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -908,4 +908,69 @@ mod test_transform_layer {
...
+		let before_scale_det = after_rotate_transform.matrix2.determinant();
+		let after_scale_det = after_scale_transform.matrix2.determinant();
+		assert!((after_scale_det - before_scale_det).abs() > 1e-5, "Scaling should change the determinant of the matrix");
+
+		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
+		let final_transform = get_layer_transform(&mut editor, layer).await.unwrap();
+		// Applying the transform keeps the same values
+		assert!(final_transform.abs_diff_eq(after_scale_transform, 1e-5), "Final transform should match the transform before committing");
+		// Verifying the whole sequence produced a different transform from original
+		assert!(!final_transform.abs_diff_eq(original_transform, 1e-5), "Final transform should be different from original transform");
```
0HyperCube, line 974: Verifying that the transform is different is not really good enough. We are trying to test that the pivots and transformations all work correctly, which was fixed in #2450.

I would expect you to build a transform based on the expected scaling and translation and then to assert that it is equal to the actual transform of the layer. Thanks.

rahat2134: @0HyperCube  I'm having difficulty with your suggestion to "build a transform based on the expected scaling and translation and then assert that it is equal to the actual transform".

**What I've tried:**
   - Tried to carefully calculate expected transforms with pivot points:
   ```rust
   // Pivot point handling
   let pivot_to_origin = DAffine2::from_translation(-pivot);
   let origin_to_pivot = DAffine2::from_translation(pivot);
   
   // Transform components
   let expected_translate = DAffine2::from_translation(translation_offset);
   let rotation = DAffine2::from_angle(rotation_angle);
   let expected_rotate = origin_to_pivot * rotation * pivot_to_origin;
   let scaling = DAffine2::from_scale(DVec2::splat(scale_factor));
   let expected_scale = origin_to_pivot * scaling * pivot_to_origin;
   
   // Combined transform
   let expected_transform = expected_scale * expected_rotate * expected_translate;
   ```
   - This still gave significantly different results from the actual transform

For example, moving to (20,30) resulted in:
```
After grab: { matrix2: DMat2 { x_axis: DVec2(1.0, 0.0), y_axis: DVec2(0.0, 1.0) }, translation: DVec2(-30.0, -20.0) }
```

Could you point me to the specific transformation logic in the codebase I should use to calculate the expected transform matrix?

Or

If i am doing something wrong!

0HyperCube: @rahat2134 You where presumably moving the mouse from (50,50) to (20,30) in order to get a translation on the object of (-30, -20)?

0HyperCube: The approach that you had at the start seems reasonable. However you will obviously need to ensure you have the correct pivot position.

rahat2134: Hey @0HyperCube Does this make any sense?
``` rust
	#[tokio::test]
	async fn test_grab_rotate_scale_chained() {
		let mut editor = EditorTestUtils::create();
		editor.new_document().await;
		editor.drag_tool(ToolType::Rectangle, 0., 0., 100., 100., ModifierKeys::empty()).await;

		let document = editor.active_document();
		let layer = document.metadata().all_layers().next().unwrap();
		editor.handle_message(NodeGraphMessage::SelectedNodesSet { nodes: vec![layer.to_node()] }).await;

		let original_transform = get_layer_transform(&mut editor, layer).await.unwrap();

		// 1. Grab operation - move from center position to a new position
		let start_pos = DVec2::new(50.0, 50.0);
		let grab_pos = DVec2::new(80.0, 70.0);
		// These coordinates appear to be transformed with:
		// 1. Inversion of sign
		// 2. Swap of x/y (or some other transformation)
		let expected_translation = DVec2::new(-20.0, -30.0);

		editor.handle_message(TransformLayerMessage::BeginGrab).await;
		editor.move_mouse(grab_pos.x, grab_pos.y, ModifierKeys::empty(), MouseKeys::NONE).await;
		editor
			.handle_message(TransformLayerMessage::PointerMove {
				slow_key: Key::Shift,
				increments_key: Key::Control,
			})
			.await;

		let after_grab_transform = get_layer_transform(&mut editor, layer).await.unwrap();
		let actual_translation = after_grab_transform.translation - original_transform.translation;

		assert!(
			(actual_translation - expected_translation).length() < 1e-5,
			"Expected translation near {:?}, got {:?}",
			expected_translation,
			actual_translation
		);

		editor.handle_message(TransformLayerMessage::BeginRotate).await;
		editor.move_mouse(150.0, 50.0, ModifierKeys::empty(), MouseKeys::NONE).await;
		editor
			.handle_message(TransformLayerMessage::PointerMove {
				slow_key: Key::Shift,
				increments_key: Key::Control,
			})
			.await;

		let after_rotate_transform = get_layer_transform(&mut editor, layer).await.unwrap();
		assert!(!after_rotate_transform.abs_diff_eq(after_grab_transform, 1e-5), "Rotation should change the transform after grab");
		// Verifying rotation matrix components have changed for more robustness
		assert!(
			!after_rotate_transform.matrix2.abs_diff_eq(after_grab_transform.matrix2, 1e-5),
			"Rotation should change matrix components"
		);

		// 3. Scale operation
		editor.handle_message(TransformLayerMessage::BeginScale).await;
		// Move to position that would create ~2x scale
		let scale_target = DVec2::new(grab_pos.x + 100.0, grab_pos.y + 100.0);
		editor.move_mouse(scale_target.x, scale_target.y, ModifierKeys::empty(), MouseKeys::NONE).await;
		editor
			.handle_message(TransformLayerMessage::PointerMove {
				slow_key: Key::Shift,
				increments_key: Key::Control,
			})
			.await;

		let after_scale_transform = get_layer_transform(&mut editor, layer).await.unwrap();
		// Expected scale factor (approximately 2.0)
		let expected_scale_factor = 2.0;
		let actual_scale_factor = (after_scale_transform.matrix2.determinant() / after_rotate_transform.matrix2.determinant()).sqrt();

		assert!(
			(actual_scale_factor - expected_scale_factor).abs() < 0.5,
			"Expected scale factor near {}, got {}",
			expected_scale_factor,
			actual_scale_factor
		);

		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
		
		// Final check is still remaining
	}
```

0HyperCube: Please just update the PR instead of putting an 80 line code block.

I'm extremly confused by:
> // These coordinates appear to be transformed with:
> // 1. Inversion of sign
> // 2. Swap of x/y (or some other transformation)

Please note that the `editor.move_mouse` function moves the mouse from whatever position it was at previously to the position specified. Since the last operation was to draw a rectangle to (100,100), the mouse will therefore be at (100,100). Moving it to (80.0, 70.0) will clearly mean a transform of (-20, -30).


Also please can you use the [`DAffine2::abs_diff_eq`](https://docs.rs/glam/latest/glam/f64/struct.DAffine2.html#method.abs_diff_eq) function as previously requested. This ensures that all components of the transform are correct.

Thanks again.

rahat2134:
@0HyperCube PTAL!!

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: I tried it again, Please have a look!

0HyperCube (approved):
Thanks for your work on this.

PR #2477, opened 2025-03-25
New node: Blur

### LINKED ISSUE #995: Gaussian blur node

calvintvu:
Closes #995

Implements both gaussian and box blurs under raster category

https://github.com/user-attachments/assets/89a2a045-5d73-4dab-84bb-962f7e1b5702


calvintvu:
Typo in commit message

TrueDoctor:
Is there a specific reason why you chose to convert this to an image data type first instead of using or inbuilt one? Also @Keavon should blur happen in srgb or linear? They should give different results, so we have to think about which one is correct / what is compatible with other tools

TrueDoctor:
It also looks like the current implementation quantizes the values to 8 bit which is most likely not what we want. We should also take SIMD friendliness into account here and potentially try to re-enable that target flag and hope that the chrome bug from two years ago has been fixed

Keavon:
We should have both linear and gamma, with linear as default. https://www.youtube.com/watch?v=LKnqECcg6Gw

calvintvu:
I've updated it to convert from linear/nonlinear before/after blurring using the functions from graphene_core::raster::Channel and changed from u8 values to f32

calvintvu:
Yes ill keep that in mind

Keavon:
A couple code review notes before I do a proper code review:
- You're repeating a whole lot of code by doing X then Y without reusing the logic, please do a self-review pass keeping DRY in mind
- Please replace whole-number decimals so they end in a decimal point (`.`) instead of point-zero (`.0`), per our style guide

Keavon:
I pushed some code review improvements. We still will need you to remove usage of the `image` crate which you shouldn't need to be doing. Once you do that, please use our existing `Color` data type's linear/gamma conversion functions. Then please test that a solid red and green has a black border when gamma is active, rather than yellow as with linear.

- Expected with gamma
  capture
- Expected with linear
  capture

Keavon:
Hi @calvintvu, just checking in to see if this is something you're still interested in finishing. Hope to hear from you! Thanks 😄

calvintvu:
> Hi @calvintvu, just checking in to see if this is something you're still interested in finishing. Hope to hear from you! Thanks 😄

Hey @Keavon yeah I can take a look and implement the changes, just busy with some other things at the moment

calvintvu:
I removed the use of the`image` crate and switched the conversion functions to the ones in color.rs (`to_linear_srgb` and `to_gamma_srgb`). The linear version has a yellow bar as expected but the gamma verson doesn't seem to have the black bar and I'm not sure how to fix that. Thanks


gamma
linear

Keavon:
!build

Keavon:
- I believe in your implementation, "Gamma" is actually linear, and linear is actually double linear.
- I also noticed the checkbox for "Box Blur" is backwards.

Keavon:
Here is a test file you can work with:

[blur-test.graphite.txt](https://github.com/user-attachments/files/19830042/blur-test.graphite.txt)

calvintvu:
Ok fixed box blur option and the colors seem to match up better 
Screenshot 2025-04-20 at 23 34 12
Screenshot 2025-04-20 at 23 34 18

Keavon:
!build

Keavon:
It's looking like we need to either premultiply or not premultiply the alpha (the opposite of what's currently being done) to avoid dark areas around the blurred rectangle edges when we have a transparent instead of a white background. We should have functions to premult and unpremult available in the Color method.

calvintvu:
Got these results after alpha premult/unpremult
Screenshot 2025-04-24 at 18 21 49
Screenshot 2025-04-24 at 18 21 41

Keavon:
Please leave a comment in #995 saying that you fixed this issue (`#2477`) for tracking purposes. Thanks.

Keavon (approved):

Keavon:
@calvintvu friendly reminder to:

> Please leave a comment in #995 saying that you fixed this issue (`#2477`) for tracking purposes. Thanks.

PR #2490, opened 2025-03-26
Add tests for drawing lines with ctrl, shift, and alt

### LINKED ISSUE #2487: Testing line tool
Testcases:
- [ ] Works when the viewport is zoomed/panned/tilted
- [ ] Works when the artboard has an offset (currently broken #2485)
- [ ] Works when the artboard has a transform before it
- [x] <kbd>shift</kbd> constrains to 15 degrees from the viewport (test when the viewport is rotated by 5 degrees)
- [x] <kbd>Ctrl</kbd> locks the angle when pressed
- [x] <kbd>Alt</kbd> drags from centre (test when combined with shift as well)
- [ ] After drawing, selected layers *only* can have their endpoints dragged.

prathmesh703:
Fixes #2487   parts

 CTRL locks the angle when pressed 
 Alt drags from centre (test when combined with shift as well)


0HyperCube (approved):
Looks good; thanks for working on these tests.

PR #2491, opened 2025-03-27
Add test for scaling objects to zero without any crash

rahat2134:
Fixes a part of #2465 

This PR adds a test case that verifies the application correctly handles scaling objects to exactly zero. This edge case is important to test as it creates matrices with zero determinants, which could potentially cause numerical instability or crashes.


rahat2134:
@0HyperCube PTAL!

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -908,4 +908,47 @@ mod test_transform_layer {
...
+
+		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
+
+		let final_transform = get_layer_transform(&mut editor, layer).await.unwrap();
+
+		// Checking the transform matrix components are valid (not NaN or infinite)
+		assert!(final_transform.matrix2.x_axis.x.is_finite(), "X-axis x component should be finite");
+		assert!(final_transform.matrix2.x_axis.y.is_finite(), "X-axis y component should be finite");
+		assert!(final_transform.matrix2.y_axis.x.is_finite(), "Y-axis x component should be finite");
+		assert!(final_transform.matrix2.y_axis.y.is_finite(), "Y-axis y component should be finite");
```
0HyperCube, line 942: You can simply do:
```rs
assert!(final_transform.is_finite(), "Matrix should be finite");
```

0HyperCube (changes requested):
Thanks for working on this. Scaling to zero should probably always work. However we've previously had problems whereby after scaling to zero and then transforming, a NaN value is used. This causes the SVG to be invalid. Please could you also test for applying another scale after setting the transform to zero? Thanks.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -982,4 +982,51 @@ mod test_transform_layer {
...
+
+		let rescale = DAffine2::from_scale(DVec2::new(2.0, 2.0));
+		editor
+			.handle_message(GraphOperationMessage::TransformSet {
+				layer,
+				transform: rescale * zero_transform,
+				transform_in: TransformIn::Local,
+				skip_rerender: false,
+			})
+			.await;
```
0HyperCube, line 1019: This isn't testing the GRS system though?

rahat2134:
CLI failure is independent of changes - https://discord.com/channels/731730685944922173/731738914812854303/1358066874314916003

0HyperCube (approved):

rahat2134:
@0HyperCube it will be merge automatically after cli test will pass?

PR #2498, opened 2025-03-29
Make the Path tool support multi-point conversion between smooth/sharp on double-click

0SlowPoke0:
Part of #1870 

Added feature:

> When double-clicking an anchor:
> 
> - Currently this swaps between smooth and sharp, but doesn't work if multiple handles are multi-selected to convert them all to smooth or sharp. If they're of mixed type, the specific one that's double clicked on should be the decider for the rest of them as to which they should all be swapped to.


Keavon:
For the record, I'll copy and paste the point of trickiness that I mentioned in Discord regarding the input conflict in the desired behavior from the first of the two bullet points addressed in this PR:

> It might be a little tricky, but can you make it still deselect all others, but if the single click is followed up with a double click, it restores the selection and performs the smooth/sharp swap?

Keavon (approved):

PR #2511, opened 2025-04-03
Fix Pen tool creates single anchor path

### LINKED ISSUE #2503: Pen tool creates single anchor path
Currently you can create a path with a single point by:
- Selecting the pen tool
- Mouse down
- Mouse up
- Switch to another tool (with a shortcut or the button)

This creates an essentially "empty" layer:
Empty layer

This is not desired. The pen tool should instead delete the new layer.

`editor/src/messages/tool/tool_messages/pen_tool.rs`

c-mateo:
This pull request fixes #2503 by adding a new field to track the last layer and implementing logic to handle layer deletion upon aborting a path with only one point.

### DIFF
--- a/editor/src/messages/tool/tool_messages/pen_tool.rs
+++ b/editor/src/messages/tool/tool_messages/pen_tool.rs
@@ -294,7 +294,6 @@ struct PenToolData {
 	handle_end: Option<DVec2>,
 	next_point: DVec2,
 	next_handle_start: DVec2,
-	last_layer: LayerNodeIdentifier,
 
 	g1_continuous: bool,
 	toggle_colinear_debounce: bool,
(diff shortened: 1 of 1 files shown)


Keavon:
I feel like this should be achievable without needing to store state. @0HyperCube could you please give your insights about if you agree or disagree with me on that?

c-mateo:
Just to mention: text tool does it the same way https://github.com/GraphiteEditor/Graphite/blob/6de235f80cb8e7b33d2b5af4cca02dc87b5345fc/editor/src/messages/tool/tool_messages/text_tool.rs#L291

Keavon:
Thanks for pointing that out. I'm always quick to raise concern when new state is introduced (engineering tip: never store more state than you have to!) but if the other tool does that for the similar use case, that's probably reasonable evidence that it's ok to do here as well.

c-mateo:
I guess it's not possible to directly modify node graph from the tools. Maybe we could add a new message to node graph to delete last added layer in order to free these tools of having to store it.

0HyperCube (changes requested):
@Keavon is correct that there is no need to store the layer again. There is a lot of other functionality that relies on the active layer, which is already determined at the top of the fsm transition function:
https://github.com/GraphiteEditor/Graphite/blob/6de235f80cb8e7b33d2b5af4cca02dc87b5345fc/editor/src/messages/tool/tool_messages/pen_tool.rs#L1092

0HyperCube (approved):
Thanks for this fix.

PR #2512, opened 2025-04-03
Invalidate cache when unexposing input

### LINKED ISSUE #2493: Unexposing an input doesn't immediately invalidate the cache
https://github.com/user-attachments/assets/1621c4ba-64f7-44a5-90fe-e39ba4831247

I expected the canvas to immediately update after toggling off the connection but as you can see it lingers for a while.

c-mateo:
Fixes  #2493 adding `NodeGraphMessage::RunDocumentGraph` to the list of responses for `NodeGraphMessage::ExposeInput` in the `node_graph_message_handler.rs` file.


0HyperCube (approved):
Looks good; thanks for the fix.

PR #2516, opened 2025-04-05
Make the 'Tangent Inverse' node accept DVec2 in lieu of the separate atan2 node

kythyria:
In this fashion it doubles as atan2; in theory the dedicated atan2 node can now be removed or replaced with a subgraph, but that might break existing files.


Keavon:
Please also delete the old atan2 node now that it's redundant (and had the wrong way of exposing X/Y).

Keavon:
!build

Keavon (approved):

PR #2518, opened 2025-04-05
Add GRS tests for different types of nested layer multi-selections

### LINKED ISSUE #2465: Testing for GRS

rahat2134:
Closes #2465 

Added a comprehensive test for Graphite's transform operations (Grab, Rotate, Scale) across different selection scenarios:
Single layer selection
Multiple layers selection
Group selection


rahat2134:
@0HyperCube PTAL!!

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -1139,4 +1140,80 @@ mod test_transform_layer {
...
+		let mut editor = EditorTestUtils::create();
+		editor.new_document().await;
+		editor.draw_rect(0., 0., 100., 100.).await;
+		editor.draw_rect(150., 0., 250., 100.).await;
+		editor.draw_rect(0., 150., 100., 250.).await;
+		editor.draw_rect(150., 150., 250., 250.).await;
+		let document = editor.active_document();
+		let layers: Vec<LayerNodeIdentifier> = document.metadata().all_layers().collect();
+
+		assert!(layers.len() >= 4);
```
0HyperCube, line 1155: Why is this `>=`? Surely you can specify exactly?

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -1139,4 +1140,80 @@ mod test_transform_layer {
...
+		editor.move_mouse(50.0, 50.0, ModifierKeys::empty(), MouseKeys::NONE).await;
+		editor
+			.handle_message(TransformLayerMessage::PointerMove {
+				slow_key: Key::Shift,
+				increments_key: Key::Control,
+			})
+			.await;
+		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
+		let final_transform = get_layer_transform(&mut editor, layers[0]).await.unwrap();
+		assert!(final_transform != original_transform, "Transform should change for single layer");
```
0HyperCube, line 1186: The fact that the transform changes is not that useful. Please test the actual transform with `DAffine2::abs_diff_eq`.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -1139,4 +1140,80 @@ mod test_transform_layer {
...
+				increments_key: Key::Control,
+			})
+			.await;
+		editor.handle_message(TransformLayerMessage::ApplyTransformOperation { final_transform: true }).await;
+		let final_transform_1 = get_layer_transform(&mut editor, layers[0]).await.unwrap();
+		let final_transform_2 = get_layer_transform(&mut editor, layers[1]).await.unwrap();
+		assert!(final_transform_1 != original_transform_1, "Transform should change for first layer in multi-selection");
+		assert!(final_transform_2 != original_transform_2, "Transform should change for second layer in multi-selection");
+
+		// Test 3: Transform group
```
0HyperCube, line 1210: I also think it would be useful if you could scale the group and then translate layers in the group. This would then test that the transforms work correctly when the parent is transformed (which tends to be rather fragile).

0HyperCube (commented):
Thanks for the contribution so far.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: Sorry for this, I will try to keep this in mind now!

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
rahat2134: @0HyperCube please have a look to new changes! If it is implemented correctly!

rahat2134:
@0HyperCube PTAL!

0HyperCube (approved):

PR #2519, opened 2025-04-06
Allow the fill bucket tool to fill a raster image

EllenGYY:
•	Adds a “Raster Fill” node when the fill bucket tool is used on a raster layer.
•	Non-destructive: deleting the node restores the image to its original state.
•	Users can adjust the “tolerance” of color similarity, based on Euclidean distance in LAB color space (currently per-layer only).

Possible improvements:
	•	Allow users to adjust and delete individual fill operations on the same layer, including their respective tolerance settings.
	•	Add support for gradient fills.

raster-fill-demo


Keavon:
Please fix the CI failure on the test `messages::tool::tool_messages::fill_tool::test_fill::ignore_raster`.

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -71,37 +76,111 @@ enum FillToolFsmState {
...
+			let tolerance = if let Some(TaggedValue::F64(tolerance)) = existing_fills[3].as_value().cloned() {
+				tolerance
+			} else {
+				1.
+			};
+
+			*self = Self { fills, start_pos, tolerance };
+		}
+
+		// TODO: Why do we overwrite the tolerance that we just set a couple lines above?
```
Keavon, line 111: @EllenGYY

node-graph/gstd/src/raster.rs
```diff
@@ -684,6 +685,92 @@ fn mandelbrot_impl(c: Vec2, max_iter: usize) -> usize {
...
+		Vec<Option<Color>>,
+		Vec<Color>,
+		Vec<Gradient>
+	)]
+	/// The fills to paint the path with.
+	#[default(Vec<Color::BLACK>)]
+	fills: Vec<F>,
+	/// The starting positions of the flood fill points in the layer's local coordinates.
+	positions: Vec<DVec2>,
+	// TODO: What is the range? I'd expect 0-255, but we should also rescale that to 0-1. But this crashes at larger values approaching 100. And tolerance should be per-position not global to the node.
```
Keavon, line 705: @EllenGYY

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -71,37 +76,111 @@ enum FillToolFsmState {
 	Filling,
 }
 
+#[derive(Clone, Debug, Default)]
+struct FillToolData {
```
Keavon, line 80: We need to set the tolerance in the tool's control bar. It should range from 0 (only the identically colored pixels) to 1 (all pixels regardless of color). We should also have a checkbox for contiguous.

editor/src/messages/tool/tool_messages/fill_tool.rs
```diff
@@ -71,37 +76,111 @@ enum FillToolFsmState {
...
+	fills: Vec<Fill>,
+	start_pos: Vec<DVec2>,
+	tolerance: f64,
```
Keavon, line 83:
```suggestion
	tolerance: Vec<f64>,
```

Keavon (changes requested):

node-graph/gstd/src/raster.rs
EllenGYY: The tolerance is calculated by the euclidean distance in the CIELAB colorspace, which is a perspecively correct way of measuring color difference. Although value of A and B is autually boundaryless, softwares usually use -128 to 128, and the L has a range of 0 to 100. I will look into it and fix the bugs that cause the crash.

Keavon:
Checking in @EllenGYY.

EllenGYY:
@Keavon As the semester is approaching its final weeks and I have several project / research deadlines, I’ll revisit and improve this function’s behavior after my finals on May 1st.

Keavon:
Sounds good, thank you for the update.

PR #2523, opened 2025-04-06
Add the settings popover menu for the Overlays toggle

seam0s-dev:
Added options on the frontend to selectively disable/enable certain types of overlays.


seam0s-dev:
@Keavon Is it good to open now?

Keavon:
You'll need to first resolve the merge conflict. Then I can give it a test if you think it's ready for production, excepting for any feedback I find when testing that you didn't notice in your own (hopefully thorough testing).

Keavon:
Ok, in that case, please mark it as ready for review with a comment about that (which you just gave) so it turns from gray to green and I can look at it in my review queue.

editor/src/messages/portfolio/document/overlays/utility_types.rs
```diff
@@ -21,6 +21,108 @@ pub fn empty_provider() -> OverlayProvider {
 	|_| Message::NoOp
 }
 
+// Types of overlays used by DocumentMessage to enable/disable select group of overlays in the frontend
+#[derive(PartialEq, Clone, Debug, serde::Serialize, serde::Deserialize, specta::Type)]
+pub enum OverlaysType {
+	All,
```
Keavon, line 27: You can probably remove `All` and replace the whole `OverlaysType` with `Option<OverlaysType>` where `None` is equivalent to `All` being false and `Some` is equivalent to true.

Keavon: Upon further thought, that's not true because you'll lose your settings when it becomes `None`. I recommend just storing a separate bool outside of `OverlaysType`.

Keavon:
Bug: when Transform Cage is disabled, but dragging the compass rose by the X or Y arrows, we're incorrectly allowed to freely drag the object anywhere (not just along the constrained axis line).

seam0s-dev:
Could you give a demo on what you meant by the bug with Anchors?

Keavon:
Sure: if you have some anchors selected with the Path tool active, turning off then back on the anchors visibility causes them to all get deselected.

seam0s-dev:
Just wanted to confirm. Thanks.

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -588,13 +592,15 @@ impl Fsm for SelectToolFsmState {
 					}
 				}
 
-				if let Some(bounds) = bounds {
-					let bounding_box_manager = tool_data.bounding_box_manager.get_or_insert(BoundingBoxManager::default());
+				if overlay_context.visibility_settings.transform_cage() {
```
c-mateo, line 595: Overlays message has important side effects. For example, it modifies `tool_data.skew_edge`, `tool_data.bounding_box_manager` and `tool_data.compass_rose`.  This line breaks snapping to an axis by grabbing the compass rose. Make sure to modify only what's needed. In this case, the condition `overlay_context.visibility_settings.transform_cage()` should be checked only to render overlays or not, like this:
```rs
if overlay_context.visibility_settings.transform_cage() {
    bounding_box_manager.render_overlays(&mut overlay_context, true);
}
```

seam0s-dev: The functionality provided by the bounding box should also be disabled with the overlays such as rotating, resizing, skewing etc.

Keavon:
One last issue: the master checkbox has stopped working (presumably as a result of your refactor to the `All` enum variant).

Keavon (approved):

PR #2561, opened 2025-04-12
Fix inability to drag axis-aligned lines

### LINKED ISSUE #2549: It's impossible to drag axis-aligned lines
If a line is drawn perfectly horizontally or vertically (even if rotated afterward), the usual visual feedback you get for hovering over the line (within distance of its stroke thickness) doesn't work as it does for lines originally drawn at other angles, and clicking on the line to select it works only within a much narrower strip along its path. And while that very narrow area can be clicked to select the layer, it can't be dragged to move the layer; this causes a box selection to occur. See the video comparing the expected behavior of a line that was drawn with a tilt from that of a line drawn horizontally:

https://github.com/user-attachments/assets/8be530d0-73b2-4738-b5c4-1ebfea425c2b

A possibly related issue that might be a debugging clue:

https://github.com/user-attachments/assets/ca5bd7fb-4223-4f1d-8303-955f1febfea1

Bonus bug fix:

The cursor always shows rotation on axis-aligned lines but it actually does correctly scale or rotate based on proximity.

https://github.com/user-attachments/assets/13e16c1b-6cec-4129-b79a-46180e936e2b

mTvare6:
Fixes #2549


Keavon:
!build

Keavon (approved):

PR #2562, opened 2025-04-13
Restore auto-panning that #2443 had broken

### LINKED ISSUE #2504: Auto panning doesn't work (regression)
Auto panning doesn't work any more for any tools.

A regression introduced by #2443.

Before (working): https://40d7cf62.graphite.pages.dev/
After (broken): https://ba64e548.graphite.pages.dev/

c-mateo:
Fixes #2504 by restoring part of the code deleted in #2443

### DIFF
--- a/frontend/wasm/src/editor_api.rs
+++ b/frontend/wasm/src/editor_api.rs
@@ -153,7 +153,8 @@ impl EditorHandle {
 							handle.send_frontend_message_to_js(message);
 						}
 
-						// Currently needed for auto panning to work. Maybe there's a better way to do this
+						// Used by auto-panning, but this could possibly be refactored in the future, see:
+						// <https://github.com/GraphiteEditor/Graphite/pull/2562#discussion_r2041102786>
 						for message in editor.handle_message(BroadcastMessage::TriggerEvent(BroadcastEvent::AnimationFrame)) {
 							handle.send_frontend_message_to_js(message);
 						}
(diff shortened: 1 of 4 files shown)


frontend/wasm/src/editor_api.rs
```diff
@@ -152,6 +152,11 @@ impl EditorHandle {
 						for message in editor.handle_message(AnimationMessage::IncrementFrameCounter) {
 							handle.send_frontend_message_to_js(message);
 						}
+
+						// Currently needed for auto panning to work. Maybe there's a better way to do this
+						for message in editor.handle_message(BroadcastMessage::TriggerEvent(BroadcastEvent::AnimationFrame)) {
+							handle.send_frontend_message_to_js(message);
+						}
```
TrueDoctor, line 159: I think it would be better to dispatch both the broadcast event and the increment frame counter message from the CurrentTime input preprocessor message handler. Having these multiple invocations is not pretty. It also somewhat obfuscates that setting the current time now has a side effect (of changing the frame time).
I'm also wondering if it even makes sense to use the animation frame here at all for non render related editor timing. We could consider adding a LogicTick  broadcast event which is triggered on a set interval timer and seperate this completely from animation related functionality

Keavon: > I'm also wondering if it even makes sense to use the animation frame here at all for non render related editor timing. We could consider adding a LogicTick broadcast event which is triggered on a set interval timer and seperate this completely from animation related functionality

(To be clear, that would be a separate discussion for a later PR, this one is just focused on un-breaking the functionality without refactoring or rearchitecting it.)

TrueDoctor: I mean the code should work, I just don't like it, but I didn't like it before either so that is not a merge blocker

Keavon (approved):

Keavon:
!build

Keavon (approved):

Issue #2488, opened 2025-03-25
Testing gradient tool


0HyperCube:
- [x] Drawing on offset documents (#2481)
- [x] <kbd>Shift</kbd> to snap to 15 degrees (#2481)
- [x] Double click to insert a gradient stop (#2527)
- [x] Dragging the stop rearranges the ordering (#2565)
- [x] Dragging an endpoint sets the correct point in the viewport (#2564)
- [ ] Select and <kbd>delete</kbd> removes a stop, which is currently broken by [this PR](https://github.com/GraphiteEditor/Graphite/issues/2445) so this also needs to be fixed

PR #2564, opened 2025-04-13
Add Gradient tool test for dragging an endpoint setting the correct point in the viewport

Sidharth-Singh10:
Closes part of  #2488


editor/src/messages/tool/tool_messages/gradient_tool.rs
```diff
@@ -709,4 +709,65 @@ mod test_gradient {
 			positions
 		);
 	}
+
+	#[tokio::test]
+	async fn dragging_endpoint_sets_correct_point() {
```
0HyperCube, line 714: It might be useful to change the transform of the layer as well as to zoom the document. This actually confirms that it works in viewport space. You can do this with something like:
```rs
editor.handle_message(NavigationMessage::CanvasZoomSet { zoom_factor: 2.0 }).await;
editor.handle_message(GraphOperationMessage::TransformSet { ... }).await;
```

Sidharth-Singh10: done

Issue #2568, opened 2025-04-14
Scatter points does not work on shapes with holes


TrueDoctor:
When using the scatter points node on shapes with holes, it fills the holes twice instead of never:

![Image](https://github.com/user-attachments/assets/94d66728-0728-4171-b3fb-3badf0801645)

PR #2569, opened 2025-04-14
Fix Poisson-disk sampling with negative space from nested subpaths

TrueDoctor:
Previously all Subpaths were considered independently for the poisson disk sampling evaluation. We now check against all Subpaths which might contain the point to fix shapes with holes such as fonts with letters with holes in them

Before:

Image

After:
image

Closes: #2568


TrueDoctor:
Fixed: 
image

Keavon (approved):
If you rename `shapes` to `subpaths` and `shape_index` to `subpath_index`, you have my approval.

PR #2570, opened 2025-04-14
Fix Pen tool unable to combine multiple selected paths by extending their endpoints or mesh points

0SlowPoke0:
- Fixes: If multiple layers are selected, the Pen tool fails to extend a path by clicking one of its endpoints.
- Bonus - adds: Auto-panning when dragging wire in graph (https://discord.com/channels/731730685944922173/881073965047636018/1361099129950048446)


Keavon:
!build

Keavon:
Thanks, this correctly makes it possible to extend endpoints. However there's also seemingly a regression(?) where it can't be connected to an endpoint of the opposite layer.

Keavon:
After combining the two layers, we end up with:

capture

But then after closing the two endpoints, we erroneously end up with:

capture

Keavon:
https://github.com/user-attachments/assets/c76fb70d-9572-4378-a7ea-5562b993967a

Keavon:
!build

Keavon (approved):

Keavon:
!build

PR #2574, opened 2025-04-15
Basic fields

### LINKED ISSUE #2497: Fields MVP
Make the "Repeat" node call its predecessor (which provides the object to be repeated) once per repetition. In each repetition, it passes a different context (`OwnedContextImpl`), where the `varargs` has a different `index: usize` as `position: DVec2` per instance. Then we use the (deduplication with stable node IDs? or caching?) if these are equivalent (because the `index` and `position` aren't accessed in the upstream nodes used as a lambda fields). TODO: More description.

0HyperCube:
Closes #2497
example


TrueDoctor:
I'm wondering if we could make a footprint based version as well. That would be really cool as well

Keavon (approved):

Keavon:
Now, finally, the possibilities are really endless!

https://github.com/user-attachments/assets/45e32bfb-30d5-4028-8748-8066027f179b

https://github.com/user-attachments/assets/7009bdd6-7070-4222-bea9-1e04bccdb786

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -116,45 +116,105 @@ fn line(_: impl Ctx, _primary: (), #[default((0., -50.))] start: DVec2, #[defaul
...
+
+					// Connect to the point above (vertical connection)
+					push_segment(current_index.checked_sub(columns as usize));
+				}
+			}
+		}
+		GridType::Isometric => {
+			// Calculate isometric grid spacing based on angles
+			let tan_a = angle_a.to_radians().tan();
+			let tan_b = angle_b.to_radians().tan();
```
0HyperCube, line 182: What do the two angles represent? You just appear to take the mean average of the tan of the two angles. For example setting the angles to `(30, 40)` yields the same as `(40, 30)` and `(tan⁻¹(tan(30)+tan(40)), tan⁻¹(tan(30)+tan(40)))`.

node-graph/gcore/src/vector/generator_nodes.rs
```diff
@@ -116,45 +116,105 @@ fn line(_: impl Ctx, _primary: (), #[default((0., -50.))] start: DVec2, #[defaul
...
+
+#[node_macro::node(category("Vector: Shape"), properties("grid_properties"))]
+fn grid<T: GridSpacing>(
 	_: impl Ctx,
 	_primary: (),
-	#[min(0.)]
-	#[default(30)]
-	#[max(90.)]
-	angle_a: f64,
+	grid_type: GridType,
```
0HyperCube, line 137: Why have several completely different sets functionality in one node? I thought the idea was to have small and compostable nodes.

editor/src/messages/portfolio/document/node_graph/node_properties.rs
```diff
@@ -1513,6 +1570,52 @@ pub(crate) fn _gpu_map_properties(document_node: &DocumentNode, node_id: NodeId,
 	vec![LayoutGroup::Row { widgets: map }]
 }
 
+pub(crate) fn grid_properties(node_id: NodeId, context: &mut NodePropertiesContext) -> Vec<LayoutGroup> {
+	let grid_type_index = 1;
+	let spacing_index = 2;
+	let angles_index = 3;
+	let rows_index = 4;
+	let columns_index = 5;
```
0HyperCube, line 1578: Why hardcode the index instead of using the fun new system?

0HyperCube: For example `generator_nodes::isometric_grid::RowsInput::INDEX`.

node-graph/gcore/src/vector/generator_nodes.rs
Keavon: Good catch, that didn't correctly match the reference I was using for the isometric grid overlay/snapping feature. I'll work on a fix.

node-graph/gcore/src/vector/generator_nodes.rs
Keavon: We prefer to design rich nodes covering a whole logically related set of similar functionality with plenty of customizability in order to reduce the number of unique nodes that a user has to learn about and—when changing from one to another—delete and recreate the alternative version of. Small and composable nodes isn't a goal since that doesn't offer the best user experience compared to fewer, richly featured nodes.

editor/src/messages/portfolio/document/node_graph/node_properties.rs
Keavon: I wasn't aware of that system, but that's awesome to hear it exists now! There are plenty of nodes using indexes directly so this should be a refactor PR that changes them all over. Thanks for making me aware of it now.

node-graph/gcore/src/vector/generator_nodes.rs
TrueDoctor: I'm wondering if we could make it easier to build and maintain document network nodes for use cases like this. I do prefer to have smaller, more composable nodes which can be modified by the user without touching the code. But I guess that is a more general discussion

PR #2589, opened 2025-04-17
Auto-generate enum type widget boilerplate for radio buttons and dropdown menus

### LINKED ISSUE #2587: Ergonomic way to define enums for use by nodes
Right now it's a laborious process to add an enum to the `TaggedValue`. We need to auto-generate ways of turning an enum into a dropdown menu or radio button input widget. And we need to consider how users will be able to define their own once they can write Rust code to create custom nodes.

I discussed some possible approaches with @TrueDoctor a couple days ago. More thought will be required once we can begin tackling this.

kythyria:
(Mostly) resolves #2587. This refactors the UI for dropdowns and radio buttons based on built-in enums.

Dropdowns and radio buttons associated with node sockets can be produced via a builder API, with this API in turn getting its data from a proc macro that operates on enums. The particular shape should facilitate extending this to user-defined enums, to values that are not associated with a socket, and to menu entries (eg, the ones to add a boolean operation).


Keavon (approved):

TrueDoctor:
The macro code looks good

Issue #2483, opened 2025-03-25
Testing spline tool


0HyperCube:
Testcases:
- [ ] Works when the viewport is zoomed/panned/tilted
- [ ] Works when the artboard has an offset
- [ ] Works when the artboard has a transform before it
- [ ] Closed splines can be drawn by clicking on the start. The tool should stop drawing. The second derivative of the curves should be continuous.
- [ ] Continuing to draw from a previous spline
- [ ] Continuing drawing when the other spline is transformed and in a group (currently broken #2482)
- [ ] Continuing to draw from some other path

PR #2591, opened 2025-04-17
Add tests for Spline tool to support continuing a previously drawn spline

Sidharth-Singh10:
Closes part of  #2483


editor/src/messages/tool/tool_messages/spline_tool.rs
```diff
@@ -534,3 +534,117 @@ fn delete_preview(tool_data: &mut SplineToolData, responses: &mut VecDeque<Messa
...
+			let distance = (*actual - *expected).length();
+			assert!(
+				distance < 1e-10,
+				"Point at index {} doesn't match: expected {:?}, got {:?} (distance: {})",
+				i,
+				expected,
+				actual,
+				distance
+			);
+		}
```
0HyperCube, line 648: Perhaps a function that extracts the viewport positions of the points and asserts that they are equal to a target slice?

0HyperCube (commented):
Looks good so far.

Sidharth-Singh10:
@0HyperCube  I've refactored the redundant code across the three open PRs.
The `assert_point_positions` function is reusable and can be shared across tests in all tools.

Would it make sense to include this in `EditorTestUtils`, or do you suggest a more appropriate place for it?

0HyperCube (approved):
Looks good; thanks for simplifying the test code.
I think the `assert_point_positions` is fine in the spline tool as it is useful for that tool only.

PR #2602, opened 2025-04-20
Fix angle logic in the Grid node to allow slanted isometric grids

0HyperCube:
Address code review comments for #2574.


Keavon (approved):
Awesome, thank you!

PR #2609, opened 2025-04-22
 Fix greedy node deletion behavior

### LINKED ISSUE #2606: Node deletion is too greedy
Deleting the selected "Equals" node:

Image

Incorrectly results in the overzealous deletion of its upstream nodes despite their usage by other nodes.

Image

This doesn't match the desired deletion-with-cleanup behavior. <Kbd>Ctrl</kbd><kbd>Delete</kbd> can be used in this situation to delete only itself, but we need the normal behavior to work in this circumstance.

Keerthi421:
This PR fixes issue #2606 where node deletion was too aggressive and would delete upstream nodes that were still being used by other nodes.
Fixes #2606 

Changes made:
-->Added a check to skip the original node in upstream traversal when determining what nodes to delete
-->Added a comprehensive check to verify if an upstream node has connections to nodes outside the deletion set
-->Modified the deletion algorithm to only delete upstream nodes that are truly sole dependents

Testing:
-->This fix preserves upstream nodes that are connected to other nodes in the graph, while still properly cleaning up nodes that are truly only dependent on the deleted node. The specific case documented in issue #2606 no longer causes the incorrect deletion of upstream nodes.

Implementation details:
-->The key improvements are in the delete_nodes function:
-->Added a check to skip the original node in upstream traversal: if upstream_id == *node_id { continue; }
-->Added a check to verify connections to other nodes: if !delete_nodes.contains(downstream_id) && !nodes_to_delete.contains(downstream_id) { can_delete = false; }
-->Added early escape from the traversal: if !can_delete { break; }
-->These changes ensure that upstream nodes are only deleted if they are truly sole dependents of the node being deleted, preserving nodes that are still in use by other parts of the graph.

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -1370,7 +1370,7 @@ impl NodeNetworkInterface {
 			selected_nodes
 				.selected_layers(self.document_metadata())
 				.map(|layer| {
-					let mut layer_path = layer.ancestors(&self.document_metadata).collect::<Vec<_>>();
+					let mut layer_path = layer.ancestors(self.document_metadata()).collect::<Vec<_>>();
 					layer_path.reverse();
 					layer_path
 				})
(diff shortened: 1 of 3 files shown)


Keavon:
Closing as likely AI generated spam.

PR #2611, opened 2025-04-22
Refactor the 'Position on Path' and 'Tangent on Path' nodes to use the Kurbo API

indierusty:


node-graph/gcore/src/vector/misc.rs
```diff
@@ -111,3 +111,7 @@ pub fn point_to_dvec2(point: Point) -> DVec2 {
 pub fn dvec2_to_point(value: DVec2) -> Point {
 	Point { x: value.x, y: value.y }
 }
+
+pub fn daffine2_to_affine(value: DAffine2) -> Affine {
+	Affine::new(value.to_cols_array())
```
Keavon, line 116: This only gets used once, it's better to just do it inline.

indierusty: It will be used in tangent_on_path and probably at more places.

Keavon: It's still only a few characters longer than the function name. I don't think it's worth keeping a function definition for something that's already so tersely inlined.

indierusty: Got it. Should I make a commit for this?

indierusty: Fixed it.

Keavon:
!build

Keavon (approved):

PR #2612, opened 2025-04-22
Add Path tool feature for angle locking upon pressing Ctrl while dragging handle over anchor

0SlowPoke0:
Part of #1870 

> Path tool - when dragging a handle:
> - Ctrl locks the current angle of the handle (https://github.com/GraphiteEditor/Graphite/pull/2160).
>   - If pressed while the non-colinear handle is over its anchor, it becomes temporarily colinear and is locked to the angle of its opposite handle, or if there is no opposite handle then it should not become colinear but it should become locked to the angle of its [shared adjacent segment angles](https://files.keavon.com/-/NovelGruesomePangolin/capture.png)


Keavon:
!build

Keavon:
!build

Keavon:
!build

Keavon (approved):

PR #2622, opened 2025-04-29
- snap any point on line to tangents instead of snapping endpoint

moOsama76:
- remove fully_constrained as per 0hypercube on Discord

Closes #

### DIFF
--- a/editor/src/messages/tool/common_functionality/snapping/alignment_snapper.rs
+++ b/editor/src/messages/tool/common_functionality/snapping/alignment_snapper.rs
@@ -88,7 +88,6 @@ impl AlignmentSnapper {
 									distance,
 									tolerance,
 									distance_to_align_target,
-									fully_constrained: false,
 									at_intersection: true,
 									alignment_target_x: Some(endpoint),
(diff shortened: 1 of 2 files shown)


Keavon:
https://github.com/user-attachments/assets/7c7145a1-86c6-4925-89d0-a344ee471685

Keavon:
Unfortunately from my testing (with a line and an ellipse), this seems really unreliable—I can barely ever get it to engage at all—and it seems to break the existing tangent snapping most of the time. Can you reproduce my results?

moosama76:
> Unfortunately from my testing (with a line and an ellipse), this seems really unreliable—I can barely ever get it to engage at all—and it seems to break the existing tangent snapping most of the time. Can you reproduce my results?

I stopped the current snapping intentionally I can revert it,
The major UX problem is caused because snapping works only inside the bb of the snapped curve (so it's optimized when we implement rtree for layer snapper), I can make it work anywhere but I will have to iterate all tangent points in the document every time mouse moves

moosama76:
Military service, see you in 8 months

Issue #2352, opened 2025-03-02
Snapping improvements/revamp


Keavon:

pratheekhassan21:
hi@Keavon,I am intrested in working this issue.

Keavon:
@pratheekhassan21 this full issue corresponds to one of the GSoC project ideas ("Snapping system overhaul" under *[More ideas](https://graphite.rs/volunteer/guide/student-projects/#more-ideas)*) in its full form, but it's big so it probably wouldn't be the ideal place to start. There are some sub-parts to it, such as [this Discord discussion](https://discord.com/channels/731730685944922173/881073965047636018/1344560084839301140) for fixing a bug that's part of the snapping system. I'd recommend starting with that debugging.

pratheekhassan21:
thank you

j-jasz:
One snapping function that seems to be missing is the ability to snap to a grid when moving path points using the 'G' key

moosama76:
Hello 👋

PR #2626, opened 2025-04-30
Improve the distribution part of the snapping system

### LINKED ISSUE #2614: Distribution snappers are positioned based wrongly on mouse position instead of tracked object
https://github.com/user-attachments/assets/71df17f9-ffe3-4f96-81c8-407cecb58600

moosama76:
Part of #2352

- Fix distribution snapper visualization going far from layers (closes #2614)
- Fix bug where center point and corner point have to be enabled for distribution snapper to work
- Fix tests
- Improving naming conventions and comments


moOsama76:
I think this issue is more complex than it looks, any change in the variables the draw those visualizations will fails tests

Keavon:
If that's because you can identify that the tests encode incorrect behavior, please fix the tests alongside your change.

moOsama76:
> If that's because you can identify that the tests encode incorrect behavior, please fix the tests alongside your change.

Sure

Keavon (approved):
Thank you!

PR #2629, opened 2025-05-03
Refactor the 'Sample Points' node to use Kurbo instead of Bezier-rs

indierusty:


node-graph/gcore/src/vector/algorithms/bezpath_algorithms.rs
```diff
@@ -81,24 +81,28 @@ enum BezPathTValue {
...
+	let segment_count = bezpath.segments().count();
+	assert!(segment_count >= 1);
 
 	match t {
 		BezPathTValue::GlobalEuclidean(t) => {
-			let lengths = bezpath.segments().map(|bezier| bezier.perimeter(PERIMETER_ACCURACY)).collect::<Vec<f64>>();
-			let total_length: f64 = lengths.iter().sum();
+			let lengths = segments_length
+				.map(|segments_length| segments_length.to_vec())
+				.unwrap_or(bezpath.segments().map(|segment| segment.perimeter(PERIMETER_ACCURACY)).collect());
```
0HyperCube, line 92: Why allocate?

indierusty: Because the `global_euclidean_to_local_euclidean` function needed &[f64] and also to calculate total_length. I tried without coverting the segments length to vect but could not figure out a way to do that.

indierusty: Can I do this without allocation. I will try.

indierusty: To calculate the lengths of the segment when pre-computed it not provided then I will have to create Vec of f64s to store the lengths right?

Keavon:
!build

Keavon (approved):

PR #2639, opened 2025-05-10
Make Select tool hover overlay in 'shallow select' mode visualize the next deeper object

### LINKED ISSUE #2560: Shallow Select mode should also display hover overlays for what a click and double-click will select
Once #2392 is fixed, the Select tool's Shallow Select mode should correctly represent what clicking and double-clicking should do. Relative to the context of what is currently selected, single clicking should select a sibling (rather than anything that's a child of that sibling which the user clicks on). We want hovering to show a rectangular bounding box of what would get selected by clicking. Inside that is the item which will get selected by double-clicking (by "digging down" a level in the nesting). We also want that to show a bounding box at 50% opacity to indicate it would become selected by double-clicking. Instead of showing this 50% opacity bounding box, we should show the actual shape outline (the same one that we have right now on hover) at 50% opacity if double-clicking would select an object (rather than a group). And if that shape would get selected by single-clicking it, then we don't show any bounding boxes but we show the shape outline at 100% opacity just as we do today. In total, there will be up to three overlays shown, with only one at 100% opacity and up to two as bounding boxes.

mTvare6:
Fixes #2560

### DIFF
--- a/editor/src/messages/portfolio/document/overlays/utility_types.rs
+++ b/editor/src/messages/portfolio/document/overlays/utility_types.rs
@@ -140,8 +140,8 @@ impl core::hash::Hash for OverlayContext {
 }
 
 impl OverlayContext {
-	pub fn quad(&mut self, quad: Quad, color_fill: Option<&str>) {
-		self.dashed_polygon(&quad.0, color_fill, None, None, None);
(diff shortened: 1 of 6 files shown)


Keavon:
!build

Keavon (approved):

PR #2644, opened 2025-05-13
Implement clipping.

mTvare6:
#### TODO

- [ ] Document upgrade code


Keavon:
A few things about stroke align:

1. If I just open a new canvas and draw an ellipse, increase its stroke weight so t's more easy to see, and change the stroke align, we end up with it not seeming to work.
2. And when it's set to *Outside*, in addition to it not working, it also crops off the edges in a rectangle https://files.keavon.com/-/UnsightlyNovelCub/capture.png
3. If I instead go to Valley of Spires, it seems to work, although *Outside* mode makes it no longer have a fill https://files.keavon.com/-/OptimisticScrawnyAnura/capture_11_.mp4

Keavon (approved):

PR #2645, opened 2025-05-14
Fix method for appending bezpath to VectorData and the 'Scatter Points' node implementation

indierusty:
Fixes mistakes I made in 'append_bezpath()' method #2608 and 'Scatter Points' #2634 node refactoring.


indierusty:
Screenshot 2025-05-15 at 11 10 46 AM
Screenshot 2025-05-15 at 11 10 40 AM

indierusty:
Though refactoring 'Scatter Points' node changed look of the 'Red Dress' demo artwork. @Keavon

indierusty:
> That's unfortunate. It seems to have affected all the Red Dress scatter points layers, not just the one (which I assumed was due to numerical instability). This leads me to believe something may be being altered more fundamentally which you could avoid by debugging the PR's changes and finding what part of the logic has resulted in a different result.

I tried debugging but could not understand why it really changes that much. I will try to find the cause again but It would be helpful if you give a quick look to the changes and maybe point out the lines which could be causing this.

Keavon:
Just from my quick glance, some of the changes in [`bezpath_algorithms.rs`](https://github.com/GraphiteEditor/Graphite/pull/2645/files#diff-0efd7039126da8c2e475b8af0172e53415cd4120fd8d9763b986d78ced91f320) look like of sus for affecting the output.

indierusty:
I tried to do the point insideness checking as it was done before using kurbo and only fixing the bug in the 'Interative Snake Pebbles', which was due to incorrect insideness checking with bouding boxes, but yet the behaviour was same to the current behaviour.

indierusty:
Fixed in #2649 and #2657.

PR #2657, opened 2025-05-20
Fix 'Scatter Points' node artifacts

indierusty:
Fixes regression introduced in #2634.

Also fixes a bug where the Poisson algorithm dart board point checking wasn't offset from its dart board space to the layer's bounding box space when testing individual points (however, testing grid squares was correctly offset). This changes the appearance of the Red Dress demo artwork, which was updated with new seed values to preserve the original's general look.

Screenshot 2025-05-20 at 3 28 55 PM
Screenshot 2025-05-20 at 3 28 17 PM


indierusty:
Sorry, I've done too many mistakes in making commits. I will take care of it by now.

Keavon (approved):

PR #2659, opened 2025-05-20
Add basic thumbnail extraction to Rawkit

geom3trik:


libraries/rawkit/src/tiff/values.rs
```diff
@@ -51,29 +53,80 @@ impl CurveLookupTable {
...
+			Self::Horizontal | Self::MirrorHorizontal | Self::Rotate180 | Self::MirrorVertical => false,
+			Self::MirrorHorizontalRotate270 | Self::Rotate90 | Self::MirrorHorizontalRotate90 | Self::Rotate270 => true,
 		}
 	}
 }
+
+#[derive(Debug, PartialEq, Eq, Clone, Copy, IntoPrimitive, TryFromPrimitive)]
+#[repr(u16)]
+#[allow(non_camel_case_types)]
+pub enum CompressionValue {
```
0HyperCube, line 85: It might be nice to comment where these values have come from?

libraries/rawkit/src/lib.rs
```diff
@@ -138,13 +149,38 @@ impl RawImage {
...
 
 		Ok(raw_image)
 	}
 
+	/// Extracts the thumbnail image from the raw file.
+	pub fn extract_thumbnail<R: Read + Seek>(reader: &mut R) -> Result<ThumbnailImage, DecoderError> {
+		let mut file = TiffRead::new(reader)?;
+		let ifd = Ifd::new_first_ifd(&mut file)?;
+
+		// TODO: ARW files Store the thumbnail offset and length in the first IFD. Add support for other file types in the future.
```
0HyperCube, line 164: Would it be worthwhile to do an assert on `metadata::identify::identify_camera_model`?

0HyperCube (commented):
Thanks for the contribution.

The code looks good (although I don't really understand the RAW formats very well). From what I can tell the transform -> orientation rename is probably beneficial for clarity.

Keavon:
What's the latest status on this? Do the code review comments need to be actioned yet? Is it awaiting further code review? And I think I need to attend to the CI runner and update it so the failure is fixed.

capture

Keavon:
@geom3trik just a reminder about my previous message in this thread :)

PR #2660, opened 2025-05-21
Add molding segments to the Path tool

### LINKED ISSUE #615: Sculpting / Molding with Path Tool
Dragging on a bezier segment whilst in the path tool should mold it (as in https://pomax.github.io/bezierinfo/#molding ) as in inkscape.

Blocked on the path tool refactor.

4adex:
Part of #1870:
- Clicking and dragging without any modifier key, as per #615, should [mold the segment](https://pomax.github.io/bezierinfo/#molding).

Closes #615.


Keavon:
!build

Keavon:
- Right-click and <kbd>Escape</kbd> need to work to cancel molding
- <kbd>Alt</kbd> should temporarily break colinearity and mold only the dragged segment, while also snapping the handles back to where they began before the molding drag began
- <kbd>C</kbd> should permanently toggle colinearity on the two adjacent anchors, without snapping the handles back like above (be sure to handle the case where they're each initially of opposite state)
- Hints should be updated for all of the above
- We can merge this but then move onto finding a better algorithm that suits our desired molding behavior better
- Please don't put any new code into Bezier-rs, find somewhere reasonable in the editor code to put that. Thanks.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -274,6 +274,167 @@ impl ClosestSegment {
...
+
+		let modification_type = handle1.set_relative_position(nc1 - start);
+		responses.add(GraphOperationMessage::Vector { layer, modification_type });
+
+		let modification_type = handle2.set_relative_position(nc2 - end);
+		responses.add(GraphOperationMessage::Vector { layer, modification_type });
+
+		// If adjacent segments have colinear handles, their direction is changed but their handle lengths is preserved
+		// TODO: Find something which is more appropriate
+		if let Some(vector_data) = document.network_interface.compute_modified_vector(self.layer()) {
```
4adex, line 313: @Keavon the part of this function has some redundant code, so I want to break it into several helper functions, but am not sure where should I place those functions. Is `utility_functions.rs` a good place to have those? But these functions might not be needed anywhere else

Keavon:
!build

Keavon:
I should update my request about <kbd>C</kbd>: it should do the same that <kbd>Alt</kbd> does now, except on a toggle basis instead of a held bass like <kbd>Alt</kbd>. Tapping <kbd>C</kbd> two times is equivalent to pressing then releasing <kbd>Alt</kbd>. But if it starts as already not colinear on both ends, then <kbd>C</kbd> does nothing and shouldn't be shown in the hints bar.

Keavon:
- If you hover and then scroll with a scroll wheel or middle click drag to pan, the line that follows your cursor when hovering over a segment doesn't go away when your cursor goes out of range, until you next move your cursor by a pixel after panning. Please fix that to make things nice and polished.
- After testing it, I've realized that <kbd>C</kbd> adds more complexity than it's worth, so let's please remove it and just stick with <kbd>Alt</kbd>.
- When <kbd>Alt</kbd> isn't valid, please make sure the hints reflect that by only showing what's valid when it's valid.

Nearly ready to merge!

Keavon:
!build

PR #2661, opened 2025-05-22
Optimize Poisson-disk point sampling by halving the number of line intersection checks with caching

Keavon:
The performance impact comes almost entirely from the square-against-segment check, so this aims to reduce it through caching the four edge lines of the square so neighboring cells can reuse the result of the shared edge line.

I ran this on a circle and found the following results for the number of times that we run the hot path Kurbo function, `PathSeg::intersect_line`, inside our `ActiveListLevel::new_filled` function:

```
102,987 (After: With Caching)
203,505 (Before: No Caching)
```

After it builds the initial dart board grid of the user's chosen point density, we also have one other location where `intersect_line` is used, when the dart board is subdivided. This only happened about 8000 times during the same above case, and because it subdivides the grid, it's harder to use a hash map in that case and there would be significantly more cache misses, so I didn't attempt to cache that part.

Further optimization work is still needed to help with avoiding creating the high-density initial dart board across the full bounding box area, by breaking it apart into a spatial data structure so large areas inside and outside the shape can be checked with fewer tests for insideness and intersection. We want to efficiently discard large areas that are fully inside and fully outside while only doing denser geometry boundary checks near the actual vector segments.

Despite the 50% reduction in the bottlenecked calls, we seem to only be reducing our AnimationFrame duration from 170 to 150 milliseconds. Besides the possibility of it getting faster over time due to the branch predictor or other optimizations that may kick in over time, I'm really not sure why a 50% reduction in the number of calls to the hot path function are only producing a 15% reduction in run time. Either way, it's a speedup so I'll take it.

### DIFF
--- a/node-graph/gcore/src/vector/algorithms/poisson_disk.rs
+++ b/node-graph/gcore/src/vector/algorithms/poisson_disk.rs
@@ -1,5 +1,6 @@
 use core::f64;
 use glam::DVec2;
+use std::collections::HashMap;
 
 const DEEPEST_SUBDIVISION_LEVEL_BEFORE_DISCARDING: usize = 8;
 
@@ -8,11 +9,12 @@ const DEEPEST_SUBDIVISION_LEVEL_BEFORE_DISCARDING: usize = 8;
 /// "Poisson Disk Point Sets by Hierarchical Dart Throwing"
 /// <https://scholarsarchive.byu.edu/facpub/237/>
 pub fn poisson_disk_sample(
+	offset: DVec2,
 	width: f64,
 	height: f64,
 	diameter: f64,
 	point_in_shape_checker: impl Fn(DVec2) -> bool,
-	square_edges_intersect_shape_checker: impl Fn(DVec2, f64) -> bool,
+	line_intersect_shape_checker: impl Fn((f64, f64), (f64, f64)) -> bool,
 	rng: impl FnMut() -> f64,
 ) -> Vec<DVec2> {
 	let mut rng = rng;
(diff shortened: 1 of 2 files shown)


node-graph/gcore/src/vector/algorithms/poisson_disk.rs
```diff
@@ -209,23 +222,64 @@ impl ActiveListLevel {
...
-				let square_edges_intersect_shape = square_edges_intersect_shape_checker(corner, square_size);
-				let square_not_outside_shape = point_in_shape || square_edges_intersect_shape;
-				let square_in_shape = !square_edges_intersect_shape && point_in_shape_checker(corner + square_size);
-				// Sometimes this fails so it is necessary to also check the bottom right corner.
-				square_not_outside_shape.then_some(ActiveSquare::new(corner, square_in_shape))
+				let corner = DVec2::new(x as f64 * square_size, y as f64 * square_size);
+				let corner_with_offset = corner + offset;
+
+				// Lazily check (and cache) if the square's edges intersect the shape, which is an expensive operation
+				let mut square_edges_intersect_shape_value = None;
```
indierusty, line 247: I think there is no need for this option value in the actual logic. Is this for any specific reason?

indierusty:
The animation frame is not reduced because we only compute if the square edges intersect with the path if the point is not in the shape.

node-graph/gcore/src/vector/algorithms/poisson_disk.rs
Keavon: It's used to indicate that the value hasn't been evaluated yet as part of the memoization logic.

indierusty: I think it could confuse developers and not using that is still fine to understand that we cache the value. Otherwise I think this is very good optimization and very well implemented.

Keavon: The other option besides memoizing the value is to not do the check lazily, or do it potentially twice, which in some cases may result in more hashmap lookups (which are constant time and cheap but not free). What did you have in mind?

indierusty: Memoizing is good as for the first time we do not have to lookup into the hashmap which is slightly efficient. So current approach will work good.

Issue #2635, opened 2025-05-07
Tracking Issue: Refactor nodes to use Kurbo instead of Bezier-rs


indierusty:
- [x] 'Offset Path' node (#2596, #2946)
- [x] 'Solidify Stroke' node (#2608)
- [x] 'Position on Path' and 'Tangent on Path' nodes (#2611)
- [x] 'Sample Points' node (#2629)
- [x] 'Scatter Points' node (#2634, #2657)
- [x] 'Bounding Box' node (#2662)
- [x] 'Morph' node (#2696)
- [x] 'Spline' node (#2701)
- [x] 'Bevel' node (#2293)
- [x] 'Area' node (#2293)
- [ ] 'Centroid' node
- [ ] 'Round Corners' node
- [ ] 'Auto-Tangents' node
- [ ] Refactor `ClickTarget` to use Kurbo
- [ ] Refactor tools code to use Kurbo's structs and functions

PR #2662, opened 2025-05-22
Refactor the 'Bounding Box' node to use Kurbo instead of Bezier-rs

indierusty:
Part of #2635.

Now the Bounding Box node uses Kurbo instead of Bezier-rs.

Also: Refactor to use kurbo's predefined default accuracy constant.


Keavon (approved):

0HyperCube:
@Keavon @indierusty I'm really confused about why you did this? It seems to significantly reduce performance since a lot of unnecessary allocations occur all of the time.

indierusty:
I refactored the function for calculating the bounding box specifically for refactoring 'Bounding Box' node implementation. We can split the function in two where first calculate tight bbox and second losse bbox that can be used by the renderer.

PR #2663, opened 2025-05-22
Implement angle locking when Ctrl is pressed over an adjacent anchor

0SlowPoke0:
Part of #1870

> If pressed while over the an adjacent anchor, it locks to handle dragging along its tangent line [(video)](https://files.keavon.com/-/InferiorIllinformedSableantelope/capture_25_.mp4), except if that anchor has no handles then it has no tangent; if it has split (non-colinear) handles then it uses the one on the side adjacent to the handle being dragged. ([source](https://github.com/GraphiteEditor/Graphite/pull/2663))


Keavon:
It's failing to detect the correct handle in this case:

https://github.com/user-attachments/assets/d54c901c-8294-45d7-9c7b-20da360dad0c

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -726,18 +727,39 @@ impl PathToolData {
...
 			.selected_shape_state
 			.iter()
 			.next()
-			.and_then(|(layer, _)| document.network_interface.compute_modified_vector(*layer))
+			.and_then(|(layer, _)| document.network_interface.compute_modified_vector(*layer).map(|vector_data| (vector_data, layer)))
 		{
+			let adjacent_anchor = check_handle_over_adjacent_anchor(handle_id, &vector_data);
+			let mut required_angle = None;
+
+			// If the handle is dragged near its adjacent anchors while holding down the Ctrl key,compute the angle based on the tangent formed with the neighboring anchor points.
```
Keavon, line 739:
```suggestion
			// If the handle is dragged over one of its adjacent anchors while holding down the Ctrl key, compute the angle based on the tangent formed with the neighboring anchor points.
```

PR #2668, opened 2025-05-24
Make the Path tool only allow selecting points that are visible

### LINKED ISSUE #2261: Reduce path editing clutter with options to hide non-selected anchors' handles

4adex:
- Only points which are visible under current `PathOverlayMode` should be selected when clicked upon.

Followup for #2264 which closes #2261.

### DIFF
--- a/editor/src/messages/tool/tool_messages/path_tool.rs
+++ b/editor/src/messages/tool/tool_messages/path_tool.rs
@@ -1108,7 +1108,7 @@ impl Fsm for PathToolFsmState {
 									if attached_segments.len() == 1 {
 										segment_endpoints.entry(attached_segments[0]).or_default().push(point);
 									}
-									// This is for the edge case of a loop where a point rather not selected can be part of two segments
+									// Handle the edge case where a point, although not explicitly selected, is shared by two segments.
 									else if !selected_anchors.contains(&point) {
 										segment_endpoints.entry(attached_segments[0]).or_default().push(point);
 										segment_endpoints.entry(attached_segments[1]).or_default().push(point);


editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1090,13 +1107,18 @@ impl Fsm for PathToolFsmState {
 								for (point, attached_segments) in selected_segments_by_point {
 									if attached_segments.len() == 1 {
 										segment_endpoints.entry(attached_segments[0]).or_default().push(point);
-									} else if !selected_anchors.contains(&point) {
+									}
+									// This is for the edge case of a loop where a point rather not selected can be part of two segments
```
Keavon, line 1111: Please fix the grammar of this sentence, I can't make sense of it currently.

4adex: Apologies for the vagueness, fixed it.

PR #2673, opened 2025-05-26
New nodes: 'Point Inside Shape' and 'Close Path'

0SlowPoke0:
New nodes:

- Point Inside Shape - returns true or false if a given DVec2 point is inside a given vector shape. [discord](https://discord.com/channels/731730685944922173/881073965047636018/1362712297340145715)
- Close Path - takes a vector path and, if its shape is open, it adds a straight segment to close it. Equivalent outcome as pressing 
Ctrl+J in the Path tool. [discord](https://discord.com/channels/731730685944922173/881073965047636018/1362712589192396852)

Also:
- Dragging a node into the vertical wire of a layer stack should no longer convert the node into a layer-type node and insert it into the stack. This ends up being a footgun far more often than it's useful. [discord](https://discord.com/channels/731730685944922173/881073965047636018/1367682239042682963)

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -1715,27 +1715,21 @@ fn bevel(_: impl Ctx, source: VectorDataTable, #[default(10.)] distance: Length)
 	result
 }
 
-#[node_macro::node(name("Close Path"), category("Vector"), path(graphene_core::vector))]
+#[node_macro::node(category("Vector"), path(graphene_core::vector))]
 fn close_path(_: impl Ctx, source: VectorDataTable) -> VectorDataTable {
 	let mut new_table = VectorDataTable::empty();
+
 	for mut source_instance in source.instance_iter() {
 		source_instance.instance.close_subpaths();
-		new_table.push(Instance {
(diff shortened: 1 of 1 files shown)


Keavon:
https://github.com/user-attachments/assets/03593bec-8399-408a-83b4-50281328eea8

node-graph/gcore/src/vector/vector_nodes.rs
```diff
@@ -1715,6 +1715,29 @@ fn bevel(_: impl Ctx, source: VectorDataTable, #[default(10.)] distance: Length)
...
+fn close_path(_: impl Ctx, source: VectorDataTable) -> VectorDataTable {
+	let mut new_table = VectorDataTable::empty();
+	for mut source_instance in source.instance_iter() {
+		source_instance.instance.close_subpaths();
+		new_table.push(Instance {
+			instance: source_instance.instance,
+			transform: source_instance.transform,
+			alpha_blending: source_instance.alpha_blending,
+			source_node_id: source_instance.source_node_id,
+		})
```
Keavon, line 1728: You don't need to construct a new instance composed of all the same fields as the current instance. Just push the current instance (`source_instance`).

PR #2678, opened 2025-05-28
Fix #[range] to act as a soft min/max

mTvare6:
Follow up to #2464


Keavon:
We still need to cap it in the frontend component because of dragging the value, which needs to bottom out and be instantly reversible by changing the mouse direction instead of having to travel the full distance it wasn't moving beyond its backend-enforced cap. I'll have to do some thinking about this.

Keavon:
This current solution disregards the backend's hard max and lets the frontend display a false state until its layout gets re-sent:

https://github.com/user-attachments/assets/b989ebad-fab4-4697-8fa9-75be1852f997

Keavon:
@otdavies requested in a chat with me:

> also, I want auto-tangents to also go negative

I believe doing that can't happen till this PR is done.

PR #2683, opened 2025-06-01
Add star/polygon gizmos and refactor the separate shape drawing tools into a unified Shape tool

0SlowPoke0:
Replaced the existing Line, Rectangle, Ellipse, and Polygon tools with a unified Shape Tool that supports the creation of all these shapes from a single interface. The old Line, Rectangle, and Ellipse tools have been aliased to use the new Shape Tool. In addition to shape creation, the new tool also supports modifying the shapes it creates.


Keavon:
Issues to address in followup work:

- After drawing any shape (except Line, apparently), or even after just hovering over the transform cage to change the cursor appearance, the crosshair cursor goes back to the regular mouse pointer. It doesn't return to the crosshair like it's supposed to indicating it will draw the next shape. (A regular mouse pointer, not a crosshair, is desired over the point radius handle gizmos.)
- Dragging a polygon or star's radius point gizmos happens at an accelerated/slowed pace rather than staying with your cursor as you drag, if that shape has a Transform node scale other than 1x.

PR #2690, opened 2025-06-05
GPU invert node

Firestar99:
image

Code below shows how much I had to change the gpu invert node. Most of the changes are due to not wanting to depend on `gcore`, as it currently doesn't compile for `no_std`. I think we could probably get away without changing it at all.
https://github.com/GraphiteEditor/Graphite/blob/bf2673428ae3260e4b55204bc930ab01ed8591d3/node-graph/gcore-shader/src/gpu_invert.rs#L4-L23

The rust-gpu shader calling it:
https://github.com/GraphiteEditor/Graphite/blob/bf2673428ae3260e4b55204bc930ab01ed8591d3/node-graph/gcore-shader/src/gpu_invert.rs#L33-L38

### DIFF
--- a/.github/workflows/build-dev-and-ci.yml
+++ b/.github/workflows/build-dev-and-ci.yml
@@ -5,8 +5,7 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
+
 env:
   CARGO_TERM_COLOR: always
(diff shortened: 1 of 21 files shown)


Firestar99:
The linker is erroring due to the wasm-specific flags somehow ending up in a native compile, probably env leaking again. Doesn't trigger locally, likely due to me using the mold linker. Will play around with it a bit more tomorrow.

Keavon:
Ready to be rebased on master now.

Firestar99:
Superseded by https://github.com/GraphiteEditor/Graphite/pull/3097 and https://github.com/GraphiteEditor/Graphite/pull/2985

PR #2692, opened 2025-06-05
Add Pen tool support for starting and ending segment drawing on existing path edges for vector meshes

4adex:
Described in detail here - 
image


Keavon:
UX concern: when we're snapping to an anchor, we also appear to be showing the new insertion point underneath the cursor which incorrectly and confusingly implies that we'd insert a point where, when in reality we are (correctly) snapping to the nearby anchor:

https://github.com/user-attachments/assets/99c05e8a-c83f-45c3-8f7b-eedb8e2f3d25

PR #2697, opened 2025-06-07
Instance tables refactor part 8: Make repeater nodes use pivot not bbox and output instance type not group; rename 'Flatten Vector Elements' to 'Flatten Path' and add 'Flatten Vector'

Keavon:
Part of #1834.

Nodes that now move the local origin instead of bounding box center, and output the instanced type rather than always being wrapped in a `GraphicGroupTable`:
- *Instance On Points*
- *Instance Repeat*
- *Repeat*
- *Circular Repeat*
- *Copy To Points*

TODO: Still needs to have the above description performed, but it involves challenging transforms I wasn't able to figure out:
- *Mirror*

Nodes that output `VectorDataTable` but now also support `VectorDataTable` as input:
- *Flatten Vector Elements* -> *Flatten Path* (**renamed** — this produces one vector instance row containing all the original geometry as subpaths)
- *Flatten Vector* (**new** — creates one `VectorDataTable` with many rows)


Keavon:
Testing:

$${\color{red}Red}$$

PR #2712, opened 2025-06-13
Add segment editing mode to the Path tool

### LINKED ISSUE #696: Points/segments/shapes vector selection mode
Feature idea: selecting with the path tool as either points or segments, equivalent to Blender's vector/edge/face modes.

4adex:
Closes #696. Resolves part 1 of #2716.


otdavies:
https://github.com/user-attachments/assets/70810b52-3f89-4cbc-bf35-cd64ee5b2930

This should only allow selecting both an edge and a point when I am holding shift. I wasn't holding shift here. (no modifier keys at all)

Awesome feature btw, good progress!

4adex:
@otdavies Yes I am working upon it, will push the latest code with some other additions.

Keavon:
Awesome start, this is great to have!

QA improvements for a follow-up PR:

- Disabling all modes shouldn't be possible.
- We want it to act more like radio buttons but where Shift can multi-select, like in Blender (but I may need to figure out a more refined design for radios vs. checkboxes when the lines between the two are blurred as they are in this case).
- Switching from segment mode to point mode should make the points connected to each segment become selected, like in Blender, and vice versa where two neighboring points cause their shared segment to become selected. And in segment only mode, selecting two adjacent segments should automatically select their shared point (but not in segment + point mode).
- Didn't we discuss a hotkey (maybe Ctrl or Alt?) to let you insert a point on a segment even if you're in segment selection mode? That should also be the same hotkey for molding when you drag instead of click, instead of currently using <kbd>A</kbd> which isn't a standard modifier key.
  - Unrelated feedback for molding: it needs to work on straight line segments, currently this doesn't work. When a segment is a straight line (both its handles are 0-length or not present because it's quadratic or linear), it should generate handles that are 1/3 the length of the straight line segment and drag those (so there is some curvature to the bend).
- Now that we have the hover highlight over segments before clicking to select them, we should have the same for anchors— that can take the form of a 1px (in all directions) rectangle with the same semitransparent blue used for the pre-selection segment hover.
- I love the dynamic contextual input hints, and it would be good to have that also happen when hovering over an anchor, not just a segment. This should also happen when currently targeted by a box or lasso selection (just like in Select mode which already does this).

The above is now tracked in #2832.

PR #2722, opened 2025-06-16
Remove old gpu nodes

Firestar99:


TrueDoctor:
I think it is unfortunate to lose so much progress but I can understand that you want to start from a clean slate and we do still have a git history of the changes. The changes themself seem reasonable

PR #2724, opened 2025-06-16
Make `gcore` a std library, remove `std`, `alloc` and more unused features

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/2729

This PR contains the code changes, for bulk renames see https://github.com/GraphiteEditor/Graphite/pull/2735

Features removed:
* `std`, `alloc` to make it std-only
  * `std` depends on `reflections`, `spectra`, `rustybuzz`, `image` and `ctor`
  * `alloc` depends on `dyn-any` and `bezier_rs`
* `gpu` was for rust-gpu compiling
* `log`, `rand_chacha` and `kurbo` are required for it to even compile


node-graph/gcore/src/types.rs
```diff
@@ -144,7 +141,7 @@ fn migrate_type_descriptor_names<'de, D: serde::Deserializer<'de>>(deserializer:
 	let name = String::deserialize(deserializer)?;
 	let name = match name.as_str() {
 		"f32" => "f64".to_string(),
-		"graphene_core::transform::Footprint" => "core::option::Option<alloc::sync::Arc<graphene_core::context::OwnedContextImpl>>".to_string(),
+		"graphene_core::transform::Footprint" => "std::option::Option<std::sync::Arc<graphene_core::context::OwnedContextImpl>>".to_string(),
```
Firestar99: I'm unsure whether this will cause breakage or not. While `core` would still compile, `alloc` shouldn't since I removed the `external crate alloc;`. Even if this particular migration code may be fine to be changed, it could indicate trouble with other documents in the wild. At least the demo artworks are unaffected.

TrueDoctor: This only affects the debug output, a breakage here is fine.

PR #2729, opened 2025-06-17
bulk rename `gcore` -> `gstd` in `editor`

Firestar99:
```rust
/// Constructs a raster image.
#[node_macro::node(category(""))]
fn image(_: impl Ctx, _primary: (), image: RasterDataTable<Color>) -> RasterDataTable<Color> {
    image
}
```
> I'm adding reexports, and the node macro generated `mod image` is conflicting with reexporting `gcore::raster::image` in `gstd`. May have to rename it, not sure to what.


TrueDoctor:
> I'm adding reexports, and the node macro generated mod image is conflicting with reexporting gcore::raster::image in gstd. May have to rename it, not sure to what.

Can you rexport `gcore::raster::*` from inside `gstd::raster`?

TrueDoctor:
And you could rename the gcore::raster module to something else to avoid coflicts. Then we don't have to rename anything in editor

node-graph/gstd/src/raster.rs
```diff
@@ -3,10 +3,10 @@ use fastnoise_lite;
 use glam::{DAffine2, DVec2, Vec2};
 use graphene_core::instances::Instance;
 use graphene_core::raster::bbox::Bbox;
-use graphene_core::raster::image::{Image, RasterDataTable};
-use graphene_core::raster::{Alpha, AlphaMut, Bitmap, BitmapMut, CellularDistanceFunction, CellularReturnType, Channel, DomainWarpType, FractalType, LinearChannel, Luminance, NoiseType, RGBMut};
+use graphene_core::raster::image::RasterDataTable;
+pub use graphene_core::raster::*;
```
Firestar99: @TrueDoctor that's the reexport I added. But since this mod already contains a `mod image` through the node macro, that mod takes priority over the glob reexport. And I can't insert a glob reexport into the image mod itself, since it's fully contained within the node macro's output.

Firestar99:
> And you could rename the gcore::raster module to something else to avoid coflicts. Then we don't have to rename anything in editor

This PR makes `editor` to only depend on `gstd` and removes it's dependency on `gcore`. Instead, all of `gcore` is reexported in `gstd`. This allows me to rip apart `gcore` without `editor` being affected, as long as I continue to reexport all nodes in `gstd`. Saving us a lot of conflicts down the line.

TrueDoctor (approved):

Issue #2375, opened 2025-03-07
Make it possible to snap the pivot gizmo when dragging it


Keavon:
Soft blocked on #2225.

PR #2730, opened 2025-06-18
Add pivot type selection with Custom Pivot, Origin (Average Point), and Origin (Active Object) to the Select tool

### LINKED ISSUE #2225: Visualize the separate concepts of tooling pivot and graphical data origin
There will be two different dot gizmos, which the Select tool's control bar shall display a dropdown to choose from. A checkbox beside these controls enables or disables the whole feature, and when disabled it uses the center of the current selection's bounding box. The reference point input widget that's presently in the control bar will become specific to the pivot mode— but when the feature's checkbox is unticked to use the bounding box center, we will show none of the 9 selector boxes as active while still letting the user click one to enable pivot mode. This would be grayed out entirely in origin mode. These same controls described above will be added to the Path tool's control bar, as well.

Currently, the pivot is used by the tooling. But that will adapt to the chosen mode once this issue is implemented, so the chosen mode's center of rotation/scaling is used by the following forms of tooling:

- G/R/S
- Resizing the corners of the transform cage when <kbd>Alt</kbd> is pressed. (Should transform cage rotation also use the bounding box center by default unless <kbd>Alt</kbd> is pressed?)

Albeit with some differences like our choice of labeling, this feature is sort of analogous to Blender's "Transform Pivot Point" menu:

Image

- **Bounding Box Center**: should behave the same in Graphite as Blender; our equivalent occurs when the mode is disabled so this won't be part of the dropdown menu; this acts equivalently to the **centered** pivot as it works today in master
- **3D Cursor**: equivalent to our pivot but it behaves differently than in Blender; this shouldn't be touched from the behavior already in master
- **Individual Origins**: should behave the same in Graphite as Blender; this is one of the origin's modes; each selected layer transforms about its own origin position
- **Median Point**: should behave the same in Graphite as Blender; this is one of the origin's modes; each selected layer transforms about the average position of their shared origins
- **Active Element**: should behave the same in Graphite as Blender; this is one of the origin's modes; transforms all the selected layers about the origin of the most recently selected layer (stretch goal, possible future follow-up work)

| | |
|:-:|-|
| <h3>Pivot</h3>Image | <p>This is the crosshair symbol. It's the temporary/transient pivot. Equivalent to Blender's 3D cursor, but isn't global to the document nor is it as persistent as the 3D cursor. It's used only for tooling, not part of the node graph data or document format. This is our current implementation, but it will need to be tweaked.</p><p><ul><li>Currently, we store the pivot permanently on each layer, but this can lead to confusion if the user comes back later and finds it in an unexpected location (especially since scaling can affect it in weird ways). This also makes the code more difficult because we have to save tooling-level data into the serialized document. Remove the per-layer, persistent storage.</li><li>Make it so each layer that the user clicks will have the pivot set to its corresponding corner/midpoint/center of the reference point input widget (located in the tool's control bar). Switching from one layer to the next keeps that chosen position of the reference point input widget.</li><li>The user can drag the pivot crosshair gizmo anywhere in the canvas, disconnecting it from the chosen position in the reference point input widget. When disconnected, clicking away from the current layer (by deselecting it, or by selecting another layer instead; however **not** by partially extending/subtracting the existing selection) resets it, going back to the last selected position of the 9-box widget.</li><li>But the user can "pin" the dragged location:<br /><br />Then, no matter what selection changes occur, that remains the location of the pivot until unpinned or until clicking any of the 9 boxes in the reference point input widget again which unpins it. Perhaps double-clicking the crosshair gizmo can pin/unpin.</li></ul></p> |
| | |
| <h3>Origin</h3>Image | <p>This is the dowel pin symbol. It's the origin (`(0, 0)` point) of the vector data's internal coordinate space. This is used when transforming in the *Transform* node's properties panel parameters, or when using the *Copy to Points* node. Its drag will be stored in the "Origin: Offset" parameter of the *Transform* node (but for now, we don't have that yet, so the origin is visualized but not draggable for now).</p><p></p><p>(This menu's scale and rotation parameters should also offer a way to edit those values factoring in the current pivot choice which would make it update its location together with the typed value for rotation or scale such that it treats the change as occurring with that temporary pivot.)</p> |

mTvare6:
Closes #2225


editor/src/consts.rs
```diff
@@ -61,6 +61,7 @@ pub const SELECTION_DRAG_ANGLE: f64 = 90.;
 pub const PIVOT_CROSSHAIR_THICKNESS: f64 = 1.;
 pub const PIVOT_CROSSHAIR_LENGTH: f64 = 9.;
 pub const PIVOT_DIAMETER: f64 = 5.;
+pub const DOWEL_PIN_RADIUS: f64 = 5.;
```
Keavon: Diameter or radius?

mTvare6: It's actually the radius here, we could change the values later

Keavon: Is the PIVOT_DIAMETER actually a radius on the line above, then? Or do they just happen to be twice/half the size of one another?

mTvare6: They happen to be twice/half of each other, I just it looked larger temporarily, we can later decide the final value

Keavon (approved):

Keavon:
Postponed to a followup PR:
- Path tool
- Origins (Individual) mode
- Draggable origins with the origin offset Transform node parameter
- Maybe: migrations for the removal of the internal Transform node's hidden parameter that was used to track the old per-layer stateful pivot reference point
- #2375

PR #2735, opened 2025-06-18
Bulk `gcore` cleanup, replace `core` and `alloc` with `std`

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/2724

This PR contains the bulk renames, for code changes see https://github.com/GraphiteEditor/Graphite/pull/2724

### DIFF
--- a/node-graph/gcore/src/context.rs
+++ b/node-graph/gcore/src/context.rs
@@ -1,7 +1,7 @@
 use crate::transform::Footprint;
-use core::any::Any;
-use core::borrow::Borrow;
-use core::panic::Location;
+use std::any::Any;
+use std::borrow::Borrow;
+use std::panic::Location;
 use std::sync::Arc;
 
 pub trait Ctx: Clone + Send {}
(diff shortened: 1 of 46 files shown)


node-graph/gcore/src/context.rs
```diff
@@ -348,7 +348,7 @@ impl OwnedContextImpl {
 
 #[derive(Default, Clone, Copy, dyn_any::DynAny)]
 pub struct ContextImpl<'a> {
-	pub(crate) footprint: Option<&'a crate::transform::Footprint>,
+	pub(crate) footprint: Option<&'a Footprint>,
 	varargs: Option<&'a [DynRef<'a>]>,
 	// This could be converted into a single enum to save extra bytes
```
TrueDoctor, line 353: Only the owned context should be moved out of the context file, the rest should if possible remain no std

Firestar99: Your comment confuses me. This PR doesn't move anything, it just changes the paths to symbols for things within the crate, to make all the absolute paths and use statements use `std` instead of `core` or `alloc`. I'll start extracting a `no_std` crate `gcore_types` out of the `gcore` crate in some later PR, yet to be submitted. And can change all the symbols back to core, as required.

TrueDoctor: This is basically a future note to you that you will need to split this file into two versions, one which uses std (and containts the `OwnedContextImpl`) and a second one which contains the rest so you'll have to change the paths you've converted to `std` in this file for this  pr back to `core`

Firestar99: I thought it was a change request and was confused at what you wanted to have changed. All good if it was just a note :D

TrueDoctor (approved):

PR #2746, opened 2025-06-23
Prep `gcore` splitup: move various symbols into their own modules

Firestar99:
Moving some symbols within `gcore` to their own modules, to allow them to remain within `gcore` and move the rest into separate crates in a later PR.


node-graph/gcore/src/color/mod.rs
```diff
@@ -0,0 +1,7 @@
+mod color;
+mod discrete_srgb;
+mod raster;
```
TrueDoctor: Why did you choose to call this module raster? I would have expected something like `color_traits`

Firestar99: Cause those symbols were previously in `gcore::raster`, let's change that :D

TrueDoctor: Alright, I'm giving you an approval so you can fix this and then merge on your own so you don't have to wait for me

Firestar99:
Removed commit "move nodes from `ops` into `math_nodes` and `debug`" so I can move `math_nodes` directly into a new crate in a later PR. 

This PR can be merged.

PR #2747, opened 2025-06-23
fix colinear handles on connected path

HWienhold:
fixes https://discord.com/channels/731730685944922173/881073965047636018/1353015950894698606

Fixes a bug when connecting two paths ending with single handle (each) and applying `colinear handles`. on the newly connected nodes. Problem arises because applying `colinear handles` on the first node will create the opposite handle. The other node then has two handles, one of which is the anchor itself, hence:

```
let mut handle_directions = handle_positions.map(|handle| handle.map(|handle| handle - anchor));
```

will contain a null-Vec.

To be fair: this slightly changes the behavior when applying `colinear handles` to other nodes that only has one handle: while before it would change that node - now it will calculate the handles based on the anchors as well.


Keavon:
!build

PR #2764, opened 2025-06-27
Extract `gmath_nodes` from `gcore::ops`, create mods `debug` and `extract_xy`

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/2760

Split out from https://github.com/GraphiteEditor/Graphite/pull/2746#issuecomment-3012398175


node-graph/gmath-nodes/src/lib.rs
```diff
@@ -0,0 +1,532 @@
...
+fn logical_not(_: impl Ctx, input: bool) -> bool {
+	!input
+}
+
+/// Constructs a bool value which may be set to true or false.
+#[node_macro::node(category("Value"))]
+fn bool_value(_: impl Ctx, _primary: (), #[name("Bool")] bool_value: bool) -> bool {
+	bool_value
+}
+
```
TrueDoctor, line 416: I'm wondering if it would make more sense to move these to the logic mod now that the mod is no longer called ops

Firestar99: We can split up the lib into smaller mods, though I'd create a new logic op and not put it into the old, since the old is almost exclusively string manipulation except for SwitchNode (which I may move here)

TrueDoctor: Alright, just wanted to make you aware of this, we can always move things later once we are a bit more settled into the new structure

PR #2765, opened 2025-06-27
Fix Path tool multi-point handle scaling behavior and double-click smooth/sharp conversion on rectangles

0SlowPoke0:
- fixed while scaling 2+ anchors ,should scale to their common center while also scaling their handles in unison.
 - double-click on anchor connected to linear segments fails to flip between smooth-sharp


Keavon:
This breaks the ability to scale the handles of a single selected anchor point with S.

Keavon (approved):

PR #2768, opened 2025-06-27
Refit on delete

### LINKED ISSUE #2651: Deleting (dissolving) a point should scale its adjacent handles to minimize curvature change
https://math.stackexchange.com/a/879213/94039

4adex:
Closes #2651


Keavon:
Please resolve conflicts, thanks!

Keavon:
Unfortunately, in my testing, this doesn't seem to work very well in even the simple case of un-splitting a path. It also seems to fail quite badly when one of the segment ends is a zero-length handle.

PR #2770, opened 2025-06-27
Fix DropdownInput widget not reactively updating when its content changes

0HyperCube:
I though the idea of svelte was that it did the fine grained reactivity automatically?

I guess maybe they fixed this in the new version?


Keavon (approved):

Keavon:
❤️

PR #2777, opened 2025-06-28
Add Color Constructor Nodes

### LINKED ISSUE #2774: Create Color from Component Values
A node that takes four numbers (int or float, doesn't matter) as R, G, B, and A and outputs a color is needed for situations where color is driven mathematically by other elements.

The best workaround I have found when alpha isn't needed is to create three gradients, one for Black->Red, Black->Green, and Black->Blue. I can then sample those gradients and use them to produce three duplicates of the geometry, each representing one color channel. Finally, I can blend them using the Lighten blendmode to get the composite image.

phailhaus:
Resolves #2774

# Adds the following:
## color crate function:
from_cmyka() function to convert CMYKA to RGBA

## Nodes:
All nodes clamp input to 0 -> 1 before passing to matching function in color crate

construct_color_rgba() accepts f32 inputs in RGBA format and outputs a color
construct_color_hsla() accepts f32 inputs in HSLA format and outputs a color
[ConstructColorDemo.zip](https://github.com/user-attachments/files/20958769/ConstructColorDemo.zip)

construct_color_grayscale() accepts a single f32 luminance input and outputs a color
construct_color_cmyka() accepts f32 inputs in CMYKA format and outputs a color

I am struggling to understand more complex node construction. These could likely get merged, but I couldn't figure out how to create options that swap out inputs.

# Demo: [ConstructColorDemo.zip](https://github.com/user-attachments/files/20958769/ConstructColorDemo.zip)
The squares in the background test the four nodes in order from the center out. The dots in the foreground demonstrate a more practical use, deriving color from the instance index in a way that can't be done with a simple gradient.

ConstructColorDemo

### DIFF
--- a/node-graph/gmath-nodes/src/lib.rs
+++ b/node-graph/gmath-nodes/src/lib.rs
@@ -463,26 +463,26 @@ fn dot_product(_: impl Ctx, vector_a: DVec2, vector_b: DVec2) -> f64 {
 }
 
 /// Constructs a color from RGBA components. Clamped to 0 -> 1.
-#[node_macro::node(category("Color"))]
-fn construct_color_rgba(_: impl Ctx, _primary: (), red: f64, green: f64, blue: f64, alpha: f64) -> Color {
+#[node_macro::node(category("Color"), name("RGBA to Color"))]
+fn rgba_to_color(_: impl Ctx, _primary: (), red: f64, green: f64, blue: f64, alpha: f64) -> Color {
 	Color::from_rgbaf32_unchecked(red.clamp(0., 1.) as f32, green.clamp(0., 1.) as f32, blue.clamp(0., 1.) as f32, alpha.clamp(0., 1.) as f32)
 }
 
 /// Constructs a color from HSLA components. Clamped to 0 -> 1.
-#[node_macro::node(category("Color"))]
-fn construct_color_hsla(_: impl Ctx, _primary: (), hue: f64, saturation: f64, lightness: f64, alpha: f64) -> Color {
+#[node_macro::node(category("Color"), name("HSLA to Color"))]
+fn hsla_to_color(_: impl Ctx, _primary: (), hue: f64, saturation: f64, lightness: f64, alpha: f64) -> Color {
 	Color::from_hsla(hue.clamp(0.0, 1.0) as f32, saturation.clamp(0., 1.) as f32, lightness.clamp(0., 1.) as f32, alpha.clamp(0., 1.) as f32)
 }
 
 /// Constructs a color from a single grayscale value. Clamped to 0 -> 1.
 #[node_macro::node(category("Color"))]
-fn construct_color_grayscale(_: impl Ctx, _primary: (), luminance: f64) -> Color {
+fn gray_to_color(_: impl Ctx, _primary: (), luminance: f64) -> Color {
 	Color::from_luminance(luminance.clamp(0., 1.) as f32)
 }
 
 /// Constructs a color from CMYKA components. Clamped to 0 -> 1.
-#[node_macro::node(category("Color"))]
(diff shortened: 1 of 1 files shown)


node-graph/gmath-nodes/src/lib.rs
```diff
@@ -462,6 +462,30 @@ fn dot_product(_: impl Ctx, vector_a: DVec2, vector_b: DVec2) -> f64 {
 	vector_a.dot(vector_b)
 }
 
+/// Constructs a color from RGBA components. Clamped to 0 -> 1.
+#[node_macro::node(category("Color"))]
+fn construct_color_rgba(_: impl Ctx, _primary: (), red: f32, green: f32, blue: f32, alpha: f32) -> Color {
```
0HyperCube: The name of these nodes is too long so that the actual colour system doesn't appear in the node graph:
Node 'Construct Colo...`

Perhaps they could be shortened e.g. `RGBA Colour`. To change the capitalisation of the name, you can use:
```rs
#[node_macro::node(category("Colour"), name("RGBA Colour"))]

```

node-graph/gmath-nodes/src/lib.rs
```diff
@@ -462,6 +462,30 @@ fn dot_product(_: impl Ctx, vector_a: DVec2, vector_b: DVec2) -> f64 {
...
+	Color::from_rgbaf32_unchecked(red.clamp(0., 1.), green.clamp(0., 1.), blue.clamp(0., 1.), alpha.clamp(0., 1.))
+}
+
+/// Constructs a color from HSLA components. Clamped to 0 -> 1.
+#[node_macro::node(category("Color"))]
+fn construct_color_hsla(_: impl Ctx, _primary: (), hue: f32, saturation: f32, lightness: f32, alpha: f32) -> Color {
```
0HyperCube: The properties panel doesn't work because it only supports f64 rather than f32. Perhaps consider having the input be an f64 and then convert?

node-graph/gmath-nodes/src/lib.rs
Keavon: "RGBA to Color" would be my preferred name for this node, and I think that would fit.

Keavon:
I'll mark this as a draft for the moment while awaiting the requested code review changes for the naming scheme and parameter types. Please mark it as ready for review once resolved, thanks!

node-graph/gmath-nodes/src/lib.rs
phailhaus: Ah, that's why those didn't populate. Fixed.

node-graph/gmath-nodes/src/lib.rs
phailhaus: > "RGBA to Color" would be my preferred name for this node, and I think that would fit.

I have updated the node names to 'RGBA to Color', 'HSLA to Color', 'Gray to Color', and 'CMYKA to Color'.

image


Would it be preferable to swap to a single 'Create Color' node? I now know how to make enums appear as dropdowns/buttons/etc, but I'm not sure if the following is possible:

1. Create five generic f64 inputs
2. Rename them in the UI based on the currently-selected color formula. IE if RGBA is selected, input 1 is R. If HSLA is selected, input 1 is H
3. Hide any inputs that aren't needed for the current formula

Alternatively, we could have R, G, B, H, S, V, L, C, M, Y, K, A inputs and then show/hide the ones not relevant to the current formula.

Keavon:
!build

Keavon:
I see there is a conflict, can you please resolve it? And once this is fully ready for review, please mark it as such so it's not a draft (I ignore drafts when working through my review queue).

Keavon:
Superseded by #3838, except CMYK which we won't merge.

PR #2789, opened 2025-06-28
Add shape-rendering SVG attribute support

### LINKED ISSUE #2775: Add Option to Disable Antialiasing
An option to disable anti-aliasing would expand usefulness for specialized situations.

Example of usecase that is broken by AA:
I've been working on a project that involves stroking paths with a series of circles that change color based on distance from the start of the path. I can then use a specialized shader in a raster pipeline to threshold the image, giving a hand-drawn line that can be 'drawn' in real time with extremely light loads on both CPU and GPU. To do this, I need every pixel of a circle to be either 100% the desired color or entirely transparent. Anything inbetween results in artifacts.

My current workflow involves copying paths out of an SVG into a python script to generate these stroked textures, but Graphite provides a very enticing alternative.

phailhaus:
Partially Resolves #2775

Adds shape_rendering attribute to the PathStyle struct, adds support for writing it out when rendering the SVG, and adds a node that allows the user to set it on vectors.

Exact behavior varies by rendering engine. Tested:
* Vello: Not supported. Does nothing.
* Chrome: Functions for normal display and export. Does *not* function when using the Rasterize node.
* Firefox: Functions for normal display, export, and when using the Rasterize node.

This isn't a clean AA on/off toggle, however. There are four options, and it's up to the renderer what it wants to do with each. Outside of the Rasterize node difference, so far I'm seeing identical behavior between Chrome and FF, but it isn't logically consistent. Here is an example:
image

The top example can be set to Auto (the default if not overridden) or Geometric Precision and get the same output as before this was implemented.
The bottom can be set to Crisp Edges or Optimize Speed to get the result shown. Notably, on the bottom, AA is off for stroke and fill edges on the flower, along with the interface between the fill and stroke on the red shape. The outer edge of the stroke is still antialiased. If I change the stroke from Inside to Center or Outside, the anti-aliasing goes away.

Useful function for a technical user, but consistency in support is lacking.


Keavon:
I was uncertain about which direction, if any, this ought to take which is why I was hesitant to merge this at the time until further thought. Seeing that it has many gotchas in different cases (as you listed across the different platforms it does and doesn't support), and also because there is some code rot at this point, I think the best course of action is to not merge this until we have a greater level of compositing infrastructure in our render pipeline across each render target platform, then we can figure out the right approach at that time. Thanks for exploring this, though! I'm sure it will be handy to reference in the future.

PR #2792, opened 2025-06-29
Fix lasso select extending the previous selection area after it was aborted

### LINKED ISSUE #2647: Lasso/box selection switching and usage over existing layers
We want to update dragging and selection such that:
- Ctrl+drag's lasso selection, upon releasing Ctrl, switches to a regular box selection, and back to lasso by holding Ctrl again. It draws a straight line between the cursor position last time the lasso was removed and when it's now restored, allowing the user to also use this as a technique for drawing straight line sections of the lasso, but also just to decide to use box selection instead of lasso at any time.
- Ctrl+dragging an *object* currently drags its layer(s) but we want to make it instead always lasso select. This lets the user perform a lasso selection over an existing layer, or a box selection but releasing Ctrl as in the previous bullet point.
- Lasso selection has a bug that should be fixed: aborting the lasso selection (RMB or Esc) then creating a new lasso selection ends up erroneously extending the previously drawn lasso path because abort fails to clear the lasso data (which also wastes memory!).

hemu:
Currently after aborting an in-progress lasso selection with RMB or ESC, starting a new lasso selection will extend the previous lasso selection instead of starting a new selection. This fix ensures the previous selection is cleared after abort.

Partly closes #2647

# before fix
https://github.com/user-attachments/assets/647fecd5-1be1-4b88-a902-6b3e739979d3

# after fix
https://github.com/user-attachments/assets/cd9a44c2-7259-4972-b1cf-0d064e4e1f04

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -1402,6 +1402,7 @@ impl Fsm for SelectToolFsmState {
 				responses.add(DocumentMessage::AbortTransaction);
 				tool_data.snap_manager.cleanup(responses);
 				tool_data.axis_align = false;
+				tool_data.lasso_polygon.clear();
 				responses.add(OverlaysMessage::Draw);
 
 				let selection = tool_data.nested_selection_behavior;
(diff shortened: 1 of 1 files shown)


Keavon:
!build

Keavon (approved):
Thank you!

PR #2794, opened 2025-06-29
Improve Path tool layer selection behavior using double-click instead of single-click

### LINKED ISSUE #2716: Improve Path tool layer and point selection behavior
- Click-dragging to box select from inside a shape (even if it's not filled) should not select all the shape's points and begin dragging them; it should allow the box selection just like outside the shape
- Our current way of clicking another layer to make it selected should not happen unless it's a double-click, and if Shift is held, it adds or removes that layer
  - Double-clicking an already selected layer, without Shift held, will select all its points. Shift-selecting will deselect it. Shift-selecting a different deselected layer will add it to the selection, with all its points initially deselected.
- TODO: more changes?

4adex:
Closes #2716


otdavies:
I am a massive fan of this change great work. The PR seems to work quite well from my testing. I do wish we could drill down selection a bit better. Though this is not a blocker on merging this PR, more just a preference,

<del>In this example I am trying to "double click" on the inner shape to attempt to swap to the smaller shape for editing. Note, neither of these shapes have any fill. This works totally fine as long as the shapes are not under each other. Ideally It would be great if it "drilled down" to shapes underneath.</del>  

**I have actually implemented this in a later branch we will add, so don't worry about it!**

https://github.com/user-attachments/assets/c21afbdd-f5ee-47e8-b324-045fedc032ea

Great work!

Keavon:
A number of the items in the spec aren't implemented as described. Also from testing, I added one detail to the spec, mentioned below. Here are the needed changes to conform to it.

> - Our current way of clicking another layer to make it selected should not happen unless it's a double click (**exception: if no layers are selected, single-click should work to select the first one**). Without Shift, it transfers the selection to the not-yet-selected double click target object. If Shift is held, it adds **or removes** that layer to the selected objects. When added, its points will always start initially deselected.
- [x] **We also want to allow single-clicking to work if no layers are selected, but otherwise double-clicking is required to select a layer. This is the item I added now to the spec.**
- [x] **Shift-double clicking needs to remove a selected layer.**
> - Double-clicking an already selected layer, without Shift held, will select all its points. Other selected layers are unaffected (they keep their point selections and layer selection doesn't transfer away).
- [x] This needs to respect point/segment/both mode. Points: current behavior. Segments: all segments but not anchors. Both: all segments and anchors.
> - **Alt + double-clicking an already selected layer will deselect all its points.**
- [x] This was supposed to be implemented to deselect all its points, not deselect the layer.
- Other:
- [x] Single-clicking inside or outside a shape should deselect all selected points (even across multiple layers). This PR breaks clicking inside, which now fails to deselect like it used to. This PR currently incorrectly deselects all layers if you click the background. The only way to deselect all layers should be double-clicking the background, never by single-clicking.

Keavon:
Next round of QA feedback based on further refinement goals after testing how this feels to use (thank you for your eye for detail in making this all work!):

- [x] We have some minor issues when you Shift-double click a single not-yet-selected layer. The selection box shows a ~20x20px square for some reason briefly, and we end up selecting the layer only very briefly and then it gets deselected when the double click is finished.
- [x] When double clicking to add a new layer to the selection, I'd like to have it restore the original selection (which got deselected because of the initial click).
- [x] Alt-double click, as re-requested above, still doesn't seem to work because all the points just get immediately deselected. Using the same approach as the previous bullet point, we should restore the selection (minus that of the double-clicked layer's points).
- [x] Similar to the previous two, we also want the double click to select all points of a layer feature to add to the selection, not replace the selection, by means of restoring the prior selection after the double click completes (but with the double clicked layer's points added).
- [x] When we single-click to select the first layer prior to any layer being selected (as you just added per the request above), we want to also allow that single click to be a double click without it causing all the points to become selected. That means a user would have to initially (single or double) click, and then double click another time to cause all the points to become selected.

Keavon (approved):

PR #2796, opened 2025-06-29
fix: Suppress verbose npm ci output in package-installer.js

Mirza-Samad-Ahmed-Baig:
Instead of opening new issue, I thought of solving the issue on my own. Thank you.

This PR resolves a minor yet impactful issue in frontend/package-installer.js that caused excessive and distracting console output during development.

The package-installer.js script is responsible for automatically running npm ci whenever package.json or package-lock.json is modified. Previously, it invoked execSync with stdio: "inherit", which resulted in the full output of npm ci—including successful installation logs—being printed directly to the console. This cluttered the terminal, especially when combined with output from vite and cargo watch, making it harder to follow relevant logs.


Keavon:
Thanks for opening your first contribution. I'll compare this if I hopefully manage to get to it in code review. But my question regarding this is, why would this bother you if the script is only supposed to run once?

Mirza-Samad-Ahmed-Baig:
Thank you. While investigating a larger issue, I came across this and decided to open a separate PR for it in the meantime.

Keavon:
Just to reiterate my question:

> why would this bother you if the script is only supposed to run once?

It's otherwise unclear to me why this matters in the slightest since it occurs only once on project setup.

Mirza-Samad-Ahmed-Baig:
You're right that the script typically runs only once during initial setup. However, while debugging an unrelated issue, I found myself frequently resetting the environment and triggering this script multiple times in quick succession. The verbose output repeatedly cluttered the terminal, making it harder to isolate logs from Vite and Cargo Watch.

By silencing the unnecessary output, the developer experience becomes a bit cleaner—especially during debugging or repeated setup cycles. That said, I understand if the team prefers the current verbosity for transparency or troubleshooting.

Keavon:
I went to check a fresh build to see what the output looks like, and the output is both useful and minimal. So I appreciate your attention to this but I don't believe it's in our interests to merge this. Thank you.

PR #2798, opened 2025-06-29
Add ability to start lasso select within object without dragging layers

hemu:
When beginning a lasso select from within an object, currently the behavior is to drag its layers. This PR updates this behavior such that a lasso select beginning within an object will remain a lasso selection over the existing layers.

Part of [#2647](https://github.com/GraphiteEditor/Graphite/issues/2647)


editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -855,8 +855,8 @@ impl Fsm for SelectToolFsmState {
 				SelectToolMessage::DragStart {
 					extend_selection,
 					remove_from_selection,
-					select_deepest,
-					lasso_select,
+					select_deepest_key,
+					lasso_select_key,
```
hemu, line 859: renamed this so it's consistent with other similar var names and it let me use `lasso_select` as a different var in logic below

PR #2799, opened 2025-06-29
Add ability to start lasso select within object without dragging layers

hemu:
When beginning a lasso select from within an object, currently the behavior is to drag its layers. This PR updates this behavior such that a lasso select beginning within an object will remain a lasso selection over the existing layers.

Refactor select_tool and path_tool transition drag logic to determine an "intent" up front. This made it easier to incorporate the type of logic needed in this PR where the lasso select intent has priority over others, but its logic is shared with box select which has the lowest priority so it was hard to capture this in the existing if/else chain.  

Part of #2647

NOTE: This is an alternate implementation compared to this simpler change: https://github.com/GraphiteEditor/Graphite/pull/2798

# Before

https://github.com/user-attachments/assets/86142094-4e4d-4457-9f6c-241df25c3ad0

# After
https://github.com/user-attachments/assets/64d97867-e38b-4699-8271-e51b4ba6af6f

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -906,12 +906,12 @@ impl Fsm for SelectToolFsmState {
 					DragStartIntent::Resize
 				} else if skew {
 					DragStartIntent::Skew
-				} else if rotate {
-					DragStartIntent::Rotate
 				} else if lasso_select {
 					DragStartIntent::Draw
 				} else if can_grab_compass_rose || intersection.is_some_and(|intersection| selected.iter().any(|selected_layer| intersection.starts_with(*selected_layer, document.metadata()))) {
 					DragStartIntent::DragWithPreviousSelection
+				} else if rotate {
+					DragStartIntent::Rotate
 				} else if let Some(intersection) = intersection {
 					DragStartIntent::DragWithNewSelection(intersection)
 				} else {


editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -502,135 +509,149 @@ impl PathToolData {
...
+						selection_info = updated_selection_info;
 					}
 				}
-				if dragging_only_handles && !self.handle_drag_toggle && !old_selection.is_empty() {
-					self.saved_points_before_handle_drag = old_selection;
-				}
 
-				if handle_drag_from_anchor {
-					if let Some((layer, point)) = shape_editor.find_nearest_point_indices(&document.network_interface, input.mouse.position, SELECTION_THRESHOLD) {
-						// Check that selected point is an anchor
```
hemu, line 550: didn't change almost any of this logic. The diff just seems large because I moved the pieces from the if/else chain into a match statement

hemu:
if this change is too involved and not preferred there is a simpler change here: https://github.com/GraphiteEditor/Graphite/pull/2798

otdavies:
I tested this and didn't find any obvious regressions @Keavon 👍

Keavon:
@hemu could you please resolve the merge conflict? I tried but it looks like it affects the structure of an if-else statement that changed. Thanks.

Also, please mark this as ready for review and comment when you've completed this. I'm marking it as a draft for now while waiting the resolution.

hemu:
looking at the other merged commits in the past week that intersect with the same files in this PR, i think too much has changed underneath and the safer thing to do at this point is redo this entirely. will close this and take another crack at in a new PR so it can get retested and re-reviewed with same level of scrutiny as a new change

Keavon:
When you have your redo ready, please push it to this branch and reopen the same PR rather than filing a new PR.

PR #2804, opened 2025-06-30
Fix Spline tool in-progress drawing being cleared on tool change

LixiviaDev:
Spline Tool: Spline won't stay unless closed or (Esc or Enter) is pressed.

My changes:
- Removed "::AbortTransaction" message, which completely cleared the new spline on "::abort".
- Added "delete_preview()" function to delete the preview point that tracks the mouse.
- Added "::EndTransaction" message to signal that we want to save the new spline as is.


Keavon:
This does not appear to solve the problem in the specific case of switching to the Select tool (with the <kbd>V</kbd> hotkey or clicking its icon). It does seem to work as expected when switching to the other tools that I've tested. Could you please take a look at solving it for that tool?

LixiviaDev:
I'm really busy these days, I'll take a look at it as soon as I can

PR #2812, opened 2025-07-01
Add copy/cut/paste/duplicate functionality for path geometry

### LINKED ISSUE #2744: Copying and pasting geometry while in the Path tool
Ctrl+C and Ctrl+V should work, while in the Path tool, directly on the geometry (points, segments, regions). If pasted while not using the Path tool, or when the Path tool is working with 0 or 2+ selected layers, it should create a new Path layer (with default fill and stroke) with that geometry.

Ctrl+D should also duplicate the selected geometry in place, as if the user had copy-and-pasted it. Currently, Ctrl+D duplicates the whole layer while in the Path tool, so this should override that while this tool is active.

4adex:
Closes #2744


Keavon:
Conflict, I'll try to review after fixed.

Keavon (approved):

Keavon:
I have followup feedback after this merges. In particular, we need to be able to paste geometry that came from a different layer into an existing separate layer rather than that resulting in a new layer being created, if one existing layer is already selected with the Path tool active.

PR #2814, opened 2025-07-01
Add X and Y offset distance overlay while dragging with the Select tool

Kavin006:


Keavon:
!build

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -837,6 +837,33 @@ impl Fsm for SelectToolFsmState {
...
+
+						let height = format!("{:.2}", document_current.y - document_start.y).trim_end_matches('0').trim_end_matches('.').to_string();
+						let y_transform = DAffine2::from_translation((quad.top_left() + quad.bottom_left()) / 2.);
+						let height_pivot = if document_current.x >= document_start.x { OverlayPivot::Start } else { OverlayPivot::End };
+						overlay_context.text(&height, COLOR_OVERLAY_BLUE, None, y_transform, 3., [height_pivot, OverlayPivot::Middle]);
+					}
+					if !tool_data.axis_align && document_start.x != document_current.x && document_start.y != document_current.y {
+						overlay_context.line(quad.top_right(), quad.bottom_right(), None, None);
+						overlay_context.line(quad.bottom_left(), quad.bottom_right(), None, None);
+					}
```
0HyperCube: This overlay code is basically the same as for the <kbd>G</kbd> translation. Please consolidate this into a function inside overlay_context to reduce code duplication.
https://github.com/GraphiteEditor/Graphite/blob/49db963ce1067bc40af822e6e46d8777a4db7281/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs#L262-L284

Kavin006: This should be resolved now.

Kavin006:
Could you please check if it has all the expected behaviours.

mTvare6:
@Kavin006 looks like the pivot stored in select tool is viewport space and doesn't respect PTZ changes.

https://github.com/user-attachments/assets/4627344f-0e2c-435e-be32-c2a15afcda4e

Kavin006:
Didn't realise this could happen. I feel like we have to add the document position of drag start to SelectToolData to fix this. This is also the case with the selection box unless its intended there.

Keavon:
It works correctly when you pan (e.g. scroll with the mouse wheel) with `G` grabbing, but just not dragging with left click as your code implemented. Could you please work on a followup PR that attempts to solve it using the best approach you can figure out? Thanks!

PR #2821, opened 2025-07-02
Separate glyphs into Vector data rows

BKSalman:
This feature was [first mentioned in the Discord server](https://discord.com/channels/731730685944922173/1384222879885037608/1389540475979042898), and is being discussed on a [dedicated thread](https://discord.com/channels/731730685944922173/1389708273728622602)


node-graph/gstd/src/text.rs
```diff
@@ -39,7 +40,23 @@ fn text<'i: 'n>(
...
+	let mut glyph_table = if glyph_vectors.is_empty() {
+		VectorDataTable::new(VectorData::default())
+	} else {
+		VectorDataTable::default()
+	};
+
+	for glyph in glyph_vectors {
+		let instance = graphene_core::instances::Instance {
+			instance: glyph,
+			transform: DAffine2::IDENTITY,
```
0HyperCube: To position the glyph within the text, do we want to apply a transform here instead of to all of the points?

0HyperCube (commented):
Works nicely; thanks for the contribution.

node-graph/gstd/src/text.rs
BKSalman: > do we want to apply a transform here instead of to all of the points?

I'm sorry, but I couldn't understand what you mean by this

0HyperCube: The text is shaped by positioning all of the glyphs. Each glyph is offset from the last one by an advance value. The `PathBuilder` struct then creates the subpaths. Currently, this is achieved by offsetting all of the coordinates within the glyph.

I was suggesting that the origin of the glyph could instead be applied here.

BKSalman: Ok, will do that and push it in a bit

BKSalman: I just rebased, and pushed some changes to do what you requested, however I'm facing some issues

in the image here the transform of the layer is set to 0,0

1. The dashed bounding box is being drawn under the text, not over it
2. The only thing that can be selected from the text is the last letter (`A` in this case), however @Keavon have already answered that, and said it will be fixed in the future
3. The `A` highlighting is not positioned on the actual rendered letter
image

I couldn't find the source of the issue yet, and wanted to give an update

0HyperCube: The source of the issue is the fact that instances<T> are broken. However this isn't relevant to this PR.

BKSalman: ok I see

what should we do, should we wait for that to fixed before merging, or merge for now (if there is no other comments)?

0HyperCube: I'm not really sure what @Keavon plans to do about the `Instances<T>` not being implemented properly. Last weekend they expressed a desire to start a refactor, however it seems that no commits in this direction have been made.

The issue is that the editor and tooling frequently use `document.metadata.transform_to_viewport(layer_id)` and related functions. However since a layer is now made of zero or more instances with different transforms (after another Keavon refactor), this assumption no longer holds.

Keavon: Still planning to start on this soon, it's ultra high priority. It's not a merge blocker for this PR.

Keavon:
Please add the checkbox to make this optional (for the moment), when you get the time, thank you!

Keavon (approved):

Issue #2447, opened 2025-03-17
Exposing a value to the graph should add and connect up an equivalent value node


Keavon:
If you click the dot to expose an input to the graph...

![Image](https://github.com/user-attachments/assets/8a44a7b9-990f-481c-945f-341969fcf14c)

(here, the top one is exposed and the other two are not)

...it removes the control from the Properties panel and puts a connector on the node in the graph.

But that connector starts out connected to nothing. And whatever value the user had typed in the Properties panel previously, isn't represented anymore.

We want to have the process of clicking to expose an input to create a value node in the graph which is set to the same value as the Properties panel had. (This will require the creation of new value nodes, since not all representable Properties panel types have a value node yet.)

As a stretch goal, we also would like the reverse to occur: if a user un-exposes a node, it should evaluate what its current value is and set that to the input in the Properties panel. Perhaps only optionally, though, because the current behavior might be desired some (most?) of the time. This aspect can wait to be figured out or we can break it out into a separate issue.

Forchapeatl:
Hello @Keavon , https://github.com/GraphiteEditor/Graphite/issues/2434 already has a PR to it , May I work on this issue ?

Keavon:
Sure, you're welcome to begin with this issue.

adamgerhant:
> And whatever value the user had typed in the Properties panel previously, isn't represented anymore

This is not correct. Exposing the input simply changes the front end display. It does not change the value stored for the input

> We want to have the process of clicking to expose an input to create a value node in the graph which is set to the same value as the Properties panel had

Value nodes don't really exist. Values are stored in the input, and the identity node is a universal value node which allows the input to be connected to many nodes. It simply needs to display its name based on the type

Issue #1919, opened 2024-08-09
Allow constructing IVec2 in graph


0HyperCube:
The artboard node takes an IVec2 as the dimensions and position. Currently it is impossible to construct this in the graph.

A node taking in two f64s and constructing an IVec2 should be added.

Keavon:
We should consider reusing the Vector2 node and letting the type system support casting from an IVec2 to a DVec2.

aybdee:
I want to try doing this to get a better feel for how graphene works, how do you suggest i go about it ?

Keavon:
@TrueDoctor is this okay to work on right now or should it wait until your refactor in #1942 is done?

Keavon:
In that case, @aybdee, you can just reference any of the value nodes. But if you want to wait like a week, the process of creating nodes will become much easier with that refactor mentioned above.

aybdee:
i think i'd wait then 
i'll try some other stuff for now

Keavon:
Sounds great. There are a lot of good options in `#code-todo-list` on Discord, that's where most of the good beginner issues are at.

lizard-demon:
Is it still impossible to create an Ivec2?

Keavon:
Yes, but automatic type conversion is right around the corner and it will be supported then.

PR #2822, opened 2025-07-02
Add a general-purpose Value node

adamgerhant:
When an exposed value input is clicked with alt pressed and dragged, a value node with that same input is created, and connected to the node. Removes all value nodes and replaced with the generic value node.

TODO: Migrate fill to be input type based to close https://github.com/GraphiteEditor/Graphite/issues/2875

https://github.com/user-attachments/assets/79a1ac3d-9a88-467d-aa44-4886607e645b

Essentially closes https://github.com/GraphiteEditor/Graphite/issues/2447
Closes https://github.com/GraphiteEditor/Graphite/issues/1919, https://github.com/GraphiteEditor/Graphite/issues/2567


editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -903,6 +903,69 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
...
+					// Disconnecting with no upstream node, create new value node.
+					let to_connector = network_interface.input_connector_from_click(ipp.mouse.position, selection_network_path);
+					if let Some(to_connector) = &to_connector {
+						let Some(input_position) = network_interface.input_position(to_connector, selection_network_path) else {
+							log::error!("Could not get input position for connector: {to_connector:?}");
+							return;
+						};
+						self.wire_in_progress_to_connector = Some(input_position);
+					}
+					// Not hovering over a node input or node output, insert the node
```
0HyperCube: Since this function is so long, I feel this self contained behaviour could be extracted to a function.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -903,6 +903,69 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
...
+					else {
+						// Disconnect if the wire was previously connected to an input
+						if let Some(disconnecting) = self.disconnecting.take() {
+							let mut position = if let Some(to_connector) = self.wire_in_progress_to_connector { to_connector } else { point };
+							// Offset to drag from center of node
+							position = position - DVec2::new(24. * 3., 24.);
```
0HyperCube: Magic 24 and 3? Perhaps add some named constants since it isn't really clear what these represent.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -903,6 +903,69 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
...
+							};
+
+							self.drag_start = Some((drag_start, false));
+							self.node_has_moved_in_drag = false;
+							self.update_node_graph_hints(responses);
+
+							let node_id = NodeId::new();
+							responses.add(NodeGraphMessage::CreateNodeFromContextMenu {
+								node_id: Some(node_id),
+								node_type: "Identity".to_string(),
```
0HyperCube: Why always the identity node? Surely the relevant value node should be preferred. For example using the identity node, it is not possible to edit the value of an f64, whereas you can with a number value node.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -903,6 +903,69 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphHandlerData<'a>> for NodeGrap
...
+								xy: Some(((position.x / 24.) as i32, (position.y / 24.) as i32)),
```
0HyperCube: More magic constants.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
adamgerhant: I agree, I made a general purpose value node to replace the identity node

editor/src/messages/portfolio/document/node_graph/node_properties.rs
```diff
@@ -1835,6 +1837,77 @@ pub fn math_properties(node_id: NodeId, context: &mut NodePropertiesContext) ->
...
+
+			let messages = vec![NodeGraphMessage::SetInputValue { node_id, input_index: 0, value }.into(), NodeGraphMessage::SendGraph.into()];
+
+			Message::Batched(messages.into_boxed_slice()).into()
+		}
+	};
+	let value_dropdown = enum_choice::<TaggedValueChoice>().dropdown_menu(choice, updater, || commit_value);
+	select_value_widgets.extend_from_slice(&[Separator::new(SeparatorType::Unrelated).widget_holder(), value_dropdown]);
+
+	let mut type_widgets = match property_from_type(node_id, 0, &input_type, (None, None, None), None, None, None, false, context) {
```
0HyperCube, line 1960: I noticed that the following types don't have a working properties: OptionalF64, OptionalDVec2, VecU64, F64Array4.

Also the VectorData says "Vector data is supplied through the node graph". However this is misleading as there is no way to actually expose the input to the graph. This is also inconsistent with the other complex types e.g. GraphicElement, which currently just displays a "-".

Keavon: The `-` is newer and we can probably just standardize on that for all the types we don't yet have widgets for. We'll someday even have a widget for vector with some kind of way to visualize the path in a box that you can click on to view larger or something (I haven't thought much about this design, but that's the notion).

Keavon:
I'm momentarily converting this to a draft while I work through my PR queue since this has merge conflicts and a test failure in CI. It appears to be an exciting change so I look forward to having the chance to review and land this. Thanks :)

Keavon:
This ultimately didn't end up being the approach we needed (see the updated description in the original issue, #2567, for more description of how this is intended to work) but hopefully that approach can be implemented soon.

@adamgerhant, your dragging the value node out of the connector like you show in the start of your video would be valuable to have split into its own PR if you are interested in landing that part, using the existing value nodes (for those which exist).

adamgerhant:
Doesnt #2567 describe a simple UI change where the drop down menu is changed to only open when clicking the expose dot?
image

Keavon:
I can describe it for you in a call but it's a whole separate system that is also type-aware.

PR #2826, opened 2025-07-03
Simplify properties sections

0HyperCube:
- No empty merge section.
- Sections go in order from first applied to last applied.


Keavon:
This may or may not be desirable, it's ultimately a design question that I don't yet feel comfortable making a decision on quite yet until after completing the bigger Properties panel redesign and figuring out the pros and cons of each. I'm currently 50%-50% on it, but would rather leave this as-is for the moment to avoid confusing people if we end up deciding to go back to how it was. This will need to be consistent with several upcoming extensions to the UI design across the Properties, Layers, and Outliner panels. Thankfully it's a simple change and we can always reopen this in the future, but presently I'd feel better about skipping merging this until I can make a more specific decision one way or the other. Thanks for bringing this up though, and it will play a larger role now in my upcoming design work now that you've raised the question about which direction is better and whether the selected layer-style node itself should be included or only its upstream chain. But tabling this for now.

PR #2828, opened 2025-07-04
Skip complex layer thumbnails

0HyperCube:
- Each graphic element gets a render complexity
- Layers with high render complexity get skipped for thumbnails

### DIFF
--- a/node-graph/gcore/src/render_complexity.rs
+++ b/node-graph/gcore/src/render_complexity.rs
@@ -1,0 +1,58 @@
+use crate::raster_types::{CPU, GPU, Raster};
+use crate::{Artboard, Color, GraphicElement, instances::Instances, vector::VectorData};
+use glam::DVec2;
+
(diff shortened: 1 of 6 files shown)


node-graph/gcore/src/render_complexity.rs
```diff
@@ -0,0 +1,58 @@
...
+	fn render_complexity(&self) -> usize {
+		(self.width * self.height) as usize
+	}
+}
+
+impl RenderComplexity for Raster<GPU> {
+	fn render_complexity(&self) -> usize {
+		(self.width() * self.height()) as usize
+	}
+}
```
TrueDoctor: you can probably set the render complexity for gpu textures to infinite as we currently can't generate thumbnails for them anyway

TrueDoctor (approved):
The code looks good, we might want to special case textures as mentioned in the comment but this looks fine either way

PR #2837, opened 2025-07-06
Add Normalize and Length nodes

celyk:
A couple standard vector operations that I've missed in Graphite.

### DIFF
--- a/node-graph/gmath-nodes/src/lib.rs
+++ b/node-graph/gmath-nodes/src/lib.rs
@@ -612,11 +612,13 @@ fn dot_product(_: impl Ctx, vector_a: DVec2, vector_b: DVec2) -> f64 {
 	vector_a.dot(vector_b)
 }
 
+/// Gets the length or magnitude of a vector.
 #[node_macro::node(category("Math: Vector"))]
 fn length(_: impl Ctx, vector: DVec2) -> f64 {
 	vector.length()
 }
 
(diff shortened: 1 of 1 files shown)


node-graph/gmath-nodes/src/lib.rs
```diff
@@ -612,6 +612,18 @@ fn dot_product(_: impl Ctx, vector_a: DVec2, vector_b: DVec2) -> f64 {
...
+/// Gets the length or magnitude of a vector.
+#[node_macro::node(category("Math: Vector"))]
+fn length(_: impl Ctx, vector: DVec2) -> f64 {
+	vector.length()
+}
+
+/// Scales the input vector to unit length while preserving it's direction. This is equivalent to dividing the input vector by it's own magnitude.
+#[node_macro::node(category("Math: Vector"))]
+fn normalize(_: impl Ctx, vector: DVec2) -> DVec2 {
+	vector.normalize()
```
0HyperCube: The `normalize` function can cause a crash if the length is zero. Consider using `normalize_or_zero`.

celyk: I agree, thanks. I'll make note of it in the comment/description

celyk: In source it goes into detail about non-finite vectors, but I think this will suffice for the user
`/// Returns zero when the input vector is zero.`

0HyperCube:
!build

0HyperCube (approved):
Looks good; thanks for the contribution.

PR #2851, opened 2025-07-10
Add gizmos for interacting with the Spiral node

0SlowPoke0:
Added turns gizmo for spirals and updated the spiral API to use inner and outer radii instead of tightness and growth


Keavon:
Can this PR be salvaged in its current state?

Keavon:
Co-authored by @Annonnymmousss in the closed branch #3734.

PR #2852, opened 2025-07-10
Fix select tool distance overlay

Kavin006:
Fixes select tool distance overlay not respecting ptz changes mentioned in https://github.com/GraphiteEditor/Graphite/pull/2814.

### DIFF
--- a/editor/src/messages/tool/tool_messages/select_tool.rs
+++ b/editor/src/messages/tool/tool_messages/select_tool.rs
@@ -2,7 +2,7 @@
 
 use super::tool_prelude::*;
 use crate::consts::*;
-use crate::messages::input_mapper::utility_types::input_mouse::ViewportPosition;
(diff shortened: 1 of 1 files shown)


Kavin006:
Also this is what I meant by the selection box also having the same behaviour. If you feel like this should also be like the grab translation box, I could try to add it to this PR.

https://github.com/user-attachments/assets/200d6163-898f-4e62-bdb1-ceb3420a7e30

Keavon:
Yes, that one has been a known issue waiting on somebody to fix it. It would be grand if you'd tackle it too!

Keavon:
I've filed one mega issue with your observations included, amongst many others: #3500.

If you would like to revive this PR, feel free to keep pushing to it and I can reopen it if you let me know to do so. But I'm closing it for now because it is likely stale.

Thanks for your earlier work on these issues, and you'd be most welcome to continue if you have the time and interest. Cheers!

PR #2860, opened 2025-07-11
Improve the Path tool's segment editing mode and make hovering manipulators react contextually

### LINKED ISSUE #2832: Path tool segment editing mode polish
From https://github.com/GraphiteEditor/Graphite/pull/2712#issuecomment-3017924338:

- [x] Disabling all modes shouldn't be possible.
- [x] We want it to act more like radio buttons but where Shift can multi-select, like in Blender (but I may need to figure out a more refined design for radios vs. checkboxes when the lines between the two are blurred as they are in this case).
- [x] Switching from segment mode to point mode should make the points connected to each segment become selected, like in Blender, and vice versa where two neighboring points cause their shared segment to become selected. And in segment only mode, ~~selecting two adjacent segments should automatically select their shared point (but not in segment + point mode)~~. selecting a segment should select both of its endpoints also (but not in segment + point mode) (request overriden in discussion on discord)
- [x] Didn't we discuss a hotkey (maybe Ctrl or Alt?) to let you insert a point on a segment even if you're in segment selection mode? That should also be the same hotkey for molding when you drag instead of click, instead of currently using <kbd>A</kbd> which isn't a standard modifier key.
- [x] Now that we have the hover highlight over segments before clicking to select them, we should have the same for anchors— that can take the form of a 1px (in all directions) rectangle with the same semitransparent blue used for the pre-selection segment hover.
- [x] I love the dynamic contextual input hints, and it would be good to have that also happen when hovering over an anchor, not just a segment. This should also happen when currently targeted by a box or lasso selection (just like in Select mode which already does this).

4adex:
Closes #2832


Keavon:
- [x] Two more issues in this video. First part shows being able to hover a point even if it's invisible. Second part shows how the new hover styling (which I updated the visuals of) turns white when it's already selected, but should only apply the visual change to the outside while keeping the inner part solid blue because it's still selected. 
  
  https://github.com/user-attachments/assets/555e9bd7-c76f-4548-a001-0debc0f629bb
  - [x] We also want the anchors/handles to light up, as if hovered, when currently within the box/lasso selection. And segments while in segment mode.
- [x] Shift clicking the point/segment mode checkboxes still lets you deselect all of them, but deselecting all shouldn't be possible.
- [x] If I have a segment selected in segment mode, switching to point mode keeps that segment selected which shouldn't be possible.
- [x] Upon inserting a point, it needs to not select the point, while in segment mode, since point selection shouldn't ever occur in that mode.

Keavon (approved):

4adex:
https://github.com/GraphiteEditor/Graphite/pull/2860#issuecomment-3066844663
All of these are also covered in this PR, double checked.

Keavon:
Can you please explain the idea behind the `frozen` flag in the CheckboxInput, why that's necessary (what consideration was taken for doing it purely in the backend?), etc.?

4adex:
I first did it purely in backend only, so these were the two approaches that I tried - 
- I used the `disabled` flag (disabled when only one mode is activated) so that you cannot turn both modes to false, as in https://130b5a88.graphite.pages.dev/ but then it also greys out the button and it doesn't look good for the UX that which mode is active and what can be actually clicked.
- Then I tried without the `disabled` flag, in which the options don't update if only that specific mode is enabled. But you can still click on the button and for a very short moment you will see the button to be off and then it turns back to white in the next frame, which worked but also was bad UX.

Hence I decided to make a flag `frozen` similar to disabled, which works as disabled button but with the primary color (without being greyed out).

Keavon:
Can you go back to the approach from your second bullet point, where it briefly flashes as disabled? We can live with that for the moment. What I need to do is adapt the UI design system to better accommodate the similarities ranging from checkboxes to radio buttons so they can share certain parts of their behavior, as needed here. The little flaw can be fixed at that time.

Keavon:
- [x] Please ensure that, while in segment editing mode, I can still drag handles.
- [x] In both segment and point mode, when I'm clicking down to select an anchor/handle or segment, we need the gray ghost-of-the-original-shape outline to not show up until dragging has begun, since this flashes when the mouse is down. Most visible on an open path like a "C" shape which still has a Fill node causing the straight line between endpoints to show this. (I realize this is somewhat unrelated but still worth mentioning here.)
- [x] After Ctrl-clicking to insert a point on a *selected* segment, the resulting two segments should also be selected. This overrides the bullet point I'd written above:
  > Upon inserting a point, it needs to not select the point, while in segment mode, since point selection shouldn't ever occur in that mode.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1645,14 +1764,72 @@ impl Fsm for PathToolFsmState {
 						};
 
 						let quad = tool_data.selection_quad(document.metadata());
-						let polygon = &tool_data.lasso_polygon;
+
+						// TODO: Calculate which points/ handles are currently in the selected region and make those have a
```
Keavon: Incomplete comment sentence

Keavon:
I pushed my code review tweaks. I'll mark this as a draft while waiting the other requests in today's comments above. We're close on this one now!

Keavon:
Can you please make this compile again following the sync with master which broke the ability to compile?

PR #2874, opened 2025-07-13
Implement basic desktop app with chromium embeded framework 

timon-schelling:
Things done:
- getting the frontend to work inside cef like it would be running in a browser 
- implement resource handler for bundling frontend files inside the desktop app binary

Known issues:
- shortcuts that include a char key do not work
- shortcuts that include a mouse press do not work
- poor performance for very large windows
- GPU process sometimes crashes during resize, this can take down the whole app if it happens to frequently
  - `[ERROR:components/viz/service/main/viz_main_impl.cc:184] Exiting GPU process due to errors during initialization`
  - `[ERROR:content/browser/gpu/gpu_process_host.cc:957] GPU process exited unexpectedly: exit_code=133`
`[FATAL:content/browser/gpu/gpu_data_manager_impl_private.cc:415] GPU process isn't usable. Goodbye.`

Other possible improvements:
- rerender only dirty regions
- use accelerated draw api
- reduce idle cpu usage


TrueDoctor (approved):

desktop/build.rs
```diff
@@ -0,0 +1,10 @@
+use std::fs::metadata;
+
+fn main() {
+	let frontend_dir = format!("{}/../frontend/dist", env!("CARGO_MANIFEST_DIR"));
+	metadata(&frontend_dir).expect("Failed to find frontend directory. Please build the frontend first.");
+	metadata(format!("{}/index.html", &frontend_dir)).expect("Failed to find index.html in frontend directory.");
```
Firestar99, line 6: Could you explain to me why this `build.rs` exists? Just to verify the frontend is build? Two scenarios:
* You are embedding the frontend/dist folder in the executable: Then the embedding process should fail during build, not the build script during check.
* You are referencing this folder from code and it will be distributed alongside the executable: Then wouldn't a simple `#[test]` protect against failure?
* Also consider just building the frontend in the build script?

Background: Every `cargo check`, `cargo test` or `cargo build` now fails if you didn't explicitly build the frontend first.

timon-schelling: It is the first case, and for me it was just a way to provide better feedback to devs. But I didn't think about your Issue. Feel free to open a pr. Or I will take a look maybe tomorrow.

Firestar99: For now I can just comment out the `build.rs`, rather you have a look cause you also know how the embedding works

PR #2876, opened 2025-07-13
Fix bug in Safari where number input field click-and-drag can't be released

### LINKED ISSUE #2038: Click-and-drag number fields can't be released in Safari
1. Make a new document.
2. Draw a circle.

seabeeberry:
Closes #2038

I put the definition of the event target and call of `requestPointerLock()` immediately within the `onDragPointerDown` handler, as this is expected for Safari to handle the pointerlock properly. I removed the event parameter from the function call for `beginDrag()` because it is no longer needed there.

I tested this to work properly for the following browsers:
- Safari
- Firefox
- Firefox Nightly
- Chrome

Edit: I confused the comment about this bug happening in Nightly as well from a different open issue.

### DIFF
--- a/editor/src/messages/layout/layout_message_handler.rs
+++ b/editor/src/messages/layout/layout_message_handler.rs
@@ -123,7 +123,10 @@ impl LayoutMessageHandler {
 				let callback_message = match action {
 					WidgetValueAction::Commit => (breadcrumb_trail_buttons.on_commit.callback)(&()),
 					WidgetValueAction::Update => {
-						let update_value = value.as_u64().expect("BreadcrumbTrailButtons update was not of type: u64");
+						let Some(update_value) = value.as_u64() else {
+							error!("BreadcrumbTrailButtons update was not of type: u64");
+							return;
+						};
 						(breadcrumb_trail_buttons.on_update.callback)(&update_value)
 					}
 				};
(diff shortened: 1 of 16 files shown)


Keavon:
!build

Keavon (approved):

Keavon:
Thank you!

Keavon:
I didn't test thoroughly enough before merging this with clicking for text entry, and I've realized this breaks things in a huge way, so I reverted this PR. Please open a new one that fixes the problem and we'll give it another try from there.

The issue is that, upon clicking a text input field, it immediately tries to begin dragging. Now that I look at the changed code more closely, that is why we have the existing method with the comment

```
// If it's a mouseup, we'll begin editing the text field.
```

so you'll need to find some other approach that keeps the existing functionality.

Thanks.

seabeeberry:
> The issue is that, upon clicking a text input field, it immediately tries to begin dragging. Now that I look at the changed code more closely, that is why we have the existing method with the comment
> 
> ```
> // If it's a mouseup, we'll begin editing the text field.
> ```
> 
> so you'll need to find some other approach that keeps the existing functionality.

Ahh, I see. Sorry, I did not think to check the text fields. I will try and find a way without breaking other functionality.

PR #2878, opened 2025-07-14
Rename the message handlers of all the tools to use the MessageHandler suffix

MohdMohsin97:


Keavon:
I've ultimately decided that the tradeoffs of this change request are not worth the attempted standardization it brought. Thanks for giving this a shot, though!

PR #2890, opened 2025-07-16
Upgrade vello to version 0.5.0 and wgpu to version 25

mTvare6:
see
* https://linebender.org/blog/tmil-14/
* https://github.com/linebender/vello/pull/803
* https://github.com/linebender/vello/pull/802

### DIFF
--- a/node-graph/wgpu-executor/src/lib.rs
+++ b/node-graph/wgpu-executor/src/lib.rs
@@ -4,13 +4,12 @@ use anyhow::Result;
 pub use context::Context;
 use dyn_any::StaticType;
 use glam::UVec2;
-use graphene_application_io::{ApplicationIo, EditorApi, SurfaceHandle};
+use graphene_application_io::{ApplicationIo, EditorApi, SurfaceHandle, SurfaceId};
 use graphene_core::{Color, Ctx};
 pub use graphene_svg_renderer::RenderContext;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use vello::Error;
 use vello::{AaConfig, AaSupport, RenderParams, Renderer, RendererOptions, Scene};
-use wgpu::TextureFormat;
 use wgpu::util::TextureBlitter;
 use wgpu::{Origin3d, SurfaceConfiguration, TextureAspect};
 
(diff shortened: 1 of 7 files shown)


TrueDoctor:
image
Yeah, vello rendering is completely broken

node-graph/wgpu-executor/src/lib.rs
```diff
@@ -105,12 +121,39 @@ impl WgpuExecutor {
 	#[cfg(target_arch = "wasm32")]
 	pub fn create_surface(&self, canvas: graphene_application_io::WasmSurfaceHandle) -> Result<SurfaceHandle<Surface>> {
 		let surface = self.context.instance.create_surface(wgpu::SurfaceTarget::Canvas(canvas.surface))?;
+		let size = UVec2::ZERO;
```
Firestar99: In wasm we're rendering to a texture of size 0...

Now I don't know why, but this code *sometimes* works for me on FF nightly anyway. (inconsistent)

Firestar99:
> Uncaptured WebGPU error: In a set_pipeline command, caused by: Render pipeline targets are incompatible with render pass, caused by: Incompatible color attachments at indices [0]: the RenderPass uses textures with formats [Some(Rgba8Unorm)] but the RenderPipeline with 'wgpu::uti::TextureBlitter::pipeline' label uses attachments with formats [Some(Bgra8Unorm)]

https://github.com/GraphiteEditor/Graphite/blob/dc27a56b1da052889cbe0bd7661f9be6db1b813c/node-graph/wgpu-executor/src/lib.rs#L166-L171

@mTvare6 I think you are misunderstanding what surface *capabilities* are. They are just reporting the many formats the surface is capable of, not what a surface "is". When you creates a texture from a surface (with `surface.configure()`), you chooses one of the texture formats the surface is capable of, and then the texture is of that format. And that chosen format is the one you need to give to `TextureBlitter`. In our case, hardcoded to `Rgba8Unorm` (which *in theory* could break if some impl doesn't support that format, but in practice all impl support `Rgba8Unorm` and `Rgba8Srgb`).

Will try to fix it

Firestar99:
@TrueDoctor feel free to retest

TrueDoctor:
> @TrueDoctor feel free to retest

Seems to work fine now

node-graph/wgpu-executor/src/lib.rs
mTvare6: I noticed it, but the other resolution was already set to zero, which is why I used that.

Firestar99:
Anything blocking this from merging? Or do you want to properly retest everything before this is merged?

PR #2891, opened 2025-07-16
Draw image frames as outlines in the outline view mode

### LINKED ISSUE #2888: Images remain opaque in outline view
Images are fully visible when in outline view-mode.
Images should either become more transparent or be only visible as outline or even better both.

4ry1337:
Render rectangle outline instead of image in outline mode

Closes #2888


Keavon:
Could you please rebase this and fix the conflict?

PR #2903, opened 2025-07-18
Render images via canvases instead of encoding and compression steps on making generating the PNG

mTvare6:


Keavon:
Marking as draft while you debug the issue with toggling Vello on, then off, then on again and nothing refreshing in the render after that point.

Keavon (approved):

Issue #2902, opened 2025-07-18
[ Suggestion ] Grids should be per-artboard / only affect active artboard


Azagwen:
I noticed the grid as it currently exists originates at the 0,0 point of a document, which is a problem if you use isometric grids on a document with multiple artboards.
Ideally grid origins should be able to be moved, for maximal freedom of creation, but in the meantime, they could at the very least only appear on the selected artboard's origin.

Keavon:
You can set the grid origin, but yes it probably should have an option to be global or dynamically move itself relative to the active artboard.

Azagwen:
> You can set the grid origin, but yes it probably should have an option to be global or dynamically move itself relative to the active artboard.

Oh right! That's great, will use that for now.
I hadn't even noticed tbh, I'm too used to having a Gizmo for that 😅

Keavon:
We should have a gizmo for it also, yeah, that will be helpful.

PR #2906, opened 2025-07-18
[ Suggestion ] Grids should be per-artboard / only affect active artboard

Biplll:
Attempts to fix issue #2902, allowing multiple ways to set grid origin for artboards.

### DIFF
--- a/GRID_ARTBOARD_ORIGINS.md
+++ b/GRID_ARTBOARD_ORIGINS.md
@@ -1,0 +1,82 @@
+# Grid Origins for Artboards
+
+## Problem
+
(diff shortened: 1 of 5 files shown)


0HyperCube:
Thanks for this PR @Biplll. Please note in the PR description if you generated your code entirely using a large language model.

I can see that you didn't bother to even try and compile the test. You also appear to have committed the reasoning steps for the LLM into the root of the project. This has lead me to suspect that the code is not human authored. Code generated entirely by an LLM may infringe on copyright of the training material.

Thanks again for the time spent on this usability tweak.

Keavon:
While we aren't opposed to LLM coding assistance, this appears to have been entirely vibe coded which is not something we allow because of its severe risk to code and product defect quality, so this won't be considered further in its present form. Thanks.

PR #2913, opened 2025-07-20
Remap Path tool point sliding to G G

### LINKED ISSUE #2715: Point sliding with G G
In the Path tool, <kbd>G</kbd><kbd>G</kbd> slides the anchor along its adjacent segment(s) while adjusting its handle(s) to best approximate the initial curvature.

4adex:
Closes #2715


Keavon:
This doesn't appear to *do* anything, from what I can tell. Did you test it? Or misunderstand what is desired?

4adex:
https://4778b953.graphite.pages.dev/
It is working in the deployement, before your code review commit @Keavon 
[Screencast from 25-07-25 09:49:36 AM IST.webm](https://github.com/user-attachments/assets/0096fa0a-2613-46a3-b6d1-5df3e87d5326)
I will fix it

Keavon:
Oh, it doesn't work on ellipses. I tested your commit before doing my code review. (It sounds like I also broke something in my code review/)

Keavon:
We also need hints during G/R/S, including (when available) the subsequent G to slide. I seem to recall we already had working G/R/S hints but we don't seem to currently in the Path tool. Maybe I was just remembering the Select tool's ones.

Keavon:
Still supposed to be a draft?

4adex:
The hints are not updated yet, but hints have been changed to dynamic ones in https://github.com/GraphiteEditor/Graphite/pull/2860 so was waiting for merger of that.

Keavon:
Feedback for a followup PR:

- When we've reached one of the ends of travel for the segment, we'd like to automatically merge the sliding point and endpoint it's touching, while also correctly dealing with the handle (which needs to remain colinear if its endpoint anchor it's merging with was).
- When sliding, we need <kbd>Enter</kbd> to confirm in addition to left click, and we need that confirmation mentioned in the hints
- We want full G/R/S hints when in G, R, or S mode just like we do with G/R/S for the Select tool
  - And when in G mode, that's when we want to show the hint for "Slide Point" upon hitting G a second time, instead of having the hint currently show G G
- We should support sliding multiple simulatneous anchors if they are on their own non-contiguous segments

PR #2914, opened 2025-07-20
Add 'Circle' to the Shape tool and its associated gizmos

0SlowPoke0:
- Add circle mode in shape tool along with the associated radius gizmo to change the radius


Keavon (changes requested):
Drawing a circle usually results in the circle being drawn from the origin, not the drag start position. Once that is fixed, we can merge, my code review is done.

PR #2915, opened 2025-07-21
Re-add upload texture

Firestar99:
Effectively just 3 commits without dependent PRs:
* https://github.com/GraphiteEditor/Graphite/pull/2915/commits/dd380f99d73cad5fc039332be0b6584f5ca59525  vello: code cleanup of resource overwrites 
* https://github.com/GraphiteEditor/Graphite/pull/2915/commits/45596c0237f011e0674e904e340df72017971f88  upload_texture: upload cpu textures as SRGBA8 
* https://github.com/GraphiteEditor/Graphite/pull/2915/commits/df9e2e19d72328d785e3e5d882638f4726a74996 vello: fix wgpu::Texture leak within vello's `context.resource_overrides` HashMap

### Texture upload node
TLDR: With @TrueDoctor we decided that SRGB8 for now is fine, but we would like to switch to f32 later and add a conversion step down to SRGB8 before we give it to vello.

We would like to have linear f32 textures, as discussed with @TrueDoctor. However, this implementation currently uses `Rgba8UnormSrgb` converted to with graphite's `SRGB8` struct, as using `Rgba32Float` causes FF nightly to emit `Uncaptured WebGPU error: In a pass parameter, caused by: Encoder is invalid`. The real fun will be debugging that, since wgpu documents the error like this:

```rust
/// Used internally by wgpu functions to indicate the encoder already
/// contained an error. This variant should usually not be seen by users of
/// the API, since an effort should be made to provide the caller with a
/// more specific reason for the encoder being invalid.
#[error("Encoder is invalid")]
Invalid,
```
https://github.com/gfx-rs/wgpu/blob/9b966bf8a39b07bfc495fb43e8bf7725c5df440c/wgpu-core/src/command/mod.rs#L971-L976

In other words, I have no clue to go on what's going wrong. Changing the image format shouldn't have triggered anything really. The only thing I can think of is the image format being unsupported for sampling, I'd have to verify this later. At least vulkan says I very much can with my hardware.

### DIFF
--- a/node-graph/gsvg-renderer/src/renderer.rs
+++ b/node-graph/gsvg-renderer/src/renderer.rs
@@ -16,10 +16,13 @@ use graphene_core::uuid::{NodeId, generate_uuid};
 use graphene_core::vector::VectorDataTable;
 use graphene_core::vector::click_target::{ClickTarget, FreePoint};
 use graphene_core::vector::style::{Fill, Stroke, StrokeAlign, ViewMode};
-use graphene_core::{AlphaBlending, Artboard, ArtboardGroupTable, GraphicElement, GraphicGroupTable};
+use graphene_core::{Artboard, ArtboardGroupTable, GraphicElement, GraphicGroupTable};
 use num_traits::Zero;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
+use std::ops::Deref;
+use std::sync::{Arc, LazyLock};
+use vello::peniko::Blob;
 #[cfg(feature = "vello")]
 use vello::*;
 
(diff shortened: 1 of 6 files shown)


node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -16,10 +16,13 @@ use graphene_core::uuid::{NodeId, generate_uuid};
...
 use graphene_core::vector::click_target::{ClickTarget, FreePoint};
 use graphene_core::vector::style::{Fill, Stroke, StrokeAlign, ViewMode};
-use graphene_core::{AlphaBlending, Artboard, ArtboardGroupTable, GraphicElement, GraphicGroupTable};
+use graphene_core::{Artboard, ArtboardGroupTable, GraphicElement, GraphicGroupTable};
 use num_traits::Zero;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
+use std::ops::Deref;
+use std::sync::{Arc, LazyLock};
+use vello::peniko::Blob;
```
Keavon: Should this also be covered by the `#[cfg(feature = "vello")]`?

Firestar99: fixed

TrueDoctor (approved):

PR #2916, opened 2025-07-21
Fixes #2038: Safari click and drag fix

### LINKED ISSUE #2038: Click-and-drag number fields can't be released in Safari

seabeeberry:
Closes #2038

For the last PR, I had put the `target.requestPointerLock();` immediately inside the `onDragPointerDown` function which made dragging possible for Safari but crashed the possibility of text input within the number fields. The current solution I've found which both fixes the dragging issues for Safari and keeps the text input possibility the same as before is moving the `target.requestPointerLock();` to the event handler definition `onMove`:

Image

This seems to work reliably on Safari, Firefox and Chrome.

Let me know if it needs any changes!

### DIFF
--- a/frontend/src/components/widgets/inputs/NumberInput.svelte
+++ b/frontend/src/components/widgets/inputs/NumberInput.svelte
@@ -326,13 +326,14 @@
 			alreadyActedGuard = true;
 			isDragging = true;
 
-			// We need to request pointer lock as immediately as possible for Safari compatibility
+			// We begin dragging if the target supports pointer lock.
+			// We need to request pointer lock as immediately as possible for Safari compatibility.
 			const target = e.target || undefined;
-			if (!(target instanceof HTMLElement)) return;
+			if (target instanceof HTMLElement) {
+				target.requestPointerLock();
+				beginDrag();
+			}
 
-			target.requestPointerLock();
-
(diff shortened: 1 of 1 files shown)


frontend/src/components/widgets/inputs/NumberInput.svelte
```diff
@@ -325,7 +325,14 @@
...
-			beginDrag(e);
+
+			// We need to request pointer lock as immediately as possible for Safari compatibility
+			const target = e.target || undefined;
+			if (!(target instanceof HTMLElement)) return;
+
+			target.requestPointerLock();
+
+			beginDrag();
 			removeEventListener("pointermove", onMove);
```
Keavon, line 337: Isn't this now leaking the `pointermove` event now that we are sometimes early returning before this line? Seems like the `target.requestPointerLock();` and `beginDrag();` lines should be inside the if statement instead of the early return in its inverse.

seabeeberry:
After testing again, the "fixed" behavior in Safari is not working anymore, even when reverting to the initial commit. I have to close this PR as it does not solve the initial fix.

Keavon:
Weird:/ good luck continuing to investigate!

PR #2917, opened 2025-07-22
Add struct field visualization to the editor message hierarchy tree visualization on the website

MohdMohsin97:


Keavon:
Marking as draft while awaiting fixes for these erroneous outputs found with a regex search for `── \n` (as mentioned on Discord):

capture
capture

Keavon:
Requested changes as written in Discord, which I'm also copying here for reference:

- [x] In `transform_message_handler.rs`, we have an example of a tuple-style message enum variant which breaks the naming convention which should require struct-style variants. It doesn't look like this was caught by the node macro changes by throwing an error. Could you please look into this?
- [x] The `Message::Frontend` variant has a child for `FrontendMessage` which is working correctly, and another child that just says \`\` which needs to be filtered out.
  
  capture

PR #2918, opened 2025-07-22
Clean up code by using Iterator::collect() when constructing instance tables

Firestar99:
Quite a few nodes editing InstanceTables have this sort of code:
```rust
let mut result_table = RasterDataTable::default();
for mut image_frame_instance in image_frame.instance_iter() {
  /// ...
  result_table.push(image_frame_instance);
}
result_table
```

Now you can write them like this:
```rust
image_frame
  .instance_iter()
  .map(|mut image_frame_instance| {
    /// ...
    image_frame_instance
  })
  .collect()
```

Steams also have the advantage that in simple cases like these, the resulting vec can use `Iterator:size_hint()` to allocate the exact amount of entries it needs, whereas the for loop variant it may go through multiple resizes and overallocate memory.

### DIFF
--- a/editor/Cargo.toml
+++ b/editor/Cargo.toml
@@ -2,7 +2,7 @@
 name = "graphite-editor"
 publish = false
 version = "0.0.0"
-rust-version = "1.85"
+rust-version = "1.88"
 authors = ["Graphite Authors <contact@graphite.rs>"]
 edition = "2024"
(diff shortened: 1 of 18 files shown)


TrueDoctor (approved):
Code looks good @Keavon can comment on this more but I think this should be good to mereg

PR #2925, opened 2025-07-24
Shaders: add `gcore-shaders` and make `graster-nodes` no-std

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/2924

* Extracts several symbols from `gcore` into `gcore-shaders` and reexports them in `gcore`
* moves anything related to `ChoiceTypeStatic` from `registry` to a separate `choice_type` mod
* Makes `graster-nodes` be no-std with an `std` feature to mask out node code

So you can now build `graster-nodes` in a no-std environment, a prerequisite for shaders:
`cargo build -p graphene-raster-nodes --no-default-features`

## Why `gcore-shaders`

Simply put, there is still way too much stuff in `gcore` for it to become a no-std crate. So we either:
* create a new `gcore-shaders` crate that contains all of the `no-std` stuff, with very little masking needed
* turn `gcore` into no_std and move everything out of it
  * either I continue refactoring everything out of gcore into several crates, but that will eat up a lot of time, and runs the risk of no working shaders in the end
  * or we move everything to `gnode`, which would end up containing all the node definitions, registry and the like anyway. Stuff could get moved out from there later.
  * either way, a lot of migration needs to be hand-written
* mask out 99% of `gcore`, and you will forget to place the masks and break the build. Which is why I strongly prefer a separate crate.


TrueDoctor:
I agree that it makes sense to have an actual no_std crate and we should have ci testing to confirm that it always builds without the std feature. We can still slim down gcore over time and let you focus on implementing gpu acceleration

TrueDoctor (approved):

Firestar99:
There won't be any CI for the moment, but as soon as shaders work you need to compile shaders to compile graphite, and you'll have implicit CI for it being no_std.

PR #2929, opened 2025-07-24
remove inner enum from raster types

Firestar99:
I just want to throw in this suggestion and see what people think. 

From an external perspective, this is equivalent to the current implementation on master.


TrueDoctor:
Yeah, I like this. That is cleaner. I think we can just probably just merge this

Firestar99:
It was just WIP cause I thought there may be some discussion on this, merging now

PR #2930, opened 2025-07-24
Desktop: Directly upload frame buffer

### LINKED ISSUE #2541: Get the viewport to render beneath the web UI
We will be rendering the web UI to an offscreen canvas for the Graphite GUI—where a transparent hole is cut out of the viewport area—then stamping that over the top of the natively rendered viewport. The OS compositor needs to allow us to draw first the viewport, then the webview, every frame. This needs to work on Windows, Mac, and Linux.

TrueDoctor:
Partly closes #2541.

Bypass copying the frame buffer data to an intermediate vec for improved performance.
image

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -12,7 +12,6 @@ use winit::event::StartCause;
 use winit::event::WindowEvent;
 use winit::event_loop::ActiveEventLoop;
 use winit::event_loop::ControlFlow;
-use winit::event_loop::EventLoopProxy;
 use winit::window::Window;
 use winit::window::WindowId;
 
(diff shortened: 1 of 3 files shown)


timon-schelling:
after disabling gpu the correct way, I get perfect frames and chromium browser level performance and responsiveness.

desktop/src/main.rs
```diff
@@ -18,7 +18,7 @@ mod dirs;
 
 #[derive(Debug)]
 pub(crate) enum CustomEvent {
-	UiUpdate(FrameBuffer),
+	UiUpdate((wgpu::Texture, u32, u32)),
```
timon-schelling: Could be something like:

```
UiUpdate {
  Texture, 
  width,
  height,
}
```

desktop/src/app.rs
```diff
@@ -22,33 +22,36 @@ pub(crate) struct WinitApp {
 	pub(crate) cef_context: cef::Context<cef::Initialized>,
 	pub(crate) window: Option<Arc<Window>>,
 	cef_schedule: Option<Instant>,
-	ui_frame_buffer: Option<FrameBuffer>,
+	_ui_frame_buffer: Option<wgpu::Texture>,
 	window_size_sender: Sender<WindowSize>,
-	_viewport_frame_buffer: Option<FrameBuffer>,
+	_viewport_frame_buffer: Option<wgpu::Texture>,
```
timon-schelling, line 26: This will not be used In the future right?
I think you should remove both frame buffers

timon-schelling (approved):

PR #2932, opened 2025-07-24
Fix the Path tool erroneously showing editable geometry overlays belonging to hidden Path nodes

0HyperCube:
If the path node is hidden, don't use it in the path tool.


Keavon:
This correctly stops showing the overlays for a disabled Path node, but unfortunately does not show any overlays for a previous Path node. Example: draw with the Pen tool, insert a new downstream Path node but disable it, now the Path tool shows no editable overlays for the original shape like expected.

Keavon (changes requested):
This correctly stops showing the overlays for a disabled Path node, but unfortunately does not show any overlays for a previous Path node. Example: draw with the Pen tool, insert a new downstream Path node but disable it, now the Path tool shows no editable overlays for the original shape like expected.

Keavon:
Could you also please look into the test failure? Thank you.

0HyperCube:
No

Keavon (approved):

PR #2935, opened 2025-07-25
Filter transform instances

0HyperCube:
image
image

### DIFF
--- a/node-graph/gcore/src/transform_nodes.rs
+++ b/node-graph/gcore/src/transform_nodes.rs
@@ -7,6 +7,44 @@ use core::f64;
 use glam::{DAffine2, DVec2};
 
 #[node_macro::node(category(""))]
+async fn transform_two(
+	ctx: impl Ctx + CloneVarArgs + ExtractAll,
+	value: impl Node<Context<'static>, Output = VectorDataTable>,
+	translate: DVec2,
+	rotate: f64,
+	scale: DVec2,
+	skew: DVec2,
+	selection: impl Node<Context<'static>, Output = bool>,
+) -> VectorDataTable {
+	let matrix = DAffine2::from_scale_angle_translation(scale, rotate, translate) * DAffine2::from_cols_array(&[1., skew.y, skew.x, 1., 0., 0.]);
+
+	let footprint = ctx.try_footprint().copied();
+
+	let mut transform_target = {
+		let mut new_ctx = OwnedContextImpl::from(ctx.clone());
+		if let Some(mut footprint) = footprint {
+			footprint.apply_transform(&matrix);
+			new_ctx = new_ctx.with_footprint(footprint);
+		}
+		value.eval(new_ctx.into_context()).await
+	};
+
+	for (index, instance) in transform_target.instance_mut_iter().enumerate() {
+		let new_ctx = OwnedContextImpl::from(ctx.clone()).with_index(index);
+
+		let should_eval = selection.eval(new_ctx.into_context()).await;
+		if should_eval {
+			info!("Applying to {index}");
+			*instance.transform = matrix * *instance.transform;
+		} else {
+			info!("Skipping index {index}");
+		}
+	}
+
+	transform_target
+}
+
+#[node_macro::node(category(""))]
 async fn transform<T: ApplyTransform + 'n + 'static>(
 	ctx: impl Ctx + CloneVarArgs + ExtractAll,
 	#[implementations(
--- a/node-graph/preprocessor/src/lib.rs
+++ b/node-graph/preprocessor/src/lib.rs
@@ -24,6 +24,36 @@ pub fn expand_network(network: &mut NodeNetwork, substitutions: &HashMap<ProtoNo
 	}
 }
 
+/// Referencing the TaggedValue::IndexOperationFilter will expand to a node that contains the evaluate_index_operation_filter
+pub fn evaluate_index_operation_filter(network: &mut NodeNetwork) {
+	let mut new_nodes = Vec::new();
+	for node in network.nodes.values_mut() {
+		for input in &mut node.inputs {
+			if !matches!(input.as_value(), Some(TaggedValue::IndexOperationFilter(_)),) {
+				continue;
+			}
+			let node_id = NodeId::new();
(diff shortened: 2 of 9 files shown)


node-graph/preprocessor/src/lib.rs
```diff
@@ -24,6 +24,36 @@ pub fn expand_network(network: &mut NodeNetwork, substitutions: &HashMap<ProtoNo
 	}
 }
 
+/// Referencing the TaggedValue::IndexOperationFilter will expand to a node that contains the evaluate_index_operation_filter
+pub fn evaluate_index_operation_filter(network: &mut NodeNetwork) {
```
adamgerhant, line 28: I don't see why we want to convert the input to a node that outputs a boolean, instead of just directly using the IndexOperationFilter value within the transform node or whatever node needs to use it

Keavon: Because `IndexOperationFilter` is a type that can be fed by a value input from the Properties panel to give it convenient index-based ranges, while more flexibility can be achieved by passing in any arbitrary field-style node input branch that evaluates to bool per instance.

node-graph/gcore/src/transform_nodes.rs
```diff
@@ -6,6 +6,44 @@ use crate::{CloneVarArgs, Context, Ctx, ExtractAll, GraphicGroupTable, OwnedCont
...
+		value.eval(new_ctx.into_context()).await
+	};
+
+	for (index, instance) in transform_target.instance_mut_iter().enumerate() {
+		let new_ctx = OwnedContextImpl::from(ctx.clone()).with_index(index);
+
+		let should_eval = selection.eval(new_ctx.into_context()).await;
+		if should_eval {
+			info!("Applying to {index}");
+			*instance.transform = matrix * *instance.transform;
```
adamgerhant: we need to know the transforms before evaluating `value`

node-graph/gcore/src/transform_nodes.rs
```diff
@@ -6,6 +6,44 @@ use crate::{CloneVarArgs, Context, Ctx, ExtractAll, GraphicGroupTable, OwnedCont
...
+			footprint.apply_transform(&matrix);
+			new_ctx = new_ctx.with_footprint(footprint);
+		}
+		value.eval(new_ctx.into_context()).await
+	};
+
+	for (index, instance) in transform_target.instance_mut_iter().enumerate() {
+		let new_ctx = OwnedContextImpl::from(ctx.clone()).with_index(index);
+
+		let should_eval = selection.eval(new_ctx.into_context()).await;
```
adamgerhant: There is no need to clone the context in order to determine if the index is in the IndexOperationFilter range.

We also cant use indices to identify what element to transform, since then adding a new item into the stack will mess up the transforms. We need a unique id for each instance row

Keavon: We have decided to temporarily live with the limitations of indices before solving our equivalent of the [topological naming problem](https://wiki.freecad.org/Topological_naming_problem/en), which will likely be the hardest problem in all of Graphite that we will have to solve (through solid engineering, heuristics, and best-effort attempts).

0HyperCube: > There is no need to clone the context in order to determine if the index is in the IndexOperationFilter range.

I want to allow arbitrary functions to determine the selection like in blender. This requires cloning the context since the `into_context` function takes the context.
image

node-graph/gcore/src/transform_nodes.rs
0HyperCube: I have decided that the benefits of the footprint (mandlebrot set node) are outweighed by the disadvantages (cannot transform with pivot based on bounds, cannot transform based on arbitrary data, cannot rasterise based on the bounding box, etc).

adamgerhant: You can rasterize based on the bounding box since it would be part of the left to right data flow. @TrueDoctor just to make sure. I also dont believe transforming on the pivot would make much sense, as the transform should always be applied in that local layer space. To move the pivot, you would need to use a path node and then an inverse transform to undo. 

But I agree it is limiting that you could not apply transforms based on the data

node-graph/gcore/src/transform_nodes.rs
adamgerhant: @Keavon would it not be possible to simply add an unique id to each instance? I agree the topological problem is an issue for the path node, but I think it could be worked around for instances, or at least improved.

@0HyperCube I agree

node-graph/preprocessor/src/lib.rs
adamgerhant: The properties panel could be for the Transform Network node, which has an internal network implementation that has a node which takes the range and converts it to a boolean, and feeds this to the Transform protonode. This allows the user to transform based on any arbitrary logic, as hypercube intented.

0HyperCube: We already have a node that converts a range value to the boolean value in this node.

node-graph/gcore/src/transform_nodes.rs
0HyperCube: Keavon & I decided that it is more important to allow users to transform things as they like rather than to support the `Footprint`s.

node-graph/gcore/src/transform_nodes.rs
0HyperCube: The issue is in preserving the ids through different transforms, whilst also avoiding conflicting ids etc.

0HyperCube:
@Keavon closing because I'm not rebasing this yet again, I only just finished rebasing it the first time. Thanks.

Keavon:
I'm always happy to rebase things, just send me a message. I'll do this one now.

Keavon:
Rebase is done. That only affected about 5 small conflicts, so I'm glad to see my name updates aren't causing too many nasty conflicts.

0HyperCube:
@Keavon if you rebase, please make sure the code actually compiles…

Keavon:
This will instead be solved by the more generalized approach of #2988. But this was a really cool prototype, thanks for exploring this approach, it was valuable to learn from.

PR #2950, opened 2025-07-28
Desktop: Cleaner CEF bidirectional message implementation

timon-schelling:
This PR refactors and improves the existing mechanisms from the native-editor branch without altering any current behavior. It introduces a thin abstraction layer for passing raw byte‑array buffers between the renderer (JavaScript) and the main/browser process, to be used in future PRs.


timon-schelling:
Thanks for rebaseing :)

PR #2951, opened 2025-07-28
Add Handler for defered execution of messages

TrueDoctor:
This avoids blocking all message processing until a graph execution has been completed


editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1435,6 +1435,20 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 						},
 					})
 				}
+				// Some parts of the editior depend on these bounds to be present
```
adamgerhant: what parts? (answer: zooming to fill bounds) ~~Ideally the editor should not be aware of the bounds at all, since it treats the top left corner of the vieport at (0,0). Only the executor should know the bounds in order to evaluate the node graph.~~

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1435,6 +1435,20 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
...
+				let bounds = if self.graph_view_overlay_open {
+					self.network_interface.all_nodes_bounding_box(&self.breadcrumb_network_path).cloned()
+				} else {
+					self.network_interface.document_bounds_document_space(true)
+				};
+				if bounds.is_some() {
+					responses.add(DeferMessage::TriggerViewportReady);
```
adamgerhant: The bounds here refer to the bounds of all layers in document space. This has nothing to do with the viewport being ready

editor/src/messages/defer/defer_message_handler.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+#[message_handler_data]
+impl MessageHandler<DeferMessage, ()> for DeferMessageHandler {
+	fn process_message(&mut self, message: DeferMessage, responses: &mut VecDeque<Message>, _: ()) {
+		match message {
+			DeferMessage::AfterGraphRun { mut messages } => {
+				self.after_graph_run.extend(messages.drain(..).map(|m| (self.current_graph_submission_id, m)));
+			}
+			DeferMessage::AfterViewportReady { messages } => {
```
adamgerhant: It should not be possible to add messages to this buffer if there are currently bounds. An if statement in here would be a good prevention.

editor/src/messages/defer/defer_message_handler.rs
```diff
@@ -0,0 +1,36 @@
+use crate::messages::prelude::*;
+
+#[derive(Debug, Default, ExtractField)]
+pub struct DeferMessageHandler {
+	after_graph_run: Vec<(u64, Message)>,
+	after_viewport_resize: Vec<Message>,
+	current_graph_submission_id: u64,
```
adamgerhant, line 7: ~~I don't see why we have to associate queued messages with their execution id, since they should always be run after the next execution.~~ (answer: If multiple graph evaluations are queued, we do not want messages which require data from a later evaluation to be run on the first evaluation )

editor/src/messages/tool/common_functionality/graph_modification_utils.rs
```diff
@@ -153,8 +153,9 @@ pub fn merge_layers(document: &DocumentMessageHandler, first_layer: LayerNodeIde
 	});
 
 	responses.add(NodeGraphMessage::RunDocumentGraph);
```
adamgerhant, line 155: It might be cleaner to have the RunDocumentGraph message contain a parameter of messages to run after evaluation

editor/src/messages/defer/defer_message_handler.rs
```diff
@@ -0,0 +1,36 @@
+use crate::messages::prelude::*;
+
+#[derive(Debug, Default, ExtractField)]
+pub struct DeferMessageHandler {
```
adamgerhant, line 4: I would prefer if we could remove this and let the executor handle it

TrueDoctor: The idea is that this is a more general purpose solution and could also be used to implement more general callback such as timer based logic which we currently use JavaScript for.

editor/src/messages/defer/defer_message_handler.rs
TrueDoctor: That is not the responsibility of this message handler and would violate architectural separation. And it is possible to use this callback even if bounds are present and the code still works

editor/src/messages/portfolio/document/document_message_handler.rs
TrueDoctor: Do you have suggestions for a better name?
You can't meaningfully interact with the viewport until the bounds are present. We could potentially rename it to `ViewportNavigationReady`

editor/src/messages/tool/common_functionality/graph_modification_utils.rs
TrueDoctor: That is also what I suggested to you in our call, but I think this is a cleaner and more extensible approach

editor/src/messages/tool/tool_messages/spline_tool.rs
```diff
@@ -360,8 +360,6 @@ impl Fsm for SplineToolFsmState {
 				tool_options.stroke.apply_stroke(tool_data.weight, layer, responses);
 				tool_data.current_layer = Some(layer);
 
-				responses.add(Message::StartBuffer);
```
Keavon, line 363: This removal is different from all the other tool code changes, which besides this one all add a DeferMessage::AfterGraphRun call. Just bringing attention to this in case it's a bug. Otherwise this PR LGTM.

Keavon (approved):

editor/src/messages/portfolio/document/document_message_handler.rs
adamgerhant: We cannot interact with the viewport until the viewport bounds are present, which comes from the JS onResize event. This is different than the document bounds, which represent the bounds of the artwork. For example, create an artwork with infinite canvas and notice this message infinitely loops

PR #2961, opened 2025-07-31
Add stroke scaling option to Stroke node

### LINKED ISSUE #2960: Allow strokes to not be affecting by a scale transformation
The following description is one possible approach to implementing this, but it has not been decided if this is how it should be done. Anyone looking to implement this will first need to clear it with @Keavon from a design angle. See [this](https://github.com/GraphiteEditor/Graphite/pull/2961#issuecomment-3239632940). The description proposed by @chbornman is as follows:

---

## Summary
Add a "Scaling" field to the Stroke node that allows users to toggle whether strokes should scale with shape transformations or maintain a constant visual width.

## Motivation
Currently, strokes always scale with shape transformations. This feature gives users control over stroke behavior, which is a common requirement in vector graphics applications. Users can choose between:
- **Scale with Shape**: Strokes scale proportionally with transformations (default)
- **Non-Scaling**: Strokes maintain constant visual width regardless of transformations

## Implementation Details
- Add `StrokeScaling` enum with `ScaleWithShape` and `NonScaling` options
- Add new "Scaling" parameter to the Stroke node
- Update document migration to handle the new parameter
- Fix migration bug that caused index out of bounds error when upgrading existing documents

chbornman:
Add a "Scaling" field to the Stroke node that allows users to toggle whether strokes should scale with shape transformations or maintain a constant visual width.

* Add StrokeScaling enum with ScaleWithShape and NonScaling options
* Add new "Scaling" parameter to the Stroke node
* Update document migration to handle the new parameter
* Fix migration bug that caused index out of bounds error
* Remove duplicate apply_stroke call in shape_tool.rs

Closes #2960

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_properties.rs
@@ -27,7 +27,7 @@ use graphene_std::vector::misc::GridType;
 use graphene_std::vector::misc::{ArcType, MergeByDistanceAlgorithm};
 use graphene_std::vector::misc::{CentroidType, PointSpacingType};
 use graphene_std::vector::style::{Fill, FillChoice, FillType, GradientStops};
-use graphene_std::vector::style::{GradientType, PaintOrder, StrokeAlign, StrokeCap, StrokeJoin};
+use graphene_std::vector::style::{GradientType, PaintOrder, StrokeAlign, StrokeCap, StrokeJoin, StrokeScaling};
 use graphene_std::{GraphicGroupTable, NodeInputDecleration};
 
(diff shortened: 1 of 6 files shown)


Keavon:
Sorry for being backed up on code review, I'm working through my backlog now.

There is still an open question as to whether we want to have this via the Stroke node (a property of the stroke to not scale, as you've implemented) or via the Transform node (an option to skip affecting strokes, but also other future things like shadows and other appearance effects). I think it is too premature to decide on that quite yet, so I'm inclined to take no action on this yet rather than committing to potentially the wrong approach and having to deal with the migrations for undoing it later on. So with that in mind, and because this is a relatively small PR that could be recreated again in the future when the time is right, my preference is to close this for the moment. Thanks for your understanding!

PR #2963, opened 2025-07-31
Fix double click in path tool not working

TrueDoctor:


Keavon (approved):

Keavon:
This introduces a regression with the Text tool, which now places the interactive text entry preview at the document origin instead of the place you clicked.

PR #2966, opened 2025-08-01
Add a Path tool action to destructively merge nearby points by distance

4adex:
- If there are selected points which are close under certain threshold, then the button becomes active and allows to merge those points into a single point.
- If vector meshes is enabled, then only two points which are endpoints of paths can be joined into one.
- If vector meshes is not enabled, then any number of points which are under the threshold can be merged into one.
- One assumption is that all the points belong to the same layer.


Keavon:
Marking as draft while awaiting:

- [ ] Add to "Action" menu (to be initially added in a separate PR to be merged first)
- [ ] Fix case where a loop from a single segment is merged when its two segment anchors become one
- [ ] Make it work on all selected points but only merge (any number of separate) clumps of those which are mutually nearby one another

Keavon:
@4adex these three todo items never got completed, right? Is there anything we could still salvage from this PR?

4adex:
Yes we can still have this completed, sorry I got busy in my acads after my semester started. But I will surely be free this winter to work on the actions panel and integrate the blender inspired actions that I previously implemented.

Keavon:
It will be great to have you back! Anything we can complete this month will likely make it into the beta launch that's scheduled for January to help Graphite make a better impression on its many soon-to-be new users with each refinement we can merge before then. Thanks!

4adex:
Yes I will be working on this, will send an update by Monday.

4adex:
#3038 will also be salvaged after this is done.

PR #2968, opened 2025-08-02
Partly fix when the "Make Path Editable" button is shown as enabled

4adex:
Follow-up fix to the feature implemented in #2900.

"Make Path editable" button should always be enabled, unless:
- There are multiple selected layers instead of just one
- The selected object isn't vector so adding the Path node at the end of its layer chain would produce a type error
- There is already a Path node at the end of the selected layer chain which is empty, meaning it has no diffs of edits (otherwise a non-empty one should still allow adding another one)

https://discord.com/channels/731730685944922173/731738914812854303/1399537281400045668


Keavon:
Marking as draft while awaiting fix:

> The last issue that needs fixing is that the Path node must be the rightmost node in a chain. Currently, the logic will gray out the button even if you have a node such as Spline following an empty Path node.

Keavon:
I'm going to merge now with that known flaw and have you resolve it in a followup PR, thanks.

Keavon (approved):

PR #2983, opened 2025-08-03
Add an error dialog that is displayed if there is an error when exporting artwork

0HyperCube:
If an error is encountered during export, the error dialogue should be shown. However currently it is not. This is due to the fact that the export dialogue puts a `FrontendMessage::DisplayDialogDismiss` after running the exporting process, causing the error to be immediately dismissed.

Also
- refactor preferences message handler to reduce duplication
- remove some unnecessary allocations


0HyperCube:
There is very little advantage to this PR.

PR #2984, opened 2025-08-03
Shaders prep2

Firestar99:


timon-schelling:
From what I can tell, your change to the nix flake makes the rust install non declarative and non reproduceable. Which tool does cargo-gpu require?

Keavon:
Before this merges, could a real/descriptive name for the PR be added please?

Firestar99:
@timon-schelling cargo-gpu assumes you installed rust via rustup and does the following:
* uses rustup to check if a toolchain is installed, and if not install it
* uses rustup to check if a list of components is installed for set toolchain, and if not install it
* uses `cargo +toolchain` to select a certain toolchain for cargo commands

The rust overlay we're currently using is not based upon rustup, and instead links `cargo` directly to the `cargo` of the toolchain specified within nix. So `cargo +toolchain` fails with `cargo` complaining that you should call it through rustup. 

Just using rustup is the easiest way to resolve it. But if you have an idea on how to install two parallel toolchains and switch between them, feel free to take this one over.

timon-schelling:
I will take a look how we can do the same thing declaratively. :) In the meantime please do not include your flake changes in this PR.

Firestar99:
Rebased it and excluded the nix commit, that one will be merged whenever #2985 is ready. @Keavon @TrueDoctor which makes this PR ready for review and merge

TrueDoctor (approved):

PR #2994, opened 2025-08-05
Refactor the node network wires to use Kurbo

indierusty:
Part of #2635


Keavon:
capture

Expected:

capture

Observed:

capture

PR #2995, opened 2025-08-05
Fix Alt-based quick measurement to show document space coordinates instead of viewport space

### LINKED ISSUE #2941: Quick measurement with Alt is incorrectly showing viewport coordinates instead of document coordinates
Draw two separated rectangles. With the Select tool active and one of them selected, hold Alt and hover over the other one. You will see a numerical distance show their separation between one another. Zoom in and measure again. You'll now see how the two are showing a different value, proportional to the viewport's new zoom level increase. This needs to be shown in document space, not viewport space, to be a useful feature.

vineetpuranik:
Fixes #2941

This PR addresses the bug where the quick measurement feature (triggered by holding Alt while hovering over a shape) displayed distances in viewport coordinates instead of document coordinates. This caused the measured distance to vary incorrectly based on zoom level.

### Changes made
- Updated the logic to compute the bounding boxes (`selected_bounds`, `hovered_bounds`) in document space using `bounding_box_document`.
- Applied the `document_to_viewport` transform only once during rendering, preserving measurement accuracy.
- Refactored variable names and added clear inline comments for maintainability.

### Before
- Measured distances between shapes would change when zooming in/out.
- Dashed overlays were drawn based on transformed (viewport) coordinates, misaligning with the actual shapes.

### After
- Measured distances remain consistent regardless of zoom level.
- Overlays are correctly drawn by transforming document-space bounding boxes at render time.

### DIFF
--- a/editor/src/messages/tool/common_functionality/measure.rs
+++ b/editor/src/messages/tool/common_functionality/measure.rs
@@ -43,18 +43,19 @@ fn draw_line_with_length(line_start: DVec2, line_end: DVec2, transform: DAffine2
 	}
 }
 
-/// Draws a dashed outline around a rectangle to visualize the AABB
+/// Draws a dashed outline around the given rectangle (assumed to be in document space).
+/// The provided transform is applied to convert coordinates (e.g., to viewport space) during rendering.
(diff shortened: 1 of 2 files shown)


Keavon:
!build

Keavon:
Please also leave a comment in the associated issue so, as a conversation participant, GitHub allows me to assign it to you.

PR #2998, opened 2025-08-05
Desktop: Fix Vello rendering colors with doubly-applied gamma by converting to linear in the shader

### LINKED ISSUE #2971: Fix Vello canvas rendering not being displayed in sRGB
Currently when using vello on Desktop, all colors look to white like if gamma was applied twice

TrueDoctor:
Vello renders it's values to an RgbaUnorm texture
but if we try to use this in the main rendering
pipeline which renders to an Srgb surface gamma
mapping is applied twice.

Closes #2971

### DIFF
--- a/desktop/src/render/composite_shader.wgsl
+++ b/desktop/src/render/composite_shader.wgsl
@@ -51,8 +51,10 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
 
 	let vp_cord = (in.tex_coords - constants.viewport_offset) * constants.viewport_scale;
 
-	let ov = textureSample(t_overlays, s_diffuse, vp_cord);
-	let vp = textureSample(t_viewport, s_diffuse, vp_cord);
+	let ov_raw = textureSample(t_overlays, s_diffuse, vp_cord);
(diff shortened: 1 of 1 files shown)


desktop/src/render/composite_shader.wgsl
```diff
@@ -51,8 +51,10 @@ fn fs_main(in: VertexOutput) -> @location(0) vec4<f32> {
 
 	let vp_cord = (in.tex_coords - constants.viewport_offset) * constants.viewport_scale;
 
-	let ov = textureSample(t_overlays, s_diffuse, vp_cord);
-	let vp = textureSample(t_viewport, s_diffuse, vp_cord);
+	let ov_raw = textureSample(t_overlays, s_diffuse, vp_cord);
```
Keavon:
```suggestion
	Vello renders its values to an `RgbaUnorm` texture, but if we try to use this in the main rendering pipeline
	which renders to an `Srgb` surface, gamma mapping is applied twice. This converts back to linear to compensate.
	let ov_raw = textureSample(t_overlays, s_diffuse, vp_cord);
```

PR #3006, opened 2025-08-05
Desktop: Open links in default browser and prevent popups

timon-schelling:
Open links in default browser and prevents cef from opening popups

logs before the second commit:
```
2025-08-05T18:11:29.118324Z ERROR graphite_desktop::cef::internal::browser_process_life_span_handler: Browser tried to open a popup at URL: https://graphite.rs/
2025-08-05T18:11:29.335806Z ERROR graphite_desktop::cef::internal::browser_process_life_span_handler: Browser tried to open a popup at URL: https://graphite.rs/
2025-08-05T18:11:48.109077Z ERROR graphite_desktop::cef::internal::browser_process_life_span_handler: Browser tried to open a popup at URL: https://graphite.rs/
2025-08-05T18:11:48.273723Z ERROR graphite_desktop::cef::internal::browser_process_life_span_handler: Browser tried to open a popup at URL: https://graphite.rs/
```


desktop/src/app.rs
```diff
@@ -55,6 +55,10 @@ impl WinitApp {
 	}
 
 	fn send_messages_to_editor(&mut self, mut responses: Vec<FrontendMessage>) {
+		if responses.is_empty() {
```
TrueDoctor: This should stay where it was previously since the extract if loops could remove messages and the only expensive part about this is sending the cef message

desktop/Cargo.toml
```diff
@@ -38,3 +38,4 @@ glam = { workspace = true }
 vello = { workspace = true }
 derivative = { workspace = true }
 rfd = { workspace = true }
+open ={ workspace = true }
```
TrueDoctor: missing whitespace
```suggestion
open = { workspace = true }
```

desktop/src/app.rs
timon-schelling: Good catch

Keavon: Please write a comment about that so it doesn't get overlooked again.

TrueDoctor (approved):

PR #3030, opened 2025-08-07
Improve tools

### LINKED ISSUE #3028: Cannot insert new point in gradient with double click
Cannot insert new point in gradient with double click. The following messages are sent:
- Gradient PointerDown
  - StartTransaction
- Gradient PointerUp
  - EndTransaction
  - Enqueue graph execution
- Doubleclick
  - InsertPoint fails because of having no document metadata.
- Execute graph

The cause of the insert not working is because the `DocumentMetadata` is blank. This is because the clone of the `NodeNetworkInterface` is manually implemented to not clone the `DocumentMetadata` and just reset it to the default empty value (for reasons known only to @adamgerhant).

Since by issuing the `StartTransaction` `EndTransaction` in series, the cloned version of the `NodeNetworkInterface` is used, the data is completely cleared. This is likely why a new graph execution is enqueued.

However for reasons of @TrueDoctor's fun new message scheduling system, the DoubleClick message is actually executed before the `DocumentMetadata` can be populated.

0HyperCube:
Closes #3028
Depends on #2983

Improvements:
- Separate some of the functionality of the select tool into separate files (only partially done).

Fixes:
- <kbd>alt</kbd> duplicating in the select tool sometimes applies an offset.
- `update_hints` didn't have access to the `tool_data` so was implemented in a hacky way in the select tool and path tool
- Offsets when drawing on non-origin artboard in:
  - line tool
  - freehand tool
  - pen tool
  - text tool
- Gradient tool double click insert stop #3028

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -52,5 +52,6 @@
 	"files.insertFinalNewline": true,
 	"files.associations": {
 		"*.graphite": "json"
-	}
+	},
+	"rust-analyzer.checkOnSave": false
 }
--- a/editor/src/messages/dialog/dialog_message.rs
+++ b/editor/src/messages/dialog/dialog_message.rs
@@ -8,8 +8,6 @@ pub enum DialogMessage {
 	ExportDialog(ExportDialogMessage),
 	#[child]
 	NewDocumentDialog(NewDocumentDialogMessage),
-	#[child]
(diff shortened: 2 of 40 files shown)


0HyperCube:
I don't think it is worth pursuing this approach any more. The tools already work at least half of the time.

PR #3031, opened 2025-08-08
Simplify Bezier-rs using poly-cool as the root finder

jneem:
This removes symmetrical_basis, which was publicly exported, and so it's a breaking change.

The big caveat here is that poly-cool is not thoroughly tested. I plan to do that next, though, and I think any issues should be fixable without breaking semver compatibility.

### DIFF
--- a/libraries/bezier-rs/src/bezier/lookup.rs
+++ b/libraries/bezier-rs/src/bezier/lookup.rs
@@ -259,14 +259,13 @@ impl Bezier {
 
 		for time in critical {
 			let distance = self.evaluate(TValue::Parametric(time)).distance_squared(point);
-			dbg!(distance, time);
 			if distance < min_dist_squared {
 				closest = time;
 				min_dist_squared = distance;
 			}
 		}
 
-		if dbg!(self.evaluate(TValue::Parametric(1.)).distance_squared(point)) < min_dist_squared {
+		if self.evaluate(TValue::Parametric(1.)).distance_squared(point) < min_dist_squared {
 			closest = 1.;
 		}
 		closest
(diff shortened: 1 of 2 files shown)


libraries/bezier-rs/Cargo.toml
```diff
@@ -13,9 +13,13 @@ homepage = "https://github.com/GraphiteEditor/Graphite/tree/master/libraries/bez
 repository = "https://github.com/GraphiteEditor/Graphite/tree/master/libraries/bezier-rs"
 documentation = "https://graphite.rs/libraries/bezier-rs/"
 
+[features]
+std = ["glam/std"]
```
jneem, line 17: I needed this to run just the bezier-rs tests (with `cargo test -p bezier-rs --features=std`), because without it there were lots of undefined references to math functions.

libraries/bezier-rs/src/bezier/lookup.rs
```diff
@@ -354,7 +354,8 @@ mod tests {
 		assert_eq!(bezier1.project(DVec2::new(100., 100.)), 1.);
 
 		let bezier2 = Bezier::from_quadratic_coordinates(0., 0., 0., 100., 100., 100.);
-		assert_eq!(bezier2.project(DVec2::new(100., 0.)), 0.);
+		assert_eq!(bezier2.project(DVec2::new(99.99, 0.)), 0.);
```
jneem, line 356: I had to change this test slightly, because the old point had two nearest points (at t=0 and t=1), and the new root-finder was flagging t=0.999999999 as the minimizer.

Keavon (approved):
Thank you!

PR #3034, opened 2025-08-09
Desktop: Add File > Save As…

timon-schelling:
- Improve save document semantics
  - **Removes name from values saved in document (No migration should be needed)**
  - Instead uses file name (without `.graphite`) as the document name
  - Document name is updated when the document path is updated
  - Document path is reset if the name is change by the user
- Add `Save as` action
  - Adds Save As Menu entry and shortcut 
  - ~~**Issue: Save as shourtcut `Ctrl + Shift + S` not working (help wanted)**~~


TrueDoctor:
@Keavon We fixed the open issue this is now waiting on your thoughts regarding how the save process flow should work and if the implemented behaviour is desired

Keavon:
Let's discuss in a call so I know more about what's to be considered.

Also, this PR has a conflict.

TrueDoctor:
@Keavon what is the status on this? I can review the code you just have to say if you are happy with the behaviour

Keavon (approved):

PR #3038, opened 2025-08-10
Add button for gluing segments together in the Path tool

4adex:
If there are two segments which are close under a certain threshold and are selected, they can be glued together by a certain toggle or button.
image


Keavon:
Can this PR be salvaged?

PR #3052, opened 2025-08-13
Desktop: Fix UI and viewport blending

### LINKED ISSUE #3043: Fix UI and Viewport blending
Image

Top window is graphite web
Right window is graphite desktop running with software rendering backend
Left window is graphite desktop running with nvidia gpu (needs to be fixed)

timon-schelling:
Closes #3043


TrueDoctor:
This currently applies srg gamma conversions multiple times. Shouldn't we just create a srgb version of the viewport value and keep the others as is?

timon-schelling:
Sure, found a way of writing that relatively clean.

Sadly this doesn't active the exact result like web. With "fade artwork" at 80% white color artwork results in #4E4E4E in web and #4C4C4C in desktop.

There is also a problem with the dots. Node graph view background dots on .graph:before are not transparent when rendered with hole punch enabled. "mix-blend-mode: screen" doesn't work like in web in that case. Cef returns [34, 34, 34, 255] (alpha always 255) for these dots when fade artwork is 80%.

timon-schelling:
image

timon-schelling:
If I understand the chromium docs correctly it will do blending in the renderer’s working/display color space (not strictly sRGB) with source-over (Porter-Duff).
This PR tries to do the same. Result is not quit the same, slight visual change when disabling vello.

Keavon:
You'd mentioned the blend mode on the parts of the graph. That won't be something we have any control over, since blend modes fundamentally have no effect at the bottom of a compositing stack. We'll have to wait for our custom-rendered node graph UI overlay before we can get that to draw similarly, but we can just make it look as decent as possible in the meantime.

TrueDoctor:
Is this ready for review / merge then?

timon-schelling:
Yes, the issue with the dots should be solved separately, should only concern frontend code.

timon-schelling:
Just noticed that overlays look more like in web when bleeding overlays and viewport in linear and then bleding that result in srgb with the ui.
Also added a early return for the common case were overlays and ui are transparent.

timon-schelling:
@TrueDoctor ready for final review and merge :)

TrueDoctor (approved):

PR #3053, opened 2025-08-14
Add workaround for a Safari bug where our number input fields can't be released from click-and-drag

### LINKED ISSUE #2038: Click-and-drag number fields can't be released in Safari

seabeeberry:
Closes #2038

I found a way to put a fallback into our `pointerUp` const (where we exit the pointerlock) to fire a `pointerLockChange` explicitly when this event is not triggered automatically (in our case when using Safari).

I've tested this on Safari, Firefox and Chrome and it seems this fallback both fixes/keeps the intended functionality without breaking the functionality of inputting per keyboard into the NumberInputs. Below a Safari demo.

https://github.com/user-attachments/assets/aa0221ff-b167-4496-aa21-8b3268822ca9

### DIFF
--- a/frontend/src/components/widgets/inputs/NumberInput.svelte
+++ b/frontend/src/components/widgets/inputs/NumberInput.svelte
@@ -365,6 +365,11 @@
 			cumulativeDragDelta = 0;
 
 			document.exitPointerLock();
+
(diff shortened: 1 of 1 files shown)


Keavon:
!build

PR #3059, opened 2025-08-17
Open document from cli and add some files to .gitignore

### LINKED ISSUE #3054: Allow opening document passed as cli arg

c-mateo:
Closes #3054

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -12,6 +12,7 @@ use graphene_std::raster::Image;
 use graphite_editor::application::Editor;
 use graphite_editor::consts::DEFAULT_DOCUMENT_NAME;
 use graphite_editor::messages::prelude::*;
+use std::env;
 use std::fs;
 use std::sync::Arc;
 use std::sync::mpsc::Sender;
@@ -221,6 +222,32 @@ impl ApplicationHandler<CustomEvent> for WinitApp {
 				self.dispatch_message(message);
 			}
 			CustomEvent::MessageReceived(message) => {
+				if let Message::Portfolio(PortfolioMessage::Init) = &message {
+					let args: Vec<String> = env::args().collect();
+
+					// Try to open a file passed as the first command line argument
+					if args.len() > 1 {
+						let event_loop_proxy = self.event_loop_proxy.clone();
+						let path = std::path::PathBuf::from(&args[1]);
+
+						if !path.exists() {
+							tracing::warn!("The first argument is not a valid path or doesn't exist: {}", path.display());
+							return;
+						}
+
+						let name = path.file_stem().and_then(|s| s.to_str()).map(|s| s.to_string());
+
+						// Load the file in a separate thread to avoid blocking the main thread
+						let _ = thread::spawn(move || {
+							let Some(content) = load_string(&path) else { return };
+							let message = PortfolioMessage::OpenDocumentFile {
+								document_name: name.unwrap_or(DEFAULT_DOCUMENT_NAME.to_string()),
+								document_serialized_content: content,
+							};
+							let _ = event_loop_proxy.send_event(CustomEvent::DispatchMessage(message.into()));
+						});
+					}
+				}
 				if let Message::InputPreprocessor(_) = &message {
 					if let Some(window) = &self.window {
 						window.request_redraw();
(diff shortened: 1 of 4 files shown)


Keavon:
Neither of your gitignore changes seem to make sense to me. Can you justify or remove them?

timon-schelling:
I would generally prefer a proper cli implementation from the beginning (with clap etc.), not a temporary solution for only implementing file opening.

@TrueDoctor I would need to implement basically all of this in my wrapper crate pr (https://github.com/GraphiteEditor/Graphite/pull/3039). Would you be fine with waiting with features that will have a easy way of implementing once the wrapper crate separation is done?

c-mateo:
> Neither of your gitignore changes seem to make sense to me. Can you justify or remove them?

I added those lines because I haven't created those folders but they were there as a result of a compilation step. Let me remove them.

Keavon:
> I would generally prefer a proper cli implementation from the beginning (with clap etc.), not a temporary solution for only implementing file opening.
> 
> @TrueDoctor I would need to implement basically all of this in my wrapper crate pr (#3039). Would you be fine with waiting with features that will have a easy way of implementing once the wrapper crate separation is done?

Are you in favor of closing this, then? I'll defer to your thinking on this.

timon-schelling:
@c-mateo are you still interested in working on this? Would need a complete rework now that 3039 is merged. 

And I would still prefer implementing this with clap.

Keavon:
I'll close this for now then, @timon-schelling's requested changes can be either a new PR or be part of this PR and reopened.

c-mateo:
Now the cli is properly implemented using clap and there's an option to select whether you desire to use cef's shared texture as Timon suggested on Discord. You can give me some ideas for more commands and options before merging

desktop/src/app.rs
```diff
@@ -91,6 +94,20 @@ impl App {
...
 				self.send_or_queue_web_message(bytes);
 			}
+			DesktopFrontendMessage::Init => {
+				tracing::info!("Initializing application...");
+				if let Some(paths) = self.cli_files.clone() {
+					for path in paths {
+						tracing::info!("Opening file from command line: {}", path.display());
+						if let Ok(content) = std::fs::read(&path) {
+							let message = DesktopWrapperMessage::OpenFile { path, content };
+							self.dispatch_desktop_wrapper_message(message);
```
timon-schelling: Not a blocker but will need to be rewritten after #3247.

desktop/src/app.rs
```diff
@@ -91,6 +94,20 @@ impl App {
 				};
 				self.send_or_queue_web_message(bytes);
 			}
+			DesktopFrontendMessage::Init => {
```
timon-schelling: We need to ensure that all auto saved documents are loaded at this point. You assume this message is handled after Portfolio::Init, this works because of the current ordering of message dispatch in desktop wrapper, but this assumption should not be made in my opinion. Maybe a special `FrontendMessage::OpenDocumetnsSuppliedAtStartup` (I'm tired, could not think of a better name) would be a better solution. Could also be used for opening documents from the URL in web.

desktop/src/cli.rs
```diff
@@ -0,0 +1,19 @@
...
+	No,
+}
+
+#[derive(Debug, Parser)]
+#[clap(name = "graphite-cli", version)]
+pub struct Cli {
+	pub files: Option<Vec<PathBuf>>,
+
+	#[arg(long, value_enum, default_value_t = UIAceleratedPainting::Auto,  help = "Enable UI accelerated painting using GPU")]
+	pub ui_accelerated_painting: UIAceleratedPainting,
```
timon-schelling: I would prefer only a `--disable-ui-accaleration` flag. Users should never run into a situation were acceleration is possible but we disable it.

timon-schelling (commented):
Generally good direction

desktop/src/app.rs
c-mateo: I barely modified that section, just changed `dispatch_desktop_wrapper_message` for `app_event_scheduler.schedule` because calling `dispatch_desktop_wrapper_message` directly prevents the native open file dialog from showing up. I'm not sure if should have done something else but the cli is working.

timon-schelling:
Looks good, I will take a closer look after the developer meetup tomorrow and probably merge. Thanks

PR #3061, opened 2025-08-17
Fixed-3045

### LINKED ISSUE #3045: Opening then closing the node graph permanently makes the artboard label overlay disappear
Notice the artboard title in its top left corner, by default it says "Artboard" but it'll say what you name its layer. (Renaming its layer also fails to update the name until the user moves the mouse pointer by at least 1px, which should also be fixed to be immediate upon hitting <kbd>Enter</kbd> to accept the new name.)

Open the node graph and close it again, notice it has gone away. Keep interacting in the document and it won't return.

The only way to get it to return is to switch to another document tab and back.

meet0806:
Closes #3045


meet0806:
@Keavon Ping for review 🙏🏼

Keavon:
!build

PR #3067, opened 2025-08-19
Improve snapping performance

TrueDoctor:
Fixes snapping performance regressions introduced in #3036 
- Use a fat line algorithm to speed up cubic cubic intersection checks
- Implement termination criteria for degenerate intersection check cases
- Implement faster nearest point calculation
- Loosen bounding boxes which are only used for pruning

### DIFF
--- a/editor/src/messages/tool/common_functionality/snapping/layer_snapper.rs
+++ b/editor/src/messages/tool/common_functionality/snapping/layer_snapper.rs
@@ -325,6 +325,43 @@ struct SnapCandidatePath {
 }
 
 impl SnapCandidatePath {
+	/// Calculates the point on the curve which lies closest to `point`.
+	///
+	/// ## Algorithm:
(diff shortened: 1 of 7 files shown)


editor/src/messages/tool/common_functionality/snapping/layer_snapper.rs
```diff
@@ -322,6 +324,99 @@ struct SnapCandidatePath {
...
+	///     0  0.25 0.5 0.75 1  (parameter t values)
+	///         ^       ^
+	///         |   |   |
+	///        min mid max
+	///    Find closest scan point
+	///
+	///    Refine left region using binary search:
+	///
+	///            ●------●------●
+	///           0.25  0.357   0.5
```
Keavon:
```suggestion
	///           0.25  0.375   0.5
```

PR #3068, opened 2025-08-20
Improve performance for finding path segment intersection using loose bbox

indierusty:


TrueDoctor:
Was superseded by #3067

PR #3070, opened 2025-08-20
Fix regressions related to path tool

4adex:
Solved these following minor regressions in the path tool -
- `Ctrl A` in path tool, should select all segments also if segment select mode is enabled. (https://discord.com/channels/731730685944922173/881073965047636018/1406378997830521102)
- The overlay of hovering segment remains after deleting the segment in segment select mode. (https://discord.com/channels/731730685944922173/881073965047636018/1406376716657754263)
- Fix the flawed implementation of `select_points_by_manipulator_id` and changed it from every location. (https://discord.com/channels/731730685944922173/881073965047636018/1399272679952875632)
- Path tool overlays (specifically the three modes introduced in https://github.com/GraphiteEditor/Graphite/pull/2264) were erroneously shown when there were duplicate layers, that has been fixed now by passing `LayerNodeIdentifier` along with the `PointId` for uniqueness of point.
- Flip Smooth Sharp had bloated code from previous implementation which used to check closest point and then change it to smooth or sharp, it was later changed to flipping for all selected points (https://github.com/GraphiteEditor/Graphite/pull/2498)
- Flip smooth sharp didn't worked for points of other layers which are also in the selection (it does now).
- Vector mesh support for `Ctrl + J` join path feature, if vector mesh enabled, can join non terminal points also.


Keavon (approved):

Keavon:
This PR introduced a regression: #3505.

PR #3081, opened 2025-08-22
Desktop: Add window buttons for linux

timon-schelling:
Implements basic window buttons for linux that look less out of place (on most desktops) than using Windows™ buttons
image


timon-schelling:
inspired by vscode buttons on linux
image
and Graphite menu bar

Keavon (approved):

PR #3089, opened 2025-08-23
Add support for adding/removing a graph's primary imports/exports separately from the secondary

adamgerhant:
Enables modifying the exposed primary input/output


Keavon:
I know this is still a draft but I'll just give a little in-progress feedback while I'm checking it out:

- The primary imports should have its reorder handle removed.
- There should be a horizontal line drawn between the primary import and the secondary imports.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -6064,11 +6122,21 @@ pub enum TypeSource {
 	Error(&'static str),
 }
 
-impl Default for TypeSource {
-	fn default() -> Self {
-		Self::Error("no source")
-	}
-}
+// /// Represents the result of a type query for an input or output connector.
```
Keavon: Should this be deleted?

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -6590,49 +6659,6 @@ struct InputTransientMetadata {
 	// types: populated for each protonode after each
 }
 
-// TODO: Eventually remove this migration document upgrade code
-fn migrate_output_names<'de, D: serde::Deserializer<'de>>(deserializer: D) -> Result<Vec<String>, D::Error> {
```
Keavon, line 6594: Is this moved elsewhere (that I can't find) or has it been fully removed? Unless its purpose has been obviated, we do still need these migrations.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
adamgerhant: Yes, these are notes for another PR to improve the editor typing system.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
adamgerhant: Removed. There is no need for these migrations, since the output name of a protonode is never visible. Also the output name should not be a type to begin with, as it is automatically set to be the type if the output name is an empty string.

PR #3090, opened 2025-08-25
Shaders: more `graster-nodes` no-std fixups

Firestar99:
prep work for shaders, does not contain any shader stuff:
* fix `cargo b -p graster-nodes --no-default-features` being broken on master
* random stuff I need for shaders to work
* change `Table<Color>` -> `Color` where only the first row is used anyway

### DIFF
--- a/node-graph/graster-nodes/src/adjustments.rs
+++ b/node-graph/graster-nodes/src/adjustments.rs
@@ -11,6 +11,8 @@ use graphene_core::table::Table;
 use graphene_core_shaders::color::Color;
 use graphene_core_shaders::context::Ctx;
 use graphene_core_shaders::registry::types::{Angle, Percentage, SignedPercentage};
+#[cfg(not(feature = "std"))]
+use num_traits::float::Float;
 
 // TODO: Implement the following:
 // Color Balance
--- a/node-graph/graster-nodes/src/adjustments.rs
+++ b/node-graph/graster-nodes/src/adjustments.rs
@@ -128,7 +128,7 @@ fn make_opaque<T: Adjust<Color>>(
 		if color.a() == 0. {
 			return color.with_alpha(1.);
 		}
-		Color::from_rgbaf32(color.r() / color.a(), color.g() / color.a(), color.b() / color.a(), 1.).unwrap()
+		Color::from_rgbaf32_unchecked(color.r() / color.a(), color.g() / color.a(), color.b() / color.a(), 1.)
 	});
 	input
 }
@@ -363,7 +363,7 @@ async fn black_and_white<T: Adjust<Color>>(
 		// TODO: Fix "Color" blend mode implementation so it matches the expected behavior perfectly (it's currently close)
 		let color = tint.with_luminance(luminance);
 
-		let color = Color::from_rgbaf32(color.r(), color.g(), color.b(), alpha_part).unwrap();
+		let color = Color::from_rgbaf32_unchecked(color.r(), color.g(), color.b(), alpha_part);
 
 		color.to_linear_srgb()
 	});
--- a/node-graph/graster-nodes/src/adjustments.rs
+++ b/node-graph/graster-nodes/src/adjustments.rs
@@ -7,6 +7,7 @@ use core::fmt::Debug;
 use graphene_core::gradient::GradientStops;
 #[cfg(feature = "std")]
 use graphene_core::raster_types::{CPU, Raster};
+#[cfg(feature = "std")]
 use graphene_core::table::Table;
 use graphene_core_shaders::color::Color;
 use graphene_core_shaders::context::Ctx;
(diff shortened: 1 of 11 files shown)


node-graph/graster-nodes/src/adjustments.rs
```diff
@@ -304,7 +307,7 @@ async fn black_and_white<T: Adjust<Color>>(
 		GradientStops,
 	)]
 	mut image: T,
-	#[default(Color::BLACK)] tint: Table<Color>,
```
Firestar99: Please verify that changing `Table<Color>` -> `Color` indeed changes absolutely nothing, since with the table you're just using the first entry of the table anyway. Affects both this `black_and_white` node and the `color_overlay` node.

Keavon: This change shouldn't occur, we do need it to take a Table<Color> since that is the only data type relating to colors that is processed within the node graph.

Firestar99: I now replaced `Table<Color>` with `Option<Color>` (instead of just `Color`) and added an `into_node!(from: Table<Color>, to: Option<Color>),`, since we already have a `impl From<Table<Color>> for Option<Color>`. So there's no need to adjust anything in the preprocessing!

However, both `Color` and `Option<Color>` makes the tint property disappear in the properties panel.

Firestar99: I tried adding `Option<Color>` to `TaggedValue` to the UI to show, but I couldn't get it to work. Would appreciate it if someone with more editor experience took a look, so I can focus on other important things.

0HyperCube: In PR #3048, Keavon removed `Color` and `Option<Color>` from the properties system and also from the TaggedValue (used to serialize properties). This seems to me to be a bad idea. I wonder what your opinions are on this matter?

0HyperCube: I'm happy to revert this PR for you @Firestar99?

Firestar99: I'd only need the properties panel to work on an `Option<Color>` (and may as well add Color as well?), just like it currently works on an `Table<Color>`. In theory, I wouldn't even need the properties panel to support `Option<Color>`, I'd only need it to recognize it as an `Table<Color>` since we can implicitly convert from that to `Option<Color>` via an `IntoNode`.

0HyperCube: It would be trivial to support `Option<Color>` and `Color` in the properties panel. These were all previously supported before Keavon decided to remove them for some reason in #3048. I do not understand why this was done.

It seems like Keavon would like nodes that just take in a single colour to instead take in a Table<Color> and then convert it to the proper type. Therefore I assume you will need to add into nodes everywhere. I'm not sure if this is possible with GPU nodes?

Firestar99: `Table` uses `Vec` from `alloc`, and we don't have `alloc` on the GPU. I could make a `GpuTable` that just like `Table` has multiple entries, but if we just need the first value anyway, why bother? Having it be `Color` or `Option<Color>` would be far simpler.

TrueDoctor: The into nodes should now be added automatically if needed. The gpu node declares it's input as `Color` or `Option<Color>` but you can still feed a `Table<Color>` as input.

0HyperCube: I agree. Even on the CPU side, there is no need for nodes that work only on a single colour to accept a table.

TrueDoctor: And this is approved by Keavon (at least for the GPU case). @0HyperCube if you want you can make a pr to just revert the removal of `Color` and `Option<Color>` from the `node_properties`

TrueDoctor: > I agree. Even on the CPU side, there is no need for nodes that work only on a single colour to accept a table.

And I think there is an argument to be made that now that we have automatic into node insertion we should revert this for nodes which only ever operate on a single color, but I assume @Keavon will have opinions on that

Firestar99:
I just noticed that I'll need them to be `Color` for them to work with the current shaders PR. But I'm working on a followup PR, which would allow us to switch them to `Option<Color>` easily.

Firestar99:
Actually, if they'd just default to black color if the Option is None, why have it be optional in the first place?

node-graph/graster-nodes/src/adjustments.rs
Keavon: Let's please do `Option<Color>` only, not `Color`, since this is somewhat temporary as discussed in our Discord conversation.

Firestar99:
@TrueDoctor are you sure you want to merge this, even though it'll break the UI for color selection of the two nodes?

Firestar99:
I removed the one problematic change of `Table<Color> -> Color` from this PR so we can handle that separately and get this merged.

TrueDoctor:
Feel free to merge at your own discretion, if the change to colors is only temporary it is fine break the ui for now and fix that in a followup pr

PR #3093, opened 2025-08-26
Fix the Select tool not selecting the target which is entirely within the selection shape

indierusty:
Bug: Selection tool do not selects the target which is entirely within the selection path.
Fix: Fixed `bezpath_is_inside_bezpath` function incorrectly checking insideness.

Fixes https://discord.com/channels/731730685944922173/881073965047636018/1407746108498772062


node-graph/gcore/src/vector/algorithms/bezpath_algorithms.rs
```diff
@@ -621,12 +622,12 @@ pub fn bezpath_is_inside_bezpath(bezpath1: &BezPath, bezpath2: &BezPath, accurac
...
 	// If the inner bezpath bounding box is larger than the outer bezpath bounding box in any direction
 	// then the inner bezpath is intersecting with or outside the outer bezpath.
-	if !outer_bbox.contains_rect(inner_bbox) {
+	if !outer_bbox.contains_rect(inner_bbox) && outer_bbox.intersect(inner_bbox).is_zero_area() {
 		return false;
 	}
 
 	// Eliminate bezpath1 if any of its anchor points are outside the bezpath2.
-	if !bezpath1.elements().iter().filter_map(|el| el.end_point()).all(|point| bezpath2.contains(point)) {
+	if !bezpath1.elements().iter().filter_map(|elm| elm.end_point()).all(|point| bezpath2.contains(point)) {
```
Keavon:
```suggestion
	if !bezpath1.elements().iter().filter_map(|el| el.end_point()).all(|point| bezpath2.contains(point)) {
```

PR #3094, opened 2025-08-26
Desktop: Make embedded resources optional

timon-schelling:
- Why: Multiple devs where confused/annoyed by an error that came from a missing frontend dist dir.
- Add a embedded_resources feature
- Add a build script that sets `cfg(embedded_resources)` if resource dir exists
- Emits warning when resource dir doesn't exist 
- embedded_resources are in a separate crate
  - reduces parse times for the desktop crate (dir is included as byte string, about 10 sec extra build time)
  - makes it very clear what the build script is used for
  - no build script in desktop
- allows specifying `GRAPHITE_RESOURCES` when build without the embedded_resources feature
- preparation for watch command for desktop
- Cleans resource handler code (moved to cef internal module etc.)


Keavon:
I'm unfamiliar with what kinds of resources these are, could you give some examples please?

timon-schelling:
> I'm unfamiliar with what kinds of resources these are, could you give some examples please?

The frontend files, `frontend/dist`.

desktop/src/cef.rs
```diff
@@ -62,6 +63,12 @@ impl WindowSize {
 	}
 }
 
+#[derive(Clone)]
+pub(crate) struct Resource {
+	pub(crate) data: Vec<u8>,
```
TrueDoctor: Would it make sense to make this a `Cow<'static, [u8]>` ?

timon-schelling: The results of a fs::read cannot be 'static,  see usage of this struct.

timon-schelling:
```rust
if let Ok(data) = std::fs::read(file_path) {
    return Some(Resource { data, mimetype });
}
```

TrueDoctor: Yes, but include dir should give you static references, right?

timon-schelling: Yes, but this pr also adds the ability to create resources from disk. Could be behind feature flag for now, but would like to have override functionality in the future.

PR #3095, opened 2025-08-26
Add node input type support for `f32` to enable usage on GPU

Firestar99:
f32 params breaks the UI, I have no idea why. Feel free to take over this PR and fix that UI.


0HyperCube:
I have fixed the UI. You just have to hook up the number input inside the `fn property_from_type` function @Firestar99.

Firestar99:
will have a look tomorrow :heart:

node-graph/graph-craft/src/proto.rs
```diff
@@ -805,9 +805,11 @@ mod test {
 		construction_network.generate_stable_node_ids();
 		assert_eq!(construction_network.nodes[0].1.identifier.name.as_ref(), "value");
 		let ids: Vec<_> = construction_network.nodes.iter().map(|(id, _)| *id).collect();
+
+		// If this assert fails: These NodeIds seem to be changing when you modify TaggedValue, just update them.
 		assert_eq!(
 			ids,
-			vec![NodeId(13743208144182721472), NodeId(4607569396187877965), NodeId(16950305885390329527), NodeId(15151181027373658932)]
+			vec![NodeId(2791689253855410677), NodeId(11246167042277902310), NodeId(1014827049498980779), NodeId(4864562752646903491)]
```
Firestar99, line 812: Is this comment I left correct? Based on git history, I see them being updated frequently

0HyperCube: Yes, I think the test network contains a `U32(u32)`, so moving the position of that will change the hash.

Firestar99: @0HyperCube what about the comment right above, which github is hiding?

0HyperCube: It seems fair. I don't really see why it matters if the node ids aren't stable across different commits. However most modifications to the TaggedValue take place under the U32 so it probably shouldn't break too much.

node-graph/graph-craft/src/document/value.rs
```diff
@@ -201,6 +202,8 @@ tagged_value! {
 	// ============
 	// STRUCT TYPES
 	// ============
+	Vec2(Vec2),
```
Keavon: Let's please call this `FVec2` because the user-facing name for `DVec2` is already "Vec2" and we need to avoid the confusion of the internal version being given the opposite name of the user-facing version. And probably the same for `Affine2` -> `FAffine2` for consistency. This also lets us discriminate between them in the future for rename purposes if we figure out how to clean this up once there is more Graphene infrastructure in place to handle numeric precision and units.

PR #3097, opened 2025-08-26
shaders: rust gpu integration

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/3027

just adds the rust-gpu integration to compile shader nodes to wgsl shaders, the runtime componentry to run these shaders can be found in github.com/GraphiteEditor/Graphite/pull/2985


timon-schelling:
I think it should be `rust-gpu cargo b -r ...` otherwise rustup will be used right? rust-gpu toolchain should not be available without wrapper script.

Firestar99:
@timon-schelling I don't quite understand what you mean... could you attach that to some piece of code? Commit [nix: use rustc_codegen_spirv.so from nix](https://github.com/GraphiteEditor/Graphite/pull/3097/commits/126f7395a841726abcf26d8f204b6a69b35db499) you made should contain the necessary switches to use the nix rust-gpu toolchain.

Resolved in voice call

timon-schelling:
I'm getting
```
warning: dropping unsupported crate type `dylib` for target `wasm32-unknown-unknown`
warning: `graphene-raster-nodes` (lib) generated 1 warning
```

@TrueDoctor take a close look at the macro changes
otherwise LGTM

Firestar99:
I can't fix that warning, that's the `crate-type = [...]` in grasternodes. rust-gpu needs  `dylib` and you also want the default `lib` for CPU stuff and wasm. Technically, this makes you compile a `dylib` on CPU targets as well, so you'll get an `*.so` in your target. And wasm doesn't support that and emits this warning. But you can't config that per-target nor pass it as a cargo arg, so we're stuck with this for now.

TrueDoctor (approved):

PR #3099, opened 2025-08-27
Fix Colors being offset when applied transforms

mTvare6:
Fixes https://discord.com/channels/731730685944922173/860830913390575646/1407882700404031710.


Keavon:
@TrueDoctor and I discussed this and we think that we have decided on a different approach than the one taken in this PR, so I will close this to help you reduce your workload for what's left in your GSoC deliverables.

PR #3103, opened 2025-08-28
Disable `proc-macros` crate from being tested due to cargo bug

Firestar99:
See cargo bug https://github.com/rust-lang/cargo/issues/15890

Disabling `proc-macros` testing is required for shaders to pass testing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,8 @@ default-members = [
 	"node-graph/node-macro",
 	"node-graph/preprocessor",
 	"node-graph/wgpu-executor",
-	"proc-macros",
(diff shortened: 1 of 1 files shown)


Firestar99:
The rust-gpu build in https://github.com/GraphiteEditor/Graphite/pull/3097 is passing, this is ready to merge!

TrueDoctor (approved):
The main disadvantage is that we now build non default targets for everyone, but I hope we'll get bug reports reasonably quickly if this causes issues for other developers

Firestar99:
Could you define "non default targets" for me?

PR #3105, opened 2025-08-28
Allow the release info build script to work without requiring Git to be installed

timon-schelling:
- Prevents panic that happens when graphite was build in an environment without git. (`len("unkonwn") < 8`)
- Adds option to set release information environment variables, useful for build environments where git is not available.
- Tell cargo when the build script needs to be run again.
- Improve readability of build script.


editor/build.rs
```diff
@@ -1,24 +1,60 @@
...
+
+	// Try to get the commit information from the environment (e.g. set by CI), otherwise fall back to Git commands.
+	let commit_date = env_or_else("GRAPHITE_GIT_COMMIT_DATE", || git_or_unknown(&["log", "-1", "--format=%cI"]));
+	let commit_hash = env_or_else("GRAPHITE_GIT_COMMIT_HASH", || git_or_unknown(&["rev-parse", "HEAD"]));
+	let commit_branch = env_or_else("GRAPHITE_GIT_COMMIT_BRANCH", || {
+		let gh = env::var("GITHUB_HEAD_REF").unwrap_or_default();
+		if !gh.trim().is_empty() {
+			gh.trim().to_string()
+		} else {
+			git_or_unknown(&["rev-parse", "--abbrev-ref", "HEAD"])
```
Keavon, line 25: Why `--abbrev-ref`?

timon-schelling: Returns a commit hash otherwise. But I could also use name-rev like you did before.

Keavon: What does it return if not a commit hash?

timon-schelling: Branch name. (Rev short name if you want to use git wording)

Keavon: We should aim to always have a commit hash so it's possible to actually look up which commit a build link belongs to. (This is also something we don't do on https://editor.graphite.rs but should do.)

timon-schelling: Commit hash is collected in the statement above. There rev-parse without --abbrev-ref.

Keavon: What is the purpose of this second one?

timon-schelling:
```rust
// Collect commit hash
        let commit_hash = env_or_else("GRAPHITE_GIT_COMMIT_HASH", || git_or_unknown(&["rev-parse", "HEAD"]));
// Collect Branch name
         let commit_branch = env_or_else("GRAPHITE_GIT_COMMIT_BRANCH", || { 
                 let gh = env::var("GITHUB_HEAD_REF").unwrap_or_default();

// Code you comment on is used to collect the branch name when it is not set by GitHub locally for now example
```

Keavon: Oh, I wasn't looking closely enough. Disregard my last two comments. Anyway, I think we're good to merge.

PR #3106, opened 2025-08-28
CI: build rust with test profile

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/3103

May decrease CI time? Nope, no idea why `cargo test` isn't reusing artifacts


Firestar99:
didn't work

PR #3109, opened 2025-08-29
Shaders: add `BufferStruct` to support bool and enums

Firestar99:
# Requires https://github.com/GraphiteEditor/Graphite/pull/2985

* Adds the `trait BufferStruct`
* support for
  * gamma_correction
  * channel_mixer
  * selective_color
  * brightness_contrast_classic
    * The original brightness_contrast node doesn't work with it's non-classic mode. **However**, we now have a "Brightness Contrast Classic" CPU node next to the "Brightness Contrast" and "Brightness Contrast Classic GPU" nodes.

### BufferStruct docs
```rust
/// A BufferStruct is a "parallel representation" of the original struct with some fundamental types remapped. This
/// struct hierarchy represents how data is stored in GPU Buffers, where all types must be [`Pod`] to allow
/// transmuting them to `&[u8]` with [`bytemuck`].
///
/// Notable type remappings (original: buffer):
/// * bool: u32 of 0 or 1
/// * any repr(u32) enum: u32 with remapping via [`num_enum`]
///
/// By adding `#[derive(ShaderStruct)]` to your struct (or enum), a parallel `{name}Buffer` struct is created with all
/// the members of the original struct, but with their types using the associated remapped types as specified by this
/// trait.
///
/// # Origin
/// I (@firestar99) copied this entire mod from my [Nanite-at-home] project, specifically the [buffer-content] crate
/// and the [buffer_struct] proc macro. The variant here has quite some modifications, to both cleaned up some of the
/// mistakes my implementation has and to customize it a bit for graphite.
///
/// Hopefully rust-gpu improves to the point where this remapping becomes unnecessary.
///
/// [Nanite-at-home]: https://github.com/Firestar99/nanite-at-home
/// [buffer-content]: https://github.com/Firestar99/nanite-at-home/tree/008dac8df656959c71efeddd2d3ddabcb801771c/rust-gpu-bindless/crates/buffer-content
/// [buffer_struct]: https://github.com/Firestar99/nanite-at-home/blob/008dac8df656959c71efeddd2d3ddabcb801771c/rust-gpu-bindless/crates/macros/src/buffer_struct.rs
///
/// # Safety
/// The associated type Transfer must be the same on all targets. Writing followed by reading back a value must result
/// in the same value.
```


node-graph/node-macro/src/buffer_struct.rs
```diff
@@ -0,0 +1,261 @@
...
+			}
+
+			fn read(from: Self::Buffer) -> Self {
+				<Self as #reexport::num_enum::FromPrimitive>::from_primitive(from)
+			}
+		}
+	})
+}
+
+/// see `BufferStruct` docs
```
TrueDoctor: Could we use a proper doc link which allows you to click on the BufferStruct here?

Firestar99: oh yea with a dev-dependency... I hope the link I made is correct, rustrover isn't linking this

Firestar99:
Added one more commit to fix shader name naming

PR #3110, opened 2025-08-30
Improve editor types

adamgerhant:
Refactor how the editor handles type queries with an improved TypeSource enum and helper functions.
Fixes the valid types list.


editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,324 @@
...
+	/// # Behavior
+	///
+	/// 1. Retrieves the primary output type from `resolved_types`.
+	/// 2. If the node is a network:
+	///    - Iterates through its exports (skipping the first/primary export).
+	///    - For each export, traverses the network until reaching a protonode or terminal condition.
+	///    - Determines the output type based on the final node/value encountered.
+	/// 3. Collects and returns all resolved types.
+	///
+	pub fn output_type(&mut self, output_connector: &OutputConnector, network_path: &[NodeId]) -> TypeSource {
```
TrueDoctor: Please don't commit doc comments for ai generated functions or let an ai write doc comments.  I can very easily tell that this comment is not accurate and I thus can't trust it to help me understand the function

Keavon: This is just relocated code.

TrueDoctor: It should still either be fixed or removed

Keavon: You are its original author 😆

Introduced in this commit that's part of #1878:

https://github.com/GraphiteEditor/Graphite/pull/1878/commits/4b6d5d3e4d787aa6c10d8e69d3ade7e179838f03#diff-6f2722f8a898a9b4a4b8d62488d215e5faa93a35c2b10499ba7796aa1ccfe130R1809-R1839

How would you like to proceed with it?

TrueDoctor: When it was added, the comment did still match the function signature which I assume was changed by someone afterwards. The mismatch should be fixed and the comment should either be removed or checked to make sure it is still accurate. The two issues I have is that this sort of detailed comment is prone to getting out of date so we should avoid them in the future but my biggest concern is the missmatch between the comment and the code because that will just lead to confusion down the line

Keavon: That's updated now. Anything else before needed in the review an approval?

TrueDoctor:
Still reviewing, I will also have to pay extra attention to whether the perf characteristics will have changed

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+			TypeSource::Compiled(compiled_type) => Some(compiled_type.nested_type()),
+			TypeSource::TaggedValue(value_type) => Some(value_type.nested_type()),
+			_ => None,
+		}
+	}
+
+	/// If Some, the type should be displayed in the imports/exports, if None it should be replaced with "import/export index _".
+	pub fn compiled_nested_type_name(&self) -> Option<String> {
+		self.compiled_nested_type().map(|ty| ty.to_string())
+	}
```
TrueDoctor: This should return &str right?

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+
+	/// Used when searching for nodes in the add Node popup.
+	pub fn add_node_string(self) -> Option<String> {
+		self.compiled_nested_type().map(|ty| format!("type:{}", ty.to_string()))
+	}
+
+	/// The type to display in the tooltip.
+	pub fn resolved_type_name(&self) -> String {
```
TrueDoctor: Also &str

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+					return Vec::new();
+				};
+				let Some(inputs_from_import) = outward_wires.get(&OutputConnector::Import(*input_index)) else {
+					log::error!("Could not get inputs from import in valid_input_types");
+					return Vec::new();
+				};
+
+				let intersection: HashSet<Type> = inputs_from_import
+					.clone()
+					.iter()
```
TrueDoctor, line 226: I don't think having a clone() followed by an `.iter()` makes sense, could you just remove the clone?

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+					.map(|input_connector| self.valid_input_types(input_connector, &nested_path).into_iter().collect::<HashSet<_>>())
+					.fold(None, |acc: Option<HashSet<Type>>, set| match acc {
+						Some(acc_set) => Some(acc_set.intersection(&set).cloned().collect()),
```
TrueDoctor, line 229: This should use the `-` operator instead of `.intersection()`

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+					return Vec::new();
+				};
+				let number_of_inputs = self.number_of_inputs(node_id, network_path);
+				implementations
+					.iter()
+					.filter_map(|(node_io, _)| {
+						let valid_implementation = (0..number_of_inputs).filter(|iterator_index| iterator_index != input_index).all(|iterator_index| {
+							let input_type = self.input_type(&InputConnector::node(*node_id, iterator_index), network_path);
+							// Value inputs are stored as concrete, so they are compared to the nested type. Node inputs are stored as fn, so they are compared to the entire type.
+							// For example a node input of (Footprint) -> Vector would not be compatible with () -> Vector
```
TrueDoctor: Is this comment still accurate?

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -0,0 +1,305 @@
...
+							return None;
+						};
+						self.random_type_for_connector(&first_input, &[network_path, &[*node_id]].concat())
+					}
+					DocumentNodeImplementation::ProtoNode(proto_node_identifier) => {
+						let Some(implementations) = NODE_REGISTRY.get(proto_node_identifier) else {
+							log::error!("Protonode {proto_node_identifier:?} not found in registry");
+							return None;
+						};
+						implementations.keys().next().and_then(|node_io| node_io.inputs.get(input_connector.input_index())).cloned()
```
TrueDoctor: Couldn't we just sort the keys to make this function deterministic? The output will still not be as useful but having it nondeterministic just seems unnecessary

TrueDoctor:
I have not done a deep dive of the logic nor have I tested this for regressions, this is basically only a code review and most of my comments are relatively minor nitpicks which should still be fixed to improve the overall code quality, but I do understand that most of these were not introduced in this pr

Keavon:
And it looks like this test failed:
`messages::tool::tool_messages::artboard_tool::test_artboard::first_artboard`

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
adamgerhant: The usages of this require an owned String and the to_cow_string method also calls .to_string() except for generics

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
adamgerhant: Its required since .valid_input_types requires a mutable borrow of self

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
adamgerhant: Im not sure how exactly type checking will work with the call arguments. I will simplify this to just check if the returned types are the same

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
TrueDoctor: I see

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
TrueDoctor: But since this is a public function we should always let the caller make the allocation if the calle could just return a reference. This avoids allocations if there are other users of this function in the future

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
adamgerhant: yes

node-graph/gcore/src/types.rs
```diff
@@ -229,8 +229,20 @@ impl PartialEq for TypeDescriptor {
...
+		self.name.cmp(&other.name)
+	}
+}
+
+impl PartialOrd for TypeDescriptor {
+	fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+		Some(self.cmp(other))
+	}
+}
+
```
TrueDoctor: I'd prefer if you used a `.min_by_key(|x|x.name)` instead of adding `Ord` to a type that does not have an obvious ordering

adamgerhant:
The complete valid types list is very expensive and will only be called when queried (hovering over an input). Keavon also wants to show both lists in the tooltip

TrueDoctor:
I guess if you only do this on hover for individual nodes the perf should be fine, but we should not call this for each node

node-graph/gcore/src/types.rs
adamgerhant: I think it could be useful to order the entire NodeIOTypes since it could be useful to have a debug option which prints all types for a node

editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
adamgerhant: function removed, default types come from the definition

TrueDoctor:
Alright, Approve from a code quality perspective. @Keavon I think we can merge this once you are happy with the functionality

PR #3111, opened 2025-08-30
Fix undo/redo history related issues with the Path tool

4adex:
Solving these following regressions - 
- Pen tool: Joining two endpoints across two separate layers (similar to Ctrl+J joining, but clicking instead) crashes the editor. (https://discord.com/channels/731730685944922173/881073965047636018/1410397377776324663)
- Path tool: Ctrl+Delete gets undone if you switch to the Select tool immediately after. (https://discord.com/channels/731730685944922173/881073965047636018/1410396669601775699)
- Path Tool: Delete adds unnecessary transactions when no point is selected but delete is pressed. (https://discord.com/channels/731730685944922173/881073965047636018/1411184562049908877)
- Other transaction related logical errors in `Delete` and its related messages in the path tool.

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -1504,7 +1504,6 @@ impl ShapeState {
 			for (segment, _, start, end) in vector.segment_bezier_iter() {
 				if state.selected_segments.contains(&segment) {
 					if start_transaction && !transaction_started {
-						log::info!("111111");
 						responses.add(DocumentMessage::AddTransaction);
 						transaction_started = true;
 					}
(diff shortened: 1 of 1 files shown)


Keavon:
Bug: deleting a point with <kbd>Delete</kbd> produces two history steps, where the intermediate step is the force deletion of the point (akin to <kbd>Ctrl</kbd><kbd>Delete</kbd>).

Keavon:
I'm also concerned that this approach is pretty hacky and not very clean. Could you think of a way to restructure things so you're not passing around conditional flags like this? Although if that's not feasible without fully rebuilding the systems here, we can live with this for now.

4adex:
> Could you think of a way to restructure things so you're not passing around conditional flags like this?

One approach is to first sanity check that under current selection some modification will be done or not (for example checking if atleast a point is selected, etc), if not then do an early return, if yes then start transaction and perform operations.
But this will require adding more checker functions in the code.

Issue #3114, opened 2025-08-31
Move persistence to native


timon-schelling:
Stop relying on the browser indexdb for persistence and user native storage instead

PR #3134, opened 2025-09-03
Desktop: Move autosave persistence to native

timon-schelling:
First part of #3114

Saving Preferences natively will be a separate PR


desktop/wrapper/src/handle_desktop_wrapper_message.rs
```diff
@@ -118,5 +118,22 @@ pub(super) fn handle_desktop_wrapper_message(dispatcher: &mut DesktopWrapperMess
...
+		DesktopWrapperMessage::LoadDocument { id, document, to_front } => {
+			let message = PortfolioMessage::OpenDocumentFileWithId {
+				document_id: id,
+				document_name: Some(document.name),
+				document_path: document.path,
+				document_serialized_content: document.content,
+				document_is_auto_saved: true,
+				document_is_saved: document.is_saved,
+				to_front,
+				select_after_open: false,
```
Keavon: This is the rebase conflict I fixed, since I introduced this `select_after_open` field in a PR I added yesterday. You may want to check if `false` is appropriate here, or if you want to feed this in as a variable (like `to_front` on the line above gets fed in as a variable) so callers of this message can decide on a case-by-case basis.

PR #3142, opened 2025-09-05
Fix a regression with duplicating or pasting vector geometry in the Path tool

indierusty:
Fix a regression introduced in the PR #3036 where Path tool incorrectly duplicates segments handles
From: https://discord.com/channels/731730685944922173/731738914812854303/1413205122162626721


Keavon:
Please fix for Ctrl+C, Ctrl+V, this only currently fixes it for Ctrl+D.

PR #3143, opened 2025-09-05
Remove unnecessary start transaction messages

4adex:
https://discord.com/channels/731730685944922173/881073965047636018/1411608076145918013


Keavon:
This does break undo for dragging the pivot gizmo around, but that's something we can live with for the moment.

PR #3149, opened 2025-09-07
Shaders: Explicitly destroy WGPU images instead of waiting on GC

Firestar99:
Issue:
> Pasting a 720p image (isometric fountain exported as png) causes excessive CPU utilization and allocator pressure on dev, has caused a wgpu OOM error (not shown here). Same firefox nightly in different tabs.

This is a fix towards that by explicitly destroying wgpu images instead of waiting for js gc to do it for us. However, it currently breaks the vello renderer when used with GPU nodes.

### DIFF
--- a/node-graph/gcore/src/raster_types.rs
+++ b/node-graph/gcore/src/raster_types.rs
@@ -134,10 +134,11 @@ pub use gpu::GPU;
 mod gpu {
 	use super::*;
 	use crate::raster_types::__private::Sealed;
+	use std::sync::Arc;
 
 	#[derive(Clone, Debug, PartialEq, Hash)]
 	pub struct GPU {
-		pub texture: wgpu::Texture,
+		pub texture: Arc<InnerTexture>,
 	}
 
 	impl Sealed for Raster<GPU> {}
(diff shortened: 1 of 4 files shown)


TrueDoctor:
Any ideas on why this might break vello? Do the textures potentially get destroyed twice when handing them off to vello?

Firestar99:
No, they get destroyed before they are rendered in vello. So vello has no texture left to read from, aka reads border color white.

TrueDoctor:
Yes I mean does deconstructing the Raster<GPU> destroy the texture or something like that?

Firestar99:
I don't *exactly* understand how you're passing data from one node to another, seems to be a lot of cloning going on...

But basically, we'd just need vello to "strongly reference" the `Raster<GPU>` instead of the `wgpu::Texture` so destroy isn't called before it's finished submitting it's draw commands (since wgpu will delay the destruction until all queued ops are finished with the images.)

TrueDoctor:
How does this work with caching? Cache nodes are allowed to clone the data that flows through them, wouldn't this lead to some sort of double free?

Firestar99:
I always think of `wgpu::Textures` as if they're an `Arc` internally, when they're clearly not...

Firestar99:
wrapped the wgpu texture in an Arc, may work, haven't tested it due to constant build failures...

TrueDoctor:
> wrapped the wgpu texture in an Arc, may work, haven't tested it due to constant build failures...

I mean it does seem to work in the build link, I think we should be good to merge then. Just confirmed that this does not break anything else on desktop

TrueDoctor (approved):

TrueDoctor:
image
I am still able to trigger an out of memory error

PR #3152, opened 2025-09-08
Add the new Typography graphical type

adamgerhant:
Adds the Typography data type, which is part of the Graphic enum and rendered to svg as `<Text/>` and rendered to Vello with `draw_glyphs`

TODO: Replace `FontCache` with `NewFontCache`, update Text node to output Typography


node-graph/gcore/src/graphic.rs
```diff
@@ -11,14 +12,34 @@ use glam::{DAffine2, DVec2};
...
+	}
+}
+
+impl<'de> serde::Deserialize<'de> for Graphic {
+	fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+	where
+		D: serde::Deserializer<'de>,
+	{
+		Ok(Graphic::Graphic(Table::new()))
+	}
```
TrueDoctor, line 42: So you just decided to not do any Serialization? Is there any reason to do this? This is a very confusing design decision and will likely lead to bugs in the future

TrueDoctor:
Did you mean to mark this as ready for review? This pr does not seem to be finished

node-graph/gcore/src/graphic.rs
adamgerhant: The Layout within the Typography cannot be serialized. Also its not meant to be a Tagged value, as it can only be produced at runtime by the text node. It only has this implemented since Graphic needs a tagged value for layer inputs. Similar to the wgpu::Texture.

adamgerhant:
> Did you mean to mark this as ready for review? This pr does not seem to be finished

Yes, this is only for use in the native node graph overlay node. There is no user integration yet to keep the scope small.

node-graph/gcore/src/bounds.rs
```diff
@@ -38,3 +38,9 @@ impl BoundingBox for GradientStops {
 		RenderBoundingBox::Infinite
 	}
 }
+impl BoundingBox for Typography {
+	fn bounding_box(&self, transform: DAffine2, _include_stroke: bool) -> RenderBoundingBox {
+		let bbox = DVec2::new(self.layout.full_width() as f64, self.layout.height() as f64);
+		RenderBoundingBox::Rectangle([transform.transform_point2(DVec2::ZERO), transform.transform_point2(bbox)])
```
TrueDoctor: should be `transform.translation`

node-graph/gcore/src/graphic.rs
TrueDoctor: You have just successfully disabled serialization / deserialization for all `Grahic` data not just `Typography`

node-graph/gcore/src/graphic.rs
```diff
@@ -507,34 +530,15 @@ pub fn migrate_graphic<'de, D: serde::Deserializer<'de>>(deserializer: D) -> Res
...
-	}
-
 	#[derive(serde::Serialize, serde::Deserialize)]
 	#[serde(untagged)]
-	enum GraphicFormat {
+	enum EitherFormat {
 		OldGraphicGroup(OldGraphicGroup),
-		OlderTableOldGraphicGroup(OlderTable<OldGraphicGroup>),
-		OldTableOldGraphicGroup(OldTable<OldGraphicGroup>),
-		OldTableGraphicGroup(OldTable<GraphicGroup>),
```
TrueDoctor: @Keavon have you approved the removal of upgrade code for older formats?

node-graph/gcore/src/render_complexity.rs
```diff
@@ -65,3 +67,9 @@ impl RenderComplexity for GradientStops {
 		1
 	}
 }
+
+impl RenderComplexity for Typography {
+	fn render_complexity(&self) -> usize {
+		1
+	}
```
TrueDoctor, line 74: This should depend on the text properties since text can be pretty perf intensive to render

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+			.field("stroke", &self.stroke)
+			.finish()
+	}
+}
+
+impl PartialEq for Typography {
+	fn eq(&self, _other: &Self) -> bool {
+		true
+	}
+}
```
TrueDoctor, line 88: What?

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+
+impl Hash for Typography {
+	fn hash<H: Hasher>(&self, state: &mut H) {
+		self.layout.len().hash(state);
+	}
+}
```
TrueDoctor: This is neither consistent with the eq implementation, nor does it make much sense as the hash function. You are hashing the number of lines so `Hello World` and `Hallo World` will have the same hash code

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+
+impl Typography {
+	pub fn to_vector(&self) -> Table<Vector> {
+		Table::new()
+	}
```
TrueDoctor: Why did you add this method? This is a footgun waiting to happen. It would be very reasonable to expect a `to_vector` method to exist on Layout, but this just does nothing?

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+}
+
+#[derive(Clone)]
+pub struct NewFontCacheWrapper(pub Arc<Mutex<NewFontCache>>);
```
TrueDoctor: I generally don't like the use of a `Mutex` here. Why couldn't you just reuse the existing text lay outing structure instead of creating a new one but leaving the old one in place?

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+
+impl fmt::Debug for NewFontCacheWrapper {
+	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+		f.debug_struct("font cache").finish()
```
TrueDoctor: Should be `"NewFontCacheWrapper"` instead of `"font cache"`

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+	}
+}
+
+impl PartialEq for NewFontCacheWrapper {
+	fn eq(&self, _other: &Self) -> bool {
+		log::error!("Font cache should not be compared");
+		false
```
TrueDoctor: why not? If something implements `PartialEq` you would expect it to actually implement equality. There would definitely be more useful comparisons you could do here instead of just returning false.
I mean, you do take great care to create a hash which you could use here for example, which would be better than nothing.

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+	}
+}
+
+unsafe impl dyn_any::StaticType for NewFontCacheWrapper {
+	type Static = NewFontCacheWrapper;
+}
```
TrueDoctor: Is there a reason you manually implement this with unsafe instead of deriving the trait?

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+
+		let source_sans_font = Font::new(SOURCE_SANS_FONT_FAMILY.to_string(), SOURCE_SANS_FONT_STYLE.to_string());
+		new.register_font(source_sans_font, SOURCE_SANS_FONT_DATA.to_vec());
+		new
+	}
+
+	pub fn register_font(&mut self, font: Font, data: Vec<u8>) {
+		match self.font_mapping.entry(font) {
+			Entry::Occupied(occupied_entry) => {
+				log::error!("Trying to register font that already is added: {:?}", occupied_entry.key());
```
TrueDoctor: I don't think this should be treated as an error. Errors should be for serious issues in the code which might threaten the editor stability, this is not one of these cases

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+			}
+			Entry::Vacant(vacant_entry) => {
+				let registered_font = self.font_context.collection.register_fonts(parley::fontique::Blob::from(data), None);
+				if registered_font.len() > 1 {
+					log::error!("Registered multiple fonts for {:?}. Only the first is accessible", vacant_entry.key());
```
TrueDoctor: see above

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+		new
+	}
+
+	pub fn register_font(&mut self, font: Font, data: Vec<u8>) {
+		match self.font_mapping.entry(font) {
+			Entry::Occupied(occupied_entry) => {
+				log::error!("Trying to register font that already is added: {:?}", occupied_entry.key());
+			}
+			Entry::Vacant(vacant_entry) => {
+				let registered_font = self.font_context.collection.register_fonts(parley::fontique::Blob::from(data), None);
```
TrueDoctor: registered_fonts

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+					Some((family_id, font_info)) => {
+						let Some(family_name) = self.font_context.collection.family_name(family_id) else {
+							log::error!("Could not get family name for font: {:?}", vacant_entry.key());
+							return;
+						};
+						let Some(font_info) = font_info.into_iter().next() else {
+							log::error!("Could not get font info for font: {:?}", vacant_entry.key());
+							return;
+						};
+						let mut hasher = FxHasher::default(); // or FxHasher::new()
```
TrueDoctor: FxHasher does not have a `new` method

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+						// Hash the Font for a unique id and add it to the cached hash
+						vacant_entry.key().hash(&mut hasher);
+						let hash_value = hasher.finish();
```
TrueDoctor:
```rust
let hash_value = FxBuildHasher.hash_one(vacant_entry.key());
```

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+						self.hash = self.hash.wrapping_add(hash_value);
```
TrueDoctor: Instead of performing an addition here, you should hash the font data + the previous hash value. `FxHasher` is very fast but not particularly collision resistant and this could cause issues. There is also the question of why you use `FxHasher` anyway, it can't really be for perf reasons since you do also compute the hash of the same data using the default hasher. If you were concerned about performance we should only compute the hash once and reuse it for both use cases but I can guarantee you that the perf is going to be completely irrelevant in this case so idk.
If your intent was to make sure the insertion order does not affect the hash value, that is fine in principle but should an explaining commend and a justification why this does not create collisions

node-graph/gcore/src/text.rs
```diff
@@ -29,3 +42,142 @@ impl From<TextAlign> for parley::Alignment {
...
+
+		builder.push_default(StyleProperty::FontStack(parley::FontStack::Single(parley::FontFamily::Named(Cow::Owned(font_family.clone())))));
+		builder.push_default(StyleProperty::FontSize(font_size));
+		builder.push_default(StyleProperty::FontWeight(font_info.weight()));
+		builder.push_default(StyleProperty::FontStyle(font_info.style()));
+		builder.push_default(StyleProperty::FontWidth(font_info.width()));
+
+		let mut layout: Layout<()> = builder.build(text);
+		layout.break_all_lines(None);
+		// layout.align(None, parley::Alignment::Start, AlignmentOptions::);
```
TrueDoctor: leftover comment

node-graph/gpath-bool/src/lib.rs
```diff
@@ -318,6 +318,7 @@ fn flatten_vector(graphic_table: &Table<Graphic>) -> Table<Vector> {
 						}
 					})
 					.collect::<Vec<_>>(),
+				Graphic::Typography(_) => Vec::new(),
```
TrueDoctor: you did even create a placeholder function which does not do anything, couldn't you use that here?

TrueDoctor (changes requested):

node-graph/gcore/src/graphic.rs
adamgerhant: This is a rebasing error where I copied the old file without checking for new changes. Ill add this back.

node-graph/gcore/src/render_complexity.rs
adamgerhant: Id need to benchmark it, but given that rendering text with <Text> elements is around 100x faster than the text -> vector -> bezpath -> svg string conversion, I don't think it would be significant. Something based on the height/width could work though.

node-graph/gcore/src/text.rs
adamgerhant: This is a dummy implementation required by TaggedValue::Graphic. It should never be called since a Graphic table should never be compared. a warn! message in here would be useful

node-graph/gcore/src/text.rs
adamgerhant: The idea is to create a hash key representing the unique state of all current fonts in the NewFontCache. This ensures the SNI of the tagged value input which supplies the cache is updated when the cache changes, and is also very fast/does not require hashing the entire font cache. I thought the FxHasher is the default hasher that is generally used, since its used in NodeNetwork and to calculate stable node ids.

node-graph/gcore/src/graphic.rs
adamgerhant: Graphic data can never be serialized since it should never be anything other than an empty table (There is no widget to modify it)

node-graph/gcore/src/text.rs
TrueDoctor: In that case Graphic should be removed from the Tagged Value enum, but this implementation is just non-sensical

node-graph/gcore/src/text.rs
TrueDoctor: The issue is that you are adding the hash codes that does not necessarily preserve the collision resistance. And FxHasher has a pretty low collision resistance to begin with so this likely amplify the issue. If you care about the position independence (and I don't think that we actually do) you could use the `DefaultHasher` here since it generates higher quality hashes and the amount of data we has is very small so perf should not be an issue. We only really use `FxHasher` if we expect to hash large amounts of data. If we cared more about collision resistance and still wanted to use FxHasher we could use a Diffie-Hellman group construction instead here by raising the generator group to the power of the hash value. But that would be a bit overkill since we don't really work in adversarial scenarios

PR #3154, opened 2025-09-08
Native background render node

adamgerhant:
Creates a network to render the node graph UI with the [generate_node_graph_overlay_network](https://github.com/GraphiteEditor/Graphite/blob/native-background-render-node/editor/src/application.rs#L37) function. Does not include a runtime or system for evaluating a network, which is currently achieved by https://github.com/GraphiteEditor/Graphite/pull/3129

depends on https://github.com/GraphiteEditor/Graphite/pull/3122


Keavon:
!build

PR #3162, opened 2025-09-10
Optimize editor performance for node selection, click target bounds, and batched messages

TrueDoctor:
Optimizations: 
- Improve `Message::Batched` to not require cloning messages
- Memoize `NodeNetwork` hash computation
- Improve performance for calculation of selected nodes
- Improve performance of click target bounding box computation


0HyperCube:
Note that this makes the bounding boxes in the select tool very inaccurate. #3181

Keavon:
Partly reverted in #3184 to fix that regression.

PR #3169, opened 2025-09-12
Optimize vector operations by adding index data structures to replace linear scans

TrueDoctor:


TrueDoctor:
Strangely this seems to improve the perf only on desktop, we might revisit this in the future but it wouldn't make sense to merge in the current state. We could potentially basically only merge the first commit as that is less likely to negatively impact performance on its own

PR #3170, opened 2025-09-12
Refresh menu for path editable

0HyperCube:
Fixes https://discord.com/channels/731730685944922173/881073965047636018/1415971652470181984:

> File > Make Path Editable is grayed out at inappropriate times, but the button in the Path tool control bar is correct.


Keavon:
Reproduction steps for a case not fixed by this:
1. Draw a shape with the Pen tool and conclude the drawing
2. With the Pen tool still active, open the File menu and see that Make Path Editable is incorrectly still grayed out
3. Switch to the Path tool and then reopen the menu, it has been updated

0HyperCube:
This is very unfortunate.

PR #3171, opened 2025-09-13
Add distance validation in merge function and clamp grid coords

### LINKED ISSUE #3155: Crash in debug mode: dragging a brush layer into a boolean operation layer
https://github.com/user-attachments/assets/38562fa1-265c-46ef-9c02-601d3d195361

It doesn't always happen in a localhost debug build, and never happens in a production build. If you have trouble reproducing it from the video instructions above, here is a file where the last step of dragging the brush layer into the boolean layer will produce the crash:

[crash.graphite.txt](https://github.com/user-attachments/files/22226091/_crash.graphite.txt)

Freeslugga:
Closes #3155 - Crash when merging brush layer into boolean layer in debug

both changes to rs files in node-graph, merge handling was improper, causing a panic with certain data types. Also clamped the grid points to avoid invalid data being passed.

### DIFF
--- a/node-graph/gpath-bool/src/lib.rs
+++ b/node-graph/gpath-bool/src/lib.rs
@@ -333,35 +333,86 @@ fn to_path(vector: &Vector, transform: DAffine2) -> Vec<path_bool::PathSegment> 
 
 fn to_path_segments(path: &mut Vec<path_bool::PathSegment>, subpath: &Subpath<PointId>, transform: DAffine2) {
 	use path_bool::PathSegment;
-	let mut global_start = None;
+
+	const EPSILON: f64 = 1e-8;
+	const COORD_LIMIT: f64 = 1.0e6;
+	#[inline]
+	fn finite_clamped(mut v: DVec2) -> Option<DVec2> {
+		if !v.x.is_finite() || !v.y.is_finite() {
+			return None;
+		}
+		v.x = v.x.clamp(-COORD_LIMIT, COORD_LIMIT);
+		v.y = v.y.clamp(-COORD_LIMIT, COORD_LIMIT);
+		Some(v)
+	}
+
+	let transform_point = |pos: DVec2| -> Option<DVec2> {
+		let p = transform.transform_point2(pos);
+		let q = p * EPSILON.recip();
+		let q = q.round();
+		let q = q * EPSILON;
+		finite_clamped(q)
+	};
+
+	let mut global_start: Option<DVec2> = None;
 	let mut global_end = DVec2::ZERO;
+	let mut any_segment = false;
 
 	for bezier in subpath.iter() {
-		const EPS: f64 = 1e-8;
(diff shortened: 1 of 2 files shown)


Freeslugga:
Currently the crash is fixed, but when you add a brush layer to the boolean group it rescales and moves the brush layer to the point where it is invisible. You are able to reset the scale and location via the properties menu and the brush re-appears.

Freeslugga:
Not sure where the transforms are taking place, I can open as a separate issue if this is merged.

Freeslugga:
@Keavon This is my first contribution to an open-source project, please let me know if you have any direct feedback to better assist my journey with contributing!

Keavon:
Thanks, I will be busy for the next day or two as I get the next release out, then I'll try to take a look. Feel free to take on more issues in the meantime.

node-graph/gpath-bool/src/lib.rs
```diff
@@ -337,35 +333,86 @@ fn to_path(vector: &Vector, transform: DAffine2) -> Vec<path_bool::PathSegment>
...
+			None => continue,
+		};
+		let p1 = match p1 {
+			Some(h) => transform_point(h),
+			None => None,
+		};
+		let p2 = match p2 {
+			Some(h) => transform_point(h),
+			None => None,
+		};
```
TrueDoctor, line 379: please keep the p1 and p2 as they where and change p0 and p3 to
```rust
let Some(p0) = transform_point(p0) else { continue;};
```
Ideally this should still only need 4 lines of code after the refactor

node-graph/gpath-bool/src/lib.rs
```diff
@@ -337,35 +333,86 @@ fn to_path(vector: &Vector, transform: DAffine2) -> Vec<path_bool::PathSegment>
...
 
-		let segment = match (p1, p2) {
-			(None, None) => PathSegment::Line(p0, p3),
-			(None, Some(p2)) | (Some(p2), None) => PathSegment::Quadratic(p0, p2, p3),
-			(Some(p1), Some(p2)) => PathSegment::Cubic(p0, p1, p2, p3),
+		let seg = match (p1, p2) {
+			(None, None) => {
+				if (p3 - p0).length_squared() == 0.0 {
+					continue;
+				}
```
TrueDoctor, line 390: This is basically just an equality check right? And since you don't use any kind of epsilon, I'm wondering if you could just write `p3 == p0`.
If not, please use `p3.distance_squared(p0) == 0.` instead.

TrueDoctor (changes requested):

Keavon:
I'm just catching up (very belatedly) on my PR review backlog to either merge or close everything possible. But I'm left with a question: this seems to me that it is a single PR composed of two unrelated things, one is an adjustment to the merge-by-distance algorithm, while the other is a fix for #3155. Is that a correct interpretation? Would you be interested in reviving this by applying True Doctor's code review feedback, rebasing onto the latest master with resolutions for the merge conflicts, and ideally also splitting up the two unrelated parts of this PR into two PRs (keep this one, and open another)? Thank you!

PR #3182, opened 2025-09-16
Fix: check if path tool is interacting with raster layer to prevent crashing.

### LINKED ISSUE #2687: Using the path edit tool on one of the corners of a raster image breaks the graph
To reproduce:

- Paste a raster image
- Select the path tool
- Click and drag on the image (can often happen accidentally)
- Observe that the node graph has an error.

This is a bug, since the tooling should avoid breaking the graph as part of #2214

KartikSharma0:
Resolves #2687 
Added a check that returns `PathToolFsmState::Ready` if the path tool is used on a raster layer.

### DIFF
--- a/.github/workflows/build-dev-and-ci.yml
+++ b/.github/workflows/build-dev-and-ci.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
   pull_request: {}
+  merge_group: {}
 env:
   CARGO_TERM_COLOR: always
   INDEX_HTML_HEAD_REPLACEMENT: <script defer data-domain="dev.graphite.art" data-api="https://graphite.art/visit/event" src="https://graphite.art/visit/script.hash.js"></script>
--- a/.github/workflows/build-linux-bundle.yml
+++ b/.github/workflows/build-linux-bundle.yml
@@ -1,10 +1,16 @@
 name: Build Linux Bundle
 
 on:
-  workflow_dispatch: {}
   push:
     branches:
       - master
+  workflow_dispatch:
+    inputs:
+      push_to_cache:
+        description: "Push to Nix Cache"
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build:
@@ -14,10 +20,20 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: DeterminateSystems/nix-installer-action@main
-      - uses: DeterminateSystems/magic-nix-cache-action@main
 
       - name: Free disk space
         run: sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc /opt/hostedtoolcache
+
+      - name: Build Nix Package
+        run: nix build .nix --no-link --print-out-paths
+
+      - name: Push to Nix Cache
+        if: github.ref == 'refs/heads/master' || inputs.push_to_cache == true
+        env:
+          NIX_CACHE_AUTH_TOKEN: ${{ secrets.NIX_CACHE_AUTH_TOKEN }}
+        run: |
+          nix run nixpkgs#cachix -- authtoken $NIX_CACHE_AUTH_TOKEN
+          nix build .nix --no-link --print-out-paths | nix run nixpkgs#cachix -- push graphite
 
       - name: Build Linux Bundle
         run: nix build .nix#graphite-bundle.tar.xz && cp ./result ./graphite-linux-bundle.tar.xz
--- a/.github/workflows/build-mac-bundle.yml
+++ b/.github/workflows/build-mac-bundle.yml
@@ -1,0 +1,157 @@
+name: Build Mac Bundle
+
+on:
+  workflow_dispatch: {}
+  push:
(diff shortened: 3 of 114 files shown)


Keavon:
!build

KartikSharma0:
Video of the fix:

https://github.com/user-attachments/assets/580c80df-8336-41c8-b68b-f8bfc82ac453

KartikSharma0:
Please let me know if the video above captures the expected behavior.

Keavon:
This does not remove anchors from the corners of the image. All it does is break the Path tool's ability to select stuff, either a box selection or even clicking on other valid points if multiple layers (including actual vector layers) are selected. Please give it another shot so it addresses the requested change (first sentence of this reply).

KartikSharma0:
So there shouldn't be any anchors displayed on the image at all when the path tool is selected? The path tool still works on the spline as shown, it's just ignoring the raster image.

KartikSharma0:
I added a check before showing the anchors/handles. Please let me know if this PR is in-line with what you expected now.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1682,7 +1689,13 @@ impl Fsm for PathToolFsmState {
 				}
 
 				// TODO: find the segment ids of which the selected points are a part of
-
+				if tool_data.is_raster_layer {
+					overlay_context.visibility_settings.anchors = false;
+					overlay_context.visibility_settings.handles = false;
+					overlay_context.visibility_settings.path = false;
```
Keavon: Changing their overlay visibility isn't going to be an appropriate solution, they need to be never rendered.

KartikSharma0: I see. In my understanding, we need to at least select a layer before we can infer what type it is, as such it becomes difficult to selectively render the overlay or not. Would you have any suggestions on how to proceed?

KartikSharma0: Like the issue is that I can definitely bypass the `transition` function in which the overlay is drawn, but I need to select the raster layer at least once, which will draw the overlay initially.

Keavon: I unfortunately don't have any additional knowledge in this area to share, so either of us would learn the same thing from reading the code. And as a new contributor, you'll learn best by reading lots of code and immersing yourself into it until you've reached a point where this should become clear.

0HyperCube: In my opinion, the cleanest solution would be for the `fn compute_modified_vector` to return `None` on raster layers. This would fix the overlays and the selections and anything else that might try to use the vector data from an image.

KartikSharma0: Thanks for the suggestion. In this case, `compute_modified_vector` takes a non-mutable reference of `&self: NodeNetworkInterface` but the function `is_raster_layer` takes a mutable reference of `network_interface: &mut NodeNetworkInterface`. If I were to call `is_raster_layer` in `compute_modified_vector`, it would require a lot of refactoring. Would you have any suggestions for a workaround?

KartikSharma0:
Okay, I simplified the code changes. I think this PR should fulfill the requirements now.

0HyperCube:
!build

TrueDoctor:
@Keavon Should we merge this?

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -375,6 +376,23 @@ impl LayoutHolder for PathTool {
 #[message_handler_data]
 impl<'a> MessageHandler<ToolMessage, &mut ToolActionMessageContext<'a>> for PathTool {
 	fn process_message(&mut self, message: ToolMessage, responses: &mut VecDeque<Message>, context: &mut ToolActionMessageContext<'a>) {
+		let target_layers: Vec<LayerNodeIdentifier> = context.document.network_interface.selected_nodes().selected_layers(context.document.metadata()).collect();
```
adamgerhant: This can be filtered before collecting

adamgerhant (changes requested):
This is not an optimal implementation since it requires an entire document click test every time a message is sent to the Path tool, including overlays and such. Ideally the check should be done in the already existing selection logic. The `PathToolMessage::SelectionChanged` message handling might be a good place to add this filter.

KartikSharma0:
@adamgerhant Thank you for the suggestion. I noticed I need to do a check for raster layers when processing a few messages other than `PathToolMessage::SelectionChanged`, I felt the cleanest approach would be to update the matching patterns in `process_message` to include all the messages that can interact with a raster layer. Now it's more optimized while maintaining functionality.

editor/src/messages/tool/tool_messages/path_tool.rs
KartikSharma0: Using filtering in updated code.

PR #3183, opened 2025-09-16
Desktop: Window resize controls wayland

timon-schelling:
This PR adds support for window resize controls on Wayland compositors that allow disabling the standard server-side window frame (i.e., when client-side decorations are enabled).

Resize controls: Implemented for all compositors that support client-side decorations.

Fallback behavior: On compositors that force server-side decorations (e.g. tiling WMs like Hyprland, Sway, etc.), we skip drawing resize borders and shadows. Since these would otherwise render inside the compositor’s frame, we delegate resizing entirely to the compositor.

No duplicate titlebars: In testing, compositors either support client-side decorations or omit drawing their own titlebar. This ensures we don’t end up with a double titlebar when rendering our own decorations.

Note: This relies on two pending PRs in winit. Until they are merged, we temporarily depend on my fork of winit.

https://github.com/rust-windowing/winit/pull/4357
https://github.com/rust-windowing/winit/pull/4361

https://github.com/user-attachments/assets/b9c8d23b-5747-407e-bdfc-6d13685f3f60


TrueDoctor (approved):
Alright, let's hope we don't have to use a fork for too long

timon-schelling:
I will try my best to get all features we need upstreamed.

PR #3188, opened 2025-09-17
Desktop: Native export

### LINKED ISSUE #3007: Export images with native Vello rendering
Currently non svg images are rendered through svg rasterization in the browser. This should be replaced with native rendering with vello. At least on desktop.

Currently non svg images are downloaded like in web, very confusing for users.

timon-schelling:
Needs #3194

Closes #3007

- [x] Png
  - [x] no transparency
  - [x] transparency
- [x] Jpeg

### DIFF
--- a/desktop/wrapper/src/messages.rs
+++ b/desktop/wrapper/src/messages.rs
@@ -135,8 +135,3 @@ pub enum Platform {
 	Mac,
 	Linux,
 }
-
-pub enum ExportType {
-	Png,
-	Jpg,
-}


node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -160,7 +160,8 @@ pub struct RenderContext {
 pub enum RenderOutputType {
 	#[default]
 	Svg,
-	Vello,
+	Texture,
+	Buffer,
```
TrueDoctor: I think this change is still unnecessary, right?

node-graph/gstd/src/render_node.rs
```diff
@@ -59,9 +58,20 @@ async fn render_intermediate<'a: 'n, T: 'static + Render + WasmNotSend + Send +
 	data.collect_metadata(&mut metadata, footprint, None);
 	let contains_artboard = data.contains_artboard();
 
-	let editor_api = editor_api.eval(None).await;
+	let use_vello = {
+		#[cfg(target_family = "wasm")]
+		{
+			let editor_api = editor_api.eval(None).await;
+			!render_params.for_export && editor_api.editor_preferences.use_vello() && matches!(render_params.render_output_type, graphene_svg_renderer::RenderOutputType::Vello)
```
TrueDoctor, line 65: I'm wondering if it wouldn't make sense if we did this in the runtime too but I guess we can do that in a different pr

editor/src/messages/frontend/frontend_message.rs
```diff
@@ -82,6 +82,16 @@ pub enum FrontendMessage {
...
+	#[cfg(feature = "gpu")]
+	#[serde(skip)]
+	TriggerExportImageBuffer {
+		data: Vec<u8>,
+		width: u32,
+		height: u32,
+		transparent: bool,
+		file_type: super::utility_types::FileType,
+		name: String,
+	},
```
TrueDoctor: Hmm, I'm wondering if this is the correct abstraction level. I'd prefer if the encoding was handled in the editor crate (e.g. when we add quality options etc. it would make it easier to keep them in sync between frontend and desktop)

PR #3193, opened 2025-09-18
Desktop: Browser console message forwarding

### LINKED ISSUE #3107: Display browser console messages on the command line
Related: #3040

timon-schelling:
Closes #3107

### DIFF
--- a/desktop/src/main.rs
+++ b/desktop/src/main.rs
@@ -27,7 +27,7 @@ fn main() {
 		// We are in a CEF subprocess
 		// This will block until the CEF subprocess quits
 		let error = cef_context_builder.execute_sub_process();
-		tracing::debug!("Cef subprocess failed with error: {error}");
+		tracing::warn!("Cef subprocess failed with error: {error}");
 		return;
 	}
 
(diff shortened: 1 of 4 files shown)


desktop/src/main.rs
```diff
@@ -27,7 +27,7 @@ fn main() {
 		// We are in a CEF subprocess
 		// This will block until the CEF subprocess quits
 		let error = cef_context_builder.execute_sub_process();
-		tracing::error!("Cef subprocess failed with error: {error}");
```
TrueDoctor, line 30: Was this included by accident or do we want this to be a debug!?

timon-schelling: I think we what this to be a debug. Cef is very good at crashing/terminating process. We don't care about all of them expect the render process. Currently this leads to useless spam. I would be fine with creating a issue to dig into the root cause but until then debug logging seems like a acceptable option.

TrueDoctor: Hmm I think I'd prefer this to be a warning then. As per convention we use `debug!` calls for debugging but basically never ship code with debug logs (except for very few exceptions). Another change I would like to see at some point is that we set the log level for packages we build ourselves to debug when not compiling in release mode so this would break anyway once we do that

TrueDoctor (approved):

PR #3194, opened 2025-09-18
Implement Convert trait to convert between CPU and GPU images

TrueDoctor:


node-graph/gcore/src/registry.rs
```diff
@@ -9,7 +9,7 @@ use std::sync::{LazyLock, Mutex};
 pub use graphene_core_shaders::registry::types;
 
 // Translation struct between macro and definition
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
timon-schelling, line 12: Intentional?

node-graph/wgpu-executor/src/texture_conversion.rs
```diff
@@ -0,0 +1,269 @@
...
+use graphene_core::table::{Table, TableRow};
+use graphene_core::transform::Footprint;
+use wgpu::util::{DeviceExt, TextureDataOrder};
+use wgpu::{Extent3d, TextureDescriptor, TextureDimension, TextureFormat, TextureUsages};
+
+/// Uploads CPU image data to a GPU texture
+///
+/// Creates a new WGPU texture with RGBA8UnormSrgb format and uploads the provided
+/// image data. The texture is configured for binding, copying, and source operations.
+fn upload_to_texture(device: &std::sync::Arc<wgpu::Device>, queue: &std::sync::Arc<wgpu::Queue>, image: &Raster<CPU>) -> wgpu::Texture {
```
timon-schelling, line 17: I'm a little annoyed that upload and download look so different, but I guess not much we can do. A function is the correct abstraction for upload and a helper struct for download.

node-graph/gcore/src/registry.rs
TrueDoctor: Yes, I needed this for debugging and there is not really a good reason to not have this derive debug

PR #3196, opened 2025-09-19
Add median filter option to blur node

youssefelzedy:
This PR adds median filtering capability to the existing blur node, providing users with a third filtering option alongside Gaussian and box blur.

## Changes Made
- Added `median` boolean parameter to the blur node
- Implemented `median_filter_algorithm()` function that processes each color channel independently
- Added `median_quickselect()` helper function using `select_nth_unstable_by()` for O(n) median calculation
- Updated node documentation to describe median filter use case
- Maintains existing gamma space support for consistency

## Use Case
Median filtering is particularly useful for:
- Removing salt-and-pepper noise from images
- Preserving sharp edges while reducing noise
- Situations where traditional blur would be too destructive to image details

## Technical Notes
- Uses quickselect algorithm (O(n) average case) instead of full sorting (O(n log n))
- Operates on square neighborhoods defined by radius parameter
- Processes RGBA channels independently to maintain color fidelity
- Integrates seamlessly with existing blur node UI and parameters

The implementation follows the same pattern as existing blur algorithms with proper gamma space handling and associated alpha processing.

Closes feature in Issue No. #912

### DIFF
--- a/node-graph/graster-nodes/src/filter.rs
+++ b/node-graph/graster-nodes/src/filter.rs
@@ -6,7 +6,7 @@ use graphene_core::raster_types::{CPU, Raster};
 use graphene_core::registry::types::PixelLength;
 use graphene_core::table::Table;
 
-/// Blurs the image with a Gaussian or blur kernel filter.
+/// Blurs the image with a Gaussian, blur kernel or Median filter.
 #[node_macro::node(category("Raster: Filter"))]
 async fn blur(
 	_: impl Ctx,
@@ -18,6 +18,8 @@ async fn blur(
 	radius: PixelLength,
 	/// Use a lower-quality box kernel instead of a circular Gaussian kernel. This is faster but produces boxy artifacts.
 	box_blur: bool,
+	/// Use a median filter instead of a blur. This is good for removing noise while preserving edges, but does not produce a smooth blur effect.
+	median: bool,
 	/// Opt to incorrectly apply the filter with color calculations in gamma space for compatibility with the results from other software.
 	gamma: bool,
 ) -> Table<Raster<CPU>> {
@@ -32,7 +34,10 @@ async fn blur(
 				image.clone()
 			} else if box_blur {
 				Raster::new_cpu(box_blur_algorithm(image.into_data(), radius, gamma))
-			} else {
+			} else if median {
+				Raster::new_cpu(median_filter_algorithm(image.into_data(), radius as u32, gamma))
+			}
+			else {
 				Raster::new_cpu(gaussian_blur_algorithm(image.into_data(), radius, gamma))
 			};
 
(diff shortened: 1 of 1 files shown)


node-graph/graster-nodes/src/filter.rs
```diff
@@ -179,3 +184,59 @@ fn box_blur_algorithm(mut original_buffer: Image<Color>, radius: f64, gamma: boo
...
+	let (width, height) = original_buffer.dimensions();
+	let mut output = Image::new(width, height, Color::TRANSPARENT);
+
+	for y in 0..height {
+		for x in 0..width {
+			// Collect pixel neighborhood
+			let mut r_vals = Vec::with_capacity(((2 * radius + 1).pow(2)) as usize);
+			let mut g_vals = Vec::with_capacity(r_vals.capacity());
+			let mut b_vals = Vec::with_capacity(r_vals.capacity());
+			let mut a_vals = Vec::with_capacity(r_vals.capacity());
```
TrueDoctor: Could these memory allocations be reused by moving them out of the loops and just clearing them?

youssefelzedy: yes, you are totally right!

node-graph/graster-nodes/src/filter.rs
```diff
@@ -179,3 +183,68 @@ fn box_blur_algorithm(mut original_buffer: Image<Color>, radius: f64, gamma: boo
...
+
+	output
+}
+
+/// Finds the median of a slice using quickselect for efficiency.
+/// This avoids the cost of full sorting (O(n log n)).
+fn median_quickselect(values: &mut [f32]) -> f32 {
+	let mid: usize = values.len() / 2;
+	// nth_unstable is like quickselect: average O(n)
+	*values.select_nth_unstable_by(mid, |a, b| a.partial_cmp(b).unwrap()).1
```
pradkrish: I am still new to rust and this repo, so I might be wrong here. could `unwrap()` potentially panic in some cases? If that happens the function wouldn't return `f32`. Also, what happens if `values` is empty?

TrueDoctor: They would only panic if NAN values are compared, which should never happen since those are not valid colors. But it will panic on empty Images

pradkrish: In that case, it's good to check for that condition.

TrueDoctor: Actually, what I meant to say is it will panic if the function is called with an empty range, but for an empty image it won't because the function will never be called so this is fine

youssefelzedy: What about replacing `partial_cmp(...).unwrap()` with `f32::total_cmp` to avoid panics on NaN values?
Also, I think for even-length slices the current logic picks only the upper median (`len/2`).
If accuracy matters, the median should be computed as the average of the two middle values instead.

TrueDoctor: That should work. Regarding the median averaging, I don't know what the expected result would be (pinging @Keavon) the disadvantage of averaging is that we can now create brightness values which never appear in the artwork. I don't know if that is important for anything. Computing the second value should be relatively cheap though, which is good

youssefelzedy: Actually, it works fine. I tested it before making a PR, but I'm not sure about the performance, and maybe I didn't recognize something else. We should wait until @Keavon makes a last review of the code.

TrueDoctor: @Keavon reminder

TrueDoctor: @youssefelzedy I think using total_cmp would probably be a good addition, we can hopefully get this merged soon, sorry for the long wait

Keavon:
Please make this a separate node.

node-graph/graster-nodes/src/filter.rs
youssefelzedy: @TrueDoctor Ok, I’ll work on this as soon as I can

youssefelzedy:
> Please make this a separate node.

@Keavon  Could you explain more? What do you mean by that?

You mean separating the node for the median filter functionality instead of adding it as an option to the existing blur node?

node-graph/graster-nodes/src/filter.rs
youssefelzedy: @TrueDoctor I did it!

Keavon:
> You mean separating the node for the median filter functionality instead of adding it as an option to the existing blur node?

Yes

youssefelzedy:
@Keavon Could you review it?

node-graph/nodes/raster/src/filter.rs
```diff
@@ -179,3 +211,69 @@ fn box_blur_algorithm(mut original_buffer: Image<Color>, radius: f64, gamma: boo
...
+		output.map_pixels(|px| px.to_linear_srgb().to_unassociated_alpha());
+	} else {
+		output.map_pixels(|px| px.to_unassociated_alpha());
+	}
+
+	output
+}
+
+/// Finds the median of a slice using quickselect for efficiency.
+/// This avoids the cost of full sorting (O(n log n)).
```
Keavon: By putting it at the end in parentheses, this comment is unclear to me whether the `(O(n log n))` refers to this function's runtime or if it refers to `the cost of full sorting`.

node-graph/nodes/raster/src/filter.rs
```diff
@@ -42,6 +42,38 @@ async fn blur(
...
+#[node_macro::node(category("Raster: Filter"))]
+async fn median_filter(
+	_: impl Ctx,
+	/// The image to be filtered.
+	image_frame: Table<Raster<CPU>>,
+	/// The radius of the filter kernel. Larger values remove more noise but may blur fine details.
+	#[range((0., 50.))]
+	#[hard_min(0.)]
+	radius: PixelLength,
+	/// Opt to incorrectly apply the filter with color calculations in gamma space for compatibility with the results from other software.
```
Keavon: We may need to test if other graphics tools use gamma or linear space for their implementations.

Keavon:
!build

node-graph/nodes/raster/src/filter.rs
youssefelzedy: You're absolutely right about the comment ambiguity. I've updated it to be clearer:

```rust
/// Finds the median of a slice using quickselect for O(n) average case performance.
/// This is more efficient than sorting the entire slice which would be O(n log n).
```

The original comment was confusing because the `(O(n log n))` at the end could be interpreted as either this function's complexity or the complexity we're avoiding. This new version explicitly states:

This function's performance: O(n) average case

node-graph/nodes/raster/src/filter.rs
youssefelzedy: Actually, we might not need to test this for the Median filter. Since gamma correction is a monotonic function, the sort order of pixel values doesn't change between Linear and Gamma space. Therefore, the median pixel selected will be mathematically identical in both, unlike arithmetic filters (e.g., Gaussian blur), where space matters.

Keavon: Ok, please remove it then.

TrueDoctor: It will differ if the median is calculated by taking the average of the middle two elements, but I guess that should usually not be an issue

youssefelzedy: Hmm, you’re right, but does it really affect the output?

youssefelzedy:
@Keavon , any updates?

Keavon:
!build

youssefelzedy:
@Keavon @TrueDoctor  any update?

TrueDoctor:
Merging this since the code itself is fine and when Keavon has more time to review the behavior, we can make a followup pr

TrueDoctor:
And again @youssefelzedy sorry for the long delay, and thanks for your patience

youssefelzedy:
@TrueDoctor No worries at all, thanks for that, and I appreciate your time.

PR #3197, opened 2025-09-19
Desktop: Forward the pointer leave event to CEF

timon-schelling:


desktop/src/cef/input.rs
```diff
@@ -11,14 +11,24 @@ use super::consts::{MULTICLICK_ALLOWED_TRAVEL, MULTICLICK_TIMEOUT, SCROLL_LINE_H
...
 			};
 			host.send_mouse_move_event(Some(&input_state.into()), 0);
 		}
+		WindowEvent::PointerLeft { position, .. } => {
+			if let Some(position) = position {
+				input_state.cursor_move(position);
+			}
+
+			let Some(host) = browser.host() else {
+				return;
```
Keavon: Make this a one-liner, please.

timon-schelling: Is from auto formating

Keavon: The one-liner version is short enough, so you can remove the semicolon and it'll stay as a one-liner.

timon-schelling: did not know that. thx

Keavon: If the line becomes long enough, it'll force it back to three lines, but it's usually worth trying to get the short version. Occasionally you'll have a bunch of long lines mixed in with one short line, in which case I usually opt for consistency to let them all be the three-liner, but otherwise prefer attempting to get the short version if nearby consistency isn't a concern.

timon-schelling:
@TrueDoctor do you understand why profiling is broken again?

TrueDoctor:
> @TrueDoctor do you understand why profiling is broken again?

I would assume a non-upgraded demo artwork

PR #3198, opened 2025-09-20
preference load message on default load

timon-schelling:
Previously, when no preferences were stored, the Load Preferences message was never called. This caused issues with hole-punch setting on desktop not being set correctly. Users had to manually go into settings and toggle Vello off and back on for it to take effect.

This PR fixes the issue by always sending a Load Preferences message from both frontends, even when there are no saved preferences. In that case, preferences are set to None to explicitly indicate that nothing was loaded (default preferences will be used).

### DIFF
--- a/frontend/src/io-managers/persistence.ts
+++ b/frontend/src/io-managers/persistence.ts
@@ -147,9 +147,7 @@ export function createPersistenceManager(editor: Editor, portfolio: PortfolioSta
 
 	async function loadPreferences() {
 		const preferences = await get<Record<string, unknown>>("preferences", graphiteStore);
-		if (!preferences) return;
-
-		editor.handle.loadPreferences(JSON.stringify(preferences));
+		editor.handle.loadPreferences(preferences ? JSON.stringify(preferences) : null);
 	}
 
 	// FRONTEND MESSAGE SUBSCRIPTIONS
--- a/frontend/wasm/src/editor_api.rs
+++ b/frontend/wasm/src/editor_api.rs
@@ -421,14 +421,18 @@ impl EditorHandle {
 	}
 
 	#[wasm_bindgen(js_name = loadPreferences)]
-	pub fn load_preferences(&self, preferences: String) {
-		let Ok(preferences) = serde_json::from_str(&preferences) else {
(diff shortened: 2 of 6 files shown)


frontend/src/io-managers/persistence.ts
```diff
@@ -147,9 +147,7 @@ export function createPersistenceManager(editor: Editor, portfolio: PortfolioSta
 
 	async function loadPreferences() {
 		const preferences = await get<Record<string, unknown>>("preferences", graphiteStore);
-		if (!preferences) return;
-
-		editor.handle.loadPreferences(JSON.stringify(preferences));
+		editor.handle.loadPreferences(preferences ? JSON.stringify(preferences) : null);
```
Keavon: Can we pass `undefined` instead of `null`? We ban the `null` keyword for style reasons. Otherwise we may need an alternate way to do that. Hopefully I can also get style linting working in CI again soon.

Keavon: Or maybe conditionally passing no argument could be another option that might work, if `undefined` doesn't?

frontend/wasm/src/editor_api.rs
```diff
@@ -421,14 +421,18 @@ impl EditorHandle {
 	}
 
 	#[wasm_bindgen(js_name = loadPreferences)]
-	pub fn load_preferences(&self, preferences: String) {
-		let Ok(preferences) = serde_json::from_str(&preferences) else {
-			log::error!("Failed to deserialize preferences");
-			return;
+	pub fn load_preferences(&self, preferences: Option<String>) {
+		let preferences = if let Some(preferences) = preferences {
```
Keavon, line 425: I think this should be a `.map()` instead of if-let-else?

Keavon: Oh, I guess you need to not be in a closure for the early return.

Keavon:
I haven't QA'd this but my review concerns are addressed so you may merge anytime.

timon-schelling:
Was waiting on rust-gpu ci issue fix. Did testing locally.

PR #3201, opened 2025-09-20
Desktop: Add rudimentary support for custom WGPU adapter selection

timon-schelling:
Temporary:
- adapter override with env `GRAPHITE_WGPU_ADAPTER` (will be replaced with cli option)
- print adapters (will be behind a `--help` like cli flag in the future)

Also selects the first DX12 adapter on windows, usually (but not in the spec) dedicated gpu > integrated gpu > software.


desktop/src/main.rs
```diff
@@ -67,3 +67,60 @@ fn main() {
 
 	event_loop.run_app(&mut app).unwrap();
 }
+
+async fn init_wgpu_context() -> WgpuContext {
```
TrueDoctor: I don't think this is the right place to put this function. I'd generall prefer if the reusable logic was moved directly into the wgpu executor (e.g. such that it can also be used by the graphene cli etc.) and for the desktop specific parts I'd like to move those to a different file in order to not clutter up the `main.rs`

timon-schelling: Forgot about graphene cli, sure. Are we sure we want the exact same adapter selection code for both? what would you like the adapters to be addressed by? Is the index (that is stable for a specific system/hardware configuration), I also thought about hashing the adapters info and using 4 bytes as a id. I will implement the printing as fmt then.

TrueDoctor: I would expose multiple function with different levels of granularity. We could have one function which does everything automagically, one which allows you to request the graphics api and one which gives you full manual control over the adapter selection. But I'd like to avoid constructign the context outside of the wgpu executor crate in case we in the future need to enable extra extensions etc. It would be a nightmare if we had to keep those in sync

desktop/src/gpu_context.rs
```diff
@@ -0,0 +1,25 @@
...
+	// TODO: add a cli flag to list adapters and exit instead of always printing
+	println!("\nAvailable WGPU adapters:\n{}", wgpu_context_builder.available_adapters_fmt().await);
+
+	// TODO: make this configurable via cli flags instead
+	let wgpu_context = match std::env::var("GRAPHITE_WGPU_ADAPTER").ok().and_then(|s| s.parse().ok()) {
+		None => wgpu_context_builder.build().await,
+		Some(adapter_index) => {
+			tracing::info!("Overriding WGPU adapter selection with adapter index {adapter_index}");
+			wgpu_context_builder
+				.build_with_adapter_selection(|adapters: &mut Vec<_>| if adapter_index < adapters.len() { Some(adapters.remove(adapter_index)) } else { None })
```
TrueDoctor: I think it would make the api cleaner if this was of type `Fn(&[Adapter]) -> Option<usize>`

desktop/src/main.rs
```diff
@@ -14,6 +12,8 @@ mod native_window;
 mod persist;
 mod render;
 
+mod gpu;
```
TrueDoctor: maybe call this `(w)gpu_context`? I thing `gpu` on its own is a bit vague

node-graph/wgpu-executor/src/context.rs
```diff
@@ -1,53 +1,152 @@
...
+			required_limits: adapter.limits(),
+			memory_hints: Default::default(),
+			trace: wgpu::Trace::Off,
+		};
+		adapter.request_device(&device_descriptor).await.ok()
+	}
+}
+#[cfg(not(target_family = "wasm"))]
+impl ContextBuilder {
+	#[cfg(not(target_family = "wasm"))]
```
TrueDoctor: this cfg does not need to be repeated right?

node-graph/wgpu-executor/src/context.rs
```diff
@@ -1,53 +1,152 @@
...
+		} else if cfg!(target_os = "windows") {
+			self.select_adapter(&instance, |adapters: &mut Vec<Adapter>| {
+				adapters.iter().position(|a| a.get_info().backend == wgpu::Backend::Dx12).map(|i| adapters.remove(i))
 			})
-			.await
-			.ok()?;
+		} else {
+			None
+		};
+		#[cfg(target_family = "wasm")]
```
TrueDoctor: I'm a bit confused by these cfg attributes, the whole block should be disabled, right?

TrueDoctor (approved):

PR #3203, opened 2025-09-21
Add tooltip to primary input to use node.reference

### LINKED ISSUE #3202: Show node name instead of custom alias on tooltip
Change the behavior of the tooltip on nodes to show the node name rather than any user alias given.

peteralieber:
Closes #3202

### DIFF
--- a/frontend/src/components/views/Graph.svelte
+++ b/frontend/src/components/views/Graph.svelte
@@ -666,7 +666,7 @@
 				<div class="primary" class:in-selected-network={$nodeGraph.inSelectedNetwork} class:no-secondary-section={exposedInputsOutputs.length === 0}>
 					<IconLabel icon={nodeIcon(node.reference)} />
 					<!-- TODO: Allow the user to edit the name, just like in the Layers panel -->
-					<TextLabel>{node.displayName}</TextLabel>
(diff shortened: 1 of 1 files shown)


peteralieber:
I would like to get descriptions of nodes in the tooltip, just like we have with the secondary inputs and outputs, but the node itself does not have a "Description" field, so that I'll leave for a follow on as it needs to populate descriptions for all the nodes, and add the field to metadata struct used in the frontend.

Keavon:
It looks like your code adds a new tooltip, but the tooltip already exists and just needs to have the correct name shown. Currently it shows the user alias but not the node type with the existing tooltip.

peteralieber:
> ser alias but not the node type with the existing tooltip.

Interesting, it seems to behave differently in my local testing. I see the node ref/type in my tooltip. I know the tooltip exists, but does it exist in a parent container or something? I did not see it or could not find the tooltip to change. I'll look again.

adamgerhant:
This has been implemented with the reference now included in parentheses if it does not match the display name
image

peteralieber:
What was the commit that implemented this? I just want to see the right way to do it, as mine apparently wasn't.

adamgerhant:
I believe here: https://github.com/GraphiteEditor/Graphite/pull/3436

PR #3209, opened 2025-09-22
Abort pen tool while performing group operation

4adex:
https://discord.com/channels/731730685944922173/1418383751293173761/1418716184609816586


Keavon:
Apparently this has been fixed at some point in the past. Ctrl+G during a Pen tool drawing state does end the interactive drawing state.

PR #3210, opened 2025-09-22
Desktop: Fix maximized window extending outside the screen on Windows

timon-schelling:
Part of #3191

We should discuss if this closes it.

On small issue remaining, due to using a custom window frame winit::Window::is_maximized always returns false on win32 and we cant update the icon correctly because maximizing looks like resizing to winit.

There are workarounds, but my preferred solution would be to introduce a WindowHandle type that holds both the window and native window handle and would have working implementations on all platforms for things like is_maximized (could depending on platform use native window handle or window to determine maximized state etc.).

That will need some work though and I would prefer merging this PR before the event on the 26th.

### DIFF
--- a/desktop/Cargo.toml
+++ b/desktop/Cargo.toml
@@ -60,6 +60,7 @@ windows = { version = "0.58.0", features = [
 	"Win32_System_LibraryLoader",
 	"Win32_UI_Controls",
 	"Win32_UI_WindowsAndMessaging",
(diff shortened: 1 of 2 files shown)


TrueDoctor:
I don't think I'm the best person to test this but I guess the code looks fine

timon-schelling:
> I don't think I'm the best person to test this but I guess the code looks fine

It was tested on Windows by keavon.

PR #3224, opened 2025-09-24
Desktop: Forward multi-clicks (double/triple/etc.-clicks) to CEF correctly

timon-schelling:


desktop/src/cef/input.rs
```diff
@@ -278,7 +300,7 @@ impl ClickTracker {
 	}
 }
 
-#[derive(Clone)]
+#[derive(Clone, PartialEq)]
```
TrueDoctor: could we derive copy here to avoid having to write `.clone()` ?

desktop/src/cef/input.rs
```diff
@@ -253,6 +253,28 @@ impl ClickTracker {
...
+					up_count: record.up_count.clone(),
+				};
+				return ClickCount::Single;
+			}
+			ElementState::Released if record.up_count == ClickCount::Double => {
+				*record = ClickRecord {
+					time: now,
+					position: position.clone(),
+					down_count: record.down_count.clone(),
+					up_count: ClickCount::Single,
```
TrueDoctor, line 263: Also could you use the struct initializer syntax here?

timon-schelling: I think no.

PR #3232, opened 2025-09-27
Add an implementation to the Rasterize node for Vello

### LINKED ISSUE #3719: No rasterize node on desktop
As seen in the image, there is no rasterize node. I am running ubuntu

Image

This is running on Ubuntu 22.04.5 LTS, running after cloning the repo and running the start-desktop command.

TrueDoctor:
Closes #3719

### DIFF
--- a/node-graph/gstd/src/wasm_application_io.rs
+++ b/node-graph/gstd/src/wasm_application_io.rs
@@ -3,22 +3,19 @@ use base64::Engine;
 pub use graph_craft::document::value::RenderOutputType;
 pub use graph_craft::wasm_application_io::*;
 use graphene_application_io::ApplicationIo;
-#[cfg(target_family = "wasm")]
 use graphene_core::gradient::GradientStops;
 #[cfg(target_family = "wasm")]
 use graphene_core::math::bbox::Bbox;
 use graphene_core::raster::image::Image;
 use graphene_core::raster_types::{CPU, Raster};
 use graphene_core::table::Table;
-#[cfg(target_family = "wasm")]
 use graphene_core::transform::Footprint;
-#[cfg(target_family = "wasm")]
 use graphene_core::vector::Vector;
 use graphene_core::{Color, Ctx};
-#[cfg(target_family = "wasm")]
 use graphene_core::{Graphic, WasmNotSend};
+use graphene_svg_renderer::Render;
 #[cfg(target_family = "wasm")]
-use graphene_svg_renderer::{Render, RenderParams, RenderSvgSegmentList, SvgRender};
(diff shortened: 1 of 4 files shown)


node-graph/gstd/src/wasm_application_io.rs
```diff
@@ -209,3 +206,71 @@ where
...
+		*row.transform = glam::DAffine2::from_translation(-aabb.start) * *row.transform;
+	}
+
+	// Create Vello scene and render context
+	let mut scene = vello::Scene::new();
+	let mut context = RenderContext::default();
+
+	// Render data to Vello scene
+	let render_params = graphene_svg_renderer::RenderParams {
+		footprint: Footprint::default(),
```
timon-schelling, line 251: Isn't this included in `..Default::default()`?

node-graph/gstd/src/wasm_application_io.rs
```diff
@@ -209,3 +206,71 @@ where
...
+		.render_vello_scene_to_texture(&scene, resolution, &context, background)
+		.await
+		.expect("Failed to render Vello scene to texture");
+
+	// Wrap the texture in a Raster<GPU>
+	let gpu_raster = Raster::new_gpu(wgpu_texture);
+
+	// Convert GPU raster to CPU raster using Convert trait
+	use graphene_core::ops::Convert;
+	let cpu_raster = gpu_raster.convert(Footprint::default(), wgpu_executor).await;
```
timon-schelling, line 269: Why don't we return a GPU texture here? Vector -> Rasterize -> Invert GPU and similar usecases.

TrueDoctor: I wanted this to match the CPU version also to make this usable for the led artwork

TrueDoctor: But we should make the two nodes compatible with each other so opening a document created in web also works on desktop

timon-schelling: Could we make two nodes one with CPU and one with GPU output, both using the same utility function to render?

timon-schelling (approved):
Feel free to merge this. ~I will at some point create a follow-up PR with a GPU node.~

node-graph/gstd/src/wasm_application_io.rs
TrueDoctor: I'd rather wait until we have proper automatic type conversion, this should ideally not be complexity we expose to the user, if we can avoid it

timon-schelling: We already do with all the oder GPU nodes

TrueDoctor: Which is something I want to fix asap

PR #3233, opened 2025-09-27
Clean up execution context for skipped node graph submissions

TrueDoctor:
Reduces memory leakage rate for the LED demo from around 500MB/h -> 30MB/h


timon-schelling (approved):
Dose this remove all `leaked` mem caused by unused executions? 30MB/H left are from some other code?

PR #3236, opened 2025-09-28
Add arrow key navigation for node catalog

### LINKED ISSUE #3235: Keyboard navigation for node catalog
Navigate through nodes using up and down arrow keys and select using enter key in the node catalog.
Useful for quickly selecting nodes after searching.

Rohan2460:
Implements navigation in node catalog using up and down arrow keys and node selection using enter key.
Selected node is emphasized.
 
Closes #3235

### DIFF
--- a/frontend/src/components/floating-menus/NodeCatalog.svelte
+++ b/frontend/src/components/floating-menus/NodeCatalog.svelte
@@ -19,6 +19,9 @@
 
 	$: nodeCategories = buildNodeCategories($nodeGraph.nodeTypes, searchTerm);
 
+	let focusedNodeIndex = 0;
+	$: flatNodeList = nodeCategories.flatMap((node) => node[1].nodes);
+
 	type NodeCategoryDetails = {
 		nodes: FrontendNodeType[];
 		open: boolean;
@@ -29,6 +32,7 @@
 		const isTypeSearch = searchTerm.toLowerCase().startsWith("type:");
 		let typeSearchTerm = "";
 		let remainingSearchTerms = [searchTerm.toLowerCase()];
+		focusedNodeIndex = 0;
 
 		if (isTypeSearch) {
 			// Extract the first word after "type:" as the type search
@@ -104,11 +108,32 @@
 			});
 	}
 
+	function keyboardNavigationHandler(e: KeyboardEvent) {
+		const listLength: number = flatNodeList.length;
+		if (listLength === 0) return;
+
+		if (e.key === "ArrowDown") {
+			focusedNodeIndex = (focusedNodeIndex + 1) % listLength;
+			e.preventDefault();
+		} else if (e.key === "ArrowUp") {
+			focusedNodeIndex = (focusedNodeIndex - 1 + listLength) % listLength;
+			e.preventDefault();
+		} else if (e.key === "Enter") {
+			const node = flatNodeList[focusedNodeIndex];
+			dispatch("selectNodeType", node.name);
+		} else {
+			return;
+		}
+
+		setTimeout(() => document.querySelector(".node-catalog button.emphasized")?.scrollIntoView({ block: "nearest" }), 0);
+	}
+
 	onMount(() => {
 		setTimeout(() => nodeSearchInput?.focus(), 0);
 	});
 </script>
 
+<svelte:window on:keydown={keyboardNavigationHandler} />
 <div class="node-catalog">
 	<TextInput placeholder="Search Nodes..." value={searchTerm} on:value={({ detail }) => (searchTerm = detail)} bind:this={nodeSearchInput} />
 	<div class="list-results" on:wheel|passive|stopPropagation>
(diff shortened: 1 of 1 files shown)


timon-schelling:
!build

frontend/src/components/floating-menus/NodeCatalog.svelte
```diff
@@ -118,7 +143,13 @@
 					<TextLabel>{nodeCategory[0]}</TextLabel>
 				</summary>
 				{#each nodeCategory[1].nodes as nodeType}
-					<TextButton {disabled} label={nodeType.name} tooltip={$nodeGraph.nodeDescriptions.get(nodeType.name)} action={() => dispatch("selectNodeType", nodeType.name)} />
+					<TextButton
+						{disabled}
+						label={nodeType.name}
+						tooltip={$nodeGraph.nodeDescriptions.get(nodeType.name)}
+						emphasized={nodeType == flatNodeList[focusedNodeIndex]}
```
timon-schelling, line 150: Image

I think emph is not ideal could you use the color used for mouse hover.

timon-schelling (commented):
Thanks for the PR, Good UX improvement. Personally looking forward to using this.

Rohan2460:
@timon-schelling I added 'focused' prop to TextButton, is that ok ?
graphite_node

timon-schelling:
!build

timon-schelling:
I will take a look once the build is done.

timon-schelling:
!build

frontend/src/components/floating-menus/NodeCatalog.svelte
```diff
@@ -104,11 +108,32 @@
 			});
 	}
 
+	function keyboardNavigationHandler(e: KeyboardEvent) {
```
timon-schelling, line 111: I'm relatively sure keavon will not like that this doesn't use the standard keyboard mapping. Should probably be handled in editor and only the visual aspect should be in the frontend. Would also allow showing the shortcuts in the bar at the bottom of the screen. You will need to look into how similar things are implemented.

Keavon: If it doesn't break editor interactions, that's an ok compromise for now to not route this through the editor backend. The intention is to fully replace the node catalog menu in the future, it has always been a placeholder.

timon-schelling: It doesn't break editor interactions. Then LGTM

timon-schelling (approved):

timon-schelling:
this can be merged right? @Keavon not sure if your comment meant approval...

frontend/src/components/floating-menus/NodeCatalog.svelte
```diff
@@ -104,11 +108,32 @@
...
+			focusedNodeIndex = (focusedNodeIndex - 1 + listLength) % listLength;
+			e.preventDefault();
+		} else if (e.key === "Enter") {
+			const node = flatNodeList[focusedNodeIndex];
+			dispatch("selectNodeType", node.name);
+		} else {
+			return;
+		}
+
+		setTimeout(() => document.querySelector(".node-catalog button.focused")?.scrollIntoView({ block: "nearest" }), 0);
```
Keavon: We do not allow querying elements except by data selectors.

frontend/src/components/widgets/buttons/TextButton.svelte
```diff
@@ -15,6 +15,7 @@
 	export let icon: IconName | undefined = undefined;
 	export let hoverIcon: IconName | undefined = undefined;
 	export let emphasized = false;
+	export let focused = false;
```
Keavon: Changing this file shouldn't be necessary. You can just use the existing `emphasized`.

Keavon:
Still need to QA and do another review round.

frontend/src/components/widgets/buttons/TextButton.svelte
timon-schelling: I suggested using the hover colors for keyboard selection, white looks to harsh in my opinion. Makes it consistent with mouse input.

Keavon: Let's stick with the white emphasized color for now because this is temporary anyways.

timon-schelling: Temp yes fair, my main concern is that ux is not that great with white emph.
1. keyboard selection is not changing the type of button and I would expect it to be more like hover or have some outline.
2. white emph that moves with keyboard selection is harder on the eyes, I find it hard to read if the color changes so rapidly.

If you don't want to change `TextButton` maybe we could add css to the temporary dialog?

Keavon: CSS in the dialog would be acceptable.

timon-schelling: @Rohan2460 Should be easy to combine that with the data selector change.

Rohan2460: Should I override `emphasized` class in the node catalog ?

timon-schelling: I think so yes

Rohan2460:
I reverted the changes to the TextButton component and overwrote the emphasized css class.

timon-schelling:
!build

Ezbaze:
I think this PR may be outdated now. Commit https://github.com/GraphiteEditor/Graphite/commit/600fb5c28f9bdedbdcaf06b5ab1804f674c6e9d7 already adds tab navigation within the node graph editor context menu, which seems to cover the same issue.

Context menu behaviour as of https://github.com/GraphiteEditor/Graphite/commit/d6c06da878beca527f1c90b4ace000fca98752b3 (latest commit as of writing this comment)

https://github.com/user-attachments/assets/4f31891b-4c40-4e8e-aa55-4da135688a4c

CC @Keavon

timon-schelling:
In my opinion, the tab selection is not sufficient, highlighting could be better and arrow key navigation is more usable. We should also think about adding a keyboard shortcut that allow users to open the dialog without touching the mouse.
Node selection (witch is inherently keyboard focused) would be faster.

Keavon:
Since this is now possible from the keyboard as Ezbaze showed in the video he posted, that's sufficient. The entire menu will be replaced by a proper context menu once that system comes online pretty soon. So this should indeed make this PR redundant. Thanks @Rohan2460 for your attempt on it!

PR #3249, opened 2025-10-03
Implement layer rearranging using pointer instead of DnD API

### LINKED ISSUE #2991: Rearranging Layers in the Layer Panel does not work

TrueDoctor:
Closes #2991 

Known issues: When dragging a layer and dropping it onto itself, we sometimes trigger a rename

### DIFF
--- a/frontend/src/components/layout/LayoutCol.svelte
+++ b/frontend/src/components/layout/LayoutCol.svelte
@@ -47,6 +47,8 @@
 	on:pointerdown
 	on:pointerenter
 	on:pointerleave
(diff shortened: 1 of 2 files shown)


TrueDoctor:
@Keavon I removed the usages of the pointer capture API and everything still works as expected

Keavon:
Superseded by #3426 which correctly preserves the web version's drag-and-drop of an image file into the Layers panel. This functionality will eventually need to be built in the desktop version using some kind of strategy we'll have to devise.

PR #3250, opened 2025-10-03
Provide shaders via artifact repo (prototype)

timon-schelling:
Build and upload graphene raster nodes shaders in a separate artifacts repo.
Needed for build environments were we can't use the rust gpu pinned rust nightly toolchain in. I'm most concerned about nixpkgs. Can also be used to significantly speed up builds (desktop on all platforms).

Currently running on every push we can think about doing this only on master, but runtime is <3m so should be fine for now.

Currently using my own repo, should preoby live in a GraphiteInfra org.

I have set the following for the repo.

vars:
ARTIFACTS_REPO_OWNER = timon-schelling
ARTIFACTS_REPO_NAME = graphite-artifacts
ARTIFACTS_REPO_BRANCH = main

secrets:
ARTIFACTS_REPO_TOKEN = github_pat_*** (access token limited to write on that one repo)

### DIFF
--- a/.github/workflows/scripts/artifact-upload.bash
+++ b/.github/workflows/scripts/artifact-upload.bash
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  cat <<EOF
+Usage: $0 <owner> <repo> <branch> <target-path> <artifact-file> <commit-message> <github-token>
+
+Arguments:
+  owner           : GitHub user or organization of the target repo
+  repo            : Target repo name
+  branch          : Branch name (e.g. main)
+  target-path     : Full path (including folders + filename) in the target repo where to upload
+  artifact-file   : Local file path to upload
+  commit-message  : Commit message for creating/updating the file
+  github-token    : GitHub token (PAT or equivalent) with write access to the target repo
+
+This will perform a GitHub API PUT to /repos/{owner}/{repo}/contents/{target-path}.
+If a file already exists at that path, it will auto-detect the SHA and update; otherwise it will create.
+EOF
+  exit 1
+}
+
+if [ $# -ne 7 ]; then
+  usage
+fi
+
+OWNER="$1"
+REPO="$2"
+BRANCH="$3"
+TARGET_PATH="$4"
+ARTIFACT_PATH="$5"
+COMMIT_MSG="$6"
+TOKEN="$7"
+
+if [ ! -f "$ARTIFACT_PATH" ]; then
+  echo "Error: artifact file not found: $ARTIFACT_PATH" >&2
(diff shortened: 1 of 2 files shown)


Firestar99:
With @TrueDoctor adding the automatic switches between the GPU and CPU variants, end users shouldn't need to use GPU nodes directly. I assume if GPU nodes aren't build, we just default to the CPU variants. So for a typical end-user who doesn't build shaders, everything should still work as expected, at most they get a failing CI later cause the shaders don't compile. Now you'd have to make sure a regular contributor changing a shader node uses the CPU implementation and not the GPU implementation from cached master, without their changes.
(In general, I'm not a fan of caching things externally from your project, due to cache invalidation concerns.)

Or is this only for building desktop on some platforms? Couldn't we have a github job / nix thing to build the shaders and pass the artifacts to the next step, entering whatever build env doesn't support rust-gpu's separate toolchain?

timon-schelling:
> Or is this only for building desktop on some platforms?

Yes and also only a PoC.

> Couldn't we have a github job / nix thing to build the shaders and pass the artifacts to the next step, entering whatever build env doesn't support rust-gpu's separate toolchain?

This essentially dose that, but nixpkgs is not part of our ci thus we need to host the artifacts needed for that build statically somewhere (probably a bucket in the end).
The source for the build and shader are build from the exact same commit, this way we ensure that they are never out of sync.

TrueDoctor:
What's the state on this? I'd prefer if we can start using this for a bit to have enough time to iron out any bugs we might encounter

timon-schelling:
I used this for all nixpkg PR updates so far, seems to work fine, just needs some cleanup, like renaming some stuff, limiting to master etc.

timon-schelling:
@TrueDoctor I guess a review from you would be helpfull.

.github/workflows/scripts/artifact-upload.bash
```diff
@@ -0,0 +1,89 @@
...
+  owner           : GitHub user or organization of the target repo
+  repo            : Target repo name
+  branch          : Branch name (e.g. main)
+  target-path     : Full path (including folders + filename) in the target repo where to upload
+  artifact-file   : Local file path to upload
+  commit-message  : Commit message for creating/updating the file
+  github-token    : GitHub token (PAT or equivalent) with write access to the target repo
+
+This will perform a GitHub API PUT to /repos/{owner}/{repo}/contents/{target-path}.
+If a file already exists at that path, it will auto-detect the SHA and update; otherwise it will create.
```
TrueDoctor:
```suggestion
If a file already exists at that path, it will auto-detect the SHA and update; otherwise it will create a new one.
```

TrueDoctor (approved):

PR #3257, opened 2025-10-05
Update cargo-gpu, remove legacy-target-specs to allow vendoring cargo dependecies

Firestar99:


timon-schelling:
changed the title because the issue also exits with cargo vendor.
```
> cargo vendor
error: failed to sync

Caused by:
  found duplicate version of package `rustc_codegen_spirv-target-specs v0.9.0` vendored from two sources:

        source 1: registry `crates-io`
        source 2: https://github.com/rust-gpu/rust-gpu?rev=c12f216121820580731440ee79ebc7403d6ea04f#c12f2161
```

PR #3274, opened 2025-10-11
Add quick measurement between objects and artboards

### LINKED ISSUE #2942: Quick measurement with Alt should also measure to the edges of artboards
Select an object and hold Alt, then hover over another object. It measures the distance they are separated. We want the same to happen for artboards, as if they were giant rectangles.

TheUnlocked:
Tweaks the hit detection logic for hover effects in the select tool to make quick measurement work on artboards.

Resolves #2942

https://github.com/user-attachments/assets/3a74599b-f98c-4b9d-9161-0d728b94a531


Keavon:
!build

PR #3279, opened 2025-10-13
Round Corners improvements

### LINKED ISSUE #3204: "Round Corners" node removes bezier curves from vectors
As I was trying to recreate one of the examples from the latest update video(the star with the cuts into it) I noticed that whenever I applied a "Round Corners" node to anything that had curves previously (circles etc) they get turned into straight lines before the node is applied. Tested both on editor.graphite.rs and dev.graphite.rs.

https://github.com/user-attachments/assets/71933a27-1193-4442-8854-c5cfa4de6c77

c-mateo:
The Round Corners node has a parameter that controls whether rounded corners are created depending on the angle between edges. I found its behavior somewhat counterintuitive.

In the image below, the minimum angle threshold is set to 90°, but the green and cyan angles are not getting rounded corners even though they are greater than or equal to 90°, while the orange angle, which is less than 90°, does.

image

Work in progress to close #3204.
Currently, it ignores vertices with handles but doesn't work well when the vertex is the first in the path. Any help would be appreciated.

image

### DIFF
--- a/node-graph/gcore/src/vector/misc.rs
+++ b/node-graph/gcore/src/vector/misc.rs
@@ -189,7 +189,10 @@ pub fn bezpath_to_manipulator_groups(bezpath: &BezPath) -> (Vec<ManipulatorGroup
 				if let Some(last_manipulators) = manipulator_groups.pop()
 					&& let Some(first_manipulators) = manipulator_groups.first_mut()
 				{
-					first_manipulators.out_handle = last_manipulators.in_handle;
+					// The popped manipulators represent the duplicate anchor produced by the closing PathEl.
+					// Its `in_handle` corresponds to the second control point of the closing curve,
+					// which should become the first manipulator's `in_handle` (not its `out_handle`).
+					first_manipulators.in_handle = last_manipulators.in_handle;
 				}
 				is_closed = true;
 				break;


c-mateo:
I tried pushing the manipulator groups directly, but the problem persisted:
```rust
for i in 0..manipulator_groups.len() {
    new_manipulator_groups.push(manipulator_groups[i]);
}
```
That pointed me to the `bezpath_to_manipulator_groups` function.
It turns out that `last_manipulators.in_handle` should be assigned to `first_manipulators.in_handle`, not to `first_manipulators.out_handle`. It works fine now.

The current solution for rounding corners is to skip producing a rounded corner for any vertex that either:
- has its own handles, or
- has neighboring vertices with handles pointing toward it.

This ensures that existing curve handles are preserved and not overridden by the corner rounding logic.
However, a more robust solution would be to properly combine rounded corners with existing curves, instead of skipping them entirely, so that rounding can coexist with vertices that have handles.

otdavies:
Hmm, having played around a bit with this, I think we'll need to potentially put off fixing this issue until we can come up with a new algorithm. Realistically you can improve it a little by avoiding anchors with handles, but it only delays things.. you end up with situations like this, since the algorithm assumes straight edges:
image

I'll think on this a bit. Thanks trying to figure this out!

c-mateo:
@otdavies, could you share the Graphite file? I'd like to take a look and see if I can propose a solution. If I can't find a fix, I'll let you know.

otdavies:
@c-mateo This is just with a box with two corners rounded, and this is the non-rounded opposite corner. I wouldn't worry a ton about it, unfortunately I think we need to come up with a slightly different algorithm.

Something that will help though, is the knowledge that when you sharpen corners, sometimes the handles are just hidden below the anchor in the same position - so you also likely need to do a check for distance of anchor to handles to ensure handles do/don't exist.

c-mateo:
@otdavies I can't reproduce this. Could you attach the file or a short screen recording showing the issue and the exact steps to reproduce it? I’ll take a quick look.

otdavies:
I can't get to reproduce now either, so might have been an editor crash or similar that I didn't catch. @c-mateo 

For now I'd pause on this as I think there needs to be a discussion on the goals. We should wait for Keavon. We want the round corners node to behave similar to other vector editors and right now it is not meeting that requirement well. Your changes aren't at all the issue, I just don't think we will be able to make the tool behave better in all cases without an overhaul to the algorithm.

If you specifically want to work towards a final version - go try some other vector editors corner rounding tools and see what ours doesn't do right. Then we can start to build in that direction. Thanks!

c-mateo:
@otdavies I'd like to continue working on this feature. Please let me know when you've decided what should be done.

lilitdn:
Any update on this?

c-mateo:
@0HyperCube has worked on another feature which supersedes the old round corners node. See #3323

I think Round Corners will be replaced and this PR will be closed when someone has a chace to review it.

PR #3281, opened 2025-10-13
Enable SVG import destination options when dragging into canvas

### LINKED ISSUE #3253: Improve SVG import UX
Dropping SVG into an existing document canvas instead of importing it as a new layer/stack could display a context menu with options:
- Import as new layer
- Import as new Document

Dropping an SVG on an empty tab space should import it as new document:

Image

kanishk6103:
Closes #3253

### DIFF
--- a/desktop/wrapper/src/intercept_frontend_message.rs
+++ b/desktop/wrapper/src/intercept_frontend_message.rs
@@ -20,22 +20,6 @@ pub(super) fn intercept_frontend_message(dispatcher: &mut DesktopWrapperMessageD
 				context: OpenFileDialogContext::Document,
 			});
 		}
-		FrontendMessage::TriggerImport => {
-			dispatcher.respond(DesktopFrontendMessage::OpenFileDialog {
-				title: "Import File".to_string(),
-				filters: vec![
-					FileFilter {
-						name: "Svg".to_string(),
-						extensions: vec!["svg".to_string()],
-					},
-					FileFilter {
-						name: "Image".to_string(),
-						extensions: vec!["png".to_string(), "jpg".to_string(), "jpeg".to_string(), "bmp".to_string()],
-					},
-				],
-				context: OpenFileDialogContext::Import,
-			});
-		}
 		FrontendMessage::TriggerSaveDocument { document_id, name, path, content } => {
 			if let Some(path) = path {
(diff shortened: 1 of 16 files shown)


kanishk6103:
Here is a video demo of the included changes:

https://github.com/user-attachments/assets/75157813-241f-45a8-898e-5adaf340faa9

@KeatonTech @timon-schelling

TrueDoctor:
> Here is a video demo of the included changes:
> 2025-10-16.21-10-43_muted.mp4
> 
I think you meant to ping @Keavon

TrueDoctor:
Also we should probably zoom in on the imported svg once it is loaded right? similar to how that is done when creating a new document

Keavon:
I can't review it yet but as a quick note, please use Title Case in menus.

timon-schelling:
Note, Drag events are not forwarded to CEF on desktop (and I would prefer keeping it that way), this will basically not work on desktop.
And other images files should work the same way.

A better flow would be:
- Frontend (desktop or web) receives a file drop and sends a OpenImportDialogMessage
- Editor responds with the corresponding dialog layout.
- User selects new document or new layer from the dialog (maybe cancel could be an option).
- Editor sends a import file message with a flag set for new document or new layer (or enum)
- Frontend reads file and responds with a import file message passing along the new document flag.



This should also include unifying import via menu and via file drop.

kanishk6103:
> Note, Drag events are not forwarded to CEF on desktop (and I would prefer keeping it that way), this will basically not work on desktop. And other images files should work the same way.
> 
> A better flow would be:
> 
> * Frontend (desktop or web) receives a file drop and sends a OpenImportDialogMessage
> * Editor responds with the corresponding dialog layout.
> * User selects new document or new layer from the dialog (maybe cancel could be an option).
> * Editor sends a import file message with a flag set for new document or new layer (or enum)
> * Frontend reads file and responds with a import file message passing along the new document flag.
> 
> This should also include unifying import via menu and via file drop.

oh, I understand, valid points. I'll try implementing this flow and make the required changes.

adamgerhant:
Any update on this @kanishk6103? Ill be marking this as a draft in the mean time

kanishk6103:
> Any update on this @kanishk6103? Ill be marking this as a draft in the mean time

Hey, I wasn't able to work on that yet, but I've got some spare time now so I'll work on it.

PR #3282, opened 2025-10-13
Ensure parent layer has a cached transform to fix #2688

### LINKED ISSUE #2688: Support correctly shifting a dragged layer into an empty layer that has a Transform node
Before #2689, which removed the dummy blank `Vector` table row, we used to be able to drag an object layer into a layer with a Transform node, and it would previously counteract the Transform node's offset to prevent a shift. Now, in this particular case, it doesn't offset it.

https://github.com/user-attachments/assets/2d6d9f94-440a-44a2-988a-83ed23fa18b8

### LINKED ISSUE #3180: Text tool inserts text with an offset when in a transformed parent layer
https://github.com/user-attachments/assets/9ef1bf98-e8f8-4784-a5ed-6388fc7820bf

c-mateo:
It's now possible to rearrange layers while keeping their position on the canvas.
This is implemented by allowing the system to pull the transform directly from the node graph when the cache holds a stale value.

I'd appreciate feedback on whether this logic should remain here or where it should be moved.

Closes #2688 and might close #3180

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -727,13 +727,36 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 						insert_index: insert_index + layer_index - insert_offset,
 					});
 
+					// If the layer is changing parent, recalc transforms so the child's viewport position is preserved.
 					if layer_to_move.parent(self.metadata()) != Some(parent) {
-						// TODO: Fix this so it works when dragging a layer into a group parent which has a Transform node, which used to work before #2689 caused this regression by removing the empty vector table row.
-						// TODO: See #2688 for this issue.
+						// --- START: ensure parent's local_transform is present in DocumentMetadata cache ---
+						let parent_node_id = parent.to_node();
+						let doc_meta = self.network_interface.document_metadata();
+						let parent_has_local = doc_meta.local_transforms.contains_key(&parent_node_id);
+						let _ = doc_meta; // release immutable borrow
+
+						if !parent_has_local {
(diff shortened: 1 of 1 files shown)


timon-schelling:
!build

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -727,13 +727,36 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
...
+								crate::messages::portfolio::document::graph_operation::utility_types::ModifyInputsContext::locate_node_in_layer_chain("Transform", parent, &self.network_interface)
+							{
+								// Read current transform from the transform node and insert into local_transforms cache
+								if let Some(document_node) = self.network_interface.document_node(&transform_node_id, &[]) {
+									let parent_transform = crate::messages::portfolio::document::graph_operation::transform_utils::get_current_transform(&document_node.inputs);
+									// Build updated maps and call update_transforms to atomically replace caches
+									let upstream_footprints = self.network_interface.document_metadata().upstream_footprints.clone();
+									let mut new_local_transforms = self.network_interface.document_metadata().local_transforms.clone();
+									new_local_transforms.insert(parent_node_id, parent_transform);
+									self.network_interface.update_transforms(upstream_footprints, new_local_transforms);
```
adamgerhant, line 750: The transforms shouldn't be modified by the editor, they are simply returned after every frame and represent the current state of the artwork. To fix this issue, the current downstream transform before moving the layer should be multiplied by the downstream transform of where the layer is being moved to, and this resulting transform should be applied after the layer move. However, this may require improving the API and system for extracting transforms from the graph, which is something that should be improved soon.

PR #3287, opened 2025-10-14
Cache calculation of bounding boxes with transforms

TrueDoctor:
Improves performance for panning etc.

### DIFF
--- a/node-graph/gcore/src/vector/click_target.rs
+++ b/node-graph/gcore/src/vector/click_target.rs
@@ -292,3 +292,165 @@ impl ClickTarget {
 		}
 	}
 }
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use crate::subpath::Subpath;
+	use glam::DVec2;
+	use std::f64::consts::PI;
+
+	#[test]
+	fn test_bounding_box_cache_fingerprint_generation() {
+		// Test that fingerprints have MSB set and use only 7 bits for data
+		let rotation1 = 0.0;
+		let rotation2 = PI / 3.0;
+		let rotation3 = PI / 2.0;
+
+		let fp1 = BoundingBoxCache::rotation_fingerprint(rotation1);
+		let fp2 = BoundingBoxCache::rotation_fingerprint(rotation2);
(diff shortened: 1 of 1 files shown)


node-graph/gcore/src/vector/click_target.rs
```diff
@@ -76,7 +85,28 @@ impl ClickTarget {
...
+				if write.len() >= 10 {
+					write.clear();
+				}
+				let bounds = subpath.bounding_box_with_transform(transform);
+
+				write.push((transform, bounds));
+				bounds
+			}
+			}
+			,
```
Keavon: Formatting looks awry.

TrueDoctor: yeah, just a draft, still need to do cleanup

node-graph/gcore/src/vector/click_target.rs
```diff
@@ -76,7 +85,28 @@ impl ClickTarget {
...
+			ClickTargetType::Subpath(ref subpath) =>
+			{
+			let read = self.bounding_box_cache.read().unwrap();
+			if let Some((_, bounds)) = read.iter().find(|(t, _)| t.to_cols_array().iter().zip(transform.to_cols_array().iter()).all(|(a, b)|a.to_bits() == b.to_bits())){
+				*bounds
+			} else {
+				std::mem::drop(read);
+				let mut write = self.bounding_box_cache.write().unwrap();
+				if write.len() >= 10 {
+					write.clear();
```
timon-schelling: Do we really want to clear the cache completely here?

Did you do performance tests on the number of max cache entries, could be interesting.

TrueDoctor: I'm using a flush when full replacement strategy which gives the same theoretical guarantees as something like an lru but is significantly easier to implement. If we wanted to improve performance,  we could try to remove any scaling factors since they don't affect the outcome and to use fingerprinting to improve the search speed

TrueDoctor: Alright, implemented the optimizations

PR #3289, opened 2025-10-15
Disallow invalid layer reorder placements as siblings of artboards

### LINKED ISSUE #3256: Lazy drag layer(s) to the bottom of the stack if there is only 1 Artboard
With only one Artboard dragging and dropping a layer on the empty space below the stack could move them to the bottom.
Right now this is only possible with Artboards.

https://github.com/user-attachments/assets/76efa70d-a3c4-46ac-93ec-0dc51bbdd53c

Sahil-Gupta584:
Closes #3256 

Demo:

https://github.com/user-attachments/assets/d3f77519-a683-4acd-8541-596ce88ca268


Keavon:
Upon watching your video demo, I realized that the ideal behavior will be rejecting the placement opportunity in places that are siblings of artboards (generalized to be describable as "invalid placements"). This would need to be given to the frontend by the backend, or maybe better, answered by the backend in response to a query by the frontend. This way, we never show the white insertion line of a width that would indicate an invalid placement, which is a flaw of this "just correct it if it's invalid" approach that I realized now.

Sahil-Gupta584:
> Upon watching your video demo, I realized that the ideal behavior will be rejecting the placement opportunity in places that are siblings of artboards (generalized to be describable as "invalid placements"). This would need to be given to the frontend by the backend, or maybe better, answered by the backend in response to a query by the frontend. This way, we never show the white insertion line of a width that would indicate an invalid placement, which is a flaw of this "just correct it if it's invalid" approach that I realized now.

makes sense.
And then what about the feat pr is raised for, it will be valid or invalid drop behaviour ?

Keavon:
I'd suggest iterating on this PR with the suggested change of approach. No need to open a separate PR.

Sahil-Gupta584:
> I'd suggest irritating on this PR with the suggested change of approach. No need to open a separate PR.

Can you tell me more clearly what i need to implement?  got little confused.

and what about the feat for which pr is raised ?

Keavon:
This PR. I'm suggesting you keep working on changes to it to implement my described updated approach (from my first comment in this discussion).

Sahil-Gupta584:
> This PR. I'm suggesting you keep working on changes to it to implement my described updated approach (from my first comment in this discussion).

Yes but what about dropping on bottom, should I consider it as valid or invalid now?

Keavon:
I'm not understanding the question. All invalid drop locations shouldn't be valid white horizontal line placements, thus pre-filtering instead of post-rejecting.

Sahil-Gupta584:
> I'm not understanding the question. All invalid drop locations shouldn't be valid white horizontal line placements, thus pre-filtering instead of post-rejecting.

Ok so you mean dropping a layer as artboard sibling should be invalid? and currently if we try to do that we do see white line indicating its valid drop even though its invalid

Keavon:
Correct.

Sahil-Gupta584:
Demo:

https://github.com/user-attachments/assets/33304b85-6ea4-495a-8d5b-87b1d777a194

Sahil-Gupta584:
Hey @Keavon you can have a look

Keavon:
Your video seems to show the line jumping around and appearing in places rather erratically. Please pay attention to small details like that and ensure it's behaving how it should. Happy to take another look at a video with your next shot (and thanks for making it easy with the video recording for me). Thanks.

Sahil-Gupta584:
Hey @Keavon, Hows its now?

https://github.com/user-attachments/assets/b12a9e3a-e83e-49cf-9372-16e0948019ef

Keavon:
That looks good! However at the bottom, it needs to check if a valid placement can exist and choose that one (the outermost if there are multiple).

Sahil-Gupta584:
> That looks good! However at the bottom, it needs to check if a valid placement can exist and choose that one (the outermost if there are multiple).

actually i dont find any reason to disallow layers/artbords placement at bottom most.
do you think there will be any case in which we have to disallow placing layer/artb at bottom most? if not then then let me remove this implementation so users can place any layers at bottom most

Keavon:
No, only artboards are valid at the bottom of the artboard layer stack.

Sahil-Gupta584:
@Keavon Is this one perfect?

https://github.com/user-attachments/assets/58a02ec7-118c-4685-8956-466ed38635cc

Sahil-Gupta584:
@Keavon @timon-schelling, just pinning.

frontend/src/components/panels/Layers.svelte
```diff
@@ -340,17 +339,47 @@
...
+					const nextLayer = layers[indexAttribute + 1];
+					if (prevLayer?.entry?.depth === 1) {
+						const prevRectTop = treeChildren?.[indexAttribute].getBoundingClientRect().top;
+						if (prevLayer?.entry?.depth === 1 && prevRectTop + 10 > clientY) {
+							isInvalidDrag = true;
+							break;
+						}
+					}
+
+					const isDraggingBtnArtBoards = nextLayer?.entry?.depth === 1 && prevLayer?.entry?.depth === 1;
```
Ayush2k02: nit : typo , should be isDraggingBetweenArtboards

frontend/src/components/panels/Layers.svelte
```diff
@@ -340,17 +339,47 @@
 
 		let markerHeight = 0;
 		const layerPanel = document.querySelector("[data-layer-panel]"); // Selects the element with the data-layer-panel attribute
+		let isInvalidDrag = false;
+
 		if (layerPanel !== null && treeChildren !== undefined && treeOffset !== undefined) {
+			const lastLayerDepth = layers[layers.length - 1]?.entry?.depth;
+			const draggingLayerDepth = layers[dataIndex]?.entry?.depth;
+
+			if (clientY > treeChildren[layers.length - 1].getBoundingClientRect().bottom - 10) {
```
Ayush2k02: why this exact 10 px threshold value is used ?

Ayush2k02: also, move it to the top in a constant

Ayush2k02 (changes requested):

frontend/src/components/panels/Layers.svelte
Sahil-Gupta584: Its for better ui feel when user drags layer on edges. other wise use sometimes still sees white line for invalid drops.

Keavon:
This generally isn't working in a lot of ways:

https://github.com/user-attachments/assets/554b1acd-d479-445f-b03b-e9bc8f5c12d9

Sahil-Gupta584:
weird let me check again

Sahil-Gupta584:
@Keavon Can you test again, tried directly using insertDepth

Keavon:
If you pass CI and mark it as ready for review, I'll take a look then.

PR #3292, opened 2025-10-16
Reduce Nix flake maintanace by eliminating need for `cargoHash` bumps

bluelinden:
This PR changes all repos in `Cargo.toml` that use an out-of-tree commit to ones where said commit is in-tree. Additionally, it removes the hash requirement for Nix builds, so the Graphite flake can now use up-to-date versions of Graphite without the need for any CI to update hashes.

Progress is being made to do the same for NPM fetching.

### DIFF
--- a/.nix/flake.nix
+++ b/.nix/flake.nix
@@ -108,7 +108,7 @@
           };
 
           
-          # TODO: Remove the need for this hash by getting individual package hashes from package-lock.json. @timon-schelling and @bluelinden tried this on 2025-10-17, but we were unsuccessful due to the bad documentation around pkgs.importNpmLock. We hope to try this again sometime.
+          # TODO: Remove the need for this hash by using individual package resolutions and hashes from package-lock.json
           npmDeps = pkgs.fetchNpmDeps {
             inherit (finalAttrs) pname version;
             src = "${finalAttrs.src}/frontend";


Keavon:
Should picomatch be a dev dependency? Please also explain its purpose and justification for usage so I can double check its necessity and authorize the new dependency.

bluelinden:
> Should picomatch be a dev dependency? Please also explain its purpose and justification for usage so I can double check its necessity and authorize the new dependency.

I have removed picomatch from the dependencies and undone all NPM-related changes I have made. @timon-schelling and I attempted to make the new install strategy work and we just couldn't. 

The PR is now in a reviewable state.

timon-schelling:
!build

timon-schelling:
!build

.nix/flake.nix
```diff
@@ -86,29 +84,31 @@
...
-          cargoDeps = pkgs.rustPlatform.fetchCargoVendor {
-            src = finalAttrs.src;
-            hash = "sha256-BVIQIZbGW19Rof0J7U2r6XFCUC52hb7+uaE1di4bV4A=";
+          cargoLock = {
+            lockFile = ../Cargo.lock;
+            allowBuiltinFetchGit = true;
           };
 
+          
+          # TODO: Remove the need for this hash by getting individual package hashes from package-lock.json. @timon-schelling and @bluelinden tried this on 2025-10-17, but we were unsuccessful due to the bad documentation around pkgs.importNpmLock. We hope to try this again sometime.
```
timon-schelling: I would prefer just: `# TODO: Remove the need for this hash by relying on package-lock.json`

bluelinden: fixed

timon-schelling:
!build

timon-schelling:
I just noticed that shader (GPU) nodes are not yet enabled.

timon-schelling:
Merging because this was also the case before this pr. I will create a separate pr.

PR #3297, opened 2025-10-20
Desktop: Bundle for Mac and Windows

timon-schelling:
depends on #3295

run to test bundle
`cargo run -p graphite-desktop-bundle -- open`

Linux bundling is not yet implemented.

`npm run start-desktop` works for all platforms.

### DIFF
--- a/.github/workflows/comment-profiling-changes.yaml
+++ b/.github/workflows/comment-profiling-changes.yaml
@@ -124,6 +124,65 @@ jobs:
               });
             }
 
+      - name: Analyze profiling changes
+        id: analyze
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+
+            function isSignificantChange(diffPct, absoluteChange, benchmarkType) {
+              const meetsPercentageThreshold = Math.abs(diffPct) > 5;
+              const meetsAbsoluteThreshold = absoluteChange > 200000;
+              const isCachedExecution = benchmarkType === 'run_cached' ||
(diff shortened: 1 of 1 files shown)


inkreasing:
Hi, great to see it compiling and running!
The window decoration buttons look a bit weird.
This is with the window not selected (This normally makes these buttons all grey on macos when the window isn't selected)
Bildschirmfoto 2025-10-20 um 23 38 35
This is with the window selected.
Bildschirmfoto 2025-10-20 um 23 38 49
So it seems that the offset of the buttons are a bit off and macos is still rendering buttons in the background.

timon-schelling:
> So it seems that the offset of the buttons are a bit off and macos is still rendering buttons in the background.

You are right, currently we have both the native buttons and the ones we draw as part of the ui displayed. The Native buttons will stay,and our custom ones will go, that will be done in a future PR.

The bigger issue is actually scaling right now.

On my device this looks like this (double buttons are also more apparent) 

image

TrueDoctor:
Does `npm run start-desktop` still work on Windows after this pr?
I still don't like that we have different crates for the final binary dependent on the platform

timon-schelling:
> Does `npm run start-desktop` still work on Windows after this pr? I still don't like that we have different crates for the final binary dependent on the platform

Yes, the main binary will, and in my opinion should, work on all platforms (on mac you would just be missing the dependencies).
I still prefer it that way because we don't have to mix platform specific build scripts (linking) and those entry points will just call start (nothing else they can do), Would you like me to leave a big comment about that at the top of the files?

PR #3304, opened 2025-10-23
New node: String Split

daniil-loban:
Features for text: 

**split**  node that allows splits a  String Value and returns a Vec of strings:
- **in** String Value: "a;b;c"  (from a csv file or a http request)   
- **param** delimeter: ";"
- **out**: vec!["a","b","c" ]

### DIFF
--- a/node-graph/gcore/src/logic.rs
+++ b/node-graph/gcore/src/logic.rs
@@ -69,11 +69,15 @@ fn string_length(_: impl Ctx, string: String) -> f64 {
 	string.chars().count() as f64
 }
 
-// Spliet a string by a specified delimeter ("1;2;3" e.t.c.) and return a vector of strings
 #[node_macro::node(category("Text"))]
-fn split(_: impl Ctx, string: String, #[default("\\n")] delimeter: String) -> Vec<String> {
-	let delimeter = delimeter.replace("\\n", "\n");
-	string.split(&delimeter).map(|x| x.to_string()).collect()
+fn string_split(_: impl Ctx, string: String, #[default("\\n")] delimeter: String, #[default(true)] delimeter_escaping: bool) -> Vec<String> {
+	let delimeter = if delimeter_escaping {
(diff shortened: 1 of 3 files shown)


TrueDoctor:
Sorry that I didn't see this pr earlier, feel free to just request a review from me in the future. I made a few changes to your pr let me know if you disagree with any of these

daniil-loban:
I agree with changes!

node-graph/gcore/src/logic.rs
```diff
@@ -69,6 +69,20 @@ fn string_length(_: impl Ctx, string: String) -> f64 {
 	string.chars().count() as f64
 }
 
+// Get an indexed part of string whitch separated a specified delimeter ("1;2;3" e.t.c.)
+#[node_macro::node(category("Text"))]
+fn substring_by_index(_: impl Ctx, string: String, #[default("\\n")] delimeter: String, index: u32) -> String {
```
TrueDoctor: I'm wondering if this node name should include split since we split by the substring and don't index by it. @Keavon opinions?

Keavon:
I suggest we allow `Vec<String>` to be a supported type so we can use the existing count and index nodes.

node-graph/gcore/src/logic.rs
daniil-loban: We split the string and take the value by index, perhaps for optimization it is better if it is 2 nodes.

daniil-loban:
> I suggest we allow `Vec<String>` to be a supported type so we can use the existing count and index nodes.

The main goal was to process strings as arrays or other structures (in the case of JSON, but the code for it is not posted, there the path to the data is used to extract the string or number), if it is possible to use node polymorphism, for example, counting elements or calculating the size of a string vector, with existing nodes (Count Elements e.t.c.), this will be cool and intuitive.

Keavon:
Yes, their monomorphized forms just have to be added to the `#[implementations(...)]` of each node that'll work with a vec of strings.

TrueDoctor:
image

TrueDoctor:
!build

PR #3306, opened 2025-10-23
Fix incorrect implementations of the Hash and PartialEq traits on Table<T>

TrueDoctor:
Previously we did not include information like the transforms for determining equality of table instances. This caused caching of values should have been different


Keavon:
@TrueDoctor shouldn't this also include `source_node_id`?

PR #3305, opened 2025-10-23
Respect device scale when calculating viewport bounds

Sambhram1:
Scale all pointer coordinates and viewport bounds by before sending them to the backend. This ensures consistent coordinate spaces across the entire rendering pipeline.

PR #3310, opened 2025-10-26
Desktop: Add support for UI scaling

### LINKED ISSUE #3055: Use os scaling factor on desktop
Respect the proper ui [scale factor](https://docs.rs/winit/latest/x86_64-unknown-linux-gnu/winit/window/struct.Window.html#method.scale_factor)

### LINKED ISSUE #3192: Integrate custom window frames for Mac

timon-schelling:
Depends on #3331

Closes #3055, closes #3192

Implements scaling.

~Fronted still reports the wrong viewport bounds like before #3305 fixes part of that, the viewport size reporting, but the overlays are scaled incorrectly (offset or not scaled).~

~Branch `desktop-scaling-demo` includes cleaned changes from #3305 and a workaround for overlays (position is correct but to small on scale > 1 devices). Testing that branch would be useful, still work in progress. May be useful to merge the result of that before this PR.~

### DIFF
--- a/desktop/src/cef/input.rs
+++ b/desktop/src/cef/input.rs
@@ -1,29 +1,25 @@
 use cef::sys::{cef_event_flags_t, cef_key_event_type_t, cef_mouse_button_type_t};
 use cef::{Browser, ImplBrowser, ImplBrowserHost, KeyEvent, KeyEventType, MouseEvent};
 use std::time::Instant;
-use winit::dpi::LogicalPosition;
+use winit::dpi::PhysicalPosition;
 use winit::event::{ButtonSource, ElementState, MouseButton, MouseScrollDelta, WindowEvent};
 
 mod keymap;
 use keymap::{ToNativeKeycode, ToVKBits};
 
-use crate::cef::CefEventHandler;
-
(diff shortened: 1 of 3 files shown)


desktop/src/cef/input.rs
```diff
@@ -1,25 +1,25 @@
...
 use keymap::{ToNativeKeycode, ToVKBits};
 
 use super::consts::{MULTICLICK_ALLOWED_TRAVEL, MULTICLICK_TIMEOUT, PINCH_ZOOM_SPEED, SCROLL_LINE_HEIGHT, SCROLL_LINE_WIDTH, SCROLL_SPEED_X, SCROLL_SPEED_Y};
 
-pub(crate) fn handle_window_event(browser: &Browser, input_state: &mut InputState, event: &WindowEvent) {
+pub(crate) fn handle_window_event(browser: &Browser, input_state: &mut InputState, event: &WindowEvent, scale: f64) {
 	match event {
 		WindowEvent::PointerMoved { position, .. } | WindowEvent::PointerEntered { position, .. } => {
-			input_state.cursor_move(position);
+			input_state.cursor_move(&position.to_logical(2.0));
```
nupsi: Why is this `to_logical(2.0)` should it also be `to_logical(scale)` like to other?

timon-schelling: Yes, thanks for catching that, left form debugging.

timon-schelling:
Extracting the viewport into its own handler will be separate pr (this pr will depend on that)

PR #3311, opened 2025-10-26
Make the Pen tool close a path by double-clicking

### LINKED ISSUE #3219: Close Pen Tool path with double click
This option is handled differently in many applications. In most open source ones double clicking while making a path with Pen Tool equivalents either does not do anything (Gimp), or ends the path drawing without closing the path (Krita, Inkscape). In "industry standard" applications however it makes a new anchor under the cursor placement and closes the path. I think the latter is a good feature to have - especially for quick editing as it does not force user to precisely point at the path start to close it.

krVatsal:
Closes the active pen path when the user intentionally double-clicks.

Closes #3219


timon-schelling:
!build

krVatsal:
Do i need to do something in order to get this pr reviewed?

timon-schelling:
No, everything fine we're just a bit behind on PRs, situation will improve when Keavon is back from vacation. Thanks for contributing :)

krVatsal:
Sure, thanks!

timon-schelling:
!build

timon-schelling:
Fails to compile now.

krVatsal:
> Fails to compile now.

The conflicts are resolved , can you review this one also.

timon-schelling:
!build

timon-schelling:
!build

timon-schelling:
https://github.com/user-attachments/assets/edaa6945-dc4d-4f2e-8978-76986b84baaf

Doesn't seem to work as expected.

krVatsal:
Can you clarify what is the desired output, drawing should end without closing the path or it should close the the shape on double click?

timon-schelling:
Closing the path, see the issue you linked.

krVatsal:
> Closing the path, see the issue you linked.

Can you review it now?

timon-schelling:
!build

timon-schelling:
LGTM, @0HyperCube any comments?

0HyperCube (approved):
Seems good.

I guess you could have bound an event for `DoubleClick` in the `input_mappings.rs` but that would confuse the logic since it happens only on mouse up.

Keavon:
Yeah, from a behavioral perspective, this needs to not close the path upon releasing the mouse from a fast click-click-drag (double-clicking the mouse down part, but dragging during the second mouse down). Currently, that scenario results in unexpected behavior after mouse up.

timon-schelling:
!build

Keavon:
This doesn't work as expected in the following case:

https://github.com/user-attachments/assets/0c6a9919-a8f2-4fe0-a0db-9da9d92bbe61

If there is clearly a single endpoint on the path being drawn, it should use that, not the starting point for the current drawing session. But if it attempts to find the endpoint but there is no single unambiguous one, because a vector mesh is involved in the currently-being-drawn subpath, then it should use the current behavior, for example this is correct:

https://github.com/user-attachments/assets/552611fb-204d-4e02-a580-1ba91656abc2

krVatsal:
Got it, will improve this.

Keavon:
I'd like it so if you accidentally double-click, you can hit Ctrl+Z and return to the drawing state you were in immediately before you double-clicked. This already works for regular segment drawing, where you can hit Ctrl+Z to undo each segment and continue where you were immediately before drawing each segment, still interactively placing the next segment. That's how we'd like to work after double-clicking and hitting Ctrl+Z.

krVatsal:
Should it behave like: when we hit Ctrl+Z(after closing path) then we can continue drawing without clicking again on the last node? If it is so, then if we keep hitting Ctrl+Z again and again it should be able to continue drawing without clicking again on the last node, which is not the current situation with pen tool after a certain number of undos.

Is this a desired condition to reduce overhead or a bug?

https://github.com/user-attachments/assets/95e4e2e6-f346-4df6-90f3-872b61fe8fd0

Keavon:
You've encountered the same bug I noticed a couple of times, but was not able to reproduce reliably. If you can help debug that and solve it also, that would be really helpful. It doesn't have to be part of this PR.

krVatsal:
> You've encountered the same bug I noticed a couple of times, but was not able to reproduce reliably. If you can help debug that and solve it also, that would be really helpful. It doesn't have to be part of this PR.

Sure i will open a new issue for this bug.

krVatsal:
Any further improvements @Keavon?

PR #3312, opened 2025-10-26
Fix snapping when drawing an artboard

### LINKED ISSUE #3263: While dragging a new Artboard only the starting point is snapped
When making new Artboard the starting point can be snapped to other objects in the document. But its no the case for the side that is actively dragged. In the screenshot below I was dragging the Artboard from top-left corner to the bottom and wanted for it to have the same height as the Artboard on the left:

Image

0HyperCube:
Closes #3263

### DIFF
--- a/editor/src/messages/tool/tool_messages/artboard_tool.rs
+++ b/editor/src/messages/tool/tool_messages/artboard_tool.rs
@@ -739,9 +739,9 @@ mod test_artboard {
 		editor.new_document().await;
 		editor.set_viewport_size(DVec2::splat(-1000.), DVec2::splat(1000.)).await; // Necessary for doing snapping since snaps outside of the viewport are discarded
 		editor.drag_tool(ToolType::Artboard, 10., 10., 20., 22., ModifierKeys::empty()).await; // Artboard to drag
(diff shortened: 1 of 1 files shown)


editor/src/messages/tool/tool_messages/artboard_tool.rs
```diff
@@ -611,29 +611,61 @@ mod test_artboard {
...
+	impl ArtboardLayoutDocument {
+		pub fn new(position: impl Into<IVec2>, dimensions: impl Into<IVec2>) -> Self {
+			Self {
+				position: position.into(),
+				dimensions: dimensions.into(),
+			}
+		}
+	}
+
+	/// Check if all of the artboards exist in any ordering
```
TrueDoctor, line 628: This could probably be simplified a bit by sorting the two vector but since this is a test this does not really matter

0HyperCube: I think that implementing `Ord` for the artboards would be a bit confusing as it doesn't really have any semantic meaning.

TrueDoctor: I mean, we could use `sort_by_key` to avoid having to implement `Ord` but we can just leave it as it is

PR #3315, opened 2025-10-27
Add tool for visualizing crate hierarchy

TrueDoctor:
Visualizes the dependency chain of our own crates

image


tools/crate-hierarchy-viz/generate-crate-viz.sh
timon-schelling, line 1: I would like this to be part of the rust code.
Also add dot as a tool in nix shell.

PR #3321, opened 2025-10-30
Fix the Spline node to maintain stable segment IDs

### LINKED ISSUE #3318: Path handles glitch out when moved
when I try to move a handle on a path, it disappears until I deselect and reselect the path
I think this started happening when I combined a "flatten nodes" node with a "repeat" node

https://github.com/user-attachments/assets/37930813-69f7-4db9-b454-09f51a69e252

this also happens on any other handle on that red path

graphite file: (github won't let me upload .graphite files so i uploaded a zip)

[\[7\] electric coil.zip](https://github.com/user-attachments/files/23210053/7.electric.coil.zip)

0HyperCube:
Closes #3318

### DIFF
--- a/node-graph/gcore/src/vector/vector_nodes.rs
+++ b/node-graph/gcore/src/vector/vector_nodes.rs
@@ -821,16 +821,17 @@ async fn vec2_to_point(_: impl Ctx, vec2: DVec2) -> Table<Vector> {
 async fn points_to_polyline(_: impl Ctx, mut points: Table<Vector>, #[default(true)] closed: bool) -> Table<Vector> {
 	for row in points.iter_mut() {
 		let mut segment_domain = SegmentDomain::new();
+		let mut next_id = SegmentId::ZERO;
 
 		let points_count = row.element.point_domain.ids().len();
 
 		if points_count > 2 {
 			(0..points_count - 1).for_each(|i| {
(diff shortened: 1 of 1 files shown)


TrueDoctor:
I lack the mental model to judge if this is a good approach. Could this lead to cases where we merge lists of segments with the same ids? And would that create issues? If the change is conceptually sound, the code looks good, I just don't know how we would expect this to behave

0HyperCube:
@TrueDoctor

The behaviour of producing small incrementing segment IDs is used by other nodes that generate geometry such as the circle and rectangle nodes.

In the `flatten_path` node, if there is a duplicate segment ID, it will deterministically modify one of the IDs through hashing `(current_index, vector_index, node_id)` and the previous segment ID:
https://github.com/GraphiteEditor/Graphite/blob/1b9a492a742990c1c3b6d5b9a7a70cedd1c325be/node-graph/gcore/src/vector/vector_nodes.rs#L1018-L1022

Then generating the new IDs in the `concat` function:
https://github.com/GraphiteEditor/Graphite/blob/1b9a492a742990c1c3b6d5b9a7a70cedd1c325be/node-graph/gcore/src/vector/vector_types.rs#L415-L421


This means that segment IDs must be:
- Unique within the current `SegmentDomain`.
- Deterministic between several runs of the same graph.

PR #3331, opened 2025-11-02
Split out viewport handling into its own message handler

timon-schelling:
Moves viewport related logic into separate message handler. IPP felt like the wrong place for that to begin with.

Intentionally breaks desktop fixed in #3310. may consider fixing right before merge, but working on both branches would have annoying merge conflicts, otherwise this should wait until #3310 is also ready for merge.


editor/src/messages/portfolio/document/overlays/utility_types_vello.rs
```diff
@@ -605,7 +618,7 @@ impl OverlayContextInternal {
 	}
 
 	fn pixel(&mut self, position: DVec2, color: Option<&str>) {
-		let size = 1.;
```
TrueDoctor, line 608: This should be reverted, as per convention we don't use the trailing zero

editor/src/messages/viewport/viewport_message.rs
```diff
@@ -0,0 +1,8 @@
+use crate::messages::prelude::*;
+
+#[impl_message(Message, Viewport)]
+#[derive(PartialEq, Clone, Debug, serde::Serialize, serde::Deserialize)]
+pub enum ViewportMessage {
+	Update { x: f64, y: f64, width: f64, height: f64, scale: f64 },
+	Trigger,
```
TrueDoctor: This should have a comment what Trigger means

editor/src/messages/viewport/viewport_message_handler.rs
```diff
@@ -0,0 +1,391 @@
...
+		logical.to_physical(self.scale)
+	}
+
+	pub fn convert_physical_to_logical_point<T: Into<PhysicalPoint>>(&self, physical: T) -> LogicalPoint {
+		physical.into().to_logical(self.scale)
+	}
+
+	pub fn convert_logical_to_physical_point<T: Into<LogicalPoint>>(&self, logical: T) -> PhysicalPoint {
+		logical.into().to_physical(self.scale)
+	}
```
TrueDoctor: I'm wondering if it would make for a nicer API if we only returned logical points, but included the scale so you could call `.to_physical()` on it or use an into conversion. It does feel award that you basically always need the viewport to perform operations on points. The advantage is of course that this helps prevent using and outdated scale, but I'm not sure how often that would be applicable

timon-schelling: You mean have a scale component in logical point and physical point etc.? I guess we can do that, if we consider possible outdated scale to be never an issue.

TrueDoctor: But yeah, was mainly an idea to reduce the api surface a bit, not sure if it is a good or a bad one. But definetly one which would create extra work so it might be best if we skip this

timon-schelling: Not that much work. I have preferred that approach from the beginning, but I thought the possible missing scale update would be a issue.

timon-schelling: Spend the time to this feels cleaner

PR #3333, opened 2025-11-03
Add lasso selection to node graph

### LINKED ISSUE #2532: Add lasso selection mode to the node graph as well
<kbd>Ctrl</kbd> dragging in the canvas creates a lasso selection instead of a box selection. For consistency, the same should occur in the node graph.

This is soft-blocked on #1922 because the current box selection is implemented in Svelte, not using the overlay system which the canvas box/lasso selection both use. This could technically be implemented using an SVG path in Svelte but that's potentially wasted effort once the above issue is completed and the node graph no longer uses Svelte.

wade-cheng:
Closes #2532.


wade-cheng:
I should say while this is on my mind right now, performance should probably be looked at for this PR. We convert the entire vec of polygon points from one unit to another once every frame.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -66,6 +67,9 @@ pub struct NodeGraphMessageHandler {
 	/// If dragging the background to create a box selection, this stores its starting point in node graph coordinates,
 	/// plus a flag indicating if it has been dragged since the mousedown began.
 	box_selection_start: Option<(DVec2, bool)>,
+	/// If dragging the background to create a lasso selection, this stores its current lasso polygon in node graph coordinates,
+	/// plus a flag indicating if it has been dragged since the mousedown began.
+	lasso_selection_curr: Option<(Vec<DVec2>, bool)>,
```
0HyperCube: nit: It is probably clearer to make a `struct { lasoo_points: Vec<DVec2>, has_dragged: bool }` rather than relying on the doc comment to name the fields. However I suppose the current implementation is consistent with the box selection.

I wonder if it is even necessary to store the `has_dragged` bool. Can't you just check if the number of points is ≥ 2?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+							.inverse()
+							.transform_point2(ipp.mouse.position);
+
+						lasso_selection_curr.push(node_graph_point);
+					}
+
+					let lasso_selection_viewport: Vec<DVec2> = lasso_selection_curr
+						.iter()
+						.map(|selection_point| network_metadata.persistent_metadata.navigation_metadata.node_graph_to_viewport.transform_point2(*selection_point))
+						.collect();
```
0HyperCube: There is no need to call `.collect()` since you use `into_iter()` later. You might have to move the frontend message to here though if the borrow checker is unhappy.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+					// WARNING WARNING WARNING: this commented-out code is copy pasted from UpdateBoxSelection above and has not been edited for lasso
+					// The mouse button was released but we missed the pointer up event
+					// if ((e.buttons & 1) === 0) {
+					// 	completeBoxSelection();
+					// 	boxSelection = undefined;
+					// } else if ((e.buttons & 2) !== 0) {
+					// 	editor.handle.selectNodes(new BigUint64Array(previousSelection));
+					// 	boxSelection = undefined;
+					// }
+
```
0HyperCube: What is this relating to? It looks like javascript.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+					let Some(network_metadata) = network_interface.network_metadata(selection_network_path) else {
+						log::error!("Could not get network metadata in UpdateBoxSelection");
+						return;
+					};
+
+					{
+						let node_graph_point = network_metadata
+							.persistent_metadata
+							.navigation_metadata
+							.node_graph_to_viewport
```
0HyperCube: nit: Probably less verbose to `let node_graph_to_viewport = network_metadata.persistent_metadata.navigation_metadata.node_graph_to_viewport;`

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+							} else {
+								nodes.insert(node_id);
+							}
+						}
+					}
+					if nodes != previous_selection {
+						responses.add(NodeGraphMessage::SelectedNodesSet {
+							nodes: nodes.into_iter().collect::<Vec<_>>(),
+						});
+					}
```
0HyperCube, line 2041: It is probably easiest to just always send this message (rather than allocating the `previous_selection` HashMap each frame).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+							points
+								.windows(2)
+								.map(|w| PathSeg::Line(Line::new(dvec2_to_point(w[0]), dvec2_to_point(w[1]))))
+								.chain(std::iter::once(PathSeg::Line(Line::new(
+									dvec2_to_point(*points.last().unwrap()),
+									dvec2_to_point(*points.first().unwrap()),
+								))))
+								.collect()
+						}
+						points_to_polygon(lasso_selection_curr)
```
0HyperCube, line 2022: Rather than allocating, consider a simplye point in polygon algorithm https://wrfranklin.org/Research/Short_Notes/pnpoly.html.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1956,6 +1989,88 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+						log::error!("Could not get selected nodes in UpdateBoxSelection");
+						return;
+					};
+					let previous_selection = selected_nodes.selected_nodes_ref().iter().cloned().collect::<HashSet<_>>();
+					let mut nodes = if shift || alt {
+						selected_nodes.selected_nodes_ref().iter().cloned().collect::<HashSet<_>>()
+					} else {
+						HashSet::new()
+					};
+					let all_nodes = network_metadata.persistent_metadata.node_metadata.keys().cloned().collect::<Vec<_>>();
```
0HyperCube, line 2010: It is quite annoying to have to call collect here, due to the very poorly designed `node_click_targets` which requires mutable access to the `NodeNetworkInterface`. However I guess there is little that can be done unless rustc adds partial borrows or someone refactors the network interface.

0HyperCube (commented):
Thanks for your work on this. It seems to work nicely.

It is good to hear you thinking about the performance. However current architectural decisions unfortunately means an excessive number of allocations are necessary.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: Oh neat, that is javascript for some reason. Yeah, the identical comment appears in UpdateBoxSelection, and got copied over when I copy-pasted the entire UpdateBoxSelection block. Didn't know if it was important so I left it.

To answer the question more explicitly, don't really know what this relates to, but your guess would be better than mine anyhow.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: Re: paragraph 2. Good point, pretty sure that's all the functionality that the dragged bool is used for in the box selection logic anyways.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: I can't intuit which is better, but I'll take your word for it. When I or whoever makes the change, I assume they should also make the identical analogous change to the box selection code?

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: The box selection seems to be working fine for many months, so it is probably safe to delete this comment (in both places).

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: I guess it doesn't matter very much either way.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: The borrow checker indeed was. And yes, moving the frontend message there fixes it.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: Hm. Perhaps the `has_dragged` field helps make it explicit that we do _something_ related to whether we have moved from the initial mousedown or not. But I agree with removing it.

Also, I took a look, and the only reason box selection can't do without the field either is because we want to `update_node_graph_hints` only _after_ moving from the initial mousedown. That is, if the initial mousedown was to update the hints, we could remove the field. But for the analogous lasso logic, we just check if the length is at least two, again.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: Hmm, I feel the disappeal of needing to allocate, but I'm not sure how point-in-polygon helps. I stole `click_targets.node_click_target.intersect_path` from... uh, probably the box selection code, and I didn't look further because I knew it worked. I'd need to think about it lots and dig deeper into the code, but first impression: aren't nodes not points? We rather need nodes (rounded rectangles) in polygon?

That is, without some digging, I currently don't know how to get the bounding boxes of all the nodes.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: implemented in 4f862018cda1f1105f7537eba0a2136c58857f02 below

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: implemented in 4f862018cda1f1105f7537eba0a2136c58857f02 below

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: implemented in 49efe270fbfee43babc412c2df174340061bad47 below

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: implemented in 0aa3860e1a5c277ac6331cdf002a63db6fd47e04 below

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
wade-cheng: Leaving it for now, then.

wade-cheng:
Made some initial tweaks, discussion above. Feel free to mark whatever resolved if you agree with changes.

I suppose the last holdup is the "point in polygon" note that I'd be stuck on.

wade-cheng:
Oh, hm. Looks like the hints only show esc/rmb to cancel, but not the extend/subtract hints that show up when using the artboard drag select tool. As you are indeed able to extend/subtract, should I add them through a commit here or leave it for a later mini PR?

Searched `is:issue state:open node drag hint` on the tracker but didn't find any matches.

Keavon:
Yes, feel free to include them in the PR here.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
0HyperCube: I guess it wouldn't matter too much.

wade-cheng:
Included :+1:

Keavon:
!build

Keavon:
On Mac, there's a discrepancy between the canvas which uses Ctrl and the node graph which uses Cmd (Accel). You should probably make *both* use *both* Ctrl and Accel. Since we'd like the node graph and the canvas to work with both Cmd and Ctrl on Mac, and Ctrl on Windows/Linux.

Keavon:
Could you please resolve the merge conflicts and address my previous comment? Thanks!

wade-cheng:
Schoolwork has ramped up again, but I'll be taking a look at this within these next seven days. Big project due this weekend. Appreciate the nudge :+1:

Keavon:
Sounds good 😸 Please just mark this as "ready for review" and reply with a comment once that's done. I'm marking it as a draft for now so I can keep it out of my active review queue.

PR #3334, opened 2025-11-03
Add major/minor grid lines to grids popover

bluelinden:
This PR adds a major/minor grid line system that allows for better usability of precise grids when zoomed farther out.

The implementation is inspired by Inkscape's.

*note: this PR contains a few segments initially written using AI and then iterated upon and fixed by me, a human. I take responsibility for any and all bugs found in the updated portions of the grid rendering code, and I am confident that the work is human-quality.


Keavon:
!build

Keavon:
!build

frontend/assets/icon-12px-solid/dot-large.svg
timon-schelling, line 1: @Keavon what this will also need to be moved to the branding repo right?
How should we handle this for now?

editor/src/messages/portfolio/document/overlays/grid_overlays.rs
```diff
@@ -238,7 +332,54 @@ pub fn overlay_options(grid: &GridSnapping) -> Vec<LayoutGroup> {
...
+			Separator::new(SeparatorType::Related).widget_holder(),
+			RadioInput::new(vec![
+				RadioEntryData::new("lines").label("Lines").icon("Grid").on_update(update_val(grid, |grid, _| {
+					grid.dot_display = false;
+				})),
+				RadioEntryData::new("dots").label("Dots").icon("GridDotted").on_update(update_val(grid, |grid, _| {
+					grid.dot_display = true;
+				})),
+			])
+			// .min_width(200)
```
timon-schelling: remove

timon-schelling:
!build

timon-schelling:
!build

Keavon:
!build

timon-schelling:
!build

timon-schelling:
Sorry for taking so long to take a look, Code looks good in general.
We need to rebase though.
(And the dot may be moved to the branding repo, not because it's copyrightable, but to keep all icons in one place. @Keavon)

I may have noticed one logic bug, lines and dots don't match the same pattern here.

image
image


And If I remember correctly I somehow got into a state where the second color selection element would not disappear with 1 1 1, but I'm not able to reproduce that, not a blocker in my opinion.

frontend/assets/icon-12px-solid/dot-large.svg
bluelinden: @Keavon this needs to be added to the branding repo

timon-schelling:
@Keavon can you please add the icon to the other repo?

Keavon:
The icon is now included as "DotThick" so you can rebase and use that now!

Keavon:
> The icon is now included as "DotThick" so you can rebase and use that now!

I will mark this as a draft until that's done. Let me know when it's ready again, thanks :)

Keavon:
!build ([build link](https://github.com/GraphiteEditor/Graphite/actions/runs/22087080804))

Keavon:
So unfortunately it is looking like some changes will be needed to how you're doing the logic for rendering the pixel-perfect dotted grids. On master, they are pixel-perfect, but these changes now result in that not being the case anymore. These have all been tested in Chrome.

Here is an example from master:

| 100% | 300% |
|-|-|
| image | image |

And with the same settings on this branch, drawn with the small dot option:

| 100% | 300% |
|-|-|
| image | image |

Basically, it is now drawing `+` shapes when before we were just filling a single pixel.

For isometric, what we had before was somewhat imperfect due to AA causing some to appear brighter than others:

| 100% | 300% |
|-|-|
| image | image |

But it was better than the new behavior which seems to be drawing a bunch of tilted lines:

| 100% | 300% |
|-|-|
| image | image |

And when thick-dot mode is set, we end up with overlapping colors in some places:

| 100% | 300% |
|-|-|
| image | image |

## What we want

For the small circle mode, we want the single illuminated pixel like we have in master for rectangular grids. For isometric grids, we want to ideally fix it to render a single illuminated pixel in the nearest pixel without that problem we're seeing with the AA. But importantly, it needs to not look like an angled dash.

For large circle mode, we should have a 3x3 circle centered about the nearest pixel to the point, with antialiasing to make it look like a circle. Same with isometric mode, staying centered about the nearest pixel-aligned position

| Small dot mode | Large dot mode |
|-|-|
| image | image |

PR #3354, opened 2025-11-09
Fix invalid popup widget ids

### LINKED ISSUE #3353: Regression: cannot add new nodes inline
This is a regression introduced in #3337.

The Image button to add a new node inline to the selected layer (found in the Properties panel, Layers panel, and node graph control bar) no longer functions when a node is clicked. For example, draw a rectangle and search this menu for "Repeat" and try clicking to add it.

> warn	editor/src/messages/layout/layout_message_handler.rs:116
> handle_widget_callback was called referencing an invalid widget ID, although the layout target was valid. `widget_id: 154848269268510619`, `layout_target: PropertiesPanel`

A related second regression that's likely caused by the same root issue:

The Overlays popover menu's checkbox options do not do anything anymore, and upon closing the menu and reopening it, they return to their originally checked states:

Image

Yet another (happy!) regression is that #2530 got fixed. Something to be aware of (but skeptical of).

0HyperCube:
Closes #3353

### DIFF
--- a/editor/src/messages/layout/utility_types/layout_widget.rs
+++ b/editor/src/messages/layout/utility_types/layout_widget.rs
@@ -523,7 +523,6 @@ impl WidgetHolder {
 					a.diff(b.clone(), &mut new_widget_path, widget_diffs);
 					new_widget_path.pop();
 				}
-				self.widget = new.widget;
 				return;
 			}
 		}


TrueDoctor:
!build

TrueDoctor (approved):

Issue #3341, opened 2025-11-04
Local-axis grab (G key) dashed line should show only the X or Y component, tilted


Keavon:
<kbd>G</kbd> lets you enable grab and then hit <kbd>X</kbd> or <kbd>Y</kbd> once for a global axis (document space) and a second time for a local axis (layer space), but the dashed line visualization is drawn incorrectly in the latter case:

https://github.com/user-attachments/assets/1e049146-e215-4abb-9f81-27f62e398836

When transforming in local space, it should tilt the dashed line with the local space as well, and draw just the X or Y component instead of both.

See [this comment](https://github.com/GraphiteEditor/Graphite/pull/3527#issuecomment-3910935758) for a little more detail.

PR #3355, opened 2025-11-09
Fix regressions with G/R/S

### LINKED ISSUE #3340: Local-axis grab (G key) doesn't work on groups
I think possibly related to #774, "local axis" transforms don't seem to respect groups of objects.

On: Commit 6d669ad7

To reproduce, create a square:

Image

Now rotate 45deg and move along local Y, 

<kbd>R</kbd><kbd>4</kbd><kbd>5</kbd><kbd>Enter</kbd><kbd>G</kbd><kbd>Y</kbd><kbd>Y</kbd><kbd>-</kbd><kbd>1</kbd><kbd>0</kbd><kbd>0</kbd>

Image

(I'm showing the original location in maroon just for illustration)

Now let's create two squares and group them with <kbd>^</kbd><kbd>G</kbd>, and then rotate **the group**:

Image

<kbd>R</kbd><kbd>4</kbd><kbd>5</kbd><kbd>Enter</kbd><kbd>G</kbd><kbd>Y</kbd><kbd>Y</kbd><kbd>-</kbd><kbd>1</kbd><kbd>0</kbd><kbd>0</kbd>

Even though the group is rotated 45deg (and the bounding box indicates this), the "local" axis appears to be the same as the global axis. (Actually in this case it appears to be the negative of the global axis, interestingly?)

(PS: I LOVE LOVE LOVE LOVE LOVE GRAPHITE!)

0HyperCube:
Closes #3340

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/transformation.rs
+++ b/editor/src/messages/portfolio/document/utility_types/transformation.rs
@@ -171,8 +171,10 @@ impl Translation {
 				Axis::Y => DVec2::Y * self.dragged_distance.dot(state.constraint_axis(self.constraint).unwrap_or_default()),
 			}
 		};
-		let displacement = displacement * document_to_viewport.matrix2.y_axis.length(); // Values are local to the viewport but scaled so values are relative to the current scale.
-		if state.is_rounded_to_intervals { displacement.round() } else { displacement }
+		let displacement_viewport = displacement * document_to_viewport.matrix2.y_axis.length(); // Values are local to the viewport but scaled so values are relative to the current scale.
+		let displacement_document = document_to_viewport.inverse().transform_vector2(displacement_viewport);
+		let displacement_document = if state.is_rounded_to_intervals { displacement_document.round() } else { displacement_document }; // It rounds in document space?
+		document_to_viewport.transform_vector2(displacement_document)
 	}
 
 	#[must_use]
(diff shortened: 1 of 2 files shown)


Keavon:
Awesome! I'm especially a fan of the lines-of-code reduction, thank you for cleaning that up. I tested this extensively for you and only found a few minor things needing a fixup before we merge this:

- There seems to be a regression where holding <kbd>Ctrl</kbd> for snapping to pixel increments (you'll want to zoom in to about 1000% to see) when using <kbd>G</kbd> no longer works, but it does work on master (demonstrated for you in this video):

https://github.com/user-attachments/assets/2ce83a80-84f7-41f6-88ce-e5a9de261c84

- When zoomed at a non-100% scale, the displacement visualization overlays move at the wrong rate for <kbd>G</kbd> translation, shown in this video taken from this branch:

https://github.com/user-attachments/assets/83731276-0444-4bd0-b2cd-575f28d4a18e

- Similarly to the previous point, this also seems to affect when the viewport is tilted and using <kbd>G</kbd> in unconstrained and locally X/Y constrained modes:

https://github.com/user-attachments/assets/09d7505c-208f-4c49-9f08-509394f7a230

- When translating with <kbd>G</kbd> locked along the local X or Y axis, we'd like the overlay to display the distance moved along that local axis (including any number typed by the user) rather than the X and Y components in document space (at the end of the video below, we see how I type numbers like 50, 150, 100, 500, etc. and it misleadingly shows that for the X and Y components instead of the diagonal/hypotenuse). That way it also matches the visualization of <kbd>S</kbd> local scaling for consistency (shown briefly at the local angle's diagonal at the beginning of the video below). See this video taken from this branch:

https://github.com/user-attachments/assets/28035095-7d4a-4878-a49d-e83208d465a1

Hopefully these are quick fixes. Looking forward to merging this soon, thanks for helping with this!

0HyperCube:
Thanks for taking the time to record these high quality video files @Keavon.

1. I have now rounded in document coordinates even when you are relative to the viewport / local space
2. I forgot to divide a thing by the viewport scale so it was broken
3. ^
4. Visualisations are part of #3341 and so out of scope for my PR.

Keavon:
!build

Keavon (approved):
Thank you!

Issue #2214, opened 2025-01-23
Prompt user to report a bug if graph becomes unrenderable due to an act of tooling


Keavon:
The graph can fail to render due to a type error, or outright crash. If this happens while a user is interacting with WYSIWYG tooling (i.e. the graph isn't open), we should consider this a bug because the tooling broke the graph. In this situation, we should indicate to users that this situation isn't supposed to happen, and request they report the bug.

To prevent users from being prompted to report bugs arising from noodling around in the graph, even if they subsequently close the graph and Ctrl+Z their way back to the broken graph, we should store if the graph was open or closed during the document history snapshot. We could also open the graph for the user automatically if they undo/redo to a point when the graph was invalid and that node graph was open at that time. If so, it should also close the graph if the user undoes/redoes past that point and returns to a valid graph, without having interacted with the automatically-opened graph.

0HyperCube:
If you set up a graph with a rectangle node connected directly to the artboard:
![A rectangle node connected directly to the artboard](https://github.com/user-attachments/assets/7ac97547-c9b3-4454-a9ae-1fbadbca7b73)

Then creating a layer with the tools (e.g. drawing a rectangle) breaks the graph:
![A broken graph since the tool inserted the layer into the primary input of the rectangle node](https://github.com/user-attachments/assets/0b190645-00e7-4dfa-b2eb-ae442c3f5aa7)

Would this count as a bug?

Keavon:
@0HyperCube Yes, the tooling should detect that situation and react accordingly. In fact that's a bug for another reason too: the Rectangle node is a generator type node, meaning it has no primary input. The tooling is erroneously attaching a primary input to it anyways. We should probably make the API for attaching an input throw an error if it's attempting to connect a primary input to a node that doesn't accept one. If this was a Path node instead of a Rectangle node, which does accept a primary input, that wouldn't really be a bug but it would still be undesired tooling behavior in this situation. We'd want the tooling to detect you have a graphic attached directly to the artboard and wrap it in a layer so it can attach it to a layer which the new layer would become a sibling of.

PR #3356, opened 2025-11-09
Fix the Artboard tool drawing the first artboard on an infinite canvas with layers present

0HyperCube:
Fixes this reproduction case:
- Create infinite canvas
- Create ≥2 rectangles
- Draw artboard
- **Broken node graph**

Fixes this reproduction case:
- Create infinite canvas
- Create one rectangle
- Draw artboard
- **Rectangle shifts**

Helps towards the goal of #2214.


timon-schelling:
!build

PR #3358, opened 2025-11-10
Fix traversal when deleting nodes

### LINKED ISSUE #3357: Deletion of a node results in incorrect upstream nodes being deleted
Deleting either of these "Add" nodes incorrectly causes the "Instance Position" and Split Vec2" nodes to also be deleted.

However, shaking the "Add" nodes to remove them from the wire does function correctly.

Image

Here is the file (remove the `.txt` extension):

[Wavy Blocks.graphite.txt](https://github.com/user-attachments/files/23443133/Wavy.Blocks.graphite.txt)

adamgerhant:
Closes #3357 
image

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -4235,6 +4235,12 @@ impl NodeNetworkInterface {
 								if !delete_nodes.contains(downstream_id) || (*input_index == 0 && self.is_layer(downstream_id, network_path)) {
 									stack.push(*downstream_id);
 								}
+								// If the traversal reaches the primary input of the node to delete then do not delete it
+								if node_id == downstream_id && *input_index == 0 {
+									can_delete = false;
+									stack = Vec::new();
+									break;
+								}
 							}
 							// If the traversal reaches the export, then the current node is not a sole dependent and cannot be deleted
 							else {


Keavon:
As [mentioned on Discord](https://discord.com/channels/731730685944922173/860830913390575646/1437543207004143667), we actually need to not delete any other nodes except the Add node in the annotated screenshot in your PR description. Sorry for the confusion. Marking as draft for the moment.

Keavon:
Basically, we should do this:

Step 1. Delete the node on its own and reconnect it using the current system
Step 2. Check if we left behind a chain of nodes from our other (secondary) inputs which now are disconnected and each one were sole dependents of the now-deleted node's secondary inputs
Step 3. Delete those if so, but if any node that wasn't directly connected to the now-deleted node is still around, that acts as the dependent to preserve those nodes

Correct cases:

| Before | After |
|-|-|
| capture | capture |
| capture | capture |
| capture | capture |

Keavon:
@adamgerhant

In this file:

[Twirl.graphite.txt](https://github.com/user-attachments/files/23760139/Twirl.graphite.txt)

Deleting the *Transform* node incorrectly deletes way too much:

https://github.com/user-attachments/assets/26fed1e4-73af-49a7-8204-d2a44d8a5bc2

Only the nodes in the red box should be deleted in this case.

adamgerhant:
Should be fixed now that traversal accounts for the nodes connected to primary input not getting deleted.

Keavon (approved):

PR #3364, opened 2025-11-10
Desktop: Enable GPU nodes

timon-schelling:


timon-schelling:
I made it a manual workflow that I can use as needed to test desktop and nix build.
I would like to merge this relatively soon and worry about using it for ci in a later PR. 
This has the clear advantage of having shaders being build for the desktop version via nix.
The shader prebuild pr Can also use this.

@TrueDoctor Can you confirm that your nix workflow still works with this?

desktop/wrapper/Cargo.toml
```diff
@@ -9,8 +9,7 @@ edition = "2024"
 rust-version = "1.87"
 
 [features]
-default = ["gpu"]
-gpu = ["graphite-editor/gpu"]
```
TrueDoctor, line 13: I'm a bit confused wouldn't this disable gpu nodes by default?

timon-schelling: no desktop enables gpu.

But I can also change that if you want.

TrueDoctor: Ah, nevermind then

PR #3370, opened 2025-11-12
Fix checkbox labels becoming unclickable after first toggle

### LINKED ISSUE #3368: Checkbox labels are now one-time use for clicking to toggle their checkbox
@TrueDoctor's PR #3337 introduced another regression (in addition to the ones reported in #3353 which @0HyperCube fixed in PR #3354) that I just found. It seems that most checkbox labels are now one-time use, meaning they can't be clicked after the first time to cause their checkbox to toggle. I presume this is because the random ID used to associate a label with a checkbox is being randomized and both widgets aren't being updated in unison.

Affected cases:
- File > New and click the "Infinite Canvas" label, which toggles once but can't be toggled again
- File > Export and click the "Transparency" label
- All the individual labels for checkboxes in the Overlays and Snapping menus:
  Image Image
- Switch to the Path tool, select a vector object's anchor point, and click the "Colinear Handles" checkbox label in the control bar

Cases which aren't affected:
- File > Preferences and click the "Zoom with Scroll", "Vello Renderer", and "Vector Meshes" checkboxes, these do work multiple times for some reason

ashishmohapatra240:
Fixes #3368

Checkbox labels could only be clicked once, then stopped working because `CheckboxId::new()` generated random IDs on every layout update. It updated `CheckboxInput` widgets with new IDs but skipped updating `TextLabel` widgets. 

I added `CheckboxId::from_string()` to generate stable, hash-based IDs, so each checkbox/label pair now maintains consistent IDs across renders.

https://github.com/user-attachments/assets/00ee985a-00f2-4171-b898-a46c0c6048ad

### DIFF
--- a/editor/src/messages/dialog/export_dialog/export_dialog_message_handler.rs
+++ b/editor/src/messages/dialog/export_dialog/export_dialog_message_handler.rs
@@ -149,7 +149,7 @@ impl LayoutHolder for ExportDialogMessageHandler {
 			DropdownInput::new(entries).selected_index(Some(index as u32)).widget_holder(),
 		];
 
-		let checkbox_id = CheckboxId::new();
+		let checkbox_id = CheckboxId::from_string("export-transparency");
 		let transparent_background = vec![
 			TextLabel::new("Transparency").table_align(true).min_width("100px").for_checkbox(checkbox_id).widget_holder(),
 			Separator::new(SeparatorType::Unrelated).widget_holder(),
--- a/editor/src/messages/dialog/new_document_dialog/new_document_dialog_message_handler.rs
+++ b/editor/src/messages/dialog/new_document_dialog/new_document_dialog_message_handler.rs
@@ -81,7 +81,7 @@ impl LayoutHolder for NewDocumentDialogMessageHandler {
 				.widget_holder(),
 		];
 
-		let checkbox_id = CheckboxId::new();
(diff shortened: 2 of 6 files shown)


Keavon:
Thank you for your attempt, but this is not an appropriate solution. The regression needs to be fixed, not worked around by fundamentally changing its approach.

TrueDoctor:
I did also think about hashing since the random id generation just isn't a a particularly sensible design. What we could do is that we do actually unse a content based hash, which includes its location in the widget tree. Then the ids would be stable which makes the diffing substantially easier and would remove some foot guns

ashishmohapatra240:
Happy to implement either approach. Just let me know which direction you'd prefer!

Keavon:
Let's please try to fix the regression for now before switching approaches quite yet.

TrueDoctor:
I think trying to fix this in the old system would take more effort than implementing a proper fix which would be to not use random ids. The random ids did work fine before had diffing, but combining both is just a very fragile system which is prone to breaking. (And I think we should do this for all widget ids, not just the checkboxes)

Keavon:
Ok, in that case, I suggest we move back to the issue instead of the PR to discuss those plans.

ashishmohapatra240:
Sure @Keavon @TrueDoctor , let's move back to the issue. 😅

PR #3371, opened 2025-11-12
Desktop: Mac fix keyboard input

### LINKED ISSUE #3336: Mac Alt+_ shortcuts don't work on desktop
On mac the shortcuts using alt only work if the menu bar item they are in is selected. This is probably due to macos translating a lot of alt combinations into weird characters (see also https://github.com/helix-editor/helix/issues/2280#issuecomment-1165256309).

From short experimentation with winit it reports "key_without_modifiers" as a platform specific extentions to the KeyEvent struct. Keyboard shortcut testing should probably be done against that.

https://docs.rs/winit/latest/x86_64-apple-darwin/winit/platform/modifier_supplement/trait.KeyEventExtModifierSupplement.html

cc @timon-schelling

timon-schelling:
and some other changes see commit messages

the texture import fix is only for mac.
they changed something about that while copying that code from us.
see https://github.com/tauri-apps/cef-rs/pull/272 for details.

Closes #3336

tested on:
- [x] mac @luca3s @timon-schelling 
- [x] windows @Keavon
- [x] linux @timon-schelling


desktop/src/cef/input.rs
```diff
@@ -64,66 +64,83 @@ pub(crate) fn handle_window_event(browser: &Browser, input_state: &mut InputStat
...
-				winit::keyboard::Key::Named(named_key) => (
-					Some(named_key),
-					match named_key {
-						winit::keyboard::NamedKey::Enter => Some('\u{000d}'),
-						_ => None,
-					},
-				),
-				winit::keyboard::Key::Character(str) => {
-					let char = str.chars().next().unwrap_or('\0');
+			let (named_key, character) = match &event.key_without_modifiers {
```
inkreasing: like i wrote on discord i don't think you can do this at this high level. This breaks @ € and similar special characters for text input.

That you have to translate this to cef events of course complicates this, but with a pure winit app i would only check against key_without_modifiers in the code for shortcuts, while the text input code still uses the regular input.

desktop/src/cef/input.rs
```diff
@@ -175,6 +192,10 @@ impl InputState {
 	fn cef_modifiers_mouse_event(&self) -> CefModifiers {
 		self.cef_modifiers(&winit::keyboard::KeyLocation::Standard, false)
 	}
+
+	fn is_unmodified(&self) -> bool {
```
inkreasing: not a maintainer, but from the function name i thought this was about modifications since the last event or frame. Not that no modifier keys are pressed.

timon-schelling: Good point I will think about that. Was thinking about modifiers to much.

desktop/src/cef/input.rs
```diff
@@ -64,66 +64,83 @@ pub(crate) fn handle_window_event(browser: &Browser, input_state: &mut InputStat
...
+						key_event.type_ = KeyEventType::from(cef_key_event_type_t::KEYEVENT_RAWKEYDOWN);
+
+						if input_state.is_unmodified() && named_key.is_none() {
+							#[cfg(not(target_os = "macos"))] // TODO: Understand why this is needed to avoid double keydown events on mac
+							host.send_key_event(Some(&key_event));
+
+							char_key_event.native_key_code = native_key_code;
+							char_key_event.windows_key_code = buf[0] as i32;
+							char_key_event.character = buf[0];
+							char_key_event.unmodified_character = buf[0];
```
inkreasing: why do you set the both of character and unmodified_character to the same? buf is the unmodified char correct?

csmoe:
Sending key event twice to cef on key press is already not that correct, isn't it? My webview wrapper just sent one event per press and it works fine on mac/windows. cefclient only does this whille handling ime related things https://github.com/chromiumembedded/cef/blob/7087113239f0ef0f929a9404b2807aa65d95b79e/tests/cefclient/browser/text_input_client_osr_mac.mm#L294.

timon-schelling:
I will take a look at your implementation.
Was never able to get cef to forward correct Inputs to the frontend without a key down event before any char event. (On windows)

Maybe I'm filling some fields differently (incorrect)?

But I'm pretty sure that there is a platform difference in some way. Especially with the option key on Mac.

timon-schelling:
Hey @csmoe just finished reimplementing Keyboard input using your webview impl as a guide.

I gave you Co-author on that commit.

I realized afterward that your code doesn't specify a license. If you'd prefer that the portions I was able to draw from not be merged into Graphite, I'll absolutely respect that. It's probably fair use, but I still want to check with you first.

timon-schelling:
@csmoe Your implementation is mostly correct for Mac but doesn't work for win and linux.
On mac both char and key down get mapped to NSKeyDown (there is no NSChar AFAIK). On win and linux both key down and char events are needed.

csmoe:
Thanks, I'll check my implementation. By the way, how can I run the desktop app locally? I tried days ago, but failed.

timon-schelling:
> Thanks, I'll check my implementation. By the way, how can I run the desktop app locally? I tried days ago, but failed.

Command would be `npm run start-desktop` for now.
Maybe come ask in the [desktop-app channel on discord](https://discord.gg/CyG8MRuK) so we don't derail this.

could you also answer my license/attribution question?

csmoe:
Feel free to use the code, I'm quite flattered.

inkreasing:
This isn't a regression in this PR, but the issue i mentioned in https://github.com/GraphiteEditor/Graphite/pull/3374, that on mac (with a german keyboard) the zoom in shortcut is cmd * still exists. It is correct in the way that the shown key and the key it reacts to are the same, but at some point this should probably be changed to + (while reacting to = physical code).

timon-schelling:
Also leaving this here for more context.

adding this in desktop wrapper fixes the not working CMD+'+' issue
but displays = like before. 

```rust
Key::FakeKeyPlus => key = Some(KeyCode::Equal),
```

supporting custom display char requires one of three things
- adding that as a feature to muda (lib used for menu creation on mac)
- forking muda and using that
- write custom NSMenu mapping code (interface with objc lib on mac directly, annoying but not impossible)

PR #3373, opened 2025-11-13
Fix Vello renderer crash when element has no stroke

### LINKED ISSUE #3361: Vello renderer crash
[This lotus pond artwork in our Discord](https://discord.com/channels/731730685944922173/840637256292106241/1431116159587061811) crashes when enabling the Vello renderer.

Download for that file (remove the `.txt` extension):
[lotus2.graphite.txt](https://github.com/user-attachments/files/23445770/lotus2.graphite.txt)

The crash error log:

> index-Xn7tNk5K.js:990 panicked at node-graph/gsvg-renderer/src/renderer.rs:899:57:
> called `Option::unwrap()` on a `None` value

The relevant line we are unwrapping:

```rs
let weight = row.element.style.stroke().unwrap().effective_width();
```

ashishmohapatra240:
Closes #3361

The Vello renderer was calling `.unwrap()` on `stroke()`, which returns an Option. When an element had no stroke defined, this caused a panic.

Used `.as_ref().map_or(0., Stroke::effective_width)` instead. This returns a stroke width of 0.0 when no stroke is present.

### DIFF
--- a/node-graph/gsvg-renderer/src/renderer.rs
+++ b/node-graph/gsvg-renderer/src/renderer.rs
@@ -904,7 +904,7 @@ impl Render for Table<Vector> {
 			let opacity = row.alpha_blending.opacity(render_params.for_mask);
 			if opacity < 1. || row.alpha_blending.blend_mode != BlendMode::default() {
 				layer = true;
-				let weight = row.element.style.stroke().unwrap().effective_width();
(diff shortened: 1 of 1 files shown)


Keavon:
Thanks. Please comment in the issue this PR fixes.

PR #3380, opened 2025-11-14
Fix regression with graph UI culling still on-screen nodes

timon-schelling:
introduced in #3331

This patch replicates the correct behavior with the new viewport api.

Without patch
image

With patch
image

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -1614,7 +1614,7 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
 					return;
 				};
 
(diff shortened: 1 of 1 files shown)


Keavon:
Thanks, I've been noticing this recently.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1614,8 +1614,7 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
 					return;
 				};
 
-				let viewport_bounds = viewport.bounds();
-				let viewport_bbox: [DVec2; 2] = viewport_bounds.into();
+				let viewport_bbox = [DVec2::new(0., 0.), viewport.size().into_dvec2()];
```
Keavon: Please use DVec::ZERO.

PR #3386, opened 2025-11-16
Fix export edge distance

adamgerhant:
Fixes an issue where the export edge distance was not correct by simplifying code and reducing duplicate state


Keavon:
@adamgerhant CI failure (formatting).

timon-schelling:
this should be moved out of the stack doing this now

PR #3388, opened 2025-11-16
Render vector mesh fills more correctly as separate subpaths

### LINKED ISSUE #3378: Render vector mesh fills correctly as separate subpaths

---

Image

</details> 

Image

Image

Image

iamthenoname:
Resolves #3378

Using the SVG renderer:

Using the Vello renderer:
image

The `render_svg` and `render_to_vello` functions have become a bit complex and hard to follow, but I plan on improving this after this change has been settled.

### DIFF
--- a/node-graph/gcore/src/vector/vector_attributes.rs
+++ b/node-graph/gcore/src/vector/vector_attributes.rs
@@ -1063,60 +1063,18 @@ impl Vector {
 		}
 
 		// Sort neighbors by angle to enable finding the "rightmost" path
-		for (point_idx, neighbors) in adjacency.iter_mut() {
-			let point_pos = self.point_domain.positions()[*point_idx];
+		for (&point_idx, neighbors) in adjacency.iter_mut() {
+			let start = self.point_domain.positions()[point_idx];
+
 			neighbors.sort_by(|a, b| {
-				let pos_a = self.point_domain.positions()[a.1];
-				let pos_b = self.point_domain.positions()[b.1];
-				let angle_a = (pos_a - point_pos).y.atan2((pos_a - point_pos).x);
-				let angle_b = (pos_b - point_pos).y.atan2((pos_b - point_pos).x);
-
-				// Compare angles
-				match angle_a.partial_cmp(&angle_b) {
-					Some(std::cmp::Ordering::Equal) | None => {
-						// Angles are equal (within floating point precision), use derivative as tiebreaker
-						const EPSILON: f64 = 1e-10;
-						if (angle_a - angle_b).abs() < EPSILON {
-							// Get segment indices to access handles
-							let seg_idx_a = self.segment_domain.ids().iter().position(|&id| id == a.0);
-							let seg_idx_b = self.segment_domain.ids().iter().position(|&id| id == b.0);
-
-							if let (Some(idx_a), Some(idx_b)) = (seg_idx_a, seg_idx_b) {
-								let handles_a = self.segment_domain.handles()[idx_a];
-								let handles_b = self.segment_domain.handles()[idx_b];
-
-								// Get start and end positions for the segments
-								let start_a = self.point_domain.positions()[self.segment_domain.start_point()[idx_a]];
-								let end_a = self.point_domain.positions()[self.segment_domain.end_point()[idx_a]];
-								let start_b = self.point_domain.positions()[self.segment_domain.start_point()[idx_b]];
-								let end_b = self.point_domain.positions()[self.segment_domain.end_point()[idx_b]];
-
-								// Determine derivative based on edge direction
-								// If reversed (a.2 == true), we're coming into the point, so use derivative_at_end
-								// Otherwise, we're leaving the point, so use derivative_at_start
-								let deriv_a = if a.2 {
-									handles_a.derivative_at_end(start_a, end_a)
-								} else {
-									handles_a.derivative_at_start(start_a, end_a)
-								};
-
-								let deriv_b = if b.2 {
-									handles_b.derivative_at_end(start_b, end_b)
-								} else {
-									handles_b.derivative_at_start(start_b, end_b)
-								};
-
-								// Compare derivative angles
-								let deriv_angle_a = deriv_a.y.atan2(deriv_a.x);
-								let deriv_angle_b = deriv_b.y.atan2(deriv_b.x);
-
-								return deriv_angle_a.partial_cmp(&deriv_angle_b).unwrap_or(std::cmp::Ordering::Equal);
-							}
-						}
-						std::cmp::Ordering::Equal
-					}
-					Some(ordering) => ordering,
-				}
+				let angles: [f64; 2] = [a, b].map(|&(segment_id, end_idx, reversed)| {
+					let end = self.point_domain.positions()[end_idx];
+					let curve = self.segment_domain.handles()[self.segment_domain.id_to_index(segment_id).unwrap()];
+					let curve = if reversed { curve.reversed() } else { curve };
+					let direction = curve.derivative_at_start(start, end);
+					direction.y.atan2(direction.x)
+				});
+				angles[0].partial_cmp(&angles[1]).unwrap_or(std::cmp::Ordering::Equal)
 			});
 		}
 
@@ -1136,18 +1094,7 @@ impl Vector {
 		}
 
 		// Filter out outer (counterclockwise) faces, keeping only inner (clockwise) faces
-		regions.retain(|face| {
-			// Always include 2-edge faces (floating point issues with area calculation)
-			if face.edges.len() == 2 {
-				return true;
-			}
-
-			let area = self.compute_signed_area(face);
-			// Keep clockwise faces (negative area), exclude counterclockwise (positive area)
-			area < 0.0
-		});
-
-		debug!("Found {} closed regions", regions.len());
+		regions.retain(|face| self.compute_signed_area(face) <= 0.0);
 		regions
 	}
 


node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+		loop {
+			iteration += 1;
+
+			if iteration > max_iterations {
+				return None;
+			}
+
+			let neighbors = adjacency.get(&current)?;
+			// Find the next edge in counterclockwise order (rightmost turn)
+			let prev_direction = self.point_domain.positions()[current] - self.point_domain.positions()[path.last()?.start];
```
liunicholas6: `neighbors` is already sorted; no need to do all this work to find `next`. If neighbors sorting is more involved (see previous comment on derivatives) then this actually matters.
```rust
let prev_index = neighbors.iter().position(|seg, _next, _rev| {*seg == prev_segment})?;
let next = neighbors[(prev_index + 1) % neighbors.size()];
```

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+		let mut path = vec![HalfEdge::new(start_segment, from_point, to_point, start_reversed)];
+		let mut current = to_point;
+		let target = from_point;
+		let mut prev_segment = start_segment;
+
+		let mut iteration = 0;
+		let max_iterations = adjacency.len() * 2;
+
+		// Follow the "rightmost" edge at each vertex to find minimal face
+		loop {
```
liunicholas6: nit (style): `for iteration in 1..=max_iterations`

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+			if path.iter().any(|e| e.end == next_point && e.id != next_segment) {
+				return None;
+			}
+
+			path.push(HalfEdge::new(next_segment, current, next_point, next_reversed));
+			prev_segment = next_segment;
+			current = next_point;
+
+			// Prevent infinite loops
+			if path.len() > adjacency.len() {
```
liunicholas6: nit: Redundant with maximum iteration check

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+		// Build adjacency list sorted by angle for proper face traversal
+		let mut adjacency: HashMap<usize, Vec<(SegmentId, usize, bool)>> = HashMap::new();
+
+		for (segment_id, start, end, _) in self.segment_domain.iter() {
+			adjacency.entry(start).or_default().push((segment_id, end, false));
+			adjacency.entry(end).or_default().push((segment_id, start, true));
+		}
+
+		// Sort neighbors by angle to enable finding the "rightmost" path
+		for (point_idx, neighbors) in adjacency.iter_mut() {
```
liunicholas6: There isn't any tiebreaker here for cases where two edges share the same endpoints. In such cases, you should sort by the derivative of the bezier curve. I think it should look something like this:

Inside impl block of BezierHandles
```rust
	pub fn derivative(&mut self, start: DVec2, end: DVec2, t: f64) -> DVec2 {
		let p0 = start;
		match self {
			BezierHandles::Cubic { handle_start, handle_end } => {
				let p1 = *handle_start;
				let p2 = *handle_end;
				let p3 = end;
				3.0 * (1.0 - t).powi(2) * (p1 - p0) + 6.0 * (1.0 - t) * t * (p2 - p1) + 3.0 * t.powi(2) * (p3 - p2)
			}
			BezierHandles::Quadratic { handle } => {
				let p1 = *handle;
				let p2 = end;
				2.0 * (1.0 - t) * (p1 - p0) + 2.0 * t * (p2 - p1)
			}
			BezierHandles::Linear => end - start,
		}
	}
```

Here
```rust
neighbors.sort(|a, b| {
    let segment_angles = [a; b].map(|(_, end_idx, _)| {
        let end = self.point_domain.positions()[end_idx];
        let segment_direction = end - start;
        segment_direction.y.atan2(segment_direction.x)
    });

    let segment_angle_cmp = segment_angles[0].partial_cmp(&segment_angles[1]).unwrap();
    if segment_angle_cmp != std::cmp::Ordering::Equal {
        return segment_angle_cmp;
    }

    let use_source_for_derivative_order = self.segment_domain.id_to_index(a.0) < self.segment_domain.id_to_index(b.0);
    let t : f64 = if use_source_for_derivative_order { 0.0 } else { 1.0 };
    let curve_angles = [a; b].map(|(segment_id, end_idx, reversed)| {
        let end = self.point_domain.positions()[end_idx];
        let forward_curve = self.segment_domain.handles()[self.segment_domain.id_to_index(segment_id)].unwrap();
        let curve_direction = if (reversed) {
            forward_curve.reversed().derivative(start, end, t);
        } else {
            forward_curve.derivative(start, end, t);
        };
        curve_direction.y.atan2(curve_direction.x)
    });
    
    curve_angles[0].partial_cmp(&curve_angles[1]).unwrap()
});
```

Note that if there is no self-intersecting geometry, then given vertex A and vertex B, the "rightmost" edge order of their shared edges should be the same. However this will not be the case if there is self-intersection -- hence the `use_source_for_derivative_order` flag, which should prevent breakage. That can be removed once intersection testing is done.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+				// Mark all half-edges as used
+				for edge in &path {
+					used_half_edges.insert((edge.id, edge.reverse));
+				}
+
+				return Some(FoundSubpath::new(path));
+			}
+
+			// Check if we've created a cycle (might not be back to start)
+			if path.iter().any(|e| e.end == next_point && e.id != next_segment) {
```
liunicholas6: nit: I don't think you should need to check that e.id != next_segment. As far as I can tell this check will only occur when traversing the outer face of a mesh a with a bridge, in which case the first part of the condition will trigger.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1023,134 @@ impl Vector {
...
+				let dir_a = self.point_domain.positions()[a.1] - self.point_domain.positions()[current];
+				let dir_b = self.point_domain.positions()[b.1] - self.point_domain.positions()[current];
+				let angle_a = angle_between(prev_direction, dir_a);
+				let angle_b = angle_between(prev_direction, dir_b);
+				angle_a.partial_cmp(&angle_b).unwrap_or(std::cmp::Ordering::Equal)
+			})?;
+
+			let (next_segment, next_point, next_reversed) = *next;
+
+			if next_point == target {
```
liunicholas6, line 1140: nit: Move after line 1144 to reuse the path-push for clarity

Keavon:
Could you please estimate (as a percentage, or feel free to explain in more detail) how much of this is LLM-authored versus authored directly by you?

iamthenoname:
I'd say around 60% was AI, and 40% was from me.

I didn't use any LLM assistance whatsoever for the rendering code.

For the algorithm itself I implemented it myself first, then had it go through an LLM to check for optimization issues (although I'll admit the LLM did most of the work here).

Admittedly I also got lazy with the follow-up changes and used the LLM to implement those as well. I've been experimenting with these tools recently and wanted to see how effective they would be.

If there's any concerns with the LLM use I'll be toning it down and redoing it myself. I apologize for not making this clear.

Keavon:
Ok, thanks for clarifying and explaining your process. I noticed the prompt files you committed and got concerned that a greater share of the implementation was written without human control than I'd initially expected. The important part is that the actual thought process is yours and the code that's produced is a representation of your engineering competency, so it's not done without a human in the loop. I'm not as concerned about its help doing trivial things like writing the actual characters of code or applying code review style change requests or tab-completing the rest of your lines of code. But beyond trivial uses like tab-completing, we ask that you always disclose LLM authorship so it is clear how that may affect quality, which is crucial information to have when approaching a code review.

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "Graphite",
-	"lockfileVersion": 2,
+	"lockfileVersion": 3,
```
Keavon: Please remove this file.

Keavon:
!build

liunicholas6:
Update: I should have sketched out a few more examples and read the blog post before reviewing.

Derivative angle should actually be the _only_ value used for the sort key for determining being "rightmost", not just the tiebreaker like I initially assumed.

I should have sketched out a few more examples and read the blog post before reviewing.

Additionally, the hack using the derivative at the end instead of the start breaks as many self-intersecting cases and can be ignored.

Apologies for the error

iamthenoname:
> Update: I should have sketched out a few more examples and read the blog post before reviewing.
> 
> Derivative angle should actually be the _only_ value used for the sort key for determining being "rightmost", not just the tiebreaker like I initially assumed.
> 
> I should have sketched out a few more examples and read the blog post before reviewing.
> 
> Additionally, the hack using the derivative at the end instead of the start breaks as many self-intersecting cases and can be ignored.
> 
> Apologies for the error

I see, I'll fix this as well. I'll be working on making tests to make sure that the algorithm is working properly.

Keavon:
@iamthenoname would you be willing to open up commit access to @liunicholas6 who was requested to collaborate with you on this PR's code? You can do that by inviting him to have write access to your fork's repo in its settings page. If you're willing, of course.

iamthenoname:
Alright, I've added @liunicholas6 as a collaborator.

node-graph/gcore/src/subpath/structs.rs
```diff
@@ -228,6 +228,24 @@ impl BezierHandles {
...
+	}
+
+	/// Returns the tangent vector at the end of the curve (t=1)
+	pub fn derivative_at_end(&self, start: DVec2, end: DVec2) -> DVec2 {
+		match self {
+			BezierHandles::Linear => end - start,
+			BezierHandles::Quadratic { handle } => 2.0 * (end - handle),
+			BezierHandles::Cubic { handle_end, .. } => 3.0 * (end - handle_end),
+		}
+	}
```
0HyperCube, line 248: Note most of the functionality of this struct has been removed (including a former derivative implementation) in favour of using `kurbo::PathSeg`. It is therefore undesirable to add this back.

node-graph/gcore/src/subpath/structs.rs
```diff
@@ -228,6 +228,24 @@ impl BezierHandles {
 			_ => self,
 		}
 	}
+
+	/// Returns the tangent vector at the start of the curve (t=0)
+	pub fn derivative_at_start(&self, start: DVec2, end: DVec2) -> DVec2 {
+		match self {
+			BezierHandles::Linear => end - start,
+			BezierHandles::Quadratic { handle } => 2.0 * (handle - start),
+			BezierHandles::Cubic { handle_start, .. } => 3.0 * (handle_start - start),
```
0HyperCube, line 237: This returns a zero vector in the case when the start handle is on top of the start anchor. This is a situation frequently created by the vector tooling in the editor. A zero vector is correct for the derivative, however it isn't correct if you want a tangent (which is what you wrote in the doc comment).

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -4,7 +4,8 @@ use crate::vector::vector_types::Vector;
 use dyn_any::DynAny;
 use glam::{DAffine2, DVec2};
 use kurbo::{CubicBez, Line, PathSeg, QuadBez};
-use std::collections::HashMap;
+use log::debug;
```
0HyperCube, line 7: This import is unused. Most crates do `#[macro_use] extern crate log;` but not this one.

It would in fact be very helpful to have some debug logging since I have no clue what the rendering code does. If you want to remove it for performance, you can put it behind a `#[cfg(feature = "logging")]` or something.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -678,6 +679,44 @@ impl FoundSubpath {
...
+				}
+			}
+		}
+
+		if self.is_closed() {
+			bezpath.close_path();
+		}
+
+		bezpath
+	}
```
0HyperCube: Why not use Vector::stroke_bezier_paths ?

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+				return true;
+			}
+		}
+
+		false
+	}
+
+	/// Compute signed area of a face using the shoelace formula.
+	/// Negative area indicates clockwise winding, positive indicates counterclockwise.
+	fn compute_signed_area(&self, face: &FoundSubpath) -> f64 {
```
0HyperCube, line 1042: This ignores all of the bézier paths, leading to a completely incorrect area (and potentially misleading winding direction). However you/your LLM never mentions this.
image

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+		area * 0.5
+	}
+
+	/// Find all minimal closed regions (faces) in a branching mesh vector.
+	pub fn find_closed_regions(&self) -> Vec<FoundSubpath> {
+		let mut regions = Vec::new();
+		let mut used_half_edges = HashSet::new();
+
+		// Build adjacency list sorted by angle for proper face traversal
+		let mut adjacency: HashMap<usize, Vec<(SegmentId, usize, bool)>> = HashMap::new();
```
0HyperCube, line 1058: It could be beneficial to add a struct since it is hard to determine what the usize and bool are.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+		}
+		area * 0.5
+	}
+
+	/// Find all minimal closed regions (faces) in a branching mesh vector.
+	pub fn find_closed_regions(&self) -> Vec<FoundSubpath> {
+		let mut regions = Vec::new();
+		let mut used_half_edges = HashSet::new();
+
+		// Build adjacency list sorted by angle for proper face traversal
```
0HyperCube, line 1057: What does it mean to sort by angle? Angle between which directions?

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+		}
+
+		// Sort neighbors by angle to enable finding the "rightmost" path
+		for (&point_idx, neighbors) in adjacency.iter_mut() {
+			let start = self.point_domain.positions()[point_idx];
+
+			neighbors.sort_by(|a, b| {
+				let angles: [f64; 2] = [a, b].map(|&(segment_id, end_idx, reversed)| {
+					let end = self.point_domain.positions()[end_idx];
+					let curve = self.segment_domain.handles()[self.segment_domain.id_to_index(segment_id).unwrap()];
```
0HyperCube, line 1072: It seems a bit wasteful to convert indexes to ids and then back to indexes in the sort by.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+					let curve = if reversed { curve.reversed() } else { curve };
+					let direction = curve.derivative_at_start(start, end);
+					direction.y.atan2(direction.x)
+				});
+				angles[0].partial_cmp(&angles[1]).unwrap_or(std::cmp::Ordering::Equal)
+			});
+		}
+
+		for (segment_id, start, end, _) in self.segment_domain.iter() {
+			for &reversed in &[false, true] {
```
0HyperCube, line 1082: There is no need to do references here since [bool;2] has `into_iter()`.

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+					continue;
+				}
+
+				if let Some(face) = self.find_minimal_face_from_edge(segment_id, from, to, reversed, &adjacency, &mut used_half_edges) {
+					regions.push(face);
+				}
+			}
+		}
+
+		// Filter out outer (counterclockwise) faces, keeping only inner (clockwise) faces
```
0HyperCube, line 1096: It isn't really clear why you'd want to do this?

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+		start_reversed: bool,
+		adjacency: &HashMap<usize, Vec<(SegmentId, usize, bool)>>,
+		used_half_edges: &mut HashSet<(SegmentId, bool)>,
+	) -> Option<FoundSubpath> {
+		let mut path = vec![HalfEdge::new(start_segment, from_point, to_point, start_reversed)];
+		let mut current = to_point;
+		let target = from_point;
+		let mut prev_segment = start_segment;
+
+		let max_iterations = adjacency.len() * 2;
```
0HyperCube, line 1116: Why would this be the max iterations? Surely you should only loop `adjacency.len()` times (since you guarantee to not use the same segment multiple times).

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -985,6 +1024,134 @@ impl Vector {
...
+				}
+				return Some(FoundSubpath::new(path));
+			}
+
+			prev_segment = next_segment;
+			current = next_point;
+		}
+
+		// If we exit the loop without returning, the path didn't close
+		None
```
0HyperCube, line 1153: Should probably be `unreachable!()` I think?

node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -719,143 +718,251 @@ impl Render for Table<Vector> {
...
 			let mask_type = if vector.style.stroke().map(|x| x.align) == Some(StrokeAlign::Inside) {
 				MaskType::Clip
 			} else {
 				MaskType::Mask
 			};
 
-			let path_is_closed = vector.stroke_bezier_paths().all(|path| path.closed());
-			let can_draw_aligned_stroke = path_is_closed && vector.style.stroke().is_some_and(|stroke| stroke.has_renderable_stroke() && stroke.align.is_not_centered());
-			let can_use_paint_order = !(row.element.style.fill().is_none() || !row.element.style.fill().is_opaque() || mask_type == MaskType::Clip);
+			// Helper closure to render a single path with given path string and closed status
```
0HyperCube, line 729: Please just make this a separate function. Very long closures are hard to reason about.

node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -719,143 +718,251 @@ impl Render for Table<Vector> {
...
+					render.leaf_tag("path", |attributes| {
+						attributes.push("d", combined_path);
+						let matrix = format_transform_matrix(element_transform);
+						if !matrix.is_empty() {
+							attributes.push("transform", matrix);
+						}
 
-				let fill_and_stroke = style.render(defs, element_transform, applied_stroke_transform, bounds_matrix, transformed_bounds_matrix, &render_params);
+						let mut style = row.element.style.clone();
+						style.clear_fill(); // Critical: Remove fill to only render stroke
```
0HyperCube, line 921: What does it mean to be "Critical"? That rather implies that the rest of the code isn't critical and so can be removed.

This is rather a junk comment since it is very obvious that `style.clear_fill()` will remove the fill.

node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -1036,70 +1147,188 @@ impl Render for Table<Vector> {
...
+								bezpath.apply_affine(Affine::new(applied_stroke_transform.to_cols_array()));
+								for element in bezpath {
+									combined_path.push(element);
+								}
+							}
 
-							scene.push_layer(peniko::Mix::Normal, 1., kurbo::Affine::IDENTITY, &rect);
-							vector_table.render_to_vello(scene, parent_transform, _context, &render_params.for_alignment(applied_stroke_transform));
-							scene.push_layer(peniko::BlendMode::new(peniko::Mix::Normal, compose), 1., kurbo::Affine::IDENTITY, &rect);
+							// Use the existing rendering logic for this fallback case
```
0HyperCube, line 1190: What do you mean by the "existing rendering logic"? In the diff, it appears to be completely new code.

node-graph/gsvg-renderer/src/renderer.rs
```diff
@@ -1036,70 +1147,188 @@ impl Render for Table<Vector> {
...
-							peniko::Compose::SrcOut
-						} else {
-							peniko::Compose::SrcIn
-						};
-
-						if wants_stroke_below {
-							scene.push_layer(peniko::Mix::Normal, 1., kurbo::Affine::IDENTITY, &rect);
-							vector_table.render_to_vello(scene, parent_transform, _context, &render_params.for_alignment(applied_stroke_transform));
-							scene.push_layer(peniko::BlendMode::new(peniko::Mix::Normal, compose), 1., kurbo::Affine::IDENTITY, &rect);
+					if is_branching {
```
0HyperCube, line 1150: I don't like having the two rendering paths separate. For example in the image below, I have appended a segment to the bottom, which ended up removing the red fill from the top shape. I don't think that the visual outcome should depend on having a ≥3 connected point anywhere in the entire network (spooky action at a distance).

two vertical stacked V shapes initially filled with red, but then empty after a random segment is appended far away

node-graph/gcore/src/vector/vector_attributes.rs
```diff
@@ -678,6 +679,44 @@ impl FoundSubpath {
...
+	pub fn to_bezpath(&self, vector: &Vector) -> kurbo::BezPath {
+		let mut bezpath = kurbo::BezPath::new();
+
+		if let Some(first_edge) = self.edges.first() {
+			let start_pos = vector.point_domain.positions()[first_edge.start];
+			bezpath.move_to(dvec2_to_point(start_pos));
+		}
+
+		for edge in &self.edges {
+			let segment_index = vector.segment_domain.ids().iter().position(|&id| id == edge.id).expect("Segment ID must exist");
```
0HyperCube, line 692: Please try to avoid O(n^2) algorithms when you could just store the segment index in the `HalfEdge` struct.

0HyperCube (commented):
A trivial case for which this does not work:

Image

[broken_fill.graphite.txt](https://github.com/user-attachments/files/23695496/broken_fill.graphite.txt)

node-graph/gsvg-renderer/src/renderer.rs
Keavon: This is the desired behavior given that the alternative is essentially arbitrary undefined behavior.

0HyperCube: @Keavon The current implementation means that appending a segment somewhere has the magic ability to alter fills anywhere else in your drawing (spooky action at a distance). This is not something (I'd expect) users would anticipate.

> arbitrary undefined behavior [sic].

I don't see what you mean. You get to define the behaviour!
In a specification, you get to leave behaviour as undefined. However the actual algorithm will deterministically produce an outcome for all inputs.

Keavon:
Superseded by #3474.

PR #3389, opened 2025-11-16
feat: Add spacing method options to Repeat and Circular Repeat nodes

### LINKED ISSUE #2551: Add options to the 'Repeat' node for the spacing with envelope/span/pitch/gap

a-khushal:
Closes #2551
image

@Keavon

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2198,6 +2198,7 @@ dependencies = [
  "text-nodes",
  "tokio",
  "url",
+ "vector-nodes",
  "wasm-bindgen",
  "web-sys",
  "wgpu-executor",
--- a/editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
+++ b/editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
@@ -2181,6 +2181,8 @@ fn static_node_properties() -> NodeProperties {
 	map.insert("grid_properties".to_string(), Box::new(node_properties::grid_properties));
 	map.insert("spiral_properties".to_string(), Box::new(node_properties::spiral_properties));
 	map.insert("sample_polyline_properties".to_string(), Box::new(node_properties::sample_polyline_properties));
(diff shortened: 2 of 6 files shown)


Keavon:
!build

Keavon:
"Width" is not supposed to be a parameter.

a-khushal:
@Keavon can you please review this now?

Keavon:
!build

a-khushal:
Hi @Keavon, any updates on the review when you get a moment?

Ezbaze:
### Repeat Node

`Span`, `Pitch`, and `Gap` work as intended.

`Envelope` does not. The initial shape should remain in the same position, but `Envelope` mode currently shifts it. See below for details:

The orange dot at (0, 0) marks the canvas center, and in both examples the rectangle starts centered on the canvas.
**Current:**
Current

**Expected:**
Expected

File (rename to `.graphite`):
[pr3389-Repeat.txt](https://github.com/user-attachments/files/24119561/pr3389-Repeat.txt)

---

### Circular Repeat Node

This PR breaks the `Circular Repeat` node.

The `Repeat` node uses `Direction` to define spacing, but `Circular Repeat` previously distributed instances evenly around the circle by default. This PR doesn’t account for that original behavior.

Video:

https://github.com/user-attachments/assets/19aad6d4-99b9-487c-8cba-138cf87d5030

File (rename to `.graphite`):
[pr3389-Circular_Repeat.txt](https://github.com/user-attachments/files/24118651/pr3389-Circular_Repeat.txt)

CC @Keavon

a-khushal:
@Ezbaze could you test this again? I’ve patched both issues you mentioned.

PR #3392, opened 2025-11-17
Comment the GH Actions run link on `!build` comments for easier troubleshooting

bluelinden:
All this does is immediately post a comment with the Actions build link when the core team triggers a build. It then edits that comment when the build finishes.

It means I don't need to go looking through the Actions tab for my build in case it fails.

Just a little development QoL patch!


TrueDoctor:
!build

bluelinden:
> !build

this actions run won't build using the new workflow - i believe Keavon said earlier that it has to be on the default branch.

Keavon (approved):

PR #3399, opened 2025-11-19
Add a command to Graphene CLI for listing protonode identifiers

TrueDoctor:
Makes the cli not panic on startup and adds a command for printing the protonode identifiers for all currently registered nodes

### DIFF
--- a/node-graph/graphene-cli/src/main.rs
+++ b/node-graph/graphene-cli/src/main.rs
@@ -6,8 +6,8 @@ use graph_craft::graphene_compiler::{Compiler, Executor};
 use graph_craft::proto::ProtoNetwork;
 use graph_craft::util::load_network;
 use graph_craft::wasm_application_io::EditorPreferences;
-use graphene_std::text::FontCache;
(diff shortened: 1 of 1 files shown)


node-graph/graphene-cli/src/main.rs
```diff
@@ -56,6 +56,7 @@ enum Command {
 		#[clap(long, short = 'l')]
 		run_loop: bool,
 	},
+	PrintNodeIdentifiers,
```
Keavon: I suggest calling this "List" not "Print".

node-graph/graphene-cli/src/main.rs
```diff
@@ -123,6 +132,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
 				tokio::time::sleep(std::time::Duration::from_millis(16)).await;
 			}
 		}
+		_ => unreachable!("all other commands should be hadled before this match statement is run")
```
Keavon: - `all` -> `All`
- `hadled` -> handled?

PR #3400, opened 2025-11-19
Add export command to Graphene CLI

TrueDoctor:

### DIFF
--- a/node-graph/graphene-cli/src/export.rs
+++ b/node-graph/graphene-cli/src/export.rs
@@ -3,7 +3,7 @@ use graph_craft::graphene_compiler::Executor;
 use graphene_std::application_io::{ExportFormat, RenderConfig};
 use graphene_std::core_types::ops::Convert;
 use graphene_std::core_types::transform::Footprint;
-use graphene_std::raster_types::{Raster, CPU, GPU};
+use graphene_std::raster_types::{CPU, GPU, Raster};
 use interpreted_executor::dynamic_executor::DynamicExecutor;
 use std::error::Error;
 use std::io::Cursor;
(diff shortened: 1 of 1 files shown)


node-graph/graphene-cli/src/main.rs
```diff
@@ -92,15 +111,24 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+
+	// Clone the application_io Arc before borrowing to extract executor
+	let application_io_for_api = application_io_arc.clone();
+
+	// Get reference to wgpu executor and clone device handle
+	let wgpu_executor_ref = application_io_arc.gpu_executor().unwrap();
+	let device = wgpu_executor_ref.context.device.clone();
 
 	let preferences = EditorPreferences { use_vello: true };
 	let editor_api = Arc::new(WasmEditorApi {
```
timon-schelling, line 129: WasmEditorApi seems to be the wrong name now.

node-graph/graphene-cli/src/main.rs
```diff
@@ -92,15 +111,24 @@ async fn main() -> Result<(), Box<dyn Error>> {
 	log::info!("creating gpu context",);
 	let mut application_io = block_on(WasmApplicationIo::new_offscreen());
 
-	if let Command::Run { image: Some(ref image_path), .. } = app.command {
```
timon-schelling, line 95: If run is not an option for now we should remove the window handling code.
That is not working at the moment AFAIK

node-graph/graphene-cli/src/main.rs
TrueDoctor: Yes, but we will need to think about how we can best restructure how the application io works. We could potentially even add a proper desktop version for the application io

node-graph/graphene-cli/src/main.rs
TrueDoctor: I think we can remove the windowing code when we rework the application io. It is already marked as deprecated

PR #3401, opened 2025-11-19
Desktop: Add Icon for Mac

timon-schelling:
depends on #3406


timon-schelling:
@Keavon fine to add Icon now?
Will probably moved to branding repo soon?
+220kb

Keavon:
We can add it to the branding repo now, add it as a frontend dev dependency, then `../../` our way out of that folder and into the `frontend/node_modules` folder to access it.

PR #3405, opened 2025-11-20
Add the 'Blend Shapes' and 'Origins to Polyline' nodes; generalize the 'Morph' node to >2 states

### LINKED ISSUE #2316: Node for tweening "topo lines" across several guide paths
This is sort of like the Morph node, but it should work despite changes in topology of the curvature and it should take multiple layers (as a GraphicGroup stack, lavender-colored data) and output either one (with a `t` value) or multiple paths that interpolate it.

In this rough hand-drawn example, the RED paths are the end guide paths, the PINK paths are the intermediate guide paths, and the GRAY lines are the tweened paths. In the real output, none of the guide paths would be directly preserved, only the interpolated versions, ensuring they are evenly spaced with each other given some density or count parameter.

Image

Keavon:
Closes #2316

This also introduces the `Progression` widget type for a `t` value along one or multiple parts (such as subpaths or vector elements):

capture

# Morph node

This changes the behavior of the Morph node to work more like the Boolean Operation node, generalizing it from two inputs to any number of inputs by reducing a group of vector shapes to a single shape along the morph progression. This would previously zip the multiple elements in the starting, and ending, states and morph each pair in unison; this is no longer supported and is a breaking change, but if only a single element was the morph source and target, that has an automatic migration to handle the change without breakage.

https://github.com/user-attachments/assets/61410ec9-27c4-4090-a0d8-8e7720e082de

# Blend Shapes node

A network node that calls Morph repeatedly and produces a chosen number of instances.

capture

capture

capture

capture

https://github.com/user-attachments/assets/7d4d592b-e53c-49be-9cab-7fa0e56e4d50

# Origins to Polyline node

A network node that creates a polyline stringing together every object in the table it is given. This is designed to be used alongside the Blend Shapes node to control the curvature of the interpolation path (such Path tool changes encoded in the Path node) using this setup:

capture

https://github.com/user-attachments/assets/064381f6-7d34-4a1f-be57-0acfd124f7a8

### DIFF
--- a/node-graph/nodes/vector/src/vector_nodes.rs
+++ b/node-graph/nodes/vector/src/vector_nodes.rs
@@ -6,9 +6,9 @@ use core_types::table::{Table, TableRow, TableRowMut};
 use core_types::transform::{Footprint, Transform};
 use core_types::{CloneVarArgs, Color, Context, Ctx, ExtractAll, ExtractVarArgs, OwnedContextImpl};
 use glam::{DAffine2, DVec2};
-use graphic_types::Graphic;
 use graphic_types::Vector;
 use graphic_types::raster_types::{CPU, GPU, Raster};
+use graphic_types::{Graphic, IntoGraphicTable};
 use kurbo::{Affine, BezPath, DEFAULT_ACCURACY, Line, ParamCurve, PathEl, PathSeg, Shape};
 use rand::{Rng, SeedableRng};
 use std::collections::hash_map::DefaultHasher;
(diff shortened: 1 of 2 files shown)


Keavon:
I need to fix the test but this is otherwise open for review, thanks @TrueDoctor @adamgerhant.

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -1557,127 +1578,147 @@ async fn morph(_: impl Ctx, source: Table<Vector>, #[expose] target: Table<Vecto
...
+	// Preserve original graphic table as upstream data so this group layer's nested layers can be edited by the tools.
+	let graphic_table_content = content.clone().into_graphic_table();
 
-	source
-		.into_iter()
-		.zip(target.into_iter())
-		.map(|(source_row, target_row)| {
-			let mut vector = Vector::default();
+	// If the input isn't a Table<Vector>, we convert it into one by flattening any Table<Graphic> content.
+	let content = content.into_flattened_vector_table();
```
adamgerhant, line 1585: I think this should be done with the "FlattenVectorNode", the input to this node should be Table<Vector>, and in the future the automatic type conversion can be added

Keavon: I thought about automatic type conversion but decided we probably don't want that, because we can't guarantee that will always be desired. Going from a less pure to more pure form of data is lossy (we throw away non-vector objects, in this case) and opinionated (do we want to flatten deeply or only one level while discarding or merging the geometry of deeper levels?) and will need to vary per node based on what makes sense for each one. This allows the node to easily reuse this functionality if needed, but not let data get destroyed by the type conversion system if it doesn't encode user intent.

PR #3407, opened 2025-11-21
Automatically show the appropriate widget for a node parameter based on its current type

adamgerhant:
Protonode widgets are currently static based on their implementation and not the actual type. For example, here I have a value of DVec2 being subtracted but it is displayed as f64.

Cannot display aliases types, which will have to be reworked to be unique types.
image


Keavon:
Breaks the Percentage Value node:

capture

PR #3412, opened 2025-11-23
Simplify node graph type error diagnostic message text

adamgerhant:
Old vs new
image
image

Potential further change: Hide error until clicked but show it on all erroring nodes. Not a single large error that jumps from node to node.
image


Keavon:
Closing this since I don't think it does what is directly needed, although I will aim to visit this in the future. I just need to design it first-hand, rather than attempt to review a proposed redesign. Thanks though :smile:

PR #3415, opened 2025-11-24
Use more specific node input indexing when displaying invalid input errors

adamgerhant:
Does not display invalid inputs in the property panel and only shows invalid inputs for the inputs which are invalid, not all.

Fixes invalid input indexing for automatically converted Into nodes.

After:
chrome_Z2GRCCfMUR
Before:
chrome_VA7eHA57sa

### DIFF
--- a/node-graph/graph-craft/src/document.rs
+++ b/node-graph/graph-craft/src/document.rs
@@ -84,7 +84,7 @@ pub struct OriginalLocation {
 	pub dependants: Vec<Vec<NodeId>>,
 	/// A list of flags indicating whether the input is exposed in the UI
 	pub inputs_exposed: Vec<bool>,
-	// For automatically inserted convert and into nodes, if there is an error display it on the node it is connect to
+	/// For automatically inserted Convert and Into nodes, if there is an error, display it on the node it is connect to.
 	pub auto_convert_index: Option<usize>,
 }
 
--- a/node-graph/graph-craft/src/proto.rs
+++ b/node-graph/graph-craft/src/proto.rs
@@ -539,13 +539,23 @@ impl ProtoNetwork {
 #[derive(Clone, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum GraphErrorType {
 	NodeNotFound(NodeId),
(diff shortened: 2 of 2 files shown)


node-graph/graph-craft/src/proto.rs
```diff
@@ -542,6 +542,8 @@ pub enum GraphErrorType {
 	UnexpectedGenerics { index: usize, inputs: Vec<Type> },
 	NoImplementations,
 	NoConstructor,
+	// The first vec represents a list of correct NodeIOTypes
+	// The second vec represents what the input index and what it expects
```
Keavon:
```suggestion
	/// The `inputs` represents a formatted list of input indices corresponding to their types.
	/// Each element in `error_inputs` represents a valid `NodeIOTypes` implementation.
	/// The inner Vec stores the inputs which need to be changed and what type each needs to be changed to.
```

node-graph/graph-craft/src/document.rs
```diff
@@ -84,6 +84,8 @@ pub struct OriginalLocation {
 	pub dependants: Vec<Vec<NodeId>>,
 	/// A list of flags indicating whether the input is exposed in the UI
 	pub inputs_exposed: Vec<bool>,
+	// For automatically inserted convert and into nodes, if there is an error display it on the node it is connect to
```
Keavon:
```suggestion
	/// For automatically inserted Convert and Into nodes, if there is an error, display it on the node it is connect to.
```

node-graph/graph-craft/src/proto.rs
```diff
@@ -756,20 +758,19 @@ impl TypingContext {
 
 		match valid_impls.as_slice() {
 			[] => {
+				let convert_node_index_offset = node.original_location.auto_convert_index.unwrap_or(0);
 				let mut best_errors = usize::MAX;
 				let mut error_inputs = Vec::new();
 				for node_io in impls.keys() {
+					// For errors on convert nodes, add to the input index to it is correct for the node it is connected to
```
Keavon:
```suggestion
					// For errors on Convert nodes, offset the input index so it correctly corresponds to the node it is connected to.
```

PR #3421, opened 2025-11-26
Fix regression that reversed indexing of the Loop Level parameter of the 'Instance Index' node

### LINKED ISSUE #3360: 'Instance Index' node's 'Loop Level' parameter counts from the wrong direction
The *Instance Index* node has a *Loop Level* parameter that lets you access the outer scopes of multiple levels of nested iteration. This is akin to `i`, `j`, and `k` index variables commonly used in nested `for` loops in programming.

As a regression, it has become reversed. 0 is supposed to correspond to the innermost loop level, but it instead is counting from the outermost loop level. This needs to be changed back to how it was designed.

Keavon:
Closes #3360


TrueDoctor:
This only adds one of the .rev() calls

Keavon:
There are multiple?

PR #3425, opened 2025-11-27
Hide cursor while dragging number inputs in Safari to approximate PointerLock, and disable it on desktop

### LINKED ISSUE #2038: Click-and-drag number fields can't be released in Safari

Keavon:
This PR, more properly this time, fixes #2038.

### DIFF
--- a/frontend/src/components/Editor.svelte
+++ b/frontend/src/components/Editor.svelte
@@ -223,6 +223,10 @@
 		user-select: none;
 	}
 
+	body.cursor-hidden * {
+		cursor: none !important;
+	}
+
 	// Needed for the viewport hole punch on desktop
 	html:has(body > .viewport-hole-punch),
 	body:has(> .viewport-hole-punch) {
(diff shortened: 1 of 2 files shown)


Keavon:
For future reference: if we try to enable pointerlock during the pointermove event, we get the exception "Pointer lock requires a user gesture." which is a result of the "engagement gesture" requirement that existed in the 2016 version of this spec (<https://www.w3.org/TR/2016/REC-pointerlock-20161027/#glossary>) but is no longer a factor in spec-complaint browsers (Chrome and Firefox). Source code for that exception: <https://github.com/WebKit/WebKit/blob/982c9baee3b84661bb70d26ccf0067001572819e/Source/WebCore/page/PointerLockController.cpp#L90>

Safari also chooses to push all content in the page down, causing a resize of the whole web app, when showing its dialog which is another reason we can't use the pointerlock API.

timon-schelling:
This did not fix the desktop (mac) part fully.
I included the necessary change in #3424.
See https://github.com/GraphiteEditor/Graphite/pull/3424/commits/23dbde4cea2757575c31db436128619bed51ec04 for explanation.

PR #3428, opened 2025-11-28
Desktop: Add app menu for Mac

timon-schelling:
Screenshot1
Screenshot2

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -280,6 +280,11 @@ impl App {
 					window.hide_others();
 				}
 			}
+			DesktopFrontendMessage::WindowShowAll => {
+				if let Some(window) = &self.window {
+					window.show_all();
+				}
+			}
 		}
 	}
 
--- a/desktop/src/window.rs
+++ b/desktop/src/window.rs
@@ -13,6 +13,7 @@ pub(crate) trait NativeWindow {
 	fn update_menu(&self, _entries: Vec<MenuItem>) {}
 	fn hide(&self) {}
 	fn hide_others(&self) {}
+	fn show_all(&self) {}
 }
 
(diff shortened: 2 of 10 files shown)


timon-schelling:
@Keavon most important confirm that the menu layout is like you intended.

Keavon:
Isn't it supposed to say "Hide Graphite" not just "Hide"?

Keavon:
Also, should we aim to put back whatever that "services" menu thing is? I don't know what it is or does, but is it likely some users may be missing it?

timon-schelling:
> Isn't it supposed to say "Hide Graphite" not just "Hide"?

I will see if apps do that or just use 'hide'

Keavon:
It includes the app name: https://developer.apple.com/design/human-interface-guidelines/the-menu-bar#App-menu

timon-schelling:
> Also, should we aim to put back whatever that "services" menu thing is? I don't know what it is or does, but is it likely some users may be missing it?

Not really, services are background tasks that can run without the app open, we don't need that, and other apps that don't use that feature also don't include the option.

timon-schelling:
@Keavon Done ready for final review and merge.

Keavon:
This is also missing the "Show All" menu item below "Hide Others". And "Quit" should say "Quit Graphite". Lastly, I think we should make the same exception that other apps make and follow Apple's guideline to remove the ellipsis after "About Graphite…" even though technically that does open a dialog. But it looks more out of place having an ellipsis (unlike all other apps), than looking out of place by missing the ellipsis (unlike all other dialog-opening actions).

Keavon:
I made the change myself (I wanted to see how that code worked).

image

Keavon (approved):

PR #3434, opened 2025-11-30
Reduce flickering

### LINKED ISSUE #3433: Reduced motion
Flickering animations such as opening the node graph should not play when `@media (prefers-reduced-motion: reduce)`.

https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion

0HyperCube:
Closes #3433 by removing transitions and animations when `@media(prefers-reduced-motion)` is active

### DIFF
--- a/frontend/index.html
+++ b/frontend/index.html
@@ -44,6 +44,11 @@
 				border: 4px solid #eee;
 				border-color: #eee transparent #eee transparent;
 				animation: spinning-loading-indicator 1s linear infinite;
+				@media (prefers-reduced-motion) {
+					animation: none;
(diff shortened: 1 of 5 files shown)


Keavon:
Thanks for the proposed change. Fades are not flickering and because they are brief, they would generally be considered as unobtrusive when it comes to accessibility considerations for users desiring reduced motion, so this doesn't solve a problem that is likely to affect many users. I am not comfortable merging code that adds complexity/maintenance burden to the CSS without clear benefit. But thank you for the suggestion.

PR #3436, opened 2025-11-30
Replace text-only tooltips with custom richly styled tooltips

### LINKED ISSUE #3202: Show node name instead of custom alias on tooltip

### LINKED ISSUE #3435: Richly styled tooltips
Image

Keavon:
Closes #3435. Also fixes #3202.

Image

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -2225,7 +2225,7 @@ impl NodeGraphMessageHandler {
 					.icon(Some("FrameAll".to_string()))
 					.tooltip_label("Preview")
 					.tooltip_description("Temporarily set the graph output to the selected node or layer. Perform the shortcut on a node or layer for quick access.")
-					.tooltip_shortcut(KeysGroup(vec![Key::Alt, Key::MouseLeft]))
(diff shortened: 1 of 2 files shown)


0HyperCube:
I like how the JSON shows up in some of the tooltips without `.to_string()` on the `.tooltip_shortcut()` calls.

Preview [{&quot;key&quot;:&quot;Alt&quot;,&quot;label&quot;:&quot;Alt&quot;},{&quot;key&quot;:&quot;MouseLeft&quot;,&quot;label&quot;:&quot;Click&quot;}]

Keavon:
Good catch, at the last minute (and I apparently didn't test it after I saw the compiler was satisfied), I thought I could remove the `.to_string()` because the function call's argument is of type `Into<String>`. Not sure why that isn't the case, but it's an easy fix to put that back. Thanks for pointing it out.

0HyperCube:
@Keavon the `.to_string()` uses display, whereas there is for some reason a custom `impl From<KeysGroup> for String` that uses the JSON.
https://github.com/GraphiteEditor/Graphite/blob/e8ebcc2c21e400de2543698ce74a35589980e6c8/editor/src/messages/input_mapper/utility_types/input_keyboard.rs#L362-L367

PR #3447, opened 2025-12-04
Restyle and refactor shortcut labels to send hints bar and welcome screen layouts from Rust

### LINKED ISSUE #739: Replace hint Rust -> JS data with a widget layout
Currently we send a bunch of recursive structs defining the layouts to JS, then Vue renders it. That should be moved to a Rust-defined layout with the new layout system, which can then reduce the amount of code serving a redundant purpose of laying out widgets.

Keavon:
Closes #739.


TrueDoctor:
Have you done perf measurements to check if this induces extra overheads?

Keavon:
No but I would expect it improves performance because now it utilizes the widget diffing system instead of, as before, sending the full hint data every update.

PR #3451, opened 2025-12-06
Replace node definition string-based lookups with DefinitionIdentifier instances

adamgerhant:
Improves the definition reference system by referencing nodes with `DefinitionIdentifier` instead of strings. This means protonodes use their identifier to reference their definition, instead of an arbitrary string.

Sets the foundation for an improved migration system and linking nodes to their definition when serializing.

 For example:
 instead of 
```rs
resolve_document_node_type("Star")
```
you would do 
```rs
resolve_document_node_type(&DefinitionIdentifier::ProtoNode(graphene_std::vector::generator_nodes::star::IDENTIFIER))
```
Another example
```rs
resolve_document_node_type("Merge")
```
to
```rs
resolve_document_node_type(&DefinitionIdentifer::Network("Merge".to_string()))
```

This PR also moves
```rs
pub reference: Option<String>,
```
from `struct DocumentNodePersistentMetadata` to its descendant `struct NodeNetworkPersistentMetadata` (but only when it's a network node, not a proto node).


editor/src/messages/portfolio/document_migration.rs
```diff
@@ -1059,86 +1055,9 @@ fn migrate_node(node_id: &NodeId, node: &DocumentNode, network_path: &[NodeId],
 		document.network_interface.set_input(&InputConnector::node(*node_id, 9), old_inputs[4].clone(), network_path);
 	}
 
-	// Rename the old "Splines from Points" node to "Spline" and upgrade it to the new "Spline" node
-	if reference == "Splines from Points" {
```
adamgerhant, line 1063: These migrations are deleted since protonodes no longer have a string reference. The only way to check for an old node is to check for the old identifier (fully qualified path), but I'm not sure what that was.

Keavon: CC @TrueDoctor for thoughts on this.

editor/src/messages/portfolio/document/node_graph/document_node_definitions/document_node_derive.rs
```diff
@@ -78,48 +40,96 @@ pub(super) fn post_process_nodes(mut custom: Vec<DocumentNodeDefinition>) -> Vec
...
-					input_metadata: fields
-						.iter()
-						.map(|f| match f.widget_override {
-							RegistryWidgetOverride::None => (f.name, f.description).into(),
-							RegistryWidgetOverride::Hidden => InputMetadata::with_name_description_override(f.name, f.description, WidgetOverride::Hidden),
-							RegistryWidgetOverride::String(str) => InputMetadata::with_name_description_override(f.name, f.description, WidgetOverride::String(str.to_string())),
-							RegistryWidgetOverride::Custom(str) => InputMetadata::with_name_description_override(f.name, f.description, WidgetOverride::Custom(str.to_string())),
-						})
-						.collect(),
-					output_names: vec![output_type.to_string()],
```
adamgerhant, line 111: Removed since the output name of a protonode should not be a single random output type, as this can change and should be based on the current graph state.

editor/src/messages/portfolio/document_migration.rs
```diff
@@ -1357,51 +1279,8 @@ fn migrate_node(node_id: &NodeId, node: &DocumentNode, network_path: &[NodeId],
 		document.network_interface.set_input(&InputConnector::node(*node_id, 2), old_inputs[3].clone(), network_path);
 	}
 
-	if reference == "Flatten Vector Elements" {
```
adamgerhant, line 1360: not possible to include this migration since it relies on changing reference to not run again, which is no longer possible.

editor/src/messages/portfolio/document_migration.rs
```diff
@@ -1357,51 +1279,8 @@ fn migrate_node(node_id: &NodeId, node: &DocumentNode, network_path: &[NodeId],
...
-		document.network_interface.replace_implementation(node_id, network_path, &mut node_template);
-
-		let old_inputs = document.network_interface.replace_inputs(node_id, network_path, &mut node_template)?;
-
-		document.network_interface.set_input(&InputConnector::node(*node_id, 0), old_inputs[0].clone(), network_path);
-
-		document.network_interface.replace_reference_name(node_id, network_path, "Flatten Path".to_string());
-	}
-
-	if reference == "Remove Handles" {
```
adamgerhant, line 1371: This migration doesnt work since "Remove Handles" is not a definition

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -1427,22 +1463,17 @@ fn static_nodes() -> Vec<DocumentNodeDefinition> {
...
-						exports: vec![NodeInput::node(NodeId(2), 0)],
+						exports: vec![NodeInput::node(NodeId(1), 0)],
 						nodes: [
 							DocumentNode {
-								inputs: vec![NodeInput::scope("editor-api")],
-								implementation: DocumentNodeImplementation::ProtoNode(ProtoNodeIdentifier::new("graphene_core::ops::IntoNode<&WgpuExecutor>")),
-								..Default::default()
-							},
-							DocumentNode {
-								inputs: vec![NodeInput::import(concrete!(Table<Raster<CPU>>), 0), NodeInput::node(NodeId(0), 0)],
```
Keavon, line 1438: @TrueDoctor we decided to remove this during our (Adam and me) code review session today because we were getting a console error about it not finding `graphene_core::ops::IntoNode<&WgpuExecutor>` and we concluded that we no longer need Into nodes now that they are inserted automatically by the compiler. Can you double-check that this is a correct assumption and you don't expect it to break stuff?

editor/src/messages/portfolio/document/node_graph/document_node_definitions/document_node_derive.rs
```diff
@@ -78,48 +40,93 @@ pub(super) fn post_process_nodes(mut custom: Vec<DocumentNodeDefinition>) -> Vec
 
 		let Some(implementations) = &node_registry.get(id) else { continue };
 
-		let valid_inputs: HashSet<_> = implementations.iter().map(|(_, node_io)| node_io.call_argument.clone()).collect();
 		let first_node_io = implementations.first().map(|(_, node_io)| node_io).unwrap_or(const { &NodeIOTypes::empty() });
 
-		let input_type = if valid_inputs.len() > 1 { &const { generic!(D) } } else { &first_node_io.call_argument };
```
Keavon, line 84: @TrueDoctor we removed this and moved its `else` case down directly into the `DocumentNode { ..., call_argument: ..., ... }` below. Please confirm this doesn't break any current functionality.

editor/src/messages/portfolio/document/node_graph/document_node_definitions/document_node_derive.rs
```diff
@@ -78,48 +40,93 @@ pub(super) fn post_process_nodes(mut custom: Vec<DocumentNodeDefinition>) -> Vec
...
-					locked: false,
-					..Default::default()
+		definitions_map.insert(
+			identifier,
+			DocumentNodeDefinition {
+				identifier: display_name,
+				node_template: NodeTemplate {
+					document_node: DocumentNode {
+						inputs,
+						call_argument: first_node_io.call_argument.clone(),
```
Keavon: This is where we moved it to.

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -75,14 +129,14 @@ pub struct DocumentNodeDefinition {
...
+// We use the once_cell to use the document node definitions throughout the editor without passing a reference
+// TODO: If dynamic node library is required, use a Mutex as well
+static DOCUMENT_NODE_TYPES: once_cell::sync::Lazy<HashMap<DefinitionIdentifier, DocumentNodeDefinition>> = once_cell::sync::Lazy::new(document_node_definitions);
 
-// TODO: Dynamic node library
 /// Defines the "signature" or "header file"-like metadata for the document nodes, but not the implementation (which is defined in the node registry).
 /// The [`DocumentNode`] is the instance while these [`DocumentNodeDefinition`]s are the "classes" or "blueprints" from which the instances are built.
-fn static_nodes() -> Vec<DocumentNodeDefinition> {
+/// Only the position can be set for protonodes within a definition. The rest of the metadata comes from the node macro in NODE_METADATA
+fn document_node_definitions() -> HashMap<DefinitionIdentifier, DocumentNodeDefinition> {
```
Keavon, line 139: @TrueDoctor please take a look at this change from `Vec` to `HashMap` and confirm that it's not an issue. Thanks.

Keavon:
Marking as a draft while a few issues with crashes on document opening are happening due to some weirdness with migrations. @adamgerhant has the files to reproduce this with in our chat.

Also, please help by fixing the tests that are failing, thanks!

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
TrueDoctor: Well, instead of concluding this, you could also have tested the node to see that this into node is still very much needed. The into node insertion only works for proto nodes, not nodes defined as document networks
image

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
TrueDoctor: That should be fine

editor/src/messages/portfolio/document/node_graph/document_node_definitions/document_node_derive.rs
TrueDoctor: Please revert this, the original code was behaving correctly. The new version will break as soon as we have multiple valid call arguments, but the current logic was correct and more future-proof

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
```diff
@@ -156,51 +160,65 @@ impl<'a> ModifyInputsContext<'a> {
 	pub fn insert_vector(&mut self, subpaths: Vec<Subpath<PointId>>, layer: LayerNodeIdentifier, include_transform: bool, include_fill: bool, include_stroke: bool) {
 		let vector = Table::new_from_element(Vector::from_subpaths(subpaths, true));
 
-		let shape = resolve_document_node_type("Path")
+		let shape = resolve_document_node_type(&DefinitionIdentifier::Network("Path".into()))
```
TrueDoctor: Would it make sense to add a utility `resolve_document_network("Path")` function to make these changes a bit less verbose?

TrueDoctor: We could also implement `From<&str>` and `From<ProtonodeIdentifer>` for `DefinitionIdentifier` Then the `resolve_document_network_type()` function could accept `T: Into<DocumentNodeIdentifier>` thus eliminating a bunch of changes from the diff and keeping things shorter.

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
```diff
@@ -325,7 +342,9 @@ impl<'a> ModifyInputsContext<'a> {
 		let backup_color_index = 2;
 		let backup_gradient_index = 3;
 
-		let Some(fill_node_id) = self.existing_node_id("Fill", true) else { return };
+		let Some(fill_node_id) = self.existing_node_id(&DefinitionIdentifier::ProtoNode(graphene_std::vector_nodes::fill::IDENTIFIER), true) else {
```
TrueDoctor: Same here, `existing_node_id` could accept `impl Into<DefinitionIdentifier>`

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
Keavon: I intentionally removed the `From` implementation because it was a footgun and allowed me to make several bugs slip through just while tweaking a bit of code during code review. This way it provides the necessary explicitness. I'm in favor of the separate constructor function in your first message, though.

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -54,10 +55,63 @@ impl NodePropertiesContext<'_> {
...
+			DefinitionIdentifier::Network(name) => name.clone(),
+			DefinitionIdentifier::ProtoNode(proto_node_identifier) => registry::NODE_METADATA
+				.lock()
+				.unwrap()
+				.get(proto_node_identifier)
+				.map(|metadata| metadata.display_name.to_string())
+				.unwrap_or_else(|| {
+					let mut last_segment = proto_node_identifier.as_str().split("::").last().unwrap_or_default().to_string();
+					last_segment = last_segment.strip_suffix("Node").unwrap_or(&last_segment).to_string();
+					last_segment
```
TrueDoctor, line 80: This branch should only be taken if the node is not registered by the node macro right? I don't think we have any nodes which are not registered by the node macro, so this should never be taken. Also if we do take it, the name would be weird, `FlattenPath` instead of `Flatten Path`

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -54,10 +55,63 @@ impl NodePropertiesContext<'_> {
...
+
+impl From<Value> for DefinitionIdentifier {
+	fn from(value: Value) -> Self {
+		match value {
+			Value::Object(mut map) => {
+				let ty = map.remove("type").unwrap().as_str().unwrap().to_owned();
+
+				match ty.as_ref() {
+					"Network" => {
+						let data = map.remove("data").unwrap().as_str().unwrap().to_owned();
```
TrueDoctor, line 94: Does this rely on serde json internals for it's wire format?

TrueDoctor: If so, are they documented somewhere / do we have stability guarantees for this?

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -2745,7 +2687,7 @@ impl DocumentNodeDefinition {
 		input_override.into_iter().enumerate().for_each(|(index, input_override)| {
 			if let Some(input_override) = input_override {
 				// Only value inputs can be overridden, since node inputs change graph structure and must be handled by the network interface
-				// assert!(matches!(input_override, NodeInput::Value { .. }), "Only value inputs are supported for input overrides");
+				debug_assert!(input_override.as_node().is_none(), "Node inputs are not supported in input overrides");
```
TrueDoctor: Why was this made a `debug_assert!`? This should happen pretty instantly so it is not some weird once in a lifetime corner case. And since our ci builds in release mode, we wouldn't be able to catch this during the pr review

editor/src/messages/portfolio/document/graph_operation/utility_types.rs
TrueDoctor: We can have the same explicitness with less code noise by adding a `Network(pub &'static str)` wrapper struct. Then it would become `resolve_document_node_type(Network("Path"))` which is still significantly less noisy, and I'd argue more explicit / consistent than having two separate functions

frontend/src/components/panels/Layers.svelte
```diff
@@ -644,15 +644,15 @@
 							{@html $nodeGraph.thumbnails.get(listing.entry.id)}
 						{/if}
 					</div>
-					{#if listing.entry.reference === "Artboard"}
+					{#if listing.entry.reference.type === "Network" && listing.entry.reference.data === "Artboard"}
```
TrueDoctor, line 647: I guess the same question about serde_json enum representation applies

editor/src/messages/portfolio/document_migration.rs
```diff
@@ -1212,8 +1210,12 @@ fn migrate_node(node_id: &NodeId, node: &DocumentNode, network_path: &[NodeId],
 	}
 
 	// Upgrade Sine, Cosine, and Tangent nodes to include a boolean input for whether the output should be in radians, which was previously the only option but is now not the default
-	if (reference == "Sine" || reference == "Cosine" || reference == "Tangent") && inputs_count == 1 {
-		let mut node_template = resolve_document_node_type(reference)?.default_node_template();
+	if (reference == DefinitionIdentifier::ProtoNode(graphene_std::math_nodes::sine::IDENTIFIER)
+		|| reference == DefinitionIdentifier::ProtoNode(graphene_std::math_nodes::cosine::IDENTIFIER)
+		|| reference == DefinitionIdentifier::ProtoNode(graphene_std::math_nodes::tangent::IDENTIFIER))
+		&& inputs_count == 1
```
TrueDoctor: I guess I would use 
```rust
use graphene_std::math_nodes;
if let DefinitionIdentifier::ProtoNode(proto_node_id) = reference
  && inputs_count == 1
   && (proto_node_id == math_nodes::sine::IDENTIFIER 
         | proto_node_id == math_nodes::cosine::IDENTIFIER 
         | proto_node_id == math_nodes::tangent::IDENTIFIER )
```

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
adamgerhant: Yes, this branch is for nodes with #[node_macro::node(skip_impl)] like the monitor node. It could have a space added before each capital letter

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
adamgerhant: #[serde(tag = "type", content = "data")] is used to serialize the DefinitionIdentifer to a Javascript object, and this converts from the serde_json version. There is some documentation here: https://serde.rs/enum-representations.html. Given the simplicity of this enum I think it will remain stable.

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
adamgerhant: I removed this assert since some of the tooling generates multiple nodes at once before they are inserted into the network, and thus needs to connect them without the `set_input` function

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
adamgerhant: Would it be possible for the node macro to add metadata to `NODE_METADATA` for nodes with `#[node_macro::node(skip_impl)]` so its purpose is to *only* not add to the registry?

TrueDoctor: That is exactly how the node macro behaves, the metadata is generated regardless of whether the `skip_impl` is used. Again the only way I see for this branch being taken is if a node is not registered using the node macro

Keavon: I was confident that was true, but then we looked into it and it seems (maybe due to a logic bug?) that it's surprisingly not actually so. Can you please double-check @adamgerhant's conclusion about this, and maybe suggest how it ought to be fixed if what I'm conveying is indeed accurate?

TrueDoctor: I mean you could have trusted my comment or double-checked by just looking at the code in `codegen.rs` but by design the `skip_impl` attribute only skips the addition of implementation and specifically not the metadata registration / implementation + codgen of the node so you add your own implementations in the node registry.
So adams first suggestion is not relevant since this entire branch is dead code. The only nodes I am aware of which are not generated with the macro are ones which would never be accessible to the user.
Regarding the second suggestion, that is how this feature has worked since I have implemented it

Keavon: I'll defer to @adamgerhant for further discussion here because I am not familiar with the specifics.

adamgerhant: I thought metadata was registered with the call to [`register_metadata`](https://github.com/GraphiteEditor/Graphite/blob/master/node-graph/node-macro/src/codegen.rs#L704), which is skipped by the early return statement on line 593. I verified this by printing out all nodes in `NODE_METADATA`, and I could not find the Monitor, Identity, or Into nodes in that HashMap.

TrueDoctor: The early return is unrelated, you likely tested an outdated git version where the monitor node was not yet defined via the node macro. If you tested it again, I'm 99% sure you'll find the monitor node in the metadata list

PR #3460, opened 2025-12-10
Improve wire path in progress

adamgerhant:
Improves various data types and code organization for translation between the editor and frontend. Fixes the in progress wire to stay vertical if dragged between layers.

Depends on #3451

https://github.com/user-attachments/assets/6f3f21f7-2ba7-47df-b036-4692917955c9


Keavon:
- [ ] Needs to not keep the state from the last time you dragged a wire, needs to reset it when you start dragging every time
- [ ] Needs to not become a thick wire if the wire is coming from a layer's top output that has more than one wire output

PR #3461, opened 2025-12-10
Desktop: Text clipboard support

timon-schelling:


timon-schelling:
@Keavon Did QA testing and comparison with master, seems to work fine now.
I suggest you spend some extra time on the two new TS functions that deal with current selection.

PR #3465, opened 2025-12-12
Add nudge-scaling for path tool for ALT and ALT + CTRL keys 

### LINKED ISSUE #2374: Nudge-scaling of points with the Path tool
With a group of points selected, we'd like to be able to use nudge-scaling to resize their bounding box.

Nudge scaling is when, in the Select tool, you hold <kbd>Alt</kbd> when nudging with the arrow keys to resize by nudging only the bottom and right (or instead, top and left if <kbd>Ctrl</kbd> is also pressed) edge while keeping the opposite edges of the bounding box stationary, thereby resizing the shape.

This feature should create a bounding box of all selected points, in document space coordinates. Points at the very top and left of the bounds should be unaffected, and points at the very bottom and left should be nudged the full amount, with intermediate points scaling proportionally. This would be equivalent to scaling them with GRS's <kbd>S</kbd> if you were able to set the pivot to the top left of the combined points' bounding box. Handles of points should behave as <kbd>S</kbd> scaling does presently.

nng98:
### Description 

This PR implements nudge-scaling for the ALT and ALT + CTRL keys for the path tool with the arrow keys. When ALT is held, using the arrow keys nudges only the bottom right points of the selected points, and when ALT and CTRL are held , only the top left points of the selected points are nudged the full amount. In both cases, points outside of these regions are unaffected. 

### Code Changes 

In `editor/src/messages/input_mapper/input_mappings.rs`, added input entries to send messages for the path tool whenever ALT or ALT + CTRL is held as a modifier to the 'PathToolMessage::NudgeSelectedPoints' message. Note that this also applies to when combinations of arrow keys are held (such as when ArrowRight and ArrowUp are held, it will produce a diagonal nudge to the top right).

In `editor/src/messages/tool/common_functionality/shape_editor.rs`, we changed the `move_selected_points_and_segments` function to have additional parameters for whether it should only move the top left or only move the bottom right points, called `bot_right` and `top_left` respectively. Additionally, we changed the function's implementation to, if one of these parameters is passed, only adjust points that are in the selected regions by continuing in the loop where points are moved if the point is outside the desired region.  

In `editor/src/messages/tool/tool_message/path_tool.rs` we changed the `PathToolMessage` enum to accept booleans `nudge_alt, nudge_ctrl` for `NudgeSelectedPoints` message. Additionally, we changed the handler for the 'PathToolMessage::NudgeSelectedPoints' to call `shape_editor.move_selected_points_and_segments` with additional parameters passed for the booleans, as described in the changes to `shape_editor.rs`.

### Demo Video 

https://github.com/user-attachments/assets/10fdb437-d2eb-462a-9cb5-439bfb31ae4f

Closes #2374


timon-schelling:
Thanks for contributing. Did you use an LLM to help on this PR? If yes specific what parts are generated?

nng98:
> Thanks for contributing. Did you use an LLM to help on this PR? If yes specific what parts are generated?

Thank you! I used an LLM to help with the change I made to `editor/src/messages/tool/common_functionality/shape_editor.rs`, specifically, I used it to help calculate the location of the center of the selected points. Otherwise, none of the other changes involved the use of LLMs.

PR #3472, opened 2025-12-13
Desktop: Forward and Backward mouse button support

timon-schelling:


timon-schelling:
Tested on Linux, Mac and Windows

PR #3474, opened 2025-12-14
Implement proper fill rendering for vector meshes

### LINKED ISSUE #3378: Render vector mesh fills correctly as separate subpaths

liunicholas6:
Closes #3378 
Didn't want to have to deal with working off of @iamthenoname repo so I made a new set of commits.
@0HyperCube Thanks for all the feedback!

### DIFF
--- a/libraries/rawkit/src/decoder/arw1.rs
+++ b/libraries/rawkit/src/decoder/arw1.rs
@@ -1,7 +1,7 @@
 use crate::tiff::Ifd;
 use crate::tiff::file::TiffRead;
 use crate::tiff::tags::SonyDataOffset;
-use crate::{RawImage, SubtractBlack, Transform};
+use crate::{OrientationValue, RawImage, SubtractBlack};
 use bitstream_io::{BE, BitRead, BitReader, Endianness};
 use std::io::{Read, Seek};
 
@@ -25,7 +25,7 @@ pub fn decode_a100<R: Read + Seek>(ifd: Ifd, file: &mut TiffRead<R>) -> RawImage
 		#[allow(unreachable_code)]
 		maximum: (1 << 12) - 1,
 		black: SubtractBlack::None,
-		transform: Transform::Horizontal,
+		orientation: OrientationValue::Horizontal,
 		camera_model: None,
 		camera_white_balance: None,
 		white_balance: None,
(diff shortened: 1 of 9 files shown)


Keavon:
!build

node-graph/libraries/vector-types/Cargo.toml
```diff
@@ -31,3 +31,4 @@ tinyvec = { workspace = true }
 
 # Optional workspace dependencies
 serde = { workspace = true, optional = true }
+fixedbitset = "0.5.7"
```
liunicholas6, line 34: Not sure what the process is for adding dependencies -- I saw fixedbitset is a dependency of petgraph so I thought it would probably be OK to bring in; I can easily swap it out for a Vec::<bool> or similar

liunicholas6:
Update -- looks like some fills still aren't working; probably going to take a lot of debug logging to figure out. Will take another look tomorrow

node-graph/libraries/vector-types/src/vector/misc.rs
```diff
@@ -246,6 +246,74 @@ pub fn pathseg_abs_diff_eq(seg1: PathSeg, seg2: PathSeg, max_abs_diff: f64) -> b
...
+	fn tangent_at_end(&self) -> DVec2 {
+		self.tangent_at_start()
+	}
+}
+
+impl Tangent for kurbo::QuadBez {
+	fn tangent_at(&self, t: f64) -> DVec2 {
+		let deriv = self.deriv();
+		let tangent = deriv.eval(t);
+		DVec2 { x: tangent.x, y: tangent.y }
```
0HyperCube: If the handle is placed on top of the anchor then this will be zero.

For reference, here is my impl of the tangent at start which attempts to avoid reporting zero length tangents:
https://github.com/GraphiteEditor/Graphite/blob/06749ecea52f3085903bca5a0b242f3a83f9538c/node-graph/libraries/vector-types/src/vector/algorithms/bezpath_algorithms.rs#L704-L724

liunicholas6: Yep -- I special-cased tangent_at_start and tangent_at_end for each type of segment but then didn't actually call into them on the match case.

I didn't think of the "find the first non-overlapping control point" strategy though; that's a good call that I addressed on revision.

node-graph/libraries/vector-types/Cargo.toml
Keavon: @TrueDoctor can you give your thoughts on this please?

Keavon:
!build

PR #3476, opened 2025-12-15
Simplify history transaction system messages

adamgerhant:
Comments the transaction system and remove unnecessary transaction messages.
Simplify the expose input message


Keavon:
This currently crashes Graphite when interacting with the artwork in basically any way (moving, drawing, etc.).

PR #3485, opened 2025-12-17
Fix regular polygon radius using inner radius correctly

### LINKED ISSUE #3484: Polygon: Use inner instead of outer radius
When creating a polygon, I often want to be able to create it with a radius based on the middle segment. For example, with hexagons:

### What the radius currently sets

Image

### What I want to set as the radius

Image

I don't know how this should be implemented. Should it be a toggle? It could be a simple custom component with a relatively simple trigonomery node also. Should it be implemented directly within the editor?

I currently don't realize what is the scope of these nodes. Should they be general? Should the provide the bare minimum to be composable into any wanted node?

Parshv-sudo:
### Description
Fixes an issue where the regular polygon node treated the radius input as an outer radius.
The radius is now interpreted as the inner radius (apothem) and correctly converted to the outer radius before generating the polygon.

###Related Issue
Fixes #3484

### Changes
- Correctly convert inner radius (apothem) to outer radius (circumradius) for regular polygons
- Remove incorrect radius scaling

### Testing
- `cargo build`
- Mathematical verification of apothem → circumradius conversion

### DIFF
--- a/node-graph/nodes/vector/src/generator_nodes.rs
+++ b/node-graph/nodes/vector/src/generator_nodes.rs
@@ -159,8 +159,9 @@ fn regular_polygon<T: AsU64>(
 	radius: f64,
 ) -> Table<Vector> {
 	let points = sides.as_u64();
-	let radius: f64 = radius * 2.;
-	Table::new_from_element(Vector::from_subpath(subpath::Subpath::new_regular_polygon(DVec2::splat(-radius), points, radius)))
+	let sides = points as f64;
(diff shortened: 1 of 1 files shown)


Keavon:
Thank you for your attempted contribution. However, this appears to be AI-generated and will not be accepted.

Parshv-sudo:
Hi, actually I am new to contributing so I don't know what to write so that's the only thing I used for AI

PR #3489, opened 2025-12-17
Use glob syntax for Cargo.toml

TrueDoctor:


Cargo.toml
```diff
@@ -12,69 +12,31 @@ members = [
...
-	"tools/crate-hierarchy-viz",
-	"tools/editor-message-tree",
-	"tools/node-docs",
+	"tools/crate-hierarchy-viz"
 ]
 default-members = [
 	"editor",
 	"frontend/wasm",
 	"libraries/dyn-any",
 	"libraries/path-bool",
```
timon-schelling, line 52: Why remove these without adding the glob pattern?

TrueDoctor: I think the reason was that all of these are dependencies of other crates, so I don't think it is necessary to have them be part of the default members. This also potentially avoids building these sub dependencies with different feature sets

PR #3482, opened 2025-12-16
Fixes the Path tool snapping to the X or Y axis with Shift held for handles (#2969)

Thunder-Blaze:
### Description
- Fixes the broken snap to X and Y axis logic when shift key is held

### Issue
Closes #2969

### Recording

https://github.com/user-attachments/assets/649386bc-39fd-49dd-ba08-a1e7b9814940

PR #3491, opened 2025-12-17
Adds a Toggle for Inner Radius in Polygon tool (#3484)

### LINKED ISSUE #3484: Polygon: Use inner instead of outer radius

Thunder-Blaze:
### Description
- Adds a Toggle to switch between inner and outer radius in a polygon

### Issue
Closes #3484 

### Notes For Maintainer
- Umm, I wasn't sure if it was better to overwrite the default radius with the improvised one or create a new var. For now, at most places I created a variable as `effective_radius`, let me know if I go with overwriting approach
- Also, the toggle could use a better name...

### Preview

https://github.com/user-attachments/assets/9f061df2-8bd0-4e0c-8772-a67674be945b

### DIFF
--- a/editor/src/messages/tool/common_functionality/gizmos/shape_gizmos/number_of_points_dial.rs
+++ b/editor/src/messages/tool/common_functionality/gizmos/shape_gizmos/number_of_points_dial.rs
@@ -13,6 +13,7 @@ use crate::messages::tool::common_functionality::shapes::shape_utility::{extract
 use glam::{DAffine2, DVec2};
 use graph_craft::document::NodeInput;
 use graph_craft::document::value::TaggedValue;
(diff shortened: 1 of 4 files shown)


Thunder-Blaze:
Umm hey @Keavon
Can you review this PR and #3482 ?
Is there something I need to fix or any reason for which it still haven't gotten reviewed ?

PR #3496, opened 2025-12-19
fix : Sync Make Path Editable state between path tool and menu bar

Annonnymmousss:
## Summary
Fix an issue where the “Make Path Editable” menu item could get stuck in a disabled state while the Path tool correctly allowed the action.

## Problem
The Menu Bar and Path tool each tracked their own `make_path_editable_is_allowed` state. The Path tool updated correctly, but the Menu Bar was never notified, leaving the menu item out of sync.

## Solution
Add a message-based update flow so the Path tool notifies the Menu Bar whenever the allowed state changes. The Menu Bar updates its state and redraws, keeping both UI entry points in sync.

## Code todo list
- https://discord.com/channels/731730685944922173/881073965047636018/1415971652470181984

## Demo
https://github.com/user-attachments/assets/8674eae2-6831-4fc0-aa39-49c8f4692742

### DIFF
--- a/editor/src/messages/menu_bar/menu_bar_message.rs
+++ b/editor/src/messages/menu_bar/menu_bar_message.rs
@@ -5,4 +5,5 @@ use crate::messages::prelude::*;
 pub enum MenuBarMessage {
 	// Messages
 	SendLayout,
+	UpdateMakePathEditableState { allowed: bool },
 }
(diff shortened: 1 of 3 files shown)


Annonnymmousss:
#3497 supersedes it.

PR #3497, opened 2025-12-19
Fix disabled state of Layer > Make Path Editable menu action getting out of sync

Annonnymmousss:
fix the issue by updating the state of menu bar as the state of path tool changes

code todo - https://discord.com/channels/731730685944922173/881073965047636018/1415971652470181984

https://github.com/user-attachments/assets/86a8972a-9e98-4c3c-9e08-5cf8430c18a6

### DIFF
--- a/editor/src/dispatcher.rs
+++ b/editor/src/dispatcher.rs
@@ -2,6 +2,7 @@ use crate::messages::debug::utility_types::MessageLoggingVerbosity;
 use crate::messages::defer::DeferMessageContext;
 use crate::messages::dialog::DialogMessageContext;
 use crate::messages::layout::layout_message_handler::LayoutMessageContext;
+use crate::messages::preferences::preferences_message_handler::PreferencesMessageContext;
 use crate::messages::prelude::*;
 use crate::messages::tool::common_functionality::utility_functions::make_path_editable_is_allowed;
 
@@ -277,7 +278,11 @@ impl Dispatcher {
 					menu_bar_message_handler.process_message(message, &mut queue, ());
 				}
 				Message::Preferences(message) => {
-					self.message_handlers.preferences_message_handler.process_message(message, &mut queue, ());
+					let context = PreferencesMessageContext {
+						tool_message_handler: &self.message_handlers.tool_message_handler,
+					};
+
+					self.message_handlers.preferences_message_handler.process_message(message, &mut queue, context);
 				}
 				Message::Tool(message) => {
 					let Some(document_id) = self.message_handlers.portfolio_message_handler.active_document_id() else {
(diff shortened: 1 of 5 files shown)


editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1575,6 +1575,7 @@ impl Fsm for PathToolFsmState {
 
 				shape_editor.set_selected_layers(target_layers);
 
+				responses.add(MenuBarMessage::SendLayout);
```
Keavon: In both of these calls, `MenuBarMessage::SendLayout` is expensive. Can you find a way to check if a re-send is necessary before making the call, so it's conditional instead of always?

Annonnymmousss: ya sure 
we can add the condition whether the state is changed or not before calling these that would be better.

Keavon: Yes, that would be ideal.

Annonnymmousss: i have added the check is it fine now?

Keavon: I think you made a mistake in the lower of the two code change areas?

Annonnymmousss: ya shit i forgot to apply the check

Annonnymmousss: also SelectionChanged does not need to store the state so removed that too

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3125,8 +3129,15 @@ impl Fsm for PathToolFsmState {
 					colinear,
 				};
 
+				let new_state = make_path_editable_is_allowed(&mut document.network_interface).is_some();
```
Keavon: This looks redundant with the line that's two lines below.

Annonnymmousss: it it seems okay now?

Keavon: Yes.

Keavon:
!build

Keavon:
You will need to set up your code editor to auto format on save to ensure you never push code with noncompliant formatting. Please resolve that on your end and push the change so CI passes. Thanks.

Annonnymmousss:
default formatter only or any specific style?

Keavon:
See https://graphite.art/volunteer/guide/project-setup/#development-tooling

Annonnymmousss:
yeah done will keep this is mind

PR #3499, opened 2025-12-20
Fix the Gradient tool's linear/radial type toggle resetting the gradient color stops

Annonnymmousss:
fix both the syncing issue and change reverting issue
code todo - https://discord.com/channels/731730685944922173/881073965047636018/1406600123571634250

https://github.com/user-attachments/assets/8779ec6c-9a4e-4562-8ec7-7dbd30cccfb4

### DIFF
--- a/.github/workflows/build-linux-bundle.yml
+++ b/.github/workflows/build-linux-bundle.yml
@@ -1,10 +1,16 @@
 name: Build Linux Bundle
 
 on:
-  workflow_dispatch: {}
   push:
     branches:
       - master
+  workflow_dispatch:
+    inputs:
+      push_to_cache:
+        description: "Push to Nix Cache"
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build:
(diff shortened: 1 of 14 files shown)


Annonnymmousss:
it creates another bug ill solve it.

editor/src/messages/portfolio/document/node_graph/node_properties.rs
```diff
@@ -1951,35 +1951,35 @@ pub(crate) fn fill_properties(node_id: NodeId, context: &mut NodePropertiesConte
...
+		let entries = [GradientType::Linear, GradientType::Radial]
+			.iter()
+			.map(|&grad_type| {
+				let gradient = gradient_for_closure.clone();
+				RadioEntryData::new(format!("{:?}", grad_type))
+					.label(format!("{:?}", grad_type))
+					.on_update(move |_| {
+						let mut new_gradient = gradient.clone();
+						new_gradient.gradient_type = grad_type;
+						Message::Batched {
```
Keavon: This makes an un-idiomatic change where the call to the `on_update` function (which receives back a `Message`) is now inlined for the message creation, for the sake of creating two batched messages. You can and should still call `on_update` and use its return value as one of the two messages that are given to the array for the batched message return value.

Annonnymmousss: done the changes.

timon-schelling:
@Keavon I think the title should be changed

Keavon:
Oh yes, I forgot that, thanks for noticing and cancelling the pending merge.

PR #3527, opened 2025-12-24
Fix localgrid grabbing

Annonnymmousss:
fix the global state showcase when grabbed even after rotating with an angle
code todo - https://discord.com/channels/731730685944922173/881073965047636018/1399533117441835080

https://github.com/user-attachments/assets/3eb7df74-f68c-4f1b-b407-13ae65fbfcc1

### DIFF
--- a/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
+++ b/editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
@@ -250,8 +250,8 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
 
 						let typed_string = (!self.typing.digits.is_empty() && self.transform_operation.can_begin_typing()).then(|| self.typing.string.clone());
 						// A transformation to the local space of the transformation from the document
-						let document_to_local = glam::DMat2::from_cols(self.state.local_transform_axes[0], self.state.local_transform_axes[1]).inverse();
-						let transform_local = document_to_local * document_to_viewport.matrix2.inverse() * translation_viewport;
+						let viewport_to_local = glam::DMat2::from_cols(self.state.local_transform_axes[0], self.state.local_transform_axes[1]).inverse();
+						let transform_local = viewport_to_local * translation_viewport / document_to_viewport.matrix2.y_axis.length();
 						overlay_context.translation_box(transform_local, quad, typed_string);
 					}
 					TransformOperation::Scaling(scale) => {


editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -537,10 +565,17 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
+							let document_space_transform = self.initial_transform * document_to_viewport.inverse();
+							let delta_pos = if self.state.is_transforming_in_local_space {
+								let [local_x_axis, local_y_axis] = self.state.local_transform_axes;
+								let local_components = DVec2::new(viewport_delta.dot(local_x_axis), viewport_delta.dot(local_y_axis));
+								document_space_transform.matrix2 * local_components
+							} else {
+								document_space_transform.transform_vector2(viewport_delta)
+							};
+							let scale = document_to_viewport.y_axis.length();
+							let delta_scaled = (if self.slow { delta_pos / SLOWING_DIVISOR } else { delta_pos }) / scale;
```
0HyperCube: Along with adding significant complexity here, you have also broken the «local x-axis constraint» feature (press <kbd>G</kbd> to grab then press <kbd>X</kbd> twice.

https://github.com/user-attachments/assets/2a427529-5721-4cce-b9cc-0ccff8896b34

I don't see why there should be any changes to the functionality of the system, since the issue concerned only changing the overlays.

Annonnymmousss: letme try this again

Annonnymmousss:
this solved the constraint breakage but now every scale is getting calculated with reference to global axis only 
can you help me?

0HyperCube:
> scale is getting calculated with reference to global axis only

The scale by pressing key <kbd>S</kbd> then <kbd>X</kbd> will scale in viewport space. Pressing <kbd>X</kbd> again will scale in local space. This is the intended behaviour.

Annonnymmousss:
image


i meant this scale the length. It is calculated with reference to global axis(

0HyperCube:
This is the issue in #3341 that you are trying to solve in this PR? I am confused since you linked an unrelated discord post…

I assume you are referring to the blue numbers in your screenshot. These are drawn by the `overlay_context.translation_box` function. If you look at the implementation, you will see that it uses the first argument, `translation`, to determine the numbers that are drawn. To change these numbers, you can transform the vector:
```rs
// A transformation to the local space of the transformation from the document
let document_to_local = glam::DMat2::from_cols(self.state.local_transform_axes[0], self.state.local_transform_axes[1]).inverse();
let transform_local = document_to_local * document_to_viewport.matrix2.inverse() * translation_viewport;
overlay_context.translation_box(transform_local, quad, typed_string);
```

Annonnymmousss:
yeah sorry for the confusion created due to terminologies ill keep this in mind for my further contributions.
and ill be working on it..

Annonnymmousss:
https://github.com/user-attachments/assets/b126fbcb-e1a0-45cb-83b0-82619ab2befe

I have made some changes i guess it fixes the issue. Please review it and let me know if any breakage is caused or any suggestions.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -234,15 +234,25 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
 						responses.add(SelectToolMessage::PivotShift {
 							offset: Some(translation_viewport),
 							flush: false,
 						});
 
 						let typed_string = (!self.typing.digits.is_empty() && self.transform_operation.can_begin_typing()).then(|| self.typing.string.clone());
-						overlay_context.translation_box(translation_viewport / document_to_viewport.matrix2.y_axis.length(), quad, typed_string);
+						// A transformation to the local space of the transformation from the document
+						let document_to_local = glam::DMat2::from_cols(self.state.local_transform_axes[0], self.state.local_transform_axes[1]).inverse();
+						let transform_local = document_to_local * document_to_viewport.matrix2.inverse() * translation_viewport;
```
0HyperCube:
```suggestion
						let viewport_to_local = glam::DMat2::from_cols(self.state.local_transform_axes[0], self.state.local_transform_axes[1]).inverse();
						let transform_local = viewport_to_local * translation_viewport / document_to_viewport.matrix2.y_axis.length();

```
I scuffed up the transformations sorry. The `local_transform_axes` are actually in viewport space.

0HyperCube (approved):
Looks good to me; thanks for your work on these changes.

I'm a bit unsure as to if the local space should be shown when transforming on the global X-axis (<kbd>G</kbd> then <kbd>X</kbd> once). It also breaks the measuring number in that case when you type some digits to choose an exact translation. This is because the exact typed number is used to avoid rounding errors.

Annonnymmousss:
okay if that would be the case I will open another PR.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
Annonnymmousss: ya i will update these changes

Annonnymmousss:
@Keavon is it ready for merge or any changes are required?

Ayush2k02:
LGTM !

Keavon:
Finally reviewing this, sorry for the wait.

From what I am seeing, this makes some changes related to what is desired but does not actually fix the exact issue. I'm a bit confused looking at the PR description and seeing what looks like the wrong #code-todo-list Discord message linked, or maybe something got mixed up there between what I posted and what got interpreted.

The only desired change that I know of is the need to show the translation in local space upon hitting the second occurrence of <kbd>X</kbd> to go from unconstrained to X-constrained to X-local-constrained space, where it currently still shows the X and Y components in global space rather than the single X component in local space:

https://github.com/user-attachments/assets/67115ed8-e7aa-426d-b1ff-b9579e77eaf6

And then as we see at the end, if I type `100`, it incorrectly shows "100" in both the X and Y components, which further shows why this is wrong and we need it to show that in local space. The equivalent of this for <kbd>S</kbd> works flawlessly:

https://github.com/user-attachments/assets/b0fd2b53-4480-4316-8d43-929caf7b5b10

I actually think the best option is to close this and let you reference this implementation to fix the problem described herein, since the current PR relates to this but (undesirably) changes other functionality while not changing this desired issue. So starting over, but using this code for reference, seems simplest to me. Feel free to reference this comment in the new PR. Thanks for your understanding.

Keavon:
That will be issue #3341.

0HyperCube:
@Keavon I don't think you have tested the correct branch. I have included (for your reference) the current behaviour of this branch (commit 49b9cd821024d0ccf7ca75cbf97f680e86fb7bbd) when pressing <kbd>G</kbd> then <kbd>X</kbd> twice then typing `100`.

https://github.com/user-attachments/assets/4005ced8-1b9e-43b5-8731-f5e8761a2fee

It is somewhat disappointing to see that you cannot even test the correct commit after two months.

Keavon:
To be clear, my videos are from master, not this branch. They are part of describing the desired behavior, which after writing about, I discovered has an existing description in #3341.

0HyperCube:
My sincerest apologies for this misunderstanding. I was attempting to demonstrate that (as far as I can tell) the desired behaviour has been correctly implemented. As per my video above, the local axis grabbing shows in the local space of the layer and typing values for the translation shows in local space.

Therefore I struggle to comprehend why you have said that this PR "changes other functionality while not changing this desired issue".

I'm sorry for my ignorance on this matter and will refrain from commenting further.

Issue #3519, opened 2025-12-22
Joining paths with the same PointIds doesn't work


0HyperCube:
## To reproduce:
- Draw two lines with the **line tool**
- Select them both in the layer panel (with <kbd>shift</kbd> + click)
- Go to path tool
- Select one point from the first line and one point from the second line
- Attempt to join with <kbd>Ctrl</kbd> + <kbd>J</kbd>

## Logic
The line tool always produces `PointId(1)` and `PointId(2)`, so when merging the two layers, the point ids of the second line must change to avoid conflict. This causes an issue with the <kbd>Ctrl</kbd>+<kbd>J</kbd> functionality since it assumes that the `PointId`s do not change.

`PointId`s are only guaranteed to be unique within the `Vector<Upstream>` because same path node must always output the same IDs (since it is a pure function with no input) and you can use the output of the path node several times. 

Therefore the Ctrl+J `shape_editor.close_selected_path` must deal with the fact that the `PointId`s it holds may change after running `merge_layers`.

Code for joining layers and adding new segment (currently mistakenly assumes that the `PointId` will not change after merging):
https://github.com/GraphiteEditor/Graphite/blob/9d26c040e977990b9fc0755933691e4619196f41/editor/src/messages/tool/common_functionality/shape_editor.rs#L461-L470

Code that resolves the new ID when there are two conflicting point ids:
https://github.com/GraphiteEditor/Graphite/blob/9d26c040e977990b9fc0755933691e4619196f41/node-graph/nodes/vector/src/vector_nodes.rs#L1247-L1251

## Reproduction video
https://github.com/user-attachments/assets/1efc75fd-da0b-48f2-9088-471c9d517a46

Annonnymmousss:
i have a doubt what if the user is again using cmd+j functionality at the same two points of different layers in which the segment is created will there be another segment between those points or only one segment should be allowed?

0HyperCube:
@Annonnymmousss <kbd>Ctrl</kbd> + <kbd>J</kbd> when two points are selected should always add an additional segment, regardless of any other state. This is the currently implemented behaviour.
You can see this functionality when you select two points on the same layer:

https://github.com/user-attachments/assets/9c5d3157-82bf-426a-92e9-9259329dd0f6

The only incorrect behaviour is described in detail in the reproduction steps.

Annonnymmousss:
okay got it, I will look towards this issue.

PR #3532, opened 2025-12-26
Fix joining path same pt Id

Annonnymmousss:
im still not getting how to create a segment in the first cmd + J only which is merging the layer but this issue is resolved..

https://github.com/user-attachments/assets/9e00ea57-72e6-4b75-ac66-85828224f892

solving #3519

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -18,7 +18,9 @@ use graphene_std::vector::algorithms::bezpath_algorithms::pathseg_compute_lookup
 use graphene_std::vector::misc::{HandleId, ManipulatorPointId, dvec2_to_point, point_to_dvec2};
 use graphene_std::vector::{HandleExt, PointId, SegmentId, Vector, VectorModificationType};
 use kurbo::{Affine, DEFAULT_ACCURACY, Line, ParamCurve, ParamCurveNearest, PathSeg, Rect, Shape};
+use std::collections::hash_map::DefaultHasher;
 use std::f64::consts::TAU;
+use std::hash::{Hash, Hasher};
 
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 pub enum SelectionChange {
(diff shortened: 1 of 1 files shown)


Annonnymmousss:
https://github.com/user-attachments/assets/33f535bb-5013-4530-93f3-0ad3d461d377

there was this issue too.

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -461,6 +463,16 @@ impl ShapeState {
...
 				merge_layers(document, layer1, layer2, responses);
+
+				let resolve_id = |layer: LayerNodeIdentifier, point_id: PointId, index: usize| {
+					let mut hasher = DefaultHasher::new();
+					(index, 0_usize, layer).hash(&mut hasher);
+					point_id.generate_from_hash(hasher.finish())
+				};
+
+				let start_point = resolve_id(layer1, start_point, 0);
+				let end_point = resolve_id(layer2, end_point, 1);
```
0HyperCube, line 474: Only ids in `layer2` will change (since the `layer1` will probably have no conflicts).

Furthermore it is possible that there is no conflict, in which case the original `PointId` must be used (like before this PR). You therefore need to check if there was some conflict by checking all of the point ids in the first path.

I don't care if this doesn't work for tables with several vector data, since I find them annoying.

0HyperCube:
@Annonnymmousss The issue in your video is the exact same as described in #3519 (the conflicting `PointId` is not updated so the wrong segment is created). However in my example, the incorrect segment wasn't visible (since it connected the point to itself).

Annonnymmousss:
working on it

Annonnymmousss:
@0HyperCube should i close this pr as the other one is approved??

0HyperCube:
I'm not a maintainer for this project so my approval on a PR is meaningless. https://docs.github.com/en/get-started/learning-about-github/access-permissions-on-github

Annonnymmousss:
ohh okayyy

Annonnymmousss:
issue fixed by PR #3534 .

PR #3534, opened 2025-12-27
Fix joining paths from different layers with Ctrl+J fails

### LINKED ISSUE #3519: Joining paths with the same PointIds doesn't work

Ayush2k02:
https://github.com/user-attachments/assets/ed71413a-589f-415c-8e15-8ce66feff474

DISCLAIMER : claude code cli extension was used in this PR ( this is free for university students )
Fixes #3519

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -435,9 +435,6 @@ impl ShapeState {
 		responses.add(NodeGraphMessage::RunDocumentGraph);
 	}
 
-	/// a two-step process: trigger reorganization, then use position-based lookup.
-	/// Helper function to connect two points by their positions after graph reorganization.
-	/// This transforms local point positions to document space and defers the connection until after the graph runs.
 	fn defer_connect_points_by_position(
 		document: &DocumentMessageHandler,
 		layer1: LayerNodeIdentifier,
@@ -475,7 +472,7 @@ impl ShapeState {
 				})
 				.into(),
 			],
-			});
+		});
 	}
 
 	fn handle_grouped_transform_close_path(document: &DocumentMessageHandler, layer: LayerNodeIdentifier, start_point: PointId, end_point: PointId, responses: &mut VecDeque<Message>) {


Keavon:
Hello and thanks for the PR. I need to mention that we do not accept PR descriptions written by AI because they increase the noise-to-signal ratio and usually serve to mislead with obscured inaccuracies. (The same applies to code, which must be written by you.)

Ayush2k02:
Oh, hey , I had this extension turned on in my IDE https://whatthediff.ai/ , will be mindful to disable it when creating PRs in this repo :) . Also regarding using AI in code, I did use claude code for targeting the files and the fix but did review that intensively before publishing .

Keavon:
Thanks for the insight into your process (we've been seeing more highly-elaborate-but-not-insightful PR descriptions like this and it's good to know some of the details about what kinds of AI tools are producing them). For code authorship, AI is acceptable for auto-completing lines you would already be typing, but we need contributions to always come from the original thought processes of its human authors who can answer to the reasoning behind each line of changes. Code can't be mysteriously changed without anyone aware of why the change was made. Thanks for working to follow our requirements to building an effective software project.

Ayush2k02:
Taking a look

Ayush2k02:
After taking a second look at the issue, here is what I came up with: The fix uses index-based tracking to handle PointId remapping:

- Vector::concat() preserves point ordering (layer1 points first, then layer2 points)
- Save each point's index before merge
- Calculate post-merge indices using layer1's point count as offset
- Look up actual PointIds at those indices after merge

Other approaches I considered:

- Position-based matching: Save coordinates, search within 250px tolerance after merge. Fragile—Flatten Path shifts coordinates unpredictably, large tolerance risks wrong matches.
- Hash-based ID prediction: Replicate collision resolution to predict new PointIds. Impractical—hash seed needs internal parameters (current_index, vector_index, node_id) inaccessible from shape_editor.rs.

Ayush2k02:
Let me know if I am missing something or if there could be any other edge case here

Ayush2k02:
https://github.com/user-attachments/assets/8c77e2c9-fd21-4837-98d9-7d17279cc181


More manual testing with data grid for when the conflicting PointId is not updated so the wrong segment is created , here it is not created .

0HyperCube (approved):
https://github.com/user-attachments/assets/d1286ac4-6a3b-42e2-b293-a3df000e3ee2

However I guess it doesn't matter too much.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -2669,6 +2674,46 @@ impl Fsm for PathToolFsmState {
...
+				// Find points by their positions (with small tolerance for floating point comparison)
+				const POSITION_TOLERANCE: f64 = 0.01;
+
+				let positions = vector.point_domain.positions();
+				let point_ids = vector.point_domain.ids();
+
+				let mut start_point_id = None;
+				let mut end_point_id = None;
+
+				for (i, &pos) in positions.iter().enumerate() {
```
Ayush2k02: I don't think this for loop is an overhead here, realistically there would'nt be more than 100-200 points max here . Thats just my guess , can definitely have a hashmap here as well

Ayush2k02:
I have a better solution for this, let me code it up 

So basically the issue with the current approach  was when layers use the repeat modifier. The merge operation flattens repeated instances, breaking the expected point array structure. 

Expected behavior:

- Layer1 has x points
- Layer2 has y points
- After merge: x + y points total

Actual behavior with repeat:

- Layer1 with 2 points × 3 repeat instances = 6 points (not 2)
- Layer2 with 2 points = 2 points
- After merge: 8 points total (not 4)

Since the old approach calculated indices using end_index = end_idx + layer1_count, it assumed layer1_count = 2, but after flattening the repeated instances, layer1 actually contributes 6 points. This causes the lookup to select the wrong points.
Solution

Instead of relying on indices (which change when instances are flattened), we now:

1. Before merge: Store the world-space coordinates of the selected points
2. After merge: Reverse-lookup the PointIDs by searching for those exact coordinates

Ayush2k02:
https://github.com/user-attachments/assets/3eed3a6e-fa93-4592-bc20-0d697dbac250


Some manual testing after the recent commit

Ayush2k02:
let me know if any other changes are required

Ayush2k02:
taking a look !

Ayush2k02:
Is this like a big limitation , this seems like a big change . I'll dig deeper if this is like a game breaker , and its happening on PROD as well , like its not related to my code changes .

0HyperCube:
@Ayush2k02 I thought that this PR was meant to "fix: joining paths from different layers with Ctrl+J".

From my very basic testing, it seems it still does not do this in some very common scenarios.

I hope that helps to inform your approach.

Ayush2k02:
Oh, are there any more edge cases that you were able to find apart from this one "It doesn't work if you have two lines in a group with a transform. You can press G with a group selected to apply a transform" , 
I'll look into this and fix it then

0HyperCube:
I have not found any more issues with the current approach. However I will be sure to test it again if you decide to modify the code.

Ayush2k02:
Hey, fixed the case with group selection. Let me know if there is anything else or if some optimizations could be made

0HyperCube:
When joining two points from the same layer, two segments are created.
The <kbd>Ctrl</kbd> + <kbd>J</kbd> is meant to insert exactly one segment between the two points.

https://github.com/user-attachments/assets/33754d77-b792-4d1d-9c88-fb48eeae27e3

Ayush2k02:
Completely missed this, I did try the same layer case with using shift and creating some lines but did'nt look well enough for the underlying line . Let me fix this quick, most likely will keep the same functionality for the group case and remove this from the same layer case . Sorry for the negligence :P

Ayush2k02:
Fixed the last issue, the check for group / folder was broken

Now I am checking if the layer has any children, means it is a group folder . The method `has_children` also seems appropriate here, hopefully there shouldn't be anymore bugs in this code path at least

0HyperCube:
It still seems to produce duplicate segments for me when joining points in the same layer inside a folder:

https://github.com/user-attachments/assets/2b05f57a-dab8-491e-b727-504c069cd078

Ayush2k02:
Okay, so added a check for if the children are more than 1 means we are inside a group , then only go down the `handle_grouped_transform_close_path` path , else directly create a segment . Have tested all the 3 cases down this path

1. simple single layer join paths
2. grouped lines which are in different layers
3. grouped image but we are in a single layer

Ayush2k02:
wait, its still not correct. Dont review yet

Ayush2k02:
Hmmm, for some reason when I connect straight lines inside a group with cmd + j it fails the first time as the point ids are getting remapped but when doing the exact same thing with creating lines from the pen tool or the spline tool , it succeeds and create the segment in the first try , weird

Ayush2k02:
Okay, so changing the reorganisation method to use `ApplyPointDelta` instead of creating a dummy segment fixed all the edge cases for me . This is now ready for review from my side . But unsure why complex paths (created with the spline/pen tool) inside a grouped folder successfully created the dummy segment on the first try, while simpler straight lines failed to do so. Something related to the pointIds remapping with `RunDocumentGraph`

Ayush2k02:
This is now reviewable from my side !

0HyperCube:
Seems to work fine now (based on my limited testing). Thanks for spending time to properly fix the <kbd>Ctrl</kbd><kbd>J</kbd> functionality.

Ayush2k02:
It'd be great if you can approve this one , if you have tested out the branch :)

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -421,6 +421,52 @@ impl ShapeState {
...
+
+	/// a two-step process: trigger reorganization, then use position-based lookup.
+	fn handle_grouped_transform_close_path(document: &DocumentMessageHandler, layer: LayerNodeIdentifier, start_point: PointId, end_point: PointId, responses: &mut VecDeque<Message>) {
+		// Get the layer's transform (handles rotation, scaling, translation)
+		let layer_transform = document.metadata().transform_to_document(layer);
+
+		let start_local_pos = document.network_interface.compute_modified_vector(layer).and_then(|v| v.point_domain.position_from_id(start_point));
+		let end_local_pos = document.network_interface.compute_modified_vector(layer).and_then(|v| v.point_domain.position_from_id(end_point));
+
+		if let (Some(start_local), Some(end_local)) = (start_local_pos, end_local_pos) {
```
0HyperCube: You might want to print a warning to help debugging edge cases.
```suggestion
		let (Some(start_local), Some(end_local)) = (start_local_pos, end_local_pos) else {
			warn!("Unable to resolve point ids for joining");
			return;
		};
```

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -421,6 +421,52 @@ impl ShapeState {
 		(point.as_handle().is_some() && self.ignore_handles) || (point.as_anchor().is_some() && self.ignore_anchors)
 	}
 
+	/// Creates a dummy modification to trigger graph reorganization.
```
0HyperCube: Perhaps the specific "graph reorganization" could be explained:
```suggestion
	/// Applies a dummy vector modification to the layer. In the case where a group containing some vector data is selected, this triggers the creation of a «Flatten Path» node.
```

editor/src/messages/tool/common_functionality/shape_editor.rs
```diff
@@ -447,28 +493,56 @@ impl ShapeState {
...
+							ToolMessage::Path(PathToolMessage::ConnectPointsByPosition {
+								layer: layer1,
+								start_position: start_pos,
+								end_position: end_pos,
+							})
+							.into(),
+						],
+					});
+				}
 			}
```
0HyperCube: This is basically duplicated from the `handle_grouped_transform_close_path` function.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -2669,6 +2674,60 @@ impl Fsm for PathToolFsmState {
...
 				self
 			}
+			(_, PathToolMessage::ConnectPointsByPosition { layer, start_position, end_position }) => {
+				// Get the merged vector
+				let Some(vector) = document.network_interface.compute_modified_vector(layer) else {
+					return self;
+				};
+
+				// Find points by their positions (with small tolerance for floating point comparison)
+				const POSITION_TOLERANCE: f64 = 0.01;
```
0HyperCube: This could be quite a big tolerance compared to some of the other values:
https://github.com/GraphiteEditor/Graphite/blob/42440c0d0bcf5735b05d8a9e5bd27187f74b1589/editor/src/messages/tool/tool_messages/pen_tool.rs#L741

Is there a reason this was chosen?

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -2669,6 +2674,60 @@ impl Fsm for PathToolFsmState {
...
+					if start_point_id.is_none() && start_distance < POSITION_TOLERANCE {
+						start_point_id = Some(point_ids[i]);
+					}
+					if end_point_id.is_none() && end_distance < POSITION_TOLERANCE {
+						end_point_id = Some(point_ids[i]);
+					}
+					if start_point_id.is_some() && end_point_id.is_some() {
+						break;
+					}
+				}
```
0HyperCube, line 2711: Could perhaps be simplified with a find expression?
```rs
let Some((start_point_id, _)) = vector
	.point_domain
	.iter()
	.find(|&(_, position)| layer_transform.transform_point2(position).distance(start_position) < POSITION_TOLERANCE)
else {
	return self;
};
```

0HyperCube (approved):
I'm not a maintainer for this project so my approval is irrelevant; however since you have requested a code review I have located some potential small improvements.

editor/src/messages/tool/common_functionality/shape_editor.rs
Ayush2k02: makes sense !

editor/src/messages/tool/tool_messages/path_tool.rs
Ayush2k02: Oh I though 0.01 was fairly small , is it not ? Let me look through the codebase for a more apt value . I'll test using various values and see what exactly do we do in such cases in the codebase .

editor/src/messages/tool/common_functionality/shape_editor.rs
Ayush2k02: I see

editor/src/messages/tool/common_functionality/shape_editor.rs
Ayush2k02: yeah, makes sense to encapsulate and create a helper function, let me do that !

editor/src/messages/tool/tool_messages/path_tool.rs
Ayush2k02: let me understand this syntax

Ayush2k02: We'd need to run this twice to find the start and the end point , so 2 loops with transforms in them , think the original approach is better ?

```
let Some((start_point_id, _)) = vector.point_domain.iter()
    .find(|&(_, position)| layer_transform.transform_point2(position).distance(start_position) < POSITION_TOLERANCE)
else {
    return self;
};
```

```
let Some((end_point_id, _)) = vector.point_domain.iter()
    .find(|&(_, position)| layer_transform.transform_point2(position).distance(end_position) < POSITION_TOLERANCE)
else {
    return self;
};
```

editor/src/messages/tool/tool_messages/path_tool.rs
Ayush2k02: done

Ayush2k02: I saw a lot of tolerances as high as 5 pixesls, and 0.01 seemed to be not visible at all, so chose 0.01 . Changed it to 1e-6 now

Ayush2k02:
Yup, seems to be working fine after the recent changes

editor/src/messages/tool/tool_messages/path_tool.rs
0HyperCube: OK.

Ayush2k02:
Any updates on this guys !

Ayush2k02:
Also thanks for all the testing and code reviews @0HyperCube  , appreciate it :)

TrueDoctor:
@0HyperCube your approval is definitely not irrelevant, please feel free to ping me once you are happy with the changes and I'll try to merge them asap. This does very much help with reducing our pr backlog, thx!

PR #3536, opened 2025-12-27
feat/drag with backspace

Sahil-Gupta584:
- **disallow invalid layer drop**
- **improve invalidDrag indicator behaviour**
- **overall layer,artboard drag behaviour impr**
- **backspace and drag for shape tools**
- **for others**
- **part of merge**


Keavon:
What is this? It appears to be spam from what I can tell, and there is no coherent explanation.

Sahil-Gupta584:
Oops is pr shouldnt be created here, my bad!

PR #3539, opened 2025-12-28
fix: rasterize node document leakage

### LINKED ISSUE #3533: Rasterize node: the rasterized image is leaking from one document to another
**Description:**
When using the Rasterize node in multiple documents which are opened at the same time, a rasterized image produced by the Rasterize node in one document is shown in another document which also uses the Rasterize node.

**Environment:**
The bug is happening in the web version of Graphite (link https://dev.graphite.art, commit 4ab75c9c) opened in the Opera web browser (v125.0.5729.49) on MS Windows 11 Home (v10.0.26200).

**Steps to reproduce:**
1. Open the web version of Graphite at https://dev.graphite.art.
2. Create a new document.
3. In the document, draw a circle in the top-left corner of the document.
4. In the Properties panel, add the Rasterize node to the circle.
5. Create another document.
6. In that new document, draw a rectangle in the top-left corner of the document.
7. In the Properties panel, add the Rasterize node to the rectangle.

Expected: the rectangle is shown in the second document.
Actual: the circle from the first document is shown in the second document, instead of the rectangle.

Ayush2k02:
Simple enough fix which skips the flawed deduplication check and generates a different id ( UUID ) for each new image

https://github.com/user-attachments/assets/5285916e-cf43-4b0c-a5fe-8e8fcb8bcace

Closes #3533

### DIFF
--- a/.github/workflows/provide-shaders.yml
+++ b/.github/workflows/provide-shaders.yml
@@ -1,0 +1,31 @@
+name: Provide Shaders
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch: {}
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v4
+      - uses: DeterminateSystems/nix-installer-action@main
+      - uses: DeterminateSystems/magic-nix-cache-action@main
+
+      - name: Build graphene raster nodes shaders
+        run: nix build .nix#raster-nodes-shaders && cp result raster_nodes_shaders_entrypoint.wgsl
+
+      - name: Upload graphene raster nodes shaders to artifacts repository
+        run: |
+          bash .github/workflows/scripts/artifact-upload.bash \
+            ${{ vars.ARTIFACTS_REPO_OWNER }} \
+            ${{ vars.ARTIFACTS_REPO_NAME }} \
+            ${{ vars.ARTIFACTS_REPO_BRANCH }} \
+            rev/${{ github.sha }}/raster_nodes_shaders_entrypoint.wgsl \
+            raster_nodes_shaders_entrypoint.wgsl \
+            "${{ github.sha }} raster_nodes_shaders_entrypoint.wgsl" \
+            ${{ secrets.ARTIFACTS_REPO_TOKEN }}
--- a/.github/workflows/scripts/artifact-upload.bash
+++ b/.github/workflows/scripts/artifact-upload.bash
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+usage() {
+  cat <<EOF
+Usage: $0 <owner> <repo> <branch> <target-path> <artifact-file> <commit-message> <github-token>
+
+Arguments:
+  owner           : GitHub user or organization of the target repo
+  repo            : Target repo name
+  branch          : Branch name (e.g. main)
+  target-path     : Full path (including folders + filename) in the target repo where to upload
+  artifact-file   : Local file path to upload
+  commit-message  : Commit message for creating/updating the file
+  github-token    : GitHub token (PAT or equivalent) with write access to the target repo
+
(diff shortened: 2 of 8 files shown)


node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -1249,6 +1251,8 @@ impl Render for Table<Raster<CPU>> {
 				let id = row.source_node_id.map(|x| x.0).unwrap_or_else(|| {
 					let mut state = DefaultHasher::new();
 					image.data().hash(&mut state);
+					// Include document_id to prevent hash collisions between documents
```
0HyperCube: You can also trivially have collisions within the same document (e.g. by rotating a rectangle with only a thin stroke and a white fill).

I think therefore that it is not viable to hash the images here at all.

Ayush2k02: oh, yeah . Makes sense

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -1246,21 +1246,17 @@ impl Render for Table<Raster<CPU>> {
 			}
 
 			if render_params.to_canvas() {
-				let id = row.source_node_id.map(|x| x.0).unwrap_or_else(|| {
```
Ayush2k02, line 1249: Don't fully understand this syntax but from what I can see this generates an ID for an image and was used in deduplication for similar images . Since we are no more deduplicating similar images and each new image is getting its own ID , figured it would be best to use a simple `generate_uuid` instead

Ayush2k02: Without doing so , its still causing the same image leakage for me

Ayush2k02:
This ones ready to be reviewed from my side

TrueDoctor:
I need a bit more time to look at this, the workaround implemented here does not feel quite right, I'm also wondering if this will effect the performane as it is currently implemented

0HyperCube:
@TrueDoctor yes it will be bad for performance.

The hashing code for images is included below:
https://github.com/GraphiteEditor/Graphite/blob/fa45efa9e29fbdc2df96d3aedbb80656509398b7/node-graph/libraries/raster-types/src/image.rs#L104-L116

However this will lead to very many conflicts when you have images such as the cases below:
ellipse_thin_stroke
rectangle_thin_stroke

Since there are 1000x1000 pixels and the vast majority are white, these two both hash to the same thing.

It is necessary for correctness that these two different images aren't treated as the same thing by the SVG rendering code. Otherwise it looks like the rasterize node is broken. Actually the rasterization works fine.

An example graphite file where this collision occurs is include below:
[rasterize_node_conflicts.graphite](https://github.com/user-attachments/files/24368191/rasterize_node_conflicts.graphite.txt)

TrueDoctor:
Yeah the hash is broken, I guess we could do an eq check after the hash calculation, that should hopefully be fast enough. But I guess we would need to benchmark if disabling the deduplication alltogether would be faster

Ayush2k02:
An example graphite file where this collision occurs is include below:
[rasterize_node_conflicts.graphite](https://github.com/user-attachments/files/24368191/rasterize_node_conflicts.graphite.txt) , like do we import this JSON somewhere to load this ourselves ?



What is this ?? and how is this used ?

0HyperCube:
@Ayush2k02 Yes, you can import this graphite file into the Graphite editor. `file -> open` and select the file.

Note that the extension is `.graphite.json` since GitHub doesn't directly support uploading files using `.graphite` extension. You can either change the extension after you download the file or just directly select file (you might have to alter the filter in your file picker).

Ayush2k02:
Made this second branch which uses the eq check approach suggested by trueDoctor , let me know if you want me to benchmark both of these to see which one is faster and also review the other PRs code as well , thanks

Ayush2k02:
Should I close this one ? Feel like the other branch has caching as well so would be better to merge that instead

TrueDoctor:
Yeah, I think we can close this one

PR #3545, opened 2025-12-30
Fix centroid calculation to return bounding box center for multiple vectors

### LINKED ISSUE #3544: Incorrect calculations of a centroid position
I noticed that the centroid for the Bounding Box node is calculated incorrectly when there are multiple shapes passed via Flatten Path.

Image

Image

daniil-loban:
I noticed that the centroid for the Bounding Box node is calculated incorrectly when there are multiple shapes passed via Flatten Path. However, the rectangle itself is calculated correctly. It seemed logical to me to use the bounding_box call to get the rectangle, since I believe its center will be the correct centroid. 

Result:
after

Further calculations may not be necessary, but I left them in just in case.

Fixes #3544

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_properties.rs
@@ -233,15 +233,7 @@ pub(crate) fn property_from_type(
 							widgets.extend_from_slice(&[
 								Separator::new(SeparatorStyle::Unrelated).widget_instance(),
 								TextLabel::new("-")
-									.tooltip_label(format!(
-										"Data Type: {}",
-										concrete_type
-											.alias
-											.as_deref()
-											// TODO: Avoid needing to remove spaces here by fixing how `alias` is generated
-											.map(|s| s.to_string().replace(" ", ""))
-											.unwrap_or_else(|| graphene_std::format_type(concrete_type.name.as_ref())),
-									))
+									.tooltip_label(format!("Data Type: {concrete_type}"))
 									.tooltip_description("This data can only be supplied through the node graph because no widget exists for its type.")
 									.widget_instance(),
 							]);
@@ -251,7 +243,7 @@ pub(crate) fn property_from_type(
 				}
 			}
 		}
-		Type::Generic(_) => vec![TextLabel::new("Generic type (not supported)").widget_instance()].into(),
+		Type::Generic(_) => vec![TextLabel::new("Generic Type (Not Supported)").widget_instance()].into(),
 		Type::Fn(_, out) => return property_from_type(node_id, index, out, number_options, unit, display_decimal_places, step, context),
 		Type::Future(out) => return property_from_type(node_id, index, out, number_options, unit, display_decimal_places, step, context),
 	};
--- a/node-graph/libraries/core-types/src/types.rs
+++ b/node-graph/libraries/core-types/src/types.rs
@@ -216,6 +216,13 @@ impl std::hash::Hash for TypeDescriptor {
 	}
 }
 
+impl std::fmt::Display for TypeDescriptor {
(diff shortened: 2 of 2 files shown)


node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -2336,6 +2336,14 @@ async fn centroid(ctx: impl Ctx + CloneVarArgs + ExtractAll, content: impl Node<
...
 	}
 
+	// Calculate the bounding box of all vectors in the table
+	let bounding_box = vector.bounding_box(DAffine2::IDENTITY, false);
+	if let RenderBoundingBox::Rectangle([min, max]) = bounding_box {
+		// Return the center of the bounding box
+		return (min + max) / 2.;
+	}
+
+	// Fallback to the original method if bounding box calculation fails
```
0HyperCube, line 2346: The bounding box only fails if there are no segments to take the bounding box. In this case, the code below isn't going to work either.

Essentially this changes the implementation to just always return the centre of the bounding box.

daniil-loban: Do I need to delete the following code in this case?

0HyperCube: As kindly pointed out by [elbertronnie](https://github.com/elbertronnie) on #3544, there was a typo in the `pathseg_to_parametric_polynomial` function that caused incorrect results. With the fix in #3562 it is now resolved:

result with star in centroid
node graph

daniil-loban: I may be wrong about the concept, but in my opinion the centroid is not where it should be.
I think that the general centroid should act according to the same principles as the single centroid — find the center.

image

Keavon: The bottom left shape has the largest area so the center of mass will be concentrated towards it, as we observe.

daniil-loban: So, I don't understand the concept. As I understand it, the concept of mass is also useful, but not in my case. I don't think this solution is very effective. Let me give you an example: if I want to place a rectangle above a group, I won't get the correct result using this concept, so I don't know how to solve this problem.

image

This won't allow you to create interesting effects with text without knowing where the center of the letter is (it doesn't matter how many vectors are in the letter)

image

Keavon: Centroid computes the center of mass. It sounds like you are expecting the center of the bounding box. You can add a Bounding Box node before Centroid to get the center of mass of the rectangular bounding box, which would have your desired effect.

daniil-loban: That's exactly how it's calculated (Discord: https://discord.com/channels/731730685944922173/1134981447539761273/1456461715872026809 )

image

image

daniil-loban: > Centroid computes the center of mass. It sounds like you are expecting the center of the bounding box. You can add a Bounding Box node before Centroid to get the center of mass of the rectangular bounding box, which would have your desired effect.

Perhaps then it's worth simply creating an additional node that will calculate the geometric center?

daniil-loban:
> Changing the centroid type to Length works (If you just want the green rectangle on top of the pink one)

Ezbaze

node-graph/nodes/vector/src/vector_nodes.rs
Keavon: That is planned, yes. Feel free to also open an issue to track that so we don't forget. Thanks!

0HyperCube: @daniil-loban If you have the `bounding box -> centroid` then it will find the centre of the bounding box.

image
image

PR #3547, opened 2025-12-30
Add issue template and issue template config

fuguesoft:
I made an issue template for bug reports.

If you're interested, it goes in `/.github/ISSUE_TEMPLATE/`. 
If you want to disable blank templates, you can use the following setting in the `config.yaml` file:

```yaml
blank_issues_enabled: false
```

You can test the template at my fork: https://github.com/fuguesoft/Graphite

### DIFF
--- a/.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
+++ b/.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
@@ -11,15 +11,15 @@ body:
       label: Does an issue for this already exist?
       description: Please search to see if an issue already exists for the bug you encountered.
       options:
-        - label: I have searched the existing issues
-        - label: I have ensured my environment (operating system, app launcher, etc) is up to date
+        - label: I have searched the existing issues.
+        - label: I have ensured my environment (operating system, app launcher, etc) is up to date.
           required: true
   - type: input
-    id: distro
+    id: os
     attributes:
       label: Operating System and Version
-      description: What is your Linux Distro? What version?
-      placeholder: Debian, Void, Gentoo, etc.
+      description: What is your Operating System? What version?
+      placeholder: Debian, Void, Gentoo, Windows, macOS etc.
     validations:
       required: true
   - type: dropdown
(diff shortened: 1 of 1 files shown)


fuguesoft:
Video Example:
https://0x0.st/P-Ty.mp4

Please let me know that other information I can provide.

.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
```diff
@@ -0,0 +1,110 @@
...
+      required: true
+  - type: input
+    id: other-install
+    attributes:
+      label: Other Install
+      description: Please tell us how the app was installed
+  - type: input
+    attributes:
+      label: Build Version
+      description: What version of Graphite2D?
```
0HyperCube: I think it is branded just as «Graphite»

.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
```diff
@@ -0,0 +1,110 @@
...
+      label: App Installation
+      description: How is the app installed?
+      multiple: false
+      options:
+        - AppImage
+        - Source Compilation
+        - Other
+      default: 0
+    validations:
+      required: true
```
0HyperCube, line 44: This is perhaps of a bit linux centric.

A lot of usage is from the web (deployed as https://dev.graphite.art/ or https://editor.graphite.art/).

There are also a number of Windows11 users who probably don't understand what a "Linux Distro"or "Desktop Environment" is.

0HyperCube (commented):
Thanks for taking the time to add this.

I'm not a maintainer so I guess Keavon will decide if they find this useful once they return from their break.

.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
fuguesoft: That is a great point that I was just looking at. I'm going to change this to a draft and update that language. Then I'll publish again.

Thanks for the review!

.github/ISSUE_TEMPLATE/01-bug-issue-template.yml
fuguesoft: okay just Graphite
Will update

Keavon:
Hi, thank you for the suggested/proposed issue template, but this kind of change needs to be done in response to solving observed issues with our process, not an external "a change for the sake of a change" even if it's well-intentioned.

fuguesoft:
Thanks for the response!
Happy New Year!

PR #3549, opened 2025-12-31
Fix/multiply alpha for table color

### LINKED ISSUE #3205: Opacity node does not work directly on colors
The opacity node works fine when applied to `Table<Vector>` but when operating on a `Table<Color>` directly it seems to do nothing, even though it's explicitly an allowed input.

Opacity Node working vs broken opacity node

garvittsingla:
Fixes #3205 
The main problem was in 
`MultiplyAlpha for Table<Color>`, where only the alpha_blending.opacity was changing but not the color opacity as mentioned itself in the issue

I changed the color alpha itself with the function `multiplied_alpha` under the `AlphaMut` trait.
My fix is this:
```rust
impl MultiplyAlpha for Table<Color> {
	fn multiply_alpha(&mut self, factor: f64) {
		for row in self.iter_mut() {
			row.element.set_alpha(factor as f32);
			row.alpha_blending.opacity *= factor as f32;
		}
	}
}
```

**Faulty behavior**

image

Change in opacity node before the fill value does not changes anything. The opacity is always full

---

**Corrected behaviour**
image

Now changing the opacity changes the color opacity

### DIFF
--- a/node-graph/nodes/blending/src/lib.rs
+++ b/node-graph/nodes/blending/src/lib.rs
@@ -1,4 +1,4 @@
-use core_types::color::Alpha;
+use core_types::color::{AlphaMut};
 use core_types::registry::types::Percentage;
 use core_types::table::Table;
 use core_types::{BlendMode, Color, Ctx};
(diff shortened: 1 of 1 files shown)


garvittsingla:
> This is an approach to fixing the issue.
> 
> I'm not sure if it is advantageous to have both the `alpha_blending.opacity` and the `element.opacity`?

@0HyperCube It makes sense, I just don't removed the alpha blending opacity, thought to have a review first 

Should i remove it?

garvittsingla:
@0HyperCube Should I delete it??

0HyperCube:
I do not mind. Do whatever you prefer.

Keavon:
Superseded by #3785.

PR #3550, opened 2025-12-31
Fix rasterize node document leakage with hashmap and eq check

### LINKED ISSUE #3533: Rasterize node: the rasterized image is leaking from one document to another

Ayush2k02:
Second approach for fixing this bug . Will be tested and benchmarked against https://github.com/GraphiteEditor/Graphite/pull/3539 to see if this is better

Closes #3533

### DIFF
--- a/node-graph/libraries/rendering/src/renderer.rs
+++ b/node-graph/libraries/rendering/src/renderer.rs
@@ -1249,10 +1249,7 @@ impl Render for Table<Raster<CPU>> {
 			if render_params.to_canvas() {
 				let mut image_copy = image.clone();
 				image_copy.data_mut().map_pixels(|p| p.to_unassociated_alpha());
-				let id = match render.image_data.entry(image_copy.into_data()) {
-					std::collections::hash_map::Entry::Occupied(occupied_entry) => *occupied_entry.get(),
-					std::collections::hash_map::Entry::Vacant(vacant_entry) => *vacant_entry.insert(generate_uuid()),
-				};
+				let id = *render.image_data.entry(image_copy.into_data()).or_insert_with(generate_uuid);
 
 				render.parent_tag(
 					"foreignObject",


node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -60,6 +60,9 @@ pub struct SvgRender {
 	pub svg_defs: String,
 	pub transform: DAffine2,
 	pub image_data: Vec<(u64, Image<Color>)>,
+	/// HashMap for O(1) lookup: content_hash -> (id, full_image_for_collision_check)
+	/// Used in deduplicating images efficiently
+	image_lookup: HashMap<u64, Vec<(u64, Image<Color>)>>,
```
0HyperCube: This ends up storing the image data twice? I don't think that this is desirable since images can have large memory requirements.

Ayush2k02: hmnmn, think I'll have the map reference the vector

TrueDoctor: This won't be allowed due to rusts ownership rules, but you can store an `Arc<Vec<…>>` in both. But I'm a bit confused why you have to do this at all, I would expect that just storing `(u64, Image<Color>)` in a `HashSet` should be all that we need, right?  The hash set / hash map already implement the eq check on hash collisions

TrueDoctor:
Alright, I've now pushed another suggestion. If the perf of this is acceptable, I think this is going to be the simplest solution which is hopefully also relatively easy to maintain. @0HyperCube thoughts?

node-graph/libraries/rendering/src/renderer.rs
Ayush2k02: That makes total sense - was redundantly reimplementing what HashMap does automatically. One follow-up question though ,  went with the manual approach because I was trying to preserve source_node_id when available (to avoid regenerating IDs for existing images). Your solution generates new UUIDs on every HashMap miss , is the source node Id not relevant here ?

Ayush2k02:
This is so much more cleaner , wow

node-graph/libraries/rendering/src/renderer.rs
TrueDoctor: The image source is just cheaper to compute if it exists, but if it does not exist the code should behave the same way. If the source ids are the same, we know that the images will also be the same without having to check them. We could also extend this with a second hashmap from source_id to image id and always first try to find the source id in the source_id hashmap (if we have a source id) and if not look up the full image. That should give us optimal perf characteristics, but I think I still prefer this simpler solution if the perf is good enough

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -1239,23 +1239,21 @@ impl Render for Table<Raster<CPU>> {
...
-					let mut image = image.data().clone();
-					image.map_pixels(|p| p.to_unassociated_alpha());
-					render.image_data.push((id, image));
-				}
+				let mut image_copy = image.clone();
+				image_copy.data_mut().map_pixels(|p| p.to_unassociated_alpha());
+				let id = match render.image_data.entry(image_copy.into_data()) {
+					std::collections::hash_map::Entry::Occupied(occupied_entry) => *occupied_entry.get(),
+					std::collections::hash_map::Entry::Vacant(vacant_entry) => *vacant_entry.insert(generate_uuid()),
+				};
```
0HyperCube: This is just [`or_insert_with`](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert_with) I guess.

TrueDoctor: Yes, I originally had more code in the vacant branch, but now that that's gone we should be able to simplify this

Ayush2k02:
Let me know if there is something which could be done from my side

TrueDoctor:
@Ayush2k02 Could you incorporate hypercubes suggestion and replace the `match` with a call to `or_insert_with` ?

Ayush2k02:
Okay, let me take a look

Ayush2k02:
Done !

Ayush2k02:
https://github.com/user-attachments/assets/4db056ea-0f3e-433c-9ddb-051f5e66b943

Some manual testing

Ayush2k02:
Screenshot 2026-01-04 at 5 11 31 PM

YAY !

PR #3553, opened 2025-12-31
Add combined parameter to box_warp function

### LINKED ISSUE #3551: Box warp doesn't treat subpaths as unified object

Image

Image

mathurojus:
Adds a `combined` parameter to control whether the box warp treats all subpaths as a single object (when true) or warps them individually (when false).

Fixes #3551


mathurojus:
@Keavon could you please approve the workflows when ready? If everything looks good, feel free to merge. Thanks!

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -785,7 +785,7 @@ async fn extrude(_: impl Ctx, mut source: Table<Vector>, direction: DVec2, joini
 }
 
 #[node_macro::node(category("Vector: Modifier"), path(core_types::vector))]
-async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Table<Vector>) -> Table<Vector> {
+async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Table<Vector>, #[expose] combined: bool) -> Table<Vector> {
```
mathurojus, line 789: @0HyperCube 
Added the implementation for the `combined` parameter. When `combined=true`, the function now computes a unified bounding box for all vectors and applies the warp uniformly. When `combined=false`, it uses individual bounding boxes (original behavior).

Does this approach look correct?

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -786,10 +785,26 @@ async fn extrude(_: impl Ctx, mut source: Table<Vector>, direction: DVec2, joini
...
 	let Some((target, target_transform)) = rectangle.get(0).map(|rect| (rect.element, rect.transform)) else {
 		return content;
 	};
+	
+    // Compute combined bounding box if needed
+    let combined_bbox = if combined {
```
Ayush2k02, line 795: consider using reduce here instead of fold

```
let combined_bbox = if combined {
    content.iter()
        .filter_map(|row| row.element.bounding_box_with_transform(row.transform))
        .reduce(|[min, max], [bbox_min, bbox_max]| {
            [min.min(bbox_min), max.max(bbox_max)]
        })
} else {
    None
};
```

since we are computing an optional aggregate here

this would eliminate the explicit Option type annotation

Ayush2k02 (changes requested):

node-graph/nodes/vector/src/vector_nodes.rs
mathurojus: @Ayush2k02 I've implemented your suggestion to use `reduce` instead of `fold` for the `combined_bbox` calculation. 

Does this look good to you? Any feedback or additional changes needed?

Keavon:
Does not compile.

PR #3554, opened 2025-12-31
fix(vector): warp groups using combined bounding box

### LINKED ISSUE #3551: Box warp doesn't treat subpaths as unified object

LilSuperUser:
Groups of paths now warp together as a unit instead of independently.

- union all subpath bounding boxes into one frame
- normalize points and handles against shared bounds
- apply bilinear interpolation into target quad
- reset transforms after 

Closes #3551

### DIFF
--- a/node-graph/nodes/vector/src/vector_nodes.rs
+++ b/node-graph/nodes/vector/src/vector_nodes.rs
@@ -1,6 +1,7 @@
 use core::f64::consts::PI;
 use core::hash::{Hash, Hasher};
 use core_types::bounds::{BoundingBox, RenderBoundingBox};
+use core_types::math::rect::Rect;
 use core_types::registry::types::{Angle, IntegerCount, Length, Multiplier, Percentage, PixelLength, PixelSize, Progression, SeedValue};
 use core_types::table::{Table, TableRow, TableRowMut};
 use core_types::transform::{Footprint, Transform};
@@ -28,6 +29,7 @@ use vector_types::vector::style::{Fill, Gradient, GradientStops, Stroke};
 use vector_types::vector::style::{PaintOrder, StrokeAlign, StrokeCap, StrokeJoin};
 use vector_types::vector::{FillId, RegionId};
 use vector_types::vector::{PointId, SegmentDomain, SegmentId, StrokeId, VectorExt};
+const NORMALIZATION_EPSILON: f64 = 1e-9;
 
 /// Implemented for types that can be converted to an iterator of vector rows.
 /// Used for the fill and stroke node so they can be used on `Table<Graphic>` or `Table<Vector>`.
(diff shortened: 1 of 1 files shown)


node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -817,20 +832,22 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
...
-
 			// Apply the warp
 			let mut result = vector.clone();
 
 			// Precompute source bounding box size for normalization
-			let source_size = source_bbox[1] - source_bbox[0];
+			let source_size = group_size;
+
+			// Safe normalization size (prevents division by zero)
+			let eps = 1e-9;
```
Ayush2k02: NIT : move this to a constant at the top of the file

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -817,20 +832,22 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
 					DVec2::new(source_bbox[0].x, source_bbox[1].y),
 				]
 			};
```
Ayush2k02: NIT: add this line back in

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -791,18 +790,34 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
 		return content;
 	};
 
+	// Compute combined bounding box across all vectors in the table
+	let mut group_bbox: Option<[DVec2; 2]> = None;
+	for row in content.iter() {
+		if let Some(bbox) = row.element.bounding_box_with_transform(*row.transform) {
+			group_bbox = Some(match group_bbox {
```
Ayush2k02, line 800: Use this here

```
group_bbox = Some(match group_bbox {
    None => bbox,
    Some(current) => Rect::combine_bounds(current, bbox),
});
```

Like use this function which is already there in the codebase

Ayush2k02 (changes requested):

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -839,13 +856,10 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
 			// Transform handles in bezier curves
 			for (_, handles, _, _) in result.handles_mut() {
 				*handles = handles.apply_transformation(|pos| {
```
Ayush2k02: NIT : add these comments back , no reason for them to be removed

Ayush2k02 (changes requested):

node-graph/nodes/vector/src/vector_nodes.rs
Ayush2k02: here

PR #3555, opened 2025-12-31
Fix Box Warp to treat multiple shapes as unified object Box warp function combined subpath

### LINKED ISSUE #3551: Box warp doesn't treat subpaths as unified object

garvittsingla:
Fixes  #3551 
The function before makes the shape overlap on each other.
Because earlier we found the minimum and then make a new rectangle(size of only one shape) which makes the both shapes combined in a single small rectangle size of only one shape

The fix is as mentioned in the issue itself
So the solution to it is to get the minimum left top and the max bottom right corner.Then warping all the shapes into that single big rectangle size of the all shapes. It makes the distance bwteen 2 shapes constant.

 **Before**
image

 **After**
Shapes maintain their relative positions when warped together.
image

### DIFF
--- a/node-graph/nodes/vector/src/vector_nodes.rs
+++ b/node-graph/nodes/vector/src/vector_nodes.rs
@@ -789,74 +789,59 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
 	let Some((target, target_transform)) = rectangle.get(0).map(|rect| (rect.element, rect.transform)) else {
 		return content;
 	};
+	//calculate the minumum and max corner of all the shapes
+	let mut min_corner = DVec2::new(f64::MAX, f64::MAX);
+	let mut max_corner = DVec2::new(f64::MIN, f64::MIN);
+
+	for row in content.iter() {
+		if let Some(bbox) = row.element.bounding_box_with_transform(*row.transform) {
+			min_corner = min_corner.min(bbox[0]);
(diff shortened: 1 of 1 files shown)


node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -790,74 +789,59 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
 	let Some((target, target_transform)) = rectangle.get(0).map(|rect| (rect.element, rect.transform)) else {
 		return content;
 	};
+	//calculate the minumum and max corner of all the shapes
```
Ayush2k02, line 792: NIT : spell check for minimum

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -790,74 +789,59 @@ async fn box_warp(_: impl Ctx, content: Table<Vector>, #[expose] rectangle: Tabl
...
-			row.element = result;
-			row.transform = DAffine2::IDENTITY;
-			row
-		})
-		.collect()
+	content
+    .into_iter()
+    .map(|mut row| {
+        let transform = row.transform;          
+        let vector = &row.element;
```
Ayush2k02, line 823: can be simplified to

let mut result = row.element.clone();

Ayush2k02 (changes requested):

PR #3565, opened 2026-01-02
Introduce QR Code node

### LINKED ISSUE #3468: QR Code Node
Novelty QR Codes (rounded corners, graphic overlay in center, gradient mask..) are pretty common nowadays, and they seem well-suited to Graphite's render pipeline, and especially its ImageMagick-style cli workflow. I propose a QR Code Generator node that returns a vector QR Code when given a string, and possibly an "Error Correction" input to specify the general number of extra error correction bits included.

One potential workaround for this is would be to make a web request to some API to generate the QR Code, and load it from there, but this requires either an internet connection (potentially undesirable for an "Offline-first" editor), or else a locally-hosted server, which of course requires extra software to be running on the machine.

Sahil-Gupta584:
Closes #3468

Do you have any suggestion ?

Current Implementation:

https://github.com/user-attachments/assets/d8c5d910-2f74-4d7e-a1d2-bc37258ac360

### DIFF
--- a/node-graph/nodes/vector/src/generator_nodes.rs
+++ b/node-graph/nodes/vector/src/generator_nodes.rs
@@ -204,26 +204,92 @@ fn qr_code(
 
 	let size = qr_code.size();
 	let mut vector = Vector::default();
-	let offset = DVec2::splat(size as f64 / -2.);
+	let offset = DVec2::splat(0.0);
 
-	for y in 0..size {
-		let mut x = 0;
-		while x < size {
-			if qr_code.get_module(x, y) {
-				let start_x = x;
-				x += 1;
-				if merge {
-					while x < size && qr_code.get_module(x, y) {
-						x += 1;
+	if merge {
+		use std::collections::{HashMap, HashSet, VecDeque};
+
+		let mut remaining: HashSet<(i32, i32)> = HashSet::new();
+		for y in 0..size {
+			for x in 0..size {
+				if qr_code.get_module(x, y) {
+					remaining.insert((x, y));
+				}
+			}
+		}
+
+		while let Some(&(start_x, start_y)) = remaining.iter().next() {
+			let mut island = HashSet::new();
+			let mut queue = VecDeque::new();
+			queue.push_back((start_x, start_y));
+			remaining.remove(&(start_x, start_y));
+
+			while let Some((x, y)) = queue.pop_front() {
+				island.insert((x, y));
+				for (dx, dy) in [(0, 1), (0, -1), (1, 0), (-1, 0)] {
+					let nx = x + dx;
+					let ny = y + dy;
+					if remaining.remove(&(nx, ny)) {
+						queue.push_back((nx, ny));
 					}
 				}
-				let end_x = x;
+			}
 
-				let corner1 = offset + DVec2::new(start_x as f64, y as f64);
-				let corner2 = offset + DVec2::new(end_x as f64, (y + 1) as f64);
-				vector.append_subpath(subpath::Subpath::new_rect(corner1, corner2), false);
-			} else {
-				x += 1;
+			let mut island_edges = HashSet::new();
+			for &(x, y) in &island {
+				for (p1, p2) in [((x, y), (x + 1, y)), ((x + 1, y), (x + 1, y + 1)), ((x + 1, y + 1), (x, y + 1)), ((x, y + 1), (x, y))] {
+					if !island_edges.remove(&(p2, p1)) {
+						island_edges.insert((p1, p2));
+					}
+				}
+			}
+
+			let mut adjacency: HashMap<(i32, i32), Vec<(i32, i32)>> = HashMap::new();
+			for (p1, p2) in island_edges {
+				adjacency.entry(p1).or_default().push(p2);
+			}
+
+			while let Some(&start_point) = adjacency.keys().next() {
+				let mut loop_points = Vec::new();
+				let mut current = start_point;
+
+				loop {
+					loop_points.push(DVec2::new(current.0 as f64, current.1 as f64));
+					let next = adjacency.get_mut(&current).and_then(|n| n.pop()).unwrap();
+					if adjacency.get(&current).map_or(false, |n| n.is_empty()) {
+						adjacency.remove(&current);
+					}
+					current = next;
+					if current == start_point {
+						break;
+					}
+				}
+
+				if loop_points.len() > 2 {
+					let mut simplified = Vec::new();
+					for i in 0..loop_points.len() {
+						let prev = loop_points[(i + loop_points.len() - 1) % loop_points.len()];
+						let curr = loop_points[i];
+						let next = loop_points[(i + 1) % loop_points.len()];
+
+						if (curr - prev).perp_dot(next - curr).abs() > 1e-6 {
+							simplified.push(curr + offset);
+						}
+					}
+					if !simplified.is_empty() {
+						vector.append_subpath(subpath::Subpath::from_anchors(simplified, true), false);
+					}
+				}
+			}
+		}
+	} else {
+		for y in 0..size {
+			for x in 0..size {
+				if qr_code.get_module(x, y) {
+					let corner1 = offset + DVec2::new(x as f64, y as f64);
+					let corner2 = corner1 + DVec2::splat(1.);
+					vector.append_subpath(subpath::Subpath::new_rect(corner1, corner2), false);
+				}
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


timon-schelling:
!build

timon-schelling:
I suggest adding a flag that allows merging adjacent tiles (should be on by default in my opinion).

Should be optional, because not merging can also be useful for example in this usecase.
image

Sahil-Gupta584:
Merged:
image
Unmerged:
image

(have renamed Merge to Merge Adjacent Tiles )

Looks good now ? @timon-schelling

timon-schelling:
That's just merging rows, not every adjacent tile. Can you fix that?

Sahil-Gupta584:
> That's just merging rows, not every adjacent tile. Can you fix that?
image

1. ig think all adjacent tiles(rows and columns) are merging already.
2. you mean there shouldnt be gap btn 4 boxed-adjacent tiles?

timon-schelling:
As you can see here in your screenshot, row are merged, collums are not merged.
I can tell because the side has indents.

TrueDoctor:
Could we just make the shape be boxes and then use a boolean union operation to merge them? Or we could just use a manual flood fill to build contiguous regions

Sahil-Gupta584:
I tried using flood fill approach ( ig first one will be heavier):

https://github.com/user-attachments/assets/5c494b89-816e-4cbb-b358-9fadff7636f3

timon-schelling:
!build

Keavon:
The node is currently nondeterministic. It produces different data every evaluation.

node-graph/nodes/vector/src/generator_nodes.rs
```diff
@@ -186,6 +186,117 @@ fn star<T: AsU64>(
 	Table::new_from_element(Vector::from_subpath(subpath::Subpath::new_star_polygon(DVec2::splat(-diameter), points, diameter, inner_diameter)))
 }
 
+/// Generates a QR code from the input text.
+#[node_macro::node(category("Vector: Shape"))]
```
Keavon:
```suggestion
#[node_macro::node(category("Vector: Shape"), name("QR Code"))]
```

node-graph/nodes/vector/src/generator_nodes.rs
```diff
@@ -186,6 +186,117 @@ fn star<T: AsU64>(
...
+fn qr_code(
+	_: impl Ctx,
+	_primary: (),
+	#[default("https://graphite.art")] text: String,
+	#[default(true)]
+	#[name("Merge Adjacent Tiles")]
+	merge: bool,
```
Keavon: Don't use `#[name(...)]` here, it should be used only to fix capitalization issues which isn't needed in this case.

Let's call this `individual_squares` which is the inverse of the current `merge`. It should be `false` by default, meaning the `#[default(true)]` line should be removed, since we normally want the squares to be merged by default unless the user chooses to have individual squares.

node-graph/nodes/vector/src/generator_nodes.rs
```diff
@@ -186,6 +186,117 @@ fn star<T: AsU64>(
...
+
+	let Ok(qr_code) = qrcodegen::QrCode::encode_text(&text, ecc) else {
+		return Table::default();
+	};
+
+	let size = qr_code.size();
+	let mut vector = Vector::default();
+	let offset = DVec2::splat(0.0);
+
+	if merge {
```
Keavon: Please move this algorithm into a helper function outside the node.

Keavon (changes requested):
See comments. The most critical one is making the node deterministic.

TrueDoctor:
@Sahil-Gupta584 Due to the use of a Hash Set which uses randomization under the hood, the flood fill algorithm will start from a different point each time, leading to visually identical but different output in each evaluation.  This should be relatively easy to fix since in some cases, the `HashSet` you used is just not necessary. You could for example replace the `HashSet` in initial calculation of the remaining pixels with a `Vec` etc., but please do yourself the favor of trying to understand the algorithm that is implemented and fix it without relying on ai tools for this task. The function is pretty sure and this might take a bit longer, but I think this would be a good learning opportunity

Sahil-Gupta584:
sure, thanks for the suggestions, looks interesting for me also!

Sahil-Gupta584:
So i tried making code more deterministic as well added few comments for understanding working.

timon-schelling:
!build

timon-schelling:
```
error[E0599]: no function or associated item named `new_rect` found for struct `vector_types::Subpath<PointId>` in the current scope
   --> node-graph/nodes/vector/src/generator_nodes.rs:208:46
    |
208 |                     vector.append_subpath(subpath::Subpath::new_rect(corner1, corner2), false);
    |                                                             ^^^^^^^^ function or associated item not found in `vector_types::Subpath<_>`
    |
```

image

not working as intended

timon-schelling:
@Sahil-Gupta584 are you interested in continuing this?

Sahil-Gupta584:
> @Sahil-Gupta584 are you interested in continuing this?

Yes of course, just need few more days more.

I will notify on discord after fixing ci and testing things.

Keavon:
!build

Keavon:
!build

TrueDoctor:
In our testing, the flood fill now seems to work, so I guess we should give this another round of review

timon-schelling:
I guess having the ecc setting be exposed users would be useful.

timon-schelling:
> I guess having the ecc setting be exposed to users would be useful.

I added this, I would be fine with merging. @TrueDoctor

timon-schelling:
!build

timon-schelling (approved):

node-graph/nodes/vector/src/merge_qr_squares.rs
```diff
@@ -0,0 +1,118 @@
...
+					}
+				}
+			}
+
+			// boundary detection
+			let mut outbound = vec![vec![0u8; size + 1]; size + 1];
+
+			for &(ix, iy) in &island {
+				if iy == 0 || remaining[iy - 1][ix] != 2 {
+					outbound[iy][ix] |= 1 << 0;
```
Ayush2k02, line 53: nit : consider adding constants here for directions

```
const DIR_TOP: u8 = 1 << 0;
const DIR_RIGHT: u8 = 1 << 1;
const DIR_BOTTOM: u8 = 1 << 2;
const DIR_LEFT: u8 = 1 << 3;
```

node-graph/nodes/vector/src/generator_nodes.rs
```diff
@@ -186,6 +186,51 @@ fn star<T: AsU64>(
...
+fn qr_code(
+	_: impl Ctx,
+	_primary: (),
+	#[default("https://graphite.art")] text: String,
+	/// Error correction level, from low (0) to high (3).
+	#[default(1)]
+	error_correction: u32,
+	#[default(false)] individual_squares: bool,
+) -> Table<Vector> {
+	let ecc = match error_correction.max(3) {
```
Ayush2k02: This should be

`let ecc = match error_correction.min(3)`

@TrueDoctor @timon-schelling @Keavon , let me know if I am wrong here , but since the intent here seems to be clamp the input to a valid range of 0 - 3 , 

by using max here , wouldn't it give wrong outputs

User Input: 0 (wants Low)
  → 0.max(3) = 3
**User gets HIGH instead of LOW**

User Input: 1 (wants Medium) [THE DEFAULT]
  → 1.max(3) = 3
**User gets HIGH instead of MEDIUM**

User Input: 2 (wants Quartile)
  → 2.max(3) = 3
**User gets HIGH instead of QUARTILE**

Ayush2k02 (changes requested):

node-graph/nodes/vector/src/generator_nodes.rs
TrueDoctor: Yes, you are right, the logic is currently wrong

timon-schelling: You are right, I used clamp before and simplified before pushing without testing again.

timon-schelling:
!build

Sahil-Gupta584:
Sorry guys couldnt be active much, having many things parallely.
Will try to update my other prs soon.

timon-schelling:
!build

Keavon:
Please leave a comment in #3565 so you can be assigned to that issue.

Sahil-Gupta584:
You mean in https://github.com/GraphiteEditor/Graphite/issues/3468 ? done!

Keavon:
Hi @Sahil-Gupta584, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

PR #3567, opened 2026-01-02
fix: move save state updates to SavedDocument handler

### LINKED ISSUE #3558: Cancelling document saving leads to content/modifications being lost
**Description:**
If cancel saving the document then there will be no warning about unsaved changes when closing the document. This can lead to losing modifications after the document is closed.

**Environment:**
The bug is happening in the desktop version of Graphite app (commit 9f25c50c) on MS Windows 11 Home (v10.0.26200).

**Steps to reproduce:**
1. Open the desktop version of Graphite app.
2. Create the new document.
3. Draw an ellipse in the document.
4. Press Ctrl+S to save the document. (-> the asterisk on the document header disappears already at this point, which is suspicious)
5. In the Save dialog, press the Cancel button.
6. Close the document.

Expected: Graphite app shows the popup with the suggestion to save the document.
Actual: Graphite app silently closes the document, so all the content in the document is lost.

The similar issue is happening when you open document, make changes, start saving it as another document, pressing the Cancel button in the Save dialog and closing the document -> all the changes in the document are lost.

theaniketgiri:
Fixes #3558 

### Problem

Previously, `set_save_state(true)` was called in the `SaveDocument` handler before the save dialog even appeared. This caused the document to be marked as saved even if the user cancelled the save dialog, leading to:
- Loss of the unsaved indicator (asterisk) in the document tab
- Users thinking their document was saved when it wasn't
- Potential loss of modifications if the user closed the document

### Solution

Moved the save state updates (`set_save_state`, `set_auto_save_state`) from the `SaveDocument` handler to the `SavedDocument` handler, which only executes after the file has actually been written to disk.

### Changes

- Removed premature `set_save_state(true)` and related calls from `SaveDocument` handler
- Added `set_save_state(true)` and `set_auto_save_state(true)` to `SavedDocument` handler
- Moved `UpdateOpenDocumentsList` response to `SavedDocument` handler

### Testing

- Verified compilation with `cargo check`
- The asterisk indicator now correctly remains visible when save is cancelled
- Document is only marked as saved after successful file write

---

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -1046,11 +1046,6 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 					self.path = None;
 				}
 
-				self.set_save_state(true);
-				responses.add(PortfolioMessage::AutoSaveActiveDocument);
-				// Update the save status of the just saved document
-				responses.add(PortfolioMessage::UpdateOpenDocumentsList);
-
 				responses.add(FrontendMessage::TriggerSaveDocument {
 					document_id,
 					name: format!("{}.{}", self.name.clone(), FILE_EXTENSION),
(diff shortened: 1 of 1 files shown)


timon-schelling:
Why the new workflow? That seems unrelated.
Why adding a pnpm lock? We don't use pnpm.

Converted to draft because this not ready for review.

theaniketgiri:
> Why the new workflow? That seems unrelated. Why adding a pnpm lock? We don't use pnpm.
> 
> Converted to draft because this not ready for review.
Hi @timon-schelling, thanks for the review!

You're absolutely right - both the workflow file and pnpm-lock were accidentally included. They were from a different branch I was working on and got mixed in when I created this PR.

I've fixed this now:

Rebased the branch onto master (2fa958aa)
Removed the unrelated workflow file
Removed the pnpm-lock file
The PR now only contains the intended changes

timon-schelling:
This breaks web functionality `DocumentMessage::SavedDocument` is only ever send from desktop.

PR #3568, opened 2026-01-02
Clean up DocumentMessageHandler by breaking out several lengthy handlers into helper functions

### LINKED ISSUE #2740: Tracking: Code Quality
## Match statements
Some files contain a single match statement that are very long (e.g. portfolio 1356 lines, document 1335 lines, select 1070, path 799). This is rather hard to manage or debug.
- Split message handlers
- Move complex logic outside of the match statement

## Cloning the document
History / transactions involve cloning the document many times. This very bad for performance and completely inflexible [#2617](https://github.com/GraphiteEditor/Graphite/issues/2617) . The [issue from 2021](https://github.com/GraphiteEditor/Graphite/issues/123) still hasn't had any progress due to the enormous refactoring that would be involved in making a system with proper actions. The current system also creates much confusion when history interacts with tools (especially the pen tool [#2262](https://github.com/GraphiteEditor/Graphite/issues/2262).
- Write a system whereby actions and their reverse can be stored instead. This will require a major refactor. I don't really want to undertake this whilst so many of the systems are in flux

## UI
UI code is verbose and poorly structured. The current system is partially immediate mode, using all of our custom types. All of the UI is then diffed in a rather fragile way. The callback system is also rather confusing causing issues like [#2530](https://github.com/GraphiteEditor/Graphite/issues/2530).
- Rewrite using a better UI library e.g. iced. Keavon wants to adopt some research project but I'm not sure this is a good idea.
- Make the API for the UI layout more like egui to improve simplicity.

## Transforms & Click targets
Transforms are completely broken in many situations. Especially with the node graph allowing transform nodes in any position. There are a series of inconsistent hacks that work around this is certain common situations. However this isn't particularly ideal. The Instances<T> only makes this more confusing. I'm not really sure how to solve this.

Click targets are very confusingly managed. This includes the path tool without a path node. This also interacts with the transform system in ways that aren't particularly robust.

The StartBuffer thing is exceedingly confusing. Each time there is an issue involving it, several hours must be spent trying to figure out what to do about it.

The relationship where the editor updates are theoretically independent of the node graph executor is very confusing since most tools rely on the node grpah being run at least most of the time. This is especially confusing when implementing tests.

I feel like the current systems and the way they interact with the node graph is very hacky. I think it would be best to discuss the requirements for the system to figure out why we have ended up in such a mess.

## Tools
- Some tools have a confusing amount of state. For example PenToolData has 31 member variables which are only set is certain situations. This has lead to many bugs.
- FSM states are inconsistent and confusing. For example the select tool fsm state for `Ready` contains the `NestedSelectionBehaviour` and the `Dragging` state contains the `has_dragged: bool` for some reason?

I feel like some of the tools need a rewrite @Keavon. However there is always a lot of random churn.

## Inputs
Rebinding inputs has been requested for a long time. However the input system has thousands of lines of macros involving many repeated things that would require a significant amount of refactoring to make a system that could be exposed to the user.

## Translations
Translations are basically impossible to implement now since there are many thousands of string literals scattered all over the code including gcore, editor, frontend svelte etc.

## Multiple documents
Having multiple documents visible at once seems to be a goal at some point. It is probably very useful for references / having a zoomed out overview etc. However we have hardcoded the active document many many times so it is basically impossible to refactor this.

## SVG
Poor integration with SVG. SVG imports are all flattened to a path rather than maintaining any structure at all. Exports end up as a mess of nested groups that is mostly unusable.

Vector editing performance is very bad with large documents (e.g. try dragging a layer in the red dress demo artwork).

theaniketgiri:
Partly closes #2740 

This PR addresses the "Match statements" section of the tracking issue by extracting three large match arms from `process_message()` into dedicated helper methods.

### Changes

Extracted the following helper methods in `DocumentMessageHandler`:

| Method | Lines Extracted | Purpose |
|--------|-----------------|---------|
| `handle_group_selected_layers` | ~50 lines | Handles layer grouping in both artboard and non-artboard workflows |
| `handle_move_selected_layers_to` | ~85 lines | Handles layer movement with proper transform preservation |
| `handle_nudge_selected_layers` | ~75 lines | Handles keyboard nudging with optional resize mode |

### Benefits

- **Improved readability:** The main match statement is now ~210 lines shorter
- **Better maintainability:** Each operation is now self-contained in its own method
- **Easier testing:** Helper methods can potentially be unit tested independently
- **Clearer intent:** Method names describe what the code does

### Before/After

**Before:**
```rust
DocumentMessage::GroupSelectedLayers { group_folder_type } => {
    // ~50 lines of inline code
}
```

**After:**
```rust
DocumentMessage::GroupSelectedLayers { group_folder_type } => {
    self.handle_group_selected_layers(group_folder_type, responses);
}
```

### Testing

- Verified compilation with `cargo check`
- Logic and behavior remain unchanged - this is a pure refactor
- Existing tests pass (note: SIGSEGV in tests is a pre-existing issue on master)

---

### DIFF
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -623,145 +623,10 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 				responses.add(OverlaysMessage::Draw);
 			}
 			DocumentMessage::GroupSelectedLayers { group_folder_type } => {
-				responses.add(DocumentMessage::AddTransaction);
-
-				let mut parent_per_selected_nodes: HashMap<LayerNodeIdentifier, Vec<NodeId>> = HashMap::new();
-				let artboards = LayerNodeIdentifier::ROOT_PARENT
-					.children(self.metadata())
-					.filter(|x| self.network_interface.is_artboard(&x.to_node(), &self.selection_network_path))
-					.collect::<Vec<_>>();
-				let selected_nodes = self.network_interface.selected_nodes();
-
-				// Non-artboard (infinite canvas) workflow
-				if artboards.is_empty() {
-					let Some(parent) = self.network_interface.deepest_common_ancestor(&selected_nodes, &self.selection_network_path, false) else {
-						return;
-					};
-					let Some(selected_nodes) = &self.network_interface.selected_nodes_in_nested_network(&self.selection_network_path) else {
-						return;
-					};
-					let insert_index = DocumentMessageHandler::get_calculated_insert_index(self.metadata(), selected_nodes, parent);
-
-					DocumentMessageHandler::group_layers(responses, insert_index, parent, group_folder_type, &mut self.network_interface);
-				}
-				// Artboard workflow
-				else {
-					for artboard in artboards {
-						let selected_descendants = artboard.descendants(self.metadata()).filter(|x| selected_nodes.selected_layers_contains(*x, self.metadata()));
-						for selected_descendant in selected_descendants {
-							parent_per_selected_nodes.entry(artboard).or_default().push(selected_descendant.to_node());
-						}
-					}
-
-					let mut new_folders: Vec<NodeId> = Vec::new();
-
-					for children in parent_per_selected_nodes.into_values() {
-						let child_selected_nodes = SelectedNodes(children);
-						let Some(parent) = self.network_interface.deepest_common_ancestor(&child_selected_nodes, &self.selection_network_path, false) else {
-							continue;
-						};
-						let insert_index = DocumentMessageHandler::get_calculated_insert_index(self.metadata(), &child_selected_nodes, parent);
-
-						responses.add(NodeGraphMessage::SelectedNodesSet { nodes: child_selected_nodes.0 });
-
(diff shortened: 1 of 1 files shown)


timon-schelling:
Why the new workflow? That seems unrelated.

Converted to draft because this not ready for review.

theaniketgiri:
> Why the new workflow? That seems unrelated.
> 
> Converted to draft because this not ready for review.

Hi @timon-schelling, thanks for the review!

You're absolutely right - both the workflow file and pnpm-lock were accidentally included. They were from a different branch I was working on and got mixed in when I created this PR.

I've fixed this now:

Rebased onto master
Removed the unrelated workflow file and pnpm-lock
The PR now only contains the refactoring changes

Keavon (approved):

Keavon:
Hi @theaniketgiri, thanks again for this code contribution to the project! We're still hoping you will respond to the request to relicense this code. Please see #4208 ASAP, thank you!

Keavon:
Hi @theaniketgiri, we're grateful of your past code contribution to Graphite, but unless we hear from you by the **end of this weekend** with agreement about relicensing your code to include the MIT license, we will begin steps to remove your code contributions **so that you will no longer be a Graphite code contributor**. We'd much rather avoid this outcome, so please respond now.

PR #3573, opened 2026-01-05
Fix Line Tool doesn't reflect endpoint changes made by Path Tool

### LINKED ISSUE #2686: Line Tool doesn't reflect endpoint changes made by Path Tool
If we create a line using the Line Tool and then move one of its endpoints using the Path Tool, switching back to the Line Tool still shows the original line (before the endpoint was moved).

https://github.com/user-attachments/assets/5ddd2afb-e81c-491b-b6b4-0d6a6de66e52

Annonnymmousss:
Closes #2686 

https://github.com/user-attachments/assets/82133a97-657d-4511-be8e-ca9f0788d307


Keavon:
While it now draws the line in the expected position, I cannot actually drag the line endpoints around with the Line tool if they have been previously moved by the Path tool.

Annonnymmousss:
@Keavon kindly see the changes ..
The Line Tool was reading endpoint data from two different sources.

PR #3579, opened 2026-01-05
use Self as much as possible, refactoring

adamnemecek:


timon-schelling:
Generally a good idea, but there are cases where using self makes the code harder to read. I suggest you split this into two commits, one with all changes that make the code easier to read the other with all changes that at least don't make it harder to read and leave out the rest.

Keavon:
Closing since no response was given by the author in a month.

PR #3585, opened 2026-01-06
Fix freehand tool showing all anchors in overlay

### LINKED ISSUE #3581: Freehand Tool displays ALL points by default (Regression)
# Issue
Using the freehand tool on the latest build shows a square selection overlay for every point being sampled, but in the last stable release, it only shows overlays for the start and end points, until selected with the path tool.

Image

This looks pretty awkward, and while it doesn't *ruin* the freehand tool, it makes it much worse to use. 

# Basic diagnosis
I ran git bisect and it looks like the issue stems from [commit c2730bd3](https://github.com/GraphiteEditor/Graphite/commit/c2730bd3a6b0d76e29919dfcde7871ad0a96295f). I also manually confirmed that the issue exists on this commit, but not the one before it.

Ayush2k02:
https://github.com/user-attachments/assets/902bf5a8-ae15-4303-803b-171d868115db

Closes #3581

### DIFF
--- a/.github/workflows/build-linux-bundle.yml
+++ b/.github/workflows/build-linux-bundle.yml
@@ -4,7 +4,7 @@ on:
   workflow_dispatch: {}
   push:
     branches:
-      - main
+      - master
 
 jobs:
   build:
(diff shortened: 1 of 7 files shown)


timon-schelling:
!build

PR #3613, opened 2026-01-08
Add artboard name to the export when exporting a single artboard

### LINKED ISSUE #3608: Export files use artboard name when specified
When the bounds of an export is set to a specific artboard (rather than All Artwork) the exported file should be named after the artboard name, rather than the file name.

I'm using artboards on Graphite to group separate but related drawings, and I got annoyed having to rename my exported files to be the same as my artboards so they wouldn't all end up as ProjectName (3) or replace each other.

My proposal is that when the bounds of an export is set to a specific artboard, the exported file should be named after the artboard name, rather than the file name.

Possible alternatives include making the export for artboards follow the "ProjectName - Artboard" format, or even checking if the artboard is named "Artboard" before applying naming logic. I'm not aware of how other apps handle this, and discussion is welcome.

Ayush2k02:
https://github.com/user-attachments/assets/f3513197-f828-4188-b9fb-3e5a6db4e4d8

I am not sure what the file name should be when this below selection option is enabled , so just using the document name for now

Screenshot 2026-01-08 at 1 53 38 PM

CLOSES #3608

### DIFF
--- a/editor/src/node_graph_executor.rs
+++ b/editor/src/node_graph_executor.rs
@@ -252,10 +252,9 @@ impl NodeGraphExecutor {
 			FileType::Png => "png",
 			FileType::Jpg => "jpg",
 		};
-		let base_name = if let Some(artboard_name) = artboard_name {
-			if artboard_count > 1 { format!("{name} - {artboard_name}") } else { name }
-		} else {
-			name
+		let base_name = match (artboard_name, artboard_count) {
+			(Some(artboard_name), count) if count > 1 => format!("{name} - {artboard_name}"),
+			_ => name,
 		};
 		let name = format!("{base_name}.{file_extension}");
 


Ayush2k02:
FIxed the video !

Keavon:
I think we want to go with `Document Name - Artboard Name` and, if only one artboard exists, just `Document Name`.

Ayush2k02:
https://github.com/user-attachments/assets/ac032571-cd3d-43e6-897d-3c40e42eebc4


After the recent commit

editor/src/node_graph_executor.rs
```diff
@@ -250,7 +252,12 @@ impl NodeGraphExecutor {
 			FileType::Png => "png",
 			FileType::Jpg => "jpg",
 		};
-		let name = format!("{name}.{file_extension}");
+		let base_name = if let Some(artboard_name) = artboard_name {
```
Ayush2k02: can have a switch statement here as well ,  something like

```
let base_name = match (artboard_name, artboard_count) {
			(Some(artboard_name), count) if count > 1 => format!("{name} - {artboard_name}"),
			_ => name,
```

Ayush2k02: think it'd be more cleaner !

Ayush2k02: Done

Ayush2k02:
Ready for review from side

TrueDoctor (approved):

PR #3617, opened 2026-01-09
fix #3604 Apply Transform node produces the wrong result on shapes with handles

### LINKED ISSUE #3604: Apply Transform node produces the wrong result on shapes with handles
Just add an *Apply Transform* node to a shape and it becomes messed up unless it has a (0, 0) translation if the shape has handles. If it's only a polyline, it is unaffected (the polygon at the end of the video below was affected until #3606 was fixed).

https://github.com/user-attachments/assets/42421c1c-01ae-4017-a5ee-179c5b6ec298

The Apply Transform node is supposed to keep the geometric form the same, just changing its geometry and transform space.

I don't recall this being broken in the past, so it might be a regression.

chintu4:
#### Issue #3604 
The issue #3604 (Apply Transform node produces the wrong result on shapes with handles) i can see in the given  video that
#### steps to reproduce issue.
1. Creating an ellipse shape using ellipse tool 
2. Selected magic button (used to apply operations on end of the layer chain.) in the right side in Properties panel.
3. Choose apply transform
4. the ellipse shape in the Artboard gets distorted.

#### fix
At "node-graph\nodes\vector\src\vector_modification_nodes.rs" file ,  apply_transform is function which is used internally in node-graph system this function shows no visible change after addition of control handler code in apply transform function.
I have added a 3 line in apply_tranform function so that tranformation from geometry to vector happen both on Achor and handlers. currently only anchor points are being handled .
 ```rust 
//control handler code 
for (handles, _start_index, _end_index) in vector.segment_domain.handles_and_points_mut() {
			*handles = handles.apply_transformation(|p| transform.transform_point2(p));
		}
``` 
the existing code only handles anchor points 
```rust 
//existing code
for (_, point) in vector.point_domain.positions_mut() {
    *point = transform.transform_point2(*point);
}
``` 
#### Verification 
I have compiled and run this project on Linux machine. I have saw that my change doesn't show any visible change in Artboard transformation happens internally. Iam getting feedback message that "The "node has no parameter" i expect this is a expected behaviour.
Here is the verification video

https://github.com/user-attachments/assets/2cd1a100-1c41-45da-8834-042865a188e5


node-graph/nodes/vector/src/vector_modification_nodes.rs
```diff
@@ -32,10 +32,14 @@ async fn apply_transform(_ctx: impl Ctx, mut vector: Table<Vector>) -> Table<Vec
 	for row in vector.iter_mut() {
 		let vector = row.element;
 		let transform = *row.transform;
-
+		//iterating over points to apply the transformation(geometry to vector space)
 		for (_, point) in vector.point_domain.positions_mut() {
 			*point = transform.transform_point2(*point);
 		}
+		//iterating over segment handles to transform them as well(geometry to vector space)
```
timon-schelling, line 39:
```suggestion
		// Iterating over segment handles to apply the transformation (geometry to vector space)
```

node-graph/nodes/vector/src/vector_modification_nodes.rs
```diff
@@ -32,10 +32,14 @@ async fn apply_transform(_ctx: impl Ctx, mut vector: Table<Vector>) -> Table<Vec
 	for row in vector.iter_mut() {
 		let vector = row.element;
 		let transform = *row.transform;
-
+		//iterating over points to apply the transformation(geometry to vector space)
```
timon-schelling, line 35:
```suggestion
		// Iterating over points to apply the transformation (geometry to vector space)
```

timon-schelling:
!build

chintu4:
Is this pr got merged @timon-schelling

Keavon:
Thank you for your attempt at this, hopefully you learned from the process. I ended up with a simpler, one-line solution that I just merged in #3687 (feel free to take a look) that is preferable to the three-line + extraneous comments proposed in this PR. Please feel free to continue submitting PRs and they will each be considered, although this one didn't end up being the approach we went with. Thanks again for taking the time and learning along the way.

PR #3618, opened 2026-01-09
Fix Exports having 0 layers

### LINKED ISSUE #3422: Export fails for documents with vector data direcly connected to canvas
https://github.com/user-attachments/assets/cf6b00d4-2cc2-49f0-9449-196584864ba6

Sahil-Gupta584:
Closes #3422

Preview:

https://github.com/user-attachments/assets/c9a12b22-24a2-40f0-bc9a-e2d1ac6617dc

Since we dont have any layer to export, only an infinite canvas is left though which we can calculate the bounding box.

I  tried to create bounding box around extreme drawing points, but still as you can see the final exported image in preview is weird, is it ok ?or what i am thinking is when export mode is all-artwork i can apply 10% (or more) padding to it.


editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -1190,7 +1191,34 @@ impl NodeNetworkInterface {
...
+			.iter()
+			.filter_map(|export| {
+				if let NodeInput::Node { node_id, .. } = export {
+					if !self.is_layer(node_id, &[]) {
+						return self.document_metadata().bounding_box_document(LayerNodeIdentifier::new_unchecked(*node_id));
+					}
+				}
+				None
+			})
+			.reduce(Quad::combine_bounds);
```
0HyperCube: This doesn't do anything since only layers are put into the bounding box. This is always `None`.

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -1190,7 +1191,34 @@ impl NodeNetworkInterface {
 				}
 				self.document_metadata.bounding_box_document(layer)
 			})
-			.reduce(Quad::combine_bounds)
+			.reduce(Quad::combine_bounds);
+
+		let root_artwork_bounds = self.document_metadata().bounding_box_document(LayerNodeIdentifier::ROOT_PARENT);
```
0HyperCube, line 1213: The `transform_to_viewport` doesn't work for the `LayerNodeIdentifier::ROOT_PARENT` for legacy reasons. This causes the bounds here to be incorrect.

https://github.com/GraphiteEditor/Graphite/blob/1b91198b28ddb5648c51a3754f2aecf75c32eafe/editor/src/messages/portfolio/document/utility_types/document_metadata.rs#L72-L76

0HyperCube (changes requested):

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1406,17 +1406,11 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 				self.network_interface.update_first_element_source_id(first_element_source_id);
 			}
 			DocumentMessage::UpdateClickTargets { click_targets } => {
-				// TODO: Allow non layer nodes to have click targets
```
adamgerhant: Dont remove todo comments

editor/src/messages/portfolio/document/document_message_handler.rs
```diff
@@ -1406,17 +1406,11 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
...
 				let layer_click_targets = click_targets
 					.into_iter()
-					.filter(|(node_id, _)|
```
adamgerhant: This doesn't need to be changed

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -1207,7 +1208,16 @@ impl NodeNetworkInterface {
...
```
adamgerhant: The current logic here is correct. We find the bounding box of all layers in the document. The ROOT_PARENT is just a conceptual placeholder used for node positioning.

adamgerhant:
This issue will be solved when click targets are returned for all data in a Table, since every graphic click target will be accessible in the editor, not just ones that pass though layers.

PR #3625, opened 2026-01-12
Desktop: Add fullscreen window mode

### LINKED ISSUE #3510: Desktop fullscreen mode with F11
Equivalent to clicking the green button on Mac, but without hiding the title bar. The maximize/unmaximize window button can show the unmaximize icon and be mapped to un-fullscreen first before returning to normal functionality.

timon-schelling:
f11 is not usable on mac for this, browsers use cmd + strg + f.
now opted for cmd + strg + f on mac.

Closes #3510

#3656 Should be merged right after this to fix
https://github.com/GraphiteEditor/Graphite/pull/3625#discussion_r2700495371

### DIFF
--- a/editor/src/messages/input_mapper/utility_types/misc.rs
+++ b/editor/src/messages/input_mapper/utility_types/misc.rs
@@ -41,19 +41,6 @@ impl Mapping {
 			InputMapperMessage::PointerShake => &self.pointer_shake,
 		}
 	}
-
-	fn associated_entries_mut(&mut self, message: &InputMapperMessage) -> &mut KeyMappingEntries {
-		match message {
-			InputMapperMessage::KeyDown(key) => &mut self.key_down[*key as usize],
-			InputMapperMessage::KeyUp(key) => &mut self.key_up[*key as usize],
-			InputMapperMessage::KeyDownNoRepeat(key) => &mut self.key_down_no_repeat[*key as usize],
-			InputMapperMessage::KeyUpNoRepeat(key) => &mut self.key_up_no_repeat[*key as usize],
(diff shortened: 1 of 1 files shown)


timon-schelling:
@Keavon I don't think you need to do a full review. Desktop changes are really strait forward. But a opinion on macro changes would be helpful.

Keavon:
- [x] When hitting F11 after being in fullscreen, we get an inset window: 
image

- [x] The title bar should show the Exit Fullscreen button in place of Minimize/Maximize/Close when fullscreen
- [x] Replace the input macro's new attribute named `active` with its inverse named `disabled`
- [x] Clean up the input macro by replacing its patching logic used for scroll preferences with `disabled`

editor/src/messages/input_mapper/input_mappings.rs
```diff
@@ -17,16 +17,19 @@ use glam::DVec2;
...
 	}
 }
 
-pub fn input_mappings() -> Mapping {
+pub fn input_mappings(zoom_with_scroll: bool) -> Mapping {
 	use InputMapperMessage::*;
 	use Key::*;
 
+	// TODO: Fix this failing to load the correct data (and throwing a console warning) because it's occurring before the value has been supplied during initialization from the JS `initAfterFrontendReady`
+	let keyboard_platform = GLOBAL_PLATFORM.get().copied().unwrap_or_default().as_keyboard_platform_layout();
```
Keavon, line 31: @timon-schelling this is a regression. Here's the error mentioned in this TODO comment I added for you:

Image

This will work when the warning is gone and you can use the web version on Mac and see *Window* > *Fullscreen* has the Mac-specific shortcut instead of <kbd>F11</kbd> which it currently shows.

PR #3628, opened 2026-01-12
Fix Text node/tool max width/height bounding box calculation

### LINKED ISSUE #3626: The box area of text is displayed wrong if EITHER max width or max height is active, but works if both are
This used to work, so it appears to be a regression from some time ago. It's old enough that it is still broken as far back as https://editor.graphite.art's latest September deploy.

1. Click (don't drag) with the Text tool and type something
2. In the Properties panel, enable the checkbox for Max Width and give it a value larger than the text
3. We should be seeing the box around the text (a dashed line with the Select tool active, a solid line with the Text tool active) update live with the Max Width, but instead we are only seeing it clamped to the rendered bounding box of the actual text
4. If you set the Alignment parameter to Right, and the Max Width is wider than the text within, you'll see it align to the right beyond the box area, because the box *should* be that far to the right to encompass it when right-aligned
5. The same occurs if only Max Height is active
6. But if we enable both Max Width and Max Height at the same time, it now behaves as expected where the bounds reflect their given values

jsjgdh:
This PR fixes a bug in which the text bound box didn't work as intended   when only one of Max Width or Max Height were set. The box would wrongly clamp to the text's rendered size instead of the specified maximum dimension. Fixes #3626.

before :- 
Screenshot 2026-01-12 212044

Now:-
Screenshot 2026-01-12 212220

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2216,6 +2216,7 @@ dependencies = [
  "text-nodes",
  "tokio",
  "url",
+ "vector-nodes",
  "wasm-bindgen",
  "web-sys",
  "wgpu-executor",
--- a/editor/src/messages/portfolio/document/node_graph/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_properties.rs
@@ -25,6 +25,7 @@ use graphene_std::raster::{
 use graphene_std::table::{Table, TableRow};
 use graphene_std::text::{Font, TextAlign};
 use graphene_std::transform::{Footprint, ReferencePoint, Transform};
+use graphene_std::vector::QRCodeErrorCorrectionLevel;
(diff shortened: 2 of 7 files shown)


Keavon:
!build

node-graph/nodes/text/src/text_context.rs
```diff
@@ -107,15 +112,23 @@ impl TextContext {
...
-
 		let Some(layout) = self.layout_text(text, font, font_cache, typesetting) else {
 			return DVec2::ZERO;
 		};
 
-		DVec2::new(layout.full_width() as f64, layout.height() as f64)
+		let layout_width = layout.full_width() as f64;
+		let layout_height = layout.height() as f64;
+
+		// For clipping tests use of the actual layout dimensions to see if text overflows
```
Ayush2k02: nit : should be " use the actual " instead of " use of the " , minor typo nothing major :)

Keavon (approved):

PR #3630, opened 2026-01-13
fix: added a len check before draining

### LINKED ISSUE #3629: Crash when hiding an 'Instance Position' or 'Pointer Position' node feeding a value
Set up this simple graph:

Image

- A 'Vec2 Value' node instead

mohan-bee:
fixes #3629 

issue: node.inputs tries to drain from the index 1 which causes the crash 
fix: added a len check before draining 

before: 

https://github.com/user-attachments/assets/798f73f0-887a-4f65-b456-88c30406b60c

after: 

https://github.com/user-attachments/assets/15806cf6-d2db-4ae5-8a90-95364ca3e4d2


mohan-bee:
@adamgerhant  review pls !

Keavon:
When we do this with a *Vec2 Value*, this error is the expected outcome:

image

In your video, we see something different:

image

This is not the expected behavior. It likely means your initial approach is masking the underlying problem instead of tracing back to its root cause.

mohan-bee:
@Keavon  i have added default unit value input so that the identity node can function properly.
image

node-graph/graph-craft/src/document.rs
```diff
@@ -791,7 +791,12 @@ impl NodeNetwork {
 			node.implementation = identity_node;
 
 			// Connect layer node to the group below
-			node.inputs.drain(1..);
+			if node.inputs.len() > 1 {
+				node.inputs.drain(1..);
+			} else if node.inputs.is_empty() {
+				node.inputs.push(NodeInput::value(TaggedValue::None, false));
```
adamgerhant, line 797: If a node with no inputs is hidden, the identity node value should be based on the output type. This can be represented by changing the `visible` field on `DocumentNode` to store an enum like
```
pub enum Visible {
    Passthrough,
    Value(Type),
}
```
Then in the editor, when hiding the node, if the node has no inputs it gets set the the Value variant with the Type based on the output type of that node. This Type is then used to construct the tagged value, instead of using TaggedValue::None.

adamgerhant (changes requested):

timon-schelling:
This doesn't really fix the problem. The review comment explains it well.

PR #3641, opened 2026-01-15
Add F2 shortcut for renaming the selected layer

### LINKED ISSUE #3636: Keyboard shortcut for renaming layer
This would help a lot (for a lot of users I guess). Photoshop and Illustrator at lest have a mechanism, where you start to edit a layer name and you press tab the cursor jumps to the next layer name (shift tab for backwards) and highlights that. This would be amazing but using Asprite for a while using the F2 button works like a charm. (There you also jump up and down from layers with the arrow buttons. This would be neat here as well with a combo button)

Ayush2k02:
This PR adds the following shortcuts , although in my opinion since graphite has layers stacked vertically , there is **no point in keeping both the tab , shft tab and the arrow keys option**, since both of them does the same job

https://github.com/user-attachments/assets/c60a75b3-2b0f-438e-bc63-4933141f3ad3

https://github.com/user-attachments/assets/85248bfa-0bad-46c4-95a3-b118f1e29265

CLOSES #3636

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_properties.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_properties.rs
@@ -4,7 +4,7 @@ use super::document_node_definitions::{NODE_OVERRIDES, NodePropertiesContext};
 use super::utility_types::FrontendGraphDataType;
 use crate::messages::layout::utility_types::widget_prelude::*;
 use crate::messages::portfolio::document::node_graph::document_node_definitions::resolve_document_node_type;
-use crate::messages::portfolio::document::utility_types::network_interface::InputConnector;
+use crate::messages::portfolio::document::utility_types::network_interface::{InputConnector, NodeNetworkInterface};
 use crate::messages::portfolio::utility_types::{FontCatalogStyle, PersistentData};
 use crate::messages::prelude::*;
 use choice::enum_choice;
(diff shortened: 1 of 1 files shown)


frontend/src/components/panels/Layers.svelte
```diff
@@ -655,8 +709,22 @@
 							placeholder={listing.entry.reference.data}
 							disabled={!listing.editingName}
 							on:blur={() => onEditLayerNameDeselect(listing)}
-							on:keydown={(e) => e.key === "Escape" && onEditLayerNameDeselect(listing)}
-							on:keydown={(e) => e.key === "Enter" && onEditLayerNameChange(listing, e)}
+							on:keydown={(e) => {
```
Ayush2k02, line 620: consolidated these handlers into one , have verified and all are working

Ayush2k02:
Any update on this guys !

Keavon:
!build

frontend/src/components/panels/Layers.svelte
```diff
@@ -491,6 +492,58 @@
...
+				onEditLayerName(selectedLayer);
+			}
+		}
+	}
+
+	function isAnyLayerBeingEdited(): boolean {
+		return layers.some((layer) => layer.editingName);
+	}
+
+	async function navigateToLayer(currentListing: LayerListingInfo, direction: "next" | "previous" | "up" | "down") {
```
Keavon: I don't think there's any reason to have four instead of two directions, when each pair are just synonyms of each other. Just name them "Up" and "Down", with the first letter capitalized because these are the JS equivalent of Rust's enum variants which are also PascalCase.

frontend/src/components/panels/Layers.svelte
```diff
@@ -491,6 +492,58 @@
...
+			if (selectedLayer) {
+				e.preventDefault();
+				onEditLayerName(selectedLayer);
+			}
+		}
+	}
+
+	function isAnyLayerBeingEdited(): boolean {
+		return layers.some((layer) => layer.editingName);
+	}
```
Keavon: A one-liner function that's only used once is rather unnecessary compared to just calling that logic inline and describing what's going on in the comment above it.

frontend/src/components/panels/Layers.svelte
```diff
@@ -491,6 +492,58 @@
...
+		// Calculate target index based on direction
+		let targetIndex: number;
+		if (direction === "next" || direction === "down") {
+			targetIndex = currentIndex + 1;
+			if (targetIndex >= layers.length) return; // Don't wrap around at the end
+		} else {
+			// previous or up
+			targetIndex = currentIndex - 1;
+			if (targetIndex < 0) return; // Don't wrap around at the beginning
+		}
```
Keavon: Move `targetIndex` declaration to a ternary statement and perform both returns, checking both conditions with a logical OR, on the following line. Full if/else blocks are too verbose here when this could all be made into two lines.

Ayush2k02:
Done ! applied the suggested changes, all of them made sense

Ayush2k02:
verified locally, working the same as before

Ayush2k02:
https://github.com/user-attachments/assets/d42284d9-16c9-4591-b68e-00ad027b48b5



Here is the recording after the commits, tested both tab , shift tab flow and the arrow keys

Keavon (approved):

PR #3343, opened 2025-11-04
Add an arrow to the Shape tool

krVatsal:
Add a new arrow shape tool

Closes #3317

PR #3650, opened 2026-01-17
Improve the Shape tool's arrow drawing controls

krVatsal:
Based on feedback on #3343

This PR adds configurable width controls for the Arrow shape tool and ensures arrow dimensions remain constant during drawing, with only the length changing.

### **Fixed Arrow Dimensions During Drawing**
Previously, all arrow dimensions scaled proportionally with the arrow length, causing the entire arrow to "zoom" as it was drawn. Now:
- **Width dimensions remain constant** (shaft, head width, head length stay fixed)
- **Only the arrow length changes** as you drag
- This creates a more natural drawing experience where the arrow stretches rather than scales

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2216,6 +2216,7 @@ dependencies = [
  "text-nodes",
  "tokio",
  "url",
+ "vector-nodes",
  "wasm-bindgen",
  "web-sys",
  "wgpu-executor",
@@ -4660,6 +4661,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "qrcodegen"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4339fc7a1021c9c1621d87f5e3505f2805c8c105420ba2f2a4df86814590c142"
+
+[[package]]
 name = "quick-error"
 version = "2.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6725,6 +6732,7 @@ dependencies = [
  "kurbo 0.12.0",
  "log",
  "node-macro",
+ "qrcodegen",
  "rand 0.9.2",
  "rustc-hash 2.1.1",
  "serde",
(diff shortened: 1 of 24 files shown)


krVatsal:
Hey @Keavon, can you guide me what could be the possible reason for this slower runtime after i updated the branch?

Keavon:
@krVatsal I will need more details to be able to help with that. Discord may be a better venue to ask for help about that.

editor/src/messages/tool/tool_messages/shape_tool.rs
```diff
@@ -906,7 +978,18 @@ impl Fsm for ShapeToolFsmState {
...
+					ShapeType::Arrow => Arrow::update_shape(
+						document,
+						input,
+						viewport,
+						layer,
+						tool_data,
+						modifier,
+						tool_options.arrow_shaft_width,
+						tool_options.arrow_head_width,
+						tool_options.arrow_head_length,
```
Keavon: This isn't how we pass the tool options to the other shapes. Too many arguments to this function, which the others don't suffer from.

editor/src/messages/tool/tool_messages/shape_tool.rs
```diff
@@ -58,6 +61,9 @@ impl Default for ShapeToolOptions {
 			spiral_type: SpiralType::Archimedean,
 			turns: 5.,
 			grid_type: GridType::Rectangular,
+			arrow_shaft_width: 14.0,
+			arrow_head_width: 32.0,
+			arrow_head_length: 28.0,
```
Keavon:
```suggestion
			arrow_shaft_width: 14.,
			arrow_head_width: 32.,
			arrow_head_length: 28.,
```

node-graph/libraries/vector-types/src/subpath/core.rs
```diff
@@ -330,14 +330,16 @@ impl<PointId: Identifier> Subpath<PointId> {
 		let head_base_distance = (length - head_length).max(0.);
 		let head_base = start + direction * head_base_distance;
 
+		// Arrow path starts at the tail (origin), traces around the shape, and returns to tail
```
Keavon:
```suggestion
		// Arrow path starts at the tail, traces around the shape, and returns to the tail
```

PR #3654, opened 2026-01-18
Fix duplicate shared upstream nodes when duplicating layers

### LINKED ISSUE #2501: Edit->Duplicate in the node graph makes extra duplicates of reused input nodes
**Repro steps:**
 
1. Create a project with one Text node supplying input to two layers (or use the attached project).
2. Drag-select the two layers and their input nodes.
   * In the attached project, use the bottom two layers, with the text "Shut up and take my money!"
3. Choose Edit -> Duplicate.
   * It seems necessary to use the menu option, NOT the Ctrl-D keyboard shortcut, as odd as that sounds.

**Expected result:** The selected nodes are duplicated exactly once, adding two new layers.

**Observed result:** In _addition_ to the expected result, we now have one copy of the original Text node for each of the original two layers. This is almost impossible to notice since they overlap perfectly. See below for a more detailed explanation.

**Suggested fix:**
* Don't make extra copies of the re-used input nodes.
* For extra points, systemically prevent the case of nodes ever having exactly the same x and y coordinates in the node graph. When this happens it is REALLY confusing.

**Details:** In the attached project, I was attempting to create a "shadow" or outline for some meme text, using multiple layers. I reused the Text and Transform nodes for both layers using the node graph editor (see screenshot).

Later, I wanted a second line of text, so I box-selected my two layers and their input nodes and chose Edit -> Duplicate. This mostly worked, apart from some messed up wiring.

Even later, I tried to edit the top line of text and was confused as to why my edits were applying only to the white text and not the black text shadow:

Image

Eventually I realised the topmost Text node is actually TWO Text nodes, perfectly overlapping so it's impossible to see. Here I've moved them slightly to demonstrate (note, the graph structure is identical to the above image):

Image

I can reproduce this by similarly duplicating the bottom two layers with the text "Shut up and take my money".

[invalidate-cache.zip](https://github.com/user-attachments/files/19523590/invalidate-cache.zip)

afrdbaig7:
This PR fixes an issue where duplicating multiple layers that share the same input Text node would create multiple duplicate Text nodes instead of reusing a single one. The problem occurred because shared upstream nodes added during traversal weren’t being tracked, allowing the same node to be collected more than once. The fix ensures each upstream node is handled only once while preserving the existing traversal order, so duplicating shared layer chains now behaves as expected. 

Fixes #2501.

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -2690,15 +2690,20 @@ impl NodeNetworkInterface {
 			return Vec::new();
 		};
 		let mut all_selected_nodes = selected_nodes.selected_nodes().cloned().collect::<Vec<_>>();
+		let mut seen = all_selected_nodes.iter().cloned().collect::<HashSet<_>>();
+
 		for selected_node_id in selected_nodes.selected_nodes() {
 			if self.is_layer(selected_node_id, network_path) {
-				let unique_upstream_chain = self
+				let upstream_chain = self
 					.upstream_flow_back_from_nodes(vec![*selected_node_id], network_path, FlowType::HorizontalFlow)
 					.skip(1)
-					.take_while(|node_id| self.is_chain(node_id, network_path))
-					.filter(|upstream_node| all_selected_nodes.iter().all(|new_selected_node| new_selected_node != upstream_node))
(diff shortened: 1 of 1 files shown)


adamgerhant:
!build ([Run ID 22248322545](https://github.com/GraphiteEditor/Graphite/actions/runs/22248322545))

PR #3656, opened 2026-01-18
Replace globals with editor environment

timon-schelling:
Depends on #3625


TrueDoctor:
Why does this still use interior mutability? I thought you wanted to make the globals message handler no longer use interior mutability and instead just pass it around. Is there a reason why it is necessary to do it like this?

Keavon:
Globals are still required by our trait impls, like `Display`.

timon-schelling:
*not required, but fixing that would include wrapping a bunch of structs and enums in wrappers that include the environment info just to be ale to impl display and similar std traits. In my opinion not worse it, but it ca be done, don't know if it would be cleaner though.

desktop/src/app.rs
```diff
@@ -50,7 +51,7 @@ impl App {
 		Window::init();
 	}
 
-	pub(crate) fn new(
+	pub(crate) fn create(
```
TrueDoctor: What was the reason for the rename of this function? If it is to indicate that it has side effects, I'm not sure if that is aided by the new name and wouldn't be better expressed as a doc comment.

timon-schelling: Sure can do that

PR #3659, opened 2026-01-20
Add selection tolerance

### LINKED ISSUE #2141: Small click target when dragging/moving a straight line or small objects

Nuhiat-Arefin:
Fixes #2141 
Modified the intersect_point function in click_target.rs. If the stroke (as measured in viewport space) is very thin(less than 5px) it calculates inflated_stroke such that the stroke as measured in viewport space is 5px. Otherwise, it keeps the value of the original stroke.
Before:

https://github.com/user-attachments/assets/cb09e06e-2dc6-4c27-8536-4239c8ff7428

After:

https://github.com/user-attachments/assets/5c6bf3cd-4eff-46e6-b8ec-8cb34b564160

With larger inflation:

https://github.com/user-attachments/assets/a51fc123-1724-48ee-84ef-9e996c19eb8d

### DIFF
--- a/node-graph/libraries/vector-types/src/vector/click_target.rs
+++ b/node-graph/libraries/vector-types/src/vector/click_target.rs
@@ -259,7 +259,15 @@ impl ClickTarget {
 
 	/// Does the click target intersect the point (accounting for stroke size)
 	pub fn intersect_point(&self, point: DVec2, layer_transform: DAffine2) -> bool {
-		let target_bounds = [point - DVec2::splat(self.stroke_width / 2.), point + DVec2::splat(self.stroke_width / 2.)];
+		// Maintain minimum 5px viewport stroke while calculating intersection point
+		let minimum_stroke: f64 = 5.;
+
+		// Scaling factor for converting to viewpoint space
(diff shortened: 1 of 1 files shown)


Keavon:
Sorry it took so long to reach this in my PR review queue! Upon taking a closer look, this doesn't solve the problem, and I believe your reproduction steps (shown in your videos) show a misunderstanding of the problem, since each of those are shapes with a (white) fill area which doesn't even apply to this situation. This implementation breaks the relationship between hover targets (when the shape that will be clicked lights up) and click targets (when clicking actually selects the shape). I also took a closer look at the original issue report and I am not sure if it actually needs to be changed from its current form. Anyway, thank you for your attempt and hopefully you learned something in the process.

PR #3661, opened 2026-01-20
Improve UX of importing vs. opening files

### LINKED ISSUE #3595: Restrictive filetype filtering on import (Linux Desktop)
Pressing File > Import on the web brings up a prompt for me to choose any image file type, but on *desktop* it limits it to just SVGs, and to import a PNG I have to change the category to image, or none, to be able to select it.

Image

It's like this for me on both the Appimage and Flatpak (but not on the web)
I'm running KDE 6.5 with Wayland on Fedora 43.

timon-schelling:
closes #3595


Keavon:
Dropping a .graphite file into the viewport of an open document:
  
  image

PR #3668, opened 2026-01-21
Desktop: Restart without UI acceleration when it fails on Linux

### LINKED ISSUE #3667: On desktop, automatically fall back to non-hardware accelerated UI on failure

timon-schelling:
Partly closes #3667

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -565,7 +565,7 @@ impl ApplicationHandler for App {
 					&& startup_time.elapsed() > Duration::from_secs(3)
 				{
 					tracing::error!("UI acceleration not working, exiting.");
-					self.exit(Some(ExitReason::UiAccalerationFailure));
+					self.exit(Some(ExitReason::UiAccelerationFailure));
 				}
 			}
 			WindowEvent::DragDropped { paths, .. } => {
@@ -654,5 +654,5 @@ impl ApplicationHandler for App {
 
 pub(crate) enum ExitReason {
 	Shutdown,
-	UiAccalerationFailure,
+	UiAccelerationFailure,
 }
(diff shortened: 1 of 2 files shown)


desktop/src/app.rs
```diff
@@ -629,3 +651,8 @@ impl ApplicationHandler for App {
 		event_loop.set_control_flow(ControlFlow::WaitUntil(wait_until));
 	}
 }
+
+pub(crate) enum ExitReason {
+	Shutdown,
+	UiAccalerationFailure,
```
Keavon: UiAccelerationFailure

desktop/src/lib.rs
```diff
@@ -77,6 +77,10 @@ pub fn start() {
 
 	let (cef_view_info_sender, cef_view_info_receiver) = std::sync::mpsc::channel();
 
+	if cli.disable_ui_acceleration {
+		println!("UI acceleration is disabled");
```
TrueDoctor, line 81: should this be an `eprintln` ?

timon-schelling: more of an info that an error, so in my opinion no.
but this will only stay for a couple of weeks anyways (until rc4).

TrueDoctor: The question is more if it should be included if I pipe the output of graphite to a file, and I'd think it should probably not. E.g. the `dbg!` macro also outputs to stderr, and I would argue that is similar in nature

timon-schelling: Then we should also change the Available GPU lines? 
I would merge this and we can decide later for all similar cases, currently most stuff ends up in stdout.

TrueDoctor: yes, we should, from `man stderr`
> Under normal circumstances every UNIX program has three streams opened for it when
       it  starts  up,  one for input, one for output, and one for printing diagnostic or
       error messages.

PR #3671, opened 2026-01-22
Desktop: Support triple-click inputs

### LINKED ISSUE #3596: Triple-click text to select paragraph (Linux Desktop)
On the web, triple-clicking text (either in a text box on the canvas, or in the UI, e.g. a layer name) will select the whole paragraph, but on desktop the third click just deselects the word.

Works as expected on [latest](https://dev.graphite.art) in Firefox, doesn't work in the Appimage or Flatpak.

he1senbrg:
#### Before
https://github.com/user-attachments/assets/e22f7ff2-8872-4d18-aab4-a7589818c429

#### After
https://github.com/user-attachments/assets/55bbbcce-2791-4507-b9e2-f706fa646be9

This PR fixes triple-clicking text not selecting the entire paragraph on desktop.

Closes #3596

### DIFF
--- a/desktop/src/cef/input/state.rs
+++ b/desktop/src/cef/input/state.rs
@@ -148,17 +148,17 @@ impl ClickTracker {
 		match state {
 			ElementState::Pressed if record.down_count == ClickCount::Triple => {
 				*record = ClickRecord {
-					down_count: ClickCount::Single,
+					down_count: ClickCount::Double,
 					..*record
 				};
-				return ClickCount::Single;
+				return ClickCount::Double;
 			}
 			ElementState::Released if record.up_count == ClickCount::Triple => {
 				*record = ClickRecord {
-					up_count: ClickCount::Single,
+					up_count: ClickCount::Double,
 					..*record
 				};
-				return ClickCount::Single;
+				return ClickCount::Double;
 			}
 			_ => {}
 		}


desktop/src/cef/input/state.rs
```diff
@@ -167,7 +168,13 @@ impl ClickTracker {
 		let within_dist = dx <= MULTICLICK_ALLOWED_TRAVEL && dy <= MULTICLICK_ALLOWED_TRAVEL;
 		let within_time = now.saturating_duration_since(prev_time) <= MULTICLICK_TIMEOUT;
 
-		let count = if within_time && within_dist { ClickCount::Double } else { ClickCount::Single };
+		let count = if prev_count == ClickCount::Double {
```
timon-schelling: Nit, this could look cleaner as a match. Logic seems fine.

he1senbrg: I have changed it to a match.

timon-schelling:
change needed because of this frontend code.
We can have a discussion if that behavior is correct in the future.

```ts
	function onPotentialDoubleClick(e: MouseEvent) {
		...
		// Allow only repeated increments of double-clicks (not 1, 3, 5, etc.)
		if (e.detail % 2 == 1) return;
```

Issue #2266, opened 2025-02-05
Update the input mapping system to not dispatch both a matched KeyDown and KeyDownNoRepeat


Keavon:
For example, if we update `input_mapping.rs` like so:

```diff
- entry!(KeyDown(KeyA); action_dispatch=ToolMessage::ActivateToolPath),
+ entry!(KeyDownNoRepeat(KeyA); action_dispatch=ToolMessage::ActivateToolPath),
```

Then <kbd>Ctrl</kbd><kbd>A</kbd> will fire both the selection of all layers...

```
entry!(KeyDown(KeyA); modifiers=[Accel], action_dispatch=DocumentMessage::SelectAllLayers),
```

...as well as the change to the Path tool, since one gets matched on `KeyDown` and another on `KeyDownNoRepeat`. If both were `KeyDown` or `KeyDownNoRepeat`, only one would match. But currently, both are matching, and we only want one to match.

After implementing this fix, the following mappings should be changed from `KeyDown` to `KeyDownNoRepeat`:

<details><summary>Mappings to change</summary>
```
// TransformLayerMessage
entry!(KeyDown(KeyX); action_dispatch=TransformLayerMessage::ConstrainX),
entry!(KeyDown(KeyY); action_dispatch=TransformLayerMessage::ConstrainY),

// PathToolMessage
entry!(KeyDown(KeyG); action_dispatch=PathToolMessage::GRS { key: KeyG }),
entry!(KeyDown(KeyR); action_dispatch=PathToolMessage::GRS { key: KeyR }),
entry!(KeyDown(KeyS); action_dispatch=PathToolMessage::GRS { key: KeyS }),

// PenToolMessage
entry!(KeyDown(KeyG); action_dispatch=PenToolMessage::GRS { grab: KeyG, rotate: KeyR, scale: KeyS }),
entry!(KeyDown(KeyR); action_dispatch=PenToolMessage::GRS { grab: KeyG, rotate: KeyR, scale: KeyS }),
entry!(KeyDown(KeyS); action_dispatch=PenToolMessage::GRS { grab: KeyG, rotate: KeyR, scale: KeyS }),

// ToolMessage
entry!(KeyDown(KeyV); action_dispatch=ToolMessage::ActivateToolSelect),
entry!(KeyDown(KeyZ); action_dispatch=ToolMessage::ActivateToolNavigate),
entry!(KeyDown(KeyI); action_dispatch=ToolMessage::ActivateToolEyedropper),
[...]

PR #3672, opened 2026-01-22
Add focus document toggle

timon-schelling:


editor/src/messages/input_mapper/input_mappings.rs
```diff
@@ -438,8 +438,8 @@ pub fn input_mappings(zoom_with_scroll: bool) -> Mapping {
 		entry!(KeyDown(Period); action_dispatch=NavigationMessage::FitViewportToSelection),
 		//
 		// PortfolioMessage
-		entry!(KeyDown(Tab); modifiers=[Control], action_dispatch=PortfolioMessage::NextDocument),
-		entry!(KeyDown(Tab); modifiers=[Control, Shift], action_dispatch=PortfolioMessage::PrevDocument),
+		entry!(KeyDownNoRepeat(Tab); modifiers=[Control], action_dispatch=PortfolioMessage::NextDocument),
+		entry!(KeyDownNoRepeat(Tab); modifiers=[Control, Shift], action_dispatch=PortfolioMessage::PrevDocument),
```
Keavon: Key repeat is expected to work when changing tabs.

editor/src/messages/portfolio/document/properties_panel/properties_panel_message_handler.rs
```diff
@@ -34,10 +34,12 @@ impl MessageHandler<PropertiesPanelMessage, PropertiesPanelMessageContext<'_>> f
 
 		match message {
 			PropertiesPanelMessage::Clear => {
-				responses.add(LayoutMessage::SendLayout {
-					layout: Layout::default(),
-					layout_target: LayoutTarget::PropertiesPanel,
-				});
+				if !properties_panel_open {
```
Keavon: This shouldn't be NOT'd, right?

Keavon:
- [x] Properties panel doesn't populate after toggling Focus Document in the case when nothing is selected.
- [x] Data panel never populates so it's completely broken.
- [x] After closing the last document to return to the welcome screen, the Properties panel doesn't depopulate.
- [x] Alt+Enter doesn't work on the welcome screen, but only sometimes. Seems to happen if you create a document, draw a layer, then close the document. Doesn't happen if you create a new document and either close it immediately or draw something then delete the layer then close it.

editor/src/messages/input_mapper/input_mappings.rs
timon-schelling: I can remove this change then, but this fixes an issue where changing tab also triggers the normal tab shortcut and the tool changes.

Keavon: "changing tab"? Confused what you mean, sorry.

timon-schelling: When switching documents with shortcut, the editor also switches between select and path tool.

Keavon: Oh, I thought we match one keyboard event to at most one resulting action?

Keavon: Ah I see, because that one is also NoRepeat:
```
entry!(KeyDownNoRepeat(Tab); action_dispatch=ToolMessage::ToggleSelectVsPath),
```
Yeah, that's a limitation that #2266 would aim to solve. I think the better option is to turn the line above into a regular KeyDown with a TODO comment to set it back once that issue is solved.

PR #3679, opened 2026-01-25
Add "Loop Level" to the Position context reader node

Keavon:


TrueDoctor:
This broke the pointer position node

PR #3681, opened 2026-01-25
Fix regression breaking Into/Convert node type coercion

### LINKED ISSUE #3677: Convert node no longer supports strings
A regression introduced by @Keavon in #3662.

Previously you could connect a number to the text (create text with text tool, open node graph, select text node, click dot next to text input)
<img width="935" height="208" alt="working number -> text" src="https://github.com/user-attachments/assets/a02ea56b-e9ce-4ff2-a09f-49a4283c5855" />

Now it doesn't work:

error

```
This input type is incompatible:
• Input 2:
…found:       f64 called with Context
…expected: string called with Context
```

The name of the `String` type is now for some reason changed:
https://github.com/GraphiteEditor/Graphite/blob/7af60e02a33a4fa22edc058d57021ced5d8d63eb/node-graph/libraries/core-types/src/types.rs#L389-L391

So no convert node is found to match it:
https://github.com/GraphiteEditor/Graphite/blob/7af60e02a33a4fa22edc058d57021ced5d8d63eb/node-graph/preprocessor/src/lib.rs#L72-L84

Keavon:
Fixes a regression from #3662. Closes #3677.

This also refactors some code messiness around how we did `Display` vs. `Debug` printing of type representations.

### DIFF
--- a/node-graph/libraries/core-types/src/types.rs
+++ b/node-graph/libraries/core-types/src/types.rs
@@ -118,11 +118,10 @@ impl NodeIOTypes {
 
 impl std::fmt::Debug for NodeIOTypes {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-		f.write_fmt(format_args!(
-			"node({}) → {}",
-			[&self.call_argument].into_iter().chain(&self.inputs).map(|input| input.to_string()).collect::<Vec<_>>().join(", "),
-			self.return_value
-		))
+		let inputs = self.inputs.iter().map(ToString::to_string).collect::<Vec<_>>().join(", ");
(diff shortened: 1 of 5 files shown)


node-graph/libraries/core-types/src/types.rs
```diff
@@ -337,15 +336,17 @@ impl Type {
 		}
 	}
 
-	pub fn to_cow_string(&self) -> Cow<'static, str> {
+	pub fn identifier_name(&self) -> String {
```
TrueDoctor, line 339: What is the reason for the changed return type here?

Keavon: This isn't a change, one unused function is deleted and another new function is added in the same location.

PR #3682, opened 2026-01-25
Fix dragging range-style integer number inputs not accumulating slow mouse movements

### LINKED ISSUE #3674: Dragging range-style number inputs constrained to integers doesn't accumulate slow mouse movements
If you drag slowly such that every mouse movement maps to less than a change of 1 in the value, those mouse movements don't count and it stays rounded to the integer, meaning you have to move the mouse quickly to be able to affect any change, and you can't move slowly for precision because those slow inputs are ignored.

krVatsal:
Closes #3674
Fixes a bug where slow mouse movements were ignored for integer inputs or when snapping is enabled.


timon-schelling:
!build

PR #3683, opened 2026-01-25
Fix solidify stroke node ignoring transforms applied before stroke node

motherboardMage:
This PR addresses the issue raised in https://discord.com/channels/731730685944922173/881073965047636018/1416281986661613669

Solidify stroke now applies the stroke's transform to the path before generating geometry and then reverses the transform. 

The rectangle on the left is 600px by 600px with transform scale at 1x and the one on the right is 100px by 100px with transform scale set to 6x on both axes.

**Before**
Screenshot 2026-01-26 at 3 02 49 AM

**After**
Screenshot 2026-01-26 at 3 02 28 AM


motherboardMage:
Thanks for your help and insights @0HyperCube !

PR #3688, opened 2026-01-26
Desktop: Only rerender the UI if it has changed

TrueDoctor:


desktop/src/app.rs
```diff
@@ -557,6 +560,7 @@ impl ApplicationHandler for App {
 						Err(RenderError::SurfaceError(e)) => tracing::error!("Render error: {:?}", e),
 					}
 					let _ = self.start_render_sender.try_send(());
+					// self.cef_context.schedule_frame();
```
timon-schelling:
```suggestion
```

PR #3689, opened 2026-01-27
Drag selected nodes from initial click

adamgerhant:
When dragging selected nodes the location of the initial mouse click is used as the reference point
Fixes offset when shaking nodes

Before

https://github.com/user-attachments/assets/7c6698be-fbe1-4621-8e38-fc07d41299f2

After

https://github.com/user-attachments/assets/3e96c9c8-bc46-4589-a32e-8d69593cd9e7


Keavon:
Some funky bugs going on regarding the transfer of selection from one to another before or as part of the shake (or in some cases, just regular drag), here's an example of one of the cases I found, but there are related ones:

https://github.com/user-attachments/assets/0233f08e-50ef-436a-b89f-26360f1afea7

Keavon:
Superseded by #4411

PR #3690, opened 2026-01-27
Fix Shift constraint to start from snapped position when dragging points (#2753)

### LINKED ISSUE #2753: Path tool dragging, then pressing Shift to constrain horiz./vert., should begin from a snapping point
At the beginning of this video, I add <kbd>Shift</kbd> to the drag and it correctly performs the horizontal/vertical snapping from its original drag location. But in the second half of the video, I demonstrate a use case where this feature is desired: when the point is dragged such that it snaps over another point, then <kbd>Shift</kbd> is pressed at that time, we would like that snapped point to become the source of the horizontal/vertical constraint. Releasing <kbd>Shift</kbd> and pressing it again while not snapped should retain its current behavior of constraining relative to the point where the drag began. As shown in the video, this feature would make it very easy to keep the top point aligned to the same X coordinate as the bottom point.

https://github.com/user-attachments/assets/e59f5531-8b8a-4a18-a5ab-520fef992aa8

afrdbaig7:
Fixes #2753.

When dragging a point and pressing Shift to constrain movement, the constraint was starting from the original drag position instead of the snapped position, which caused an unexpected jump.

This PR fixes that by locking the constraint origin when Shift is first pressed and using the snapped mouse position, so constrained movement behaves consistently even after snapping or axis changes.

The behavior when starting a drag with Shift, releasing and re-pressing Shift, and normal dragging remains unchanged. All existing tests pass.

### DIFF
--- a/editor/src/messages/tool/tool_messages/path_tool.rs
+++ b/editor/src/messages/tool/tool_messages/path_tool.rs
@@ -590,6 +590,10 @@ struct PathToolData {
 	last_clicked_point_was_selected: bool,
 	last_clicked_segment_was_selected: bool,
 	snapping_axis: Option<Axis>,
+	/// The origin point for horizontal/vertical constraints when Shift is pressed.
+	/// When `None`, defaults to `drag_start_pos`. When `Some`, uses the snapped position
+	/// if Shift was pressed while the point was snapped to another point.
+	constraint_origin: Option<DVec2>,
 	alt_clicked_on_anchor: bool,
 	alt_dragging_from_anchor: bool,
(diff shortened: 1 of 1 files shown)


Keavon:
This is letting me move it around while the snap is happening, so I'm not sure what's going on. Draw a bunch of random layers to create a bunch of snapping points, then drag an anchor over another anchor and hold Shift, then keep dragging and you'll see that it doesn't remain constrained to the horizontal or vertical line defined by the snapping point that was set upon pressing Shift in that moment the key was depressed.

Issue #3500, opened 2025-12-20
Robustness when panning/tilting/zooming during dragging actions


Keavon:
The user can PTZ...
- Pan (middle mouse drag, or scroll using the scroll wheel)
- Tilt (<kbd>Alt</kbd> middle mouse drag)
- Zoom (<kbd>Ctrl</kbd> middle mouse drag, or <kbd>Ctrl</kbd> scroll with the scroll wheel)
...while performing various dragging operations and those currently don't follow the original document-space start position coordinates:

- [ ] Drag offset (Select tool):
  
  https://github.com/user-attachments/assets/87e63c93-cff5-40e1-a058-75e4461c1e02

- [ ] Selection box (Select tool and Path tool):
  
  https://github.com/user-attachments/assets/20ad0109-e828-4493-a04e-13d97ed8d635

- [ ] Grab transformation (<kbd>G</kbd> with an object selected using the Select tool or points selected using the Path tool):
  
  https://github.com/user-attachments/assets/e4699542-8589-419a-889f-5185fe71c62a
  
  https://github.com/user-attachments/assets/93e193c3-e53e-47ee-95f1-ac4745d3fe05

- [ ] Point rotate transformation when tilted (<kbd>R</kbd> with points selected using the Path tool):
  
  https://github.com/user-attachments/assets/bd0fe282-395e-4d1b-87d0-51964413a928

- [ ] Scale transformation when tilted (<kbd>S</kbd> with an object selected using the Select tool or points selected using the Path tool):
  
  https://github.com/user-attachments/assets/2186ce09-d487-40b5-9e76-71a5f5be981c
  
  https://github.com/user-attachments/assets/7be92f8c-8460-4360-9507-80f4b97237f6

- [ ] Grab transformation when tilted <kbd>G</kbd> with an object selected using the Select tool or points selected using the Path tool):
  
  https://github.com/user-attachments/assets/df0f672d-acf5-45e0-8d72-5d27e36937b7
  
  https://github.com/user-attachments/assets/bea3ab95-a230-4b83-8e17-05183ac7f2d6

- [ ] Duplication incorrectly still occurs when <kbd>Alt</kbd> middle mouse dragging to tilt the viewport while dragging (once tilting begins, duplication should be canceled):
  
  https://github.com/user-attachments/assets/8d73573d-2552-4cdb-a9f3-f5adb200a926

- [ ] The gray original path ghost outline stays frozen in place under PTZ
  
[...]

PR #3692, opened 2026-01-27
Abort tools when tilt in viewport

krVatsal:
Fixes one of the issues in #3500
Implemented a fix to ensure that entering viewport tilt mode (Alt + MMB) correctly aborts any active tool operation.


Keavon:
Which task within the tracking issue does this solve?

Also from my testing, this didn't seem to work as I'm interpreting it from your description (but since I don't know which requested problem this aims to resolve, I can't say with certainty that I'm testing the right problem).

Once you respond, you may mark this as not a draft so I can revisit it.

krVatsal:
Fixes: `Duplication incorrectly still occurs when Alt middle mouse dragging to tilt the viewport while dragging (once tilting begins, duplication should be canceled)`

Before:

https://github.com/user-attachments/assets/d70eda5d-bc8f-4a59-8f9a-3c6c288df274

After:

https://github.com/user-attachments/assets/bdd34094-de98-475d-ab7b-b8b30b16795c

PR #3697, opened 2026-01-28
Adding Accel and Ctrl

Nuhiat-Arefin:
On Mac, the canvas uses Ctrl and the node graph uses Cmd (Accel). Trying to modify so that both use both Ctrl and Accel.


Keavon:
This brute-force approach is not a suitable solution. Which issue number or Discord message does this resolve? I don't see one referenced. Feel free to reply and we can reopen this, but I'm going to close this PR for the moment since it is presently far from being able to be merged. Thanks.

Nuhiat-Arefin:
> This brute-force approach is not a suitable solution. Which issue number or Discord message does this resolve? I don't see one referenced. Feel free to reply and we can reopen this, but I'm going to close this PR for the moment since it is presently far from being able to be merged. Thanks.

I thought using Ctrl in canvas and Cmd in node graph is quiet confusing. But yes,  brute force shouldn't work.

PR #3708, opened 2026-02-02
Add JSON get node

TrueDoctor:


timon-schelling:
Personalty I would prefer if it just returned "" on errors but this is also fine I guess.

TrueDoctor:
We could also log an error to the console and return an empty string but I since this can fail for a bunch of reasons I'd like to give some form of feedback to the user

PR #3711, opened 2026-02-02
Expose the secondary input for Logical And and Or nodes

he1senbrg:
### Before
Before

### After
After

Fixes https://discord.com/channels/731730685944922173/881073965047636018/1467300857027756307


TrueDoctor:
Could you also add this for other logic nodes where having a non exposed parameter obviously does not make sense (e.g. the `or` node)?

he1senbrg:
I have exposed the other logic nodes. In retrospect it was dumb of me not to do that, lol.

logical_nodes

Keavon:
This is not desired on any nodes except AND and OR.

he1senbrg:
Okay got it, I have updated it.

update

PR #3715, opened 2026-02-03
Adjust ruler tick alignment when document is tilted

### LINKED ISSUE #1482: Stretch and tilt ruler increments when canvas is tilted

GiacomoViolini:
Closes #1482

### DIFF
--- a/editor/src/messages/frontend/frontend_message.rs
+++ b/editor/src/messages/frontend/frontend_message.rs
@@ -247,6 +247,7 @@ pub enum FrontendMessage {
 		spacing: f64,
 		interval: f64,
 		visible: bool,
+		tilt: f64,
 	},
 	UpdateDocumentScrollbars {
 		position: (f64, f64),
--- a/editor/src/messages/portfolio/document/document_message_handler.rs
+++ b/editor/src/messages/portfolio/document/document_message_handler.rs
@@ -1012,6 +1012,7 @@ impl MessageHandler<DocumentMessage, DocumentMessageContext<'_>> for DocumentMes
 				}
 
 				let ruler_spacing = ruler_interval * ruler_scale;
+
 				responses.add(FrontendMessage::UpdateDocumentRulers {
 					origin: ruler_origin.into(),
 					spacing: ruler_spacing,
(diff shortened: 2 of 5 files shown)


Keavon:
!build

Keavon:
This doesn't seem to be working, does it?

GiacomoViolini:
Maybe I misunderstood some of the requirements.


https://github.com/user-attachments/assets/8af4b133-9303-4b22-912e-d2400fcb0cc4

Should I change something about the ruler behavior when tilted?

The current build doesn't include the interval width change based on the angle by the way.

Keavon:
If the current build doesn't include the actual requested feature, what do all the code changes do? I was not able to notice any difference from normal, so I am confused.

GiacomoViolini:
For example:
This is the current ruler at 45
image
This the ruler with the changes
image

As you can see it is tilted, following the tilt angle. If you retrigger the build you should see the latest changes (I don't know if I'm allowed to trigger it myself).

Let me know if you were expecting something different from the issue

Keavon:
!build

jsjgdh:
There appears to be a mismatch, as shown in the image. As you can see 0 of the ruler is not aligned after the tilt .

image

Issue #2595, opened 2025-04-18
Path tool box selection starts on mousedown at a far-off location when animation is playing


Keavon:
1. Switch to the Path tool
2. Hit play to start animation playback
3. Click but don't drag or release— notice how the box selection has gone from some far-off location to the current cursor location
4. Pan the document and repeat the above steps— notice how the starting point of the erroneous box changes in relation to how you pan

henriquevmac:
Hey, as I mentioned previously on discord I will be attempting to work on this issue

Lamoara:
I'll be working on this one assuming no one is doing it

PR #3718, opened 2026-02-04
Fix selection box positioning and alignment when document is PTZ'd

### LINKED ISSUE #2391: Box/lasso selection should be compatible with PTZ
Box selection with the Select tool and Path tool doesn't follow the canvas when panning (via the mouse vertical/horizontal scroll wheel) or zooming (with Ctrl + or Ctrl -). The starting point of the rectangle should be locked down to the canvas location, not follow the viewport. This works correctly with the auto-panning feature (dragging the box selection off the edges of the viewport), so that code can be referenced. The actual box selection must still be rounded to viewport coordinates so the box selection is drawn cleanly (without fuzzy antialiasing). The video below shows the incorrect present behavior.

https://github.com/user-attachments/assets/6687128d-0123-4feb-bd10-88c36440971d

krVatsal:
Fixes #2391

When performing a box selection with the Select tool, panning (via scroll wheel) would cause the selection rectangle's starting point to move with the viewport instead of staying fixed on the canvas.

Solution
Store the drag start position in document coordinates (drag_start_document) rather than viewport coordinates
Transform this document anchored point back to viewport space when computing the selection box
Remove the viewport shift compensation for the start point during auto panning since the start is now properly anchored to document space

Still confused about the expected outcome for zooming, can you clarify more for what should be the output for it?

https://github.com/user-attachments/assets/5e6bacee-f307-48a7-a822-5a8fdf4368fb


Keavon:
> Still confused about the expected outcome for zooming, can you clarify more for what should be the output for it?

In all cases, the document-space start point and document-space current pointer position should determine the bounding box shown, as well as the current document tilt. You can always take those three parameters (start point, current point, current tilt) and draw a rectangle matching the document space coordinates of them (but transformed into viewport space for the final drawing of the rectangle, since overlays are in viewport space).

krVatsal:
> > Still confused about the expected outcome for zooming, can you clarify more for what should be the output for it?
> 
> In all cases, the document-space start point and document-space current pointer position should determine the bounding box shown, as well as the current document tilt. You can always take those three parameters (start point, current point, current tilt) and draw a rectangle matching the document space coordinates of them (but transformed into viewport space for the final drawing of the rectangle, since overlays are in viewport space).

Thanks for the clarification! I think the current implementation already handles this correctly for both panning and zooming.

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -440,12 +441,19 @@ impl SelectToolData {
 		self.selection_mode.unwrap_or(SelectionMode::Touched)
 	}
 
-	pub fn selection_box(&self) -> [DVec2; 2] {
-		if self.drag_current == self.drag_start {
+	pub fn selection_box(&self, metadata: &DocumentMetadata) -> [DVec2; 2] {
+		// If we have a document-anchored start point, transform it to viewport
```
Keavon: Why is this an "if" scenario? Why isn't it always the case that the start point is document-anchored?

editor/src/messages/tool/tool_messages/select_tool.rs
```diff
@@ -402,6 +402,7 @@ struct SelectToolData {
 	snap_candidates: Vec<SnapCandidatePoint>,
 	auto_panning: AutoPanning,
 	drag_start_center: ViewportPosition,
+	drag_start_document: Option<DVec2>,
```
Keavon: See other comment. I'm possibly a bit concerned that this is an `Option`. When would it be `None`? Can you make sure that this is built (if it's not already) in such a way that it makes invalid state unrepresentable?

Keavon:
Select tool behavior is good from my QA test. Just see my concern from my code review comments. It might represent an opportunity to simplify the code if there is some kind of extra conditionality going on right now.

The equivalent implementation is missing from the Path tool.

There is also one small bug I've noticed that this implementation seems to consistent trigger, which is that the box selection (upon initially clicking down but not yet dragging) draws a tiny square (#2595). I've observed this in a few other occasions but now your branch appears to be making this happen all the time, which is good since that should make it easier to debug and fix the root cause.

https://github.com/user-attachments/assets/6da18eac-0858-432d-a4ac-2bcdf5ff98fc

Please mark as Ready for Review when you respond so I don't miss this while it's a draft.

editor/src/messages/tool/tool_messages/select_tool.rs
krVatsal: because the tool is not always in a dragging state (e.g., when it is `[Ready]`). However, the variants themselves (like `[Dragging]`, `[Drawing]`, `[ResizingBounds]`, etc.) now store `[drag_start]` as a mandatory, non-optional field(in the new commit i am pushing in some time).

editor/src/messages/tool/tool_messages/select_tool.rs
krVatsal: Thanks for pointing out, you are correct it should always be document anchored i have fixed this in next commit.

Keavon:
Can you please help me understand how you went from +29-18 lines of changes before my code review that requested a potential simplification, to a staggering +196-89 after the new commit? It is too many changes for me to try to read and understand, but this doesn't make sense to me at all and it makes me think it was either written by an LLM or— actually I don't have any other theories that could explain how my change requests resulted in 6x the changes from before then. And this doesn't even include the requested Path tool changes, which would be the only thing I requested that I would expect a ~doubling (not 6x'ing) of the code. But before you apply this to the Path tool, let's make sure we first figure out what the problem is here that I'm pointing out.

krVatsal:
> Can you please help me understand how you went from +29-18 lines of changes before my code review that requested a potential simplification, to a staggering +196-89 after the new commit? It is too many changes for me to try to read and understand, but this doesn't make sense to me at all and it makes me think it was either written by an LLM or— actually I don't have any other theories that could explain how my change requests resulted in 6x the changes from before then. And this doesn't even include the requested Path tool changes, which would be the only thing I requested that I would expect a ~doubling (not 6x'ing) of the code. But before you apply this to the Path tool, let's make sure we first figure out what the problem is here that I'm pointing out.

Yeah, i was a bit blank about how should i simplify this further, that's why i took help of llm apologies for that. I'll look into this again and update with a much cleaner version. Thankyou for your patience.

Keavon:
Undisclosed LLM usage is a violation of our acceptable AI usage policy. Consider this your one warning.

krVatsal:
Sure, this won't happen again.

Keavon:
Do you have any updates on reverting this to the original state and then taking action on my original comments?

krVatsal:
Yeah was working on this pr actually my mid semester exams are going on so there was some blockage of time, will update this by EOD.

krVatsal:
I think i should open a new PR as this one is getting a bit messy for me to manage.

Keavon:
Please do not, there is no reason to do so.

node_modules/.vite/deps/_metadata.json
Ayush2k02, line 1: Is this file really needed for this PR ??

Ayush2k02 (changes requested):

node_modules/.vite/deps/package.json
Ayush2k02, line 1: same for this

Ayush2k02 (changes requested):

PR #3721, opened 2026-02-05
Desktop: Fix drag followed by a click causing a double click

### LINKED ISSUE #3714: On desktop, a drag followed by a click causes a double click
Drag and release, then immediately click once. This is interpreted as a double click on desktop. For example, this can be done with the window frame to cause it to maximize after dragging it then clicking once; or this can be done with a box selection in the node graph that releases on a node and then, by single-clicking, ends up double-clicking into the sub-graph of that node.

This occurs both before and after #3671.

he1senbrg:
### Before
https://github.com/user-attachments/assets/d10dc173-94f3-4262-9f06-20ce027095fd

### After
https://github.com/user-attachments/assets/dfdd37e0-2cfc-4df5-a2e7-0762b54e1264

Closes #3714

### DIFF
--- a/desktop/src/cef/input/state.rs
+++ b/desktop/src/cef/input/state.rs
@@ -149,7 +149,9 @@ impl ClickTracker {
 			ElementState::Released => (record.up_count, record.up_position),
 		};
 
-		let within_dist = position.x.abs_diff(prev_position.x) <= MULTICLICK_ALLOWED_TRAVEL && position.y.abs_diff(prev_position.y) <= MULTICLICK_ALLOWED_TRAVEL;
+		let dx = position.x.abs_diff(prev_position.x);
+		let dy = position.y.abs_diff(prev_position.y);
+		let within_dist = dx <= MULTICLICK_ALLOWED_TRAVEL && dy <= MULTICLICK_ALLOWED_TRAVEL;
 
 		let count = match (prev_count, within_time, within_dist) {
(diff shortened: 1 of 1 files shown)


timon-schelling:
I think an easier and cleaner solution would be resetting both up_ and down_count whenever xy_delta is to big. down position is already in the record and should not be duplicated.

he1senbrg:
@timon-schelling I might be wrong, but from what I understood the last time a value is stored in `position` would be when a click was released, right? And when the next click happens within the conditions it registers as a double click. So, even if we reset up_ and down_count wouldn't the next click still register as double as the previous one was single (either via reset or because of drag)?

timon-schelling:
I would need to spend more than 2 min looking at the code, but from what I remember it should be possible to split up and down positions and use both of them to figure out the xy delta with each and abort the double click when either the down or up position is to far (fine to add that to the record).

Keavon:
Marking as draft while awaiting changes.

he1senbrg:
@timon-schelling Sorry for the delay. Can i remove `position` from `ClickRecord` and add `down_position` and `up_position` to it so that these can be individually tracked?

he1senbrg:
@timon-schelling Could you please review the changes

Keavon:
I'd like to give this a QA test before merging.

timon-schelling:
@Keavon ok, I think now I'm fine with the patch.

Issue #3509, opened 2025-12-22
Use tiled rendering with Vello


Keavon:
We should stitch together multiple tiles if we go above Vello's approximate maximum of 4096x4096.

Issue #1607, opened 2024-02-15
Fix caching with resolution-aware data


Keavon:
Urgent

0HyperCube:
@Keavon I'm not really sure how we can do this. If we have `f(Footprint) = disk sampled points` and want to avoid doing the disk sampling many times, then how can we be sure that changing the footprint won't impact the result?

If we actually implemented the cull node, then changing the footprint by panning around would certainly impact the result - so if the shape was initially outside the viewport and we cached the empty result, then that would not be ideal.

TrueDoctor:
> If we actually implemented the cull node, then changing the footprint by panning around would certainly impact the result - so if the shape was initially outside the viewport and we cached the empty result, then that would not be ideal.

The idea of resolution aware caching would be that we only reuse the cached result if the outcome would be the same and otherwise recompute. The click targets are a bit more tricky to update but that's the basic idea

0HyperCube:
@TrueDoctor How do we know if the "outcome would be the same" without computing the outcome?

Keavon:
@TrueDoctor wrote in Discord:

> There are basically two possible ways of going about this:
> 
> 1. Fix the performance and actually make everything fast enough to be executed in real time
> 2. Change how click targets are propagated: 
>      If instead of just storing the click targets in the monitor nodes, we instead attach the click targets to the data type, we can modify the click target in the cache node.
>      When the user then pans around in the viewport and we know that the data we have cached is basically still correct, we could modify the click target in the cache node to reflect the proper location.
> 
> But I honestly think that we really should not need to rely on caching for purely vector data. There are a bunch of potential optimizations we can do which should could speed this up (e.g. eliminating  unnecessary clones etc.)

0HyperCube:
@TrueDoctor / @Keavon

For 1, this is quite challenging given that poisson disk sampling inherently requires quite a lot of allocation and computation. A popular algorithm for fast poisson disk sampling is https://www.cs.ubc.ca/~rbridson/docs/bridson-siggraph07-poissondisk.pdf. I have also noticed that the code for sampling based on Euclidean distance is very inefficient and I might try and improve that sometime. However even with the current total caching currently we get a frame time of 80ms or ~12fps which doesn't really give a lot of room for these algorithms to run in real time.

For 2, I'm not sure how this is relevant as click targets are stored in document space so don't change as the user pans around the document.

TrueDoctor:
@0HyperCube 
[...]

PR #3722, opened 2026-02-05
Cache Vello render output as stitchable textures

TrueDoctor:
Cache render output as texture and only rerender necessary parts while panning
Closes: #3509, part of #1607.

### DIFF
--- a/editor/src/messages/dialog/preferences_dialog/preferences_dialog_message_handler.rs
+++ b/editor/src/messages/dialog/preferences_dialog/preferences_dialog_message_handler.rs
@@ -349,7 +349,7 @@ impl PreferencesDialogMessageHandler {
 				let render_tile_resolution_description = "
 					Maximum X or Y resolution per render tile. Larger tiles may improve performance but can cause flickering or missing content in complex artwork if set too high.\n\
 					\n\
-					*Default: 1024 px.*
+					*Default: 1280 px.*
 					"
 				.trim();
 				let render_tile_resolution_label = vec![
--- a/node-graph/graph-craft/src/wasm_application_io.rs
+++ b/node-graph/graph-craft/src/wasm_application_io.rs
@@ -358,7 +358,8 @@ impl graphene_application_io::GetEditorPreferences for EditorPreferences {
 	}
 
 	fn max_render_region_area(&self) -> u32 {
-		self.max_render_region_size.pow(2)
+		let size = self.max_render_region_size.min(u32::MAX.isqrt());
+		size.pow(2)
 	}
 }
 
@@ -366,7 +367,7 @@ impl Default for EditorPreferences {
 	fn default() -> Self {
 		Self {
 			vello_preference: VelloPreference::Auto,
-			max_render_region_size: 1024,
+			max_render_region_size: 1280,
 		}
 	}
 }


TrueDoctor:
We don't store tiles, so I think the title is technically wrong / misleading

editor/src/messages/dialog/preferences_dialog/preferences_dialog_message_handler.rs
```diff
@@ -281,7 +288,47 @@ impl PreferencesDialogMessageHandler {
 					.widget_instance(),
 			];
 
-			rows.extend_from_slice(&[header, node_graph_wires_label, graph_wire_style, use_vello, brush_tool]);
+			let max_region_size_description =
+				"Maximum render region size in total pixels (width × height). Larger values require fewer render passes. If you see rendering artifacts or nothing at all, configure a smaller value.";
```
timon-schelling: Can we split this explanation in two parts, first explaining for non technical users and a second part going deeper and explaining the exact effect?

editor/src/node_graph_executor/runtime.rs
```diff
@@ -328,6 +328,7 @@ impl NodeRuntime {
 
 							executor.context.queue.submit([encoder.finish()]);
 							surface_texture.present();
+							// image_texture.texture.destroy();
```
timon-schelling:
```suggestion
```

node-graph/interpreted-executor/src/util.rs
```diff
@@ -51,12 +51,23 @@ pub fn wrap_network_in_scope(mut network: NodeNetwork, editor_api: Arc<WasmEdito
...
+				},
 				DocumentNode {
 					call_argument: concrete!(graphene_std::application_io::RenderConfig),
-					inputs: vec![NodeInput::node(NodeId(1), 0)],
+					inputs: vec![NodeInput::node(NodeId(2), 0)],
 					implementation: DocumentNodeImplementation::ProtoNode(graphene_std::render_node::create_context::IDENTIFIER),
 					context_features: graphene_std::ContextDependencies {
-						extract: ContextFeatures::empty(),
+						// We add the extract index annotation here to force the compiler to add a context nullification node before this node so the render context is properly nullified so the render cache node can do its's work
+						extract: ContextFeatures::INDEX,
```
timon-schelling, line 70: could we add a separate ContextFeature for this purpose?

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -0,0 +1,582 @@
...
+	for y in tile_start_y..tile_end_y {
+		for x in tile_start_x..tile_end_x {
+			tiles.push(TileCoord { x, y });
+		}
+	}
+	tiles
+}
+
+#[inline]
+pub fn tile_world_start(tile: &TileCoord, scale: f64) -> DVec2 {
```
timon-schelling:
```suggestion
pub fn tile_space_origin(tile: &TileCoord, scale: f64) -> DVec2 {
```

timon-schelling: not sure, just something to think about

timon-schelling: also for similar cases

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -0,0 +1,582 @@
...
+use graphene_application_io::{ApplicationIo, ImageTexture};
+use rendering::{RenderOutputType as RenderOutputTypeRequest, RenderParams};
+use std::collections::HashSet;
+use std::hash::Hash;
+use std::sync::{Arc, Mutex};
+
+use crate::render_node::RenderOutputType;
+
+pub const TILE_SIZE: u32 = 256;
+pub const MAX_CACHE_MEMORY_BYTES: usize = 512 * 1024 * 1024;
```
timon-schelling, line 18: Maybe this could be a user setting?

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -0,0 +1,582 @@
...
+	output_resolution: UVec2,
+	logical_scale: f64,
+	physical_scale: f64,
+	exec: &wgpu_executor::WgpuExecutor,
+) -> (wgpu::Texture, rendering::RenderMetadata) {
+	let device = &exec.context.device;
+	let queue = &exec.context.queue;
+
+	// TODO: Use texture pool to reuse existing unused textures instead of allocating fresh ones every time
+	let output_texture = device.create_texture(&wgpu::TextureDescriptor {
```
timon-schelling, line 507: Could we use target texture here? Only needs to be recreated if viewport size changes?

timon-schelling (commented):
Generally LGTM, didn't check the cache as thoroughly. Can do that if you want but the approach is good.

Keavon (approved):
Good to go from a UI perspective and to the extent of the regression testing I've done. Please merge when you feel it's ready, ideally after marking Timon's comments as resolved after considering them.

node-graph/interpreted-executor/src/util.rs
TrueDoctor: I think, if it is only used here, this workaround is fine. I'd like to not pollute the context features with this. One thing I did think about is to have a `cached` property on a node which would allow a node to request to have its output cached

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -0,0 +1,582 @@
...
+
+	let all_regions: Vec<_> = cache_query.cached_regions.into_iter().chain(new_regions.into_iter()).collect();
+
+	// If no regions, fall back to direct render
+	if all_regions.is_empty() {
+		let context = OwnedContextImpl::empty().with_footprint(*footprint).with_vararg(Box::new(render_params.clone()));
+		return data.eval(context.into_context()).await;
+	}
+
+	let exec = editor_api.application_io.as_ref().unwrap().gpu_executor().unwrap();
```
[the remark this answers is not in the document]
TrueDoctor: We know that the executor will be available if we request a vello render

node-graph/nodes/gstd/src/render_cache.rs
TrueDoctor: Not sure what you mean by this. We should not overwrite a texture that is in use by someone else, in which case we would return a new texture until the old one becomes unused again, which is what the texture pool would be

Keavon:
Regression found: #3807

PR #3723, opened 2026-02-05
fixed shape at non-100% viewport zoom to apply scale correctly

### LINKED ISSUE #2582: Drawing a shape at non-100% viewport zoom wrongly applies the scale to the Transform node not the shape
1. Zoom out so you're not at 100% viewport zoom.
2. Draw a rectangle.
3. Notice the Transform node has a scale value other than `1` corresponding to the zoom level of the viewport.
4. This should be `1` and the viewport zoom should have been applied to the calculation of the rectangle dimensions instead.

This occurs with Rectangle, Ellipse, Polygon, and Star— but not Line.

krVatsal:
Fixes #2582 
When drawing shapes (Rectangle, Ellipse, Circle, Polygon, Star, Arc, Spiral) at a viewport zoom other than 100%, the Transform node would receive a scale value corresponding to the zoom level instead of 1.

The fix converts viewport-space dimensions to document-space before setting shape parameters and uses `TransformIn::Local` instead of `TransformIn::Viewport` . This matches how the Line tool already worked correctly.


Keavon:
- This is seemingly not working for Ellipse.
- We get a crash when drawing a Spiral.

krVatsal:
> * We get a crash when drawing a Spiral.
Hey, can you explain a bit what do you mean by crash, like it works fine on my localhost for spiral.

krVatsal:
After analyzing the test failures, I realized that ellipse is intentionally designed to work in viewport space (visual/screen space), not document space(which was my solution for this scale issue). This is correct behavior for canvas rotation scenarios. The zoom-related fix likely only applies to shapes like Rectangle that use document space (`[TransformIn::Local]`), which I already fixed in my previous fix.

If there is something that i am getting wrong or if i should follow any other method please suggest.
@Keavon

Keavon:
Ellipse should work the same as Rectangle in that regard.

krVatsal:
Can you guide me on this for ellipse one like when i try to fix the scale bug in ellipse using viewport both these rotation tests fails, while this is not the case for other shapes as they don't have any such test cases.

Keavon:
You are allowed to modify test cases if you can determine that they enclosed an expectation of incorrect or outdated behavior.

PR #3727, opened 2026-02-06
Fix the Path tool adding a Path node but not applying the change until a second attempt due to wrong segment IDs

### LINKED ISSUE #2068: Preserve point IDs without a path node
When there is no path node, the `compute_modified_vector` function (included below) relies on the `click_targets`. However click targets don't store PointIds, SegementIds, or colinear handles. This leads to a couple of issues:
- If you select a point on ellipse (no dragging) the colinear check box won't be filled.
- If the segments are in an odd order (e.g. with the bevel node), inserting a point in a segment crashes.

https://github.com/GraphiteEditor/Graphite/blob/c3a3c4c907419e2ebb9a382546a1d7b66f57cdc0/editor/src/messages/portfolio/document/utility_types/network_interface.rs#L2826-L2831

### LINKED ISSUE #2687: Using the path edit tool on one of the corners of a raster image breaks the graph

### LINKED ISSUE #3481: Segment IDs are wrong when no Path node exists, causing Path tool to fail before Path node is added
https://github.com/GraphiteEditor/Graphite/blob/b4fc3d05808c780829f1a0b8c193cac68149d586/editor/src/messages/portfolio/document/utility_types/network_interface.rs#L3047-L3068

It tries to get a path node and then falls back on the `document_metadata.click_targets`. `SegmentId`s are not stored in the click targets. Therefore this code will just **incorrectly** assume incremental `SegmentId`s (the actual generation of `SegmentId`s occurs inside `Vector::from_target_types` in `vector.append_subpath`).

This means that trying to delete a segment or modify a handle when the `SegmentId`s are not contiguous (e.g. after a boolean operation) will result in confusing results (targeting the wrong segment or no segment at all). From this [Discord report](https://discord.com/channels/731730685944922173/1134981447539761273/1448263107326775316).

See reproduction steps: https://github.com/GraphiteEditor/Graphite/issues/3481#issuecomment-3855946622

Ayush2k02:
https://github.com/user-attachments/assets/0551f0b3-b45d-49fc-a307-0ad323cc9784

Closes #3481, closes #2687, closes #2068

### DIFF
--- a/editor/src/messages/portfolio/document/document_message.rs
+++ b/editor/src/messages/portfolio/document/document_message.rs
@@ -211,7 +211,7 @@ pub enum DocumentMessage {
 		clip_targets: HashSet<NodeId>,
 	},
 	UpdateVectorData {
-		vector_data: HashMap<NodeId, Vector>,
+		vector_data: HashMap<NodeId, Arc<Vector>>,
 	},
 	Undo,
 	UngroupSelectedLayers,
--- a/editor/src/messages/portfolio/document/document_message.rs
+++ b/editor/src/messages/portfolio/document/document_message.rs
@@ -17,6 +17,7 @@ use graphene_std::raster::Image;
 use graphene_std::transform::Footprint;
 use graphene_std::vector::click_target::ClickTarget;
 use graphene_std::vector::style::RenderMode;
+use graphene_std::vector::Vector;
 
 #[impl_message(Message, PortfolioMessage, Document)]
 #[derive(derivative::Derivative, Clone, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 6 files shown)


Ayush2k02:
Hey , @0HyperCube  . Can you review this one please, looks good to me

0HyperCube (approved):
Seems like a reasonable fix; it seems to work well.

Ayush2k02:
Yup, all working after the recent rebase against master

Keavon:
!build

Keavon (approved):

Keavon:
This also apparently fixes #2687 by accident. If you comment in that issue, I can assign it to you also.

PR #3729, opened 2026-02-06
Pan and Zoom on Document when in Node Graph

KshAbo:
https://github.com/user-attachments/assets/e773a7aa-2062-4789-b208-1804168e51e0

Adds:  Alt+MMB dragging to pan the artwork canvas in node graph
           Ctrl+Alt+MMB drag to zoom the artwork canvas in node graph
           
Resolves: https://discord.com/channels/731730685944922173/881073965047636018/1364204626433409046


Keavon:
!build

Keavon:
Update: I started reviewing this but I'm concerned that its implementation is far more complicated than it needs to be, although without solving it myself, I can't be certain. I think the best approach is to have someone else who's had more time getting the experience with the code base to create a second implementation, and if that second attempt ends up being simpler, we merge that one instead.

I'm going to convert this to a draft for now because of that concern and so I can continue reviewing other PRs without this still being *Ready for Review* like the ones I'm working through. The path forward is likely to wait on a second PR for this as a point of comparison.

PR #3731, opened 2026-02-07
Fix Shape tool type dropdown not persisting selection and not excluding Line/Rectangle/Ellipse

Ayush2k02:
This regression was introduced in #2914 (commit 523132da) 

In the SetShape handler, this line was added:
`responses.add(ShapeToolMessage::UpdateOptions(ShapeOptionsUpdate:`

since the setShape was being called with polygon always , the added UpdateOptions call overwrites options.shape_type

below is the call
[responses.add(ShapeToolMessage::SetShape { shape: ShapeType::Polygon });](https://github.com/GraphiteEditor/Graphite/blob/20e12edd45e84a70091a8c143b79a94f29ed98cc/editor/src/messages/tool/tool_message_handler.rs#L80C5-L80C77)

Below are the two main fixes from this PR

- Remove UpdateOptions from SetShape handler - Now SetShape only updates current_shape and calls update_dynamic_hints directly for hint updates, without touching options.shape_type.
- Add SyncShapeWithOptions message - When activating the Shape tool, sync current_shape to match options.shape_type. This ensures the drawing state matches the dropdown selection, especially when returning from Line/Rectangle/Ellipse alias tools.

https://github.com/user-attachments/assets/44b540e3-b03d-4109-870b-90e4ee17b56e

Fixes the issue from this discussion 
https://discord.com/channels/731730685944922173/731738914812854303/1469590633361707204

### DIFF
--- a/editor/src/messages/tool/tool_message_handler.rs
+++ b/editor/src/messages/tool/tool_message_handler.rs
@@ -1,5 +1,5 @@
 use super::common_functionality::shape_editor::ShapeState;
-use super::common_functionality::shapes::shape_utility::ShapeType::{self, Ellipse, Line, Rectangle};
+use super::common_functionality::shapes::shape_utility::ShapeType::{Ellipse, Line, Rectangle};
 use super::utility_types::{ToolActionMessageContext, ToolFsmState, tool_message_to_tool_type};
 use crate::application::generate_uuid;
 use crate::messages::layout::utility_types::widget_prelude::*;


editor/src/messages/tool/tool_messages/shape_tool.rs
```diff
@@ -82,23 +82,42 @@ pub enum ShapeOptionsUpdate {
 #[derive(PartialEq, Clone, Debug, serde::Serialize, serde::Deserialize, specta::Type)]
 pub enum ShapeToolMessage {
 	// Standard messages
-	Overlays { context: OverlayContext },
+	Overlays {
```
Ayush2k02: These formatting changes are required else the CI formatting is failing , the versions I am using locally are 
- rustfmt 1.8.0-stable
- rustc 1.92.0

Ayush2k02: whereas the CI uses rustc 1.93.0

Ayush2k02:
Screenshot 2026-02-09 at 10 09 56 AM

Keavon:
Thanks for fixing these two bugs. But I have a concern, which is that your implementation involved new logic, when in both cases, this already *used* to work and they each broke as a regression at some point in the past few months. I would expect an implementation should track down what broke and then fix what existing logic already performed the functionality for these two issues, rather than leaving behind the other logic in some unknown state and building new logic. Could you please spend some time to go back and check which commit broke both of these issues and post those PRs or commit hashes in a comment here, and then investigate what specific changes within each PR resulted in the regression, and then checking what minimal changes are required to fix the issues and seeing if the ones you did here are redundant?

Ayush2k02:
Yeah, makes sense to fix that. Let me do that

Keavon:
Any updates? Just checking how it's going, since this fix will be one of the more important ones to have merged.

Ayush2k02:
yeah, will get this done today

Ayush2k02:
https://github.com/GraphiteEditor/Graphite/pull/2914

Its this one, from where the bug originated. Can share the bisect logs if you want. I'll fix this now

Ayush2k02:
Have updated the video and PR description, this is reviewable from my side now

Keavon:
!build

Ayush2k02:
any updates regarding this @Keavon  !

Keavon:
!build ([Run ID 22617451730](https://github.com/GraphiteEditor/Graphite/actions/runs/22617451730))

PR #3734, opened 2026-02-08
Add gizmos for interacting with the Spiral node

Annonnymmousss:
chore : Refactoring PR #2851 for current code base with some fixes

https://github.com/user-attachments/assets/756fb3e1-e6d1-44f5-8595-6906489feed2


Keavon:
I moved your fixes back into the original PR #2851, so it can merge with you both as co-authors.

PR #3745, opened 2026-02-10
Simplify layer structure encoding and decoding

### LINKED ISSUE #3742: Simplify layer structure hierarchy plumbing
See `fn serialize_structure` and `fn serialize_root`:

https://github.com/GraphiteEditor/Graphite/blob/master/editor/src/messages/portfolio/document/document_message_handler.rs#L1894-L1958

They implement an over-engineered encoding of the hierarchical depths of a list of layers.

And see `function newUpdateDocumentLayerStructure`:

https://github.com/GraphiteEditor/Graphite/blob/master/frontend/src/components/panels/Layers.svelte#L138-L190

It implements the decoding of that over-engineered encoding into a JS format.

We want to replace this complexity with a much simpler flat structure of structs that become JS objects using the same way this happens with all our other data types sent from Rust to JS (via serde and [`messages.ts`](https://github.com/GraphiteEditor/Graphite/blob/master/frontend/src/messages.ts)).

Instead of encoding things "cleverly" using a custom schema documented in that verbose comment in the first code snippet link, where negative signs correspond to hierarchy level shifting and such-and-such, it should just use the simplest form that directly describes the same concept but without the bit fiddling and custom schema.

krVatsal:
Fixes #3742
This PR replaces the bit-packed encoding with a flat list of entries, each containing:

`layerId`: The unique ID of the layer.
`depth`: The nesting depth (0 = root).
`isLastInParent`: Whether this is the last child in its parent (for efficient UI rendering).

This format is:
-More compact than a tree/bti-packed, but still human-readable and easy to debug.
-Easier to serialize/deserialize using standard `serde` and class-transformer pipelines.
-Faster to process in the frontend, as it avoids recursion and allows for simple, index-based hierarchy reconstruction.

Tests
Added Rust unit tests:
-`test_serialize_root_empty_document`
-`test_serialize_root_flat_structure`
-`test_serialize_root_excludes_collapsed_children`

### DIFF
--- a/desktop/src/cef/context/builder.rs
+++ b/desktop/src/cef/context/builder.rs
@@ -1,9 +1,9 @@
 use std::path::{Path, PathBuf};
 
 use cef::args::Args;
-use cef::sys::{CEF_API_VERSION_LAST, cef_resultcode_t};
+use cef::sys::{CEF_API_VERSION_LAST, cef_log_severity_t, cef_resultcode_t};
 use cef::{
-	App, BrowserSettings, CefString, Client, DictionaryValue, ImplCommandLine, ImplRequestContext, RequestContextSettings, SchemeHandlerFactory, Settings, WindowInfo, api_hash,
+	App, BrowserSettings, CefString, Client, DictionaryValue, ImplCommandLine, ImplRequestContext, LogSeverity, RequestContextSettings, SchemeHandlerFactory, Settings, WindowInfo, api_hash,
 	browser_host_create_browser_sync, execute_process,
 };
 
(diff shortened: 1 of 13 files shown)


Keavon:
Thanks for attempting this along with @Ayush2k02 since it was valuable to see the two approaches compared side-by-side. I ended up preferring the simplicity of #3744 even though it does involve sending the nested structure from Rust (I don't think this is likely to be a performance concern, since your approach which matches the original logic more closely does still have to build a nested structure in JS after, likely negating any perf benefits of a flat structure). Both seem like good implementations, but it always is best to pick the simplest. Thanks :)

PR #3755, opened 2026-02-13
Improve CI times and make merge queue only trigger one CI run

TrueDoctor:
• Splits `rust-fmt` into a separate job so authors get feedback on their PRs more quickly
• Frontend linting can't easily be split the same way since the linting depends on the wasm module to be present
• Removes the explicit build step from the ci pipeline (we should catch the build failures in the other phases)
• Splits the action into two separate ones, one for the deploy when pushing to master and one for regular ci jobs
• Don't run the entire rust build for pushes to master, these have been checked in the pr beforehand. When force pushing to master manually, the authors should test that their code compiles themselves, that is not what the ci is for.
• Adds the skip-merge-queue action to avoid running the ci twice when a merge queue is used


timon-schelling (approved):
LGTM, we could do dev builds for PRs and only release builds on master

Keavon:
> LGTM, we could do dev builds for PRs and only release builds on master

Do you mean for build links? I think we should always use release optimizations for build links.

PR #3759, opened 2026-02-13
Fix incorrect flattening behavior for hidden nodes

### LINKED ISSUE #3629: Crash when hiding an 'Instance Position' or 'Pointer Position' node feeding a value

JustJ01:
Closes #3629

When a node like **Instance Position** or **Pointer Position** is hidden, the flattening logic incorrectly replaces it with a default `()` value.  

This causes:

- Type mismatches in downstream nodes (Ex `ConvertNode<DVec2>`)
- Graph evaluation failures
- Panics such as:
  - `Cannot construct default value for hidden node`
  - `entered unreachable code: tried to resolve not flattened node`

Before:

### Expected Behavior

Hiding a node should:
- Preserve its output type
- Forward a compatible input when possible
- Never inject `()` unless the node’s real output type is `()`

---
## Solution

https://github.com/user-attachments/assets/4a53e410-a5fa-47a0-8211-d7a796a186c5

Updated `flatten_with_fns` handling for `Visible::Value(output_type)`.

### New behavior:

1. **If a compatible input exists**
   - Replace the node with an identity node
   - Forward that input directly
   - Preserve type integrity

2. **If no compatible input exists**
   - Generate a default value using:
     ```rust
     TaggedValue::from_type(&output_type)
     ```
   - Insert a correctly typed constant node

3. **Prevent incorrect `()` propagation**
   - The logic no longer defaults to `()` unless the node’s actual output type is unit.

---

## Result

- Hiding `Instance Position` no longer crashes.
- `ConvertNode<DVec2>` no longer receives `()`.

---

### DIFF
--- a/editor/src/messages/portfolio/document/utility_types/network_interface.rs
+++ b/editor/src/messages/portfolio/document/utility_types/network_interface.rs
@@ -10,18 +10,16 @@ use crate::messages::portfolio::document::graph_operation::utility_types::Modify
 use crate::messages::portfolio::document::node_graph::document_node_definitions::{DefinitionIdentifier, resolve_document_node_type};
 use crate::messages::portfolio::document::node_graph::utility_types::{Direction, FrontendClickTargets, FrontendGraphDataType, FrontendGraphInput, FrontendGraphOutput};
 use crate::messages::portfolio::document::overlays::utility_functions::text_width;
-use crate::messages::portfolio::document::utility_types::network_interface::resolved_types::ResolvedDocumentNodeTypes;
-use crate::messages::portfolio::document::utility_types::network_interface::resolved_types::TypeSource;
+use crate::messages::portfolio::document::utility_types::network_interface::resolved_types::{ResolvedDocumentNodeTypes, TypeSource};
 use crate::messages::portfolio::document::utility_types::wires::{GraphWireStyle, WirePath, WirePathUpdate, build_vector_wire};
 use crate::messages::tool::common_functionality::graph_modification_utils;
 use crate::messages::tool::tool_messages::tool_prelude::NumberInputMode;
 use deserialization::deserialize_node_persistent_metadata;
 use glam::{DAffine2, DVec2, IVec2};
 use graph_craft::Type;
-use graph_craft::concrete;
 use graph_craft::document::value::TaggedValue;
 use graph_craft::document::{DocumentNode, DocumentNodeImplementation, NodeId, NodeInput, NodeNetwork, OldDocumentNodeImplementation, OldNodeNetwork};
-use graph_craft::document::{HiddenNodeInput, Visible};
+use graph_craft::document::{Hidden, HiddenNodeInput};
 use graphene_std::ContextDependencies;
 use graphene_std::math::quad::Quad;
 use graphene_std::subpath::Subpath;
@@ -1052,7 +1050,7 @@ impl NodeNetworkInterface {
 			log::error!("Could not get node in is_visible");
 			return false;
 		};
-		node.visible.is_visible()
+		node.visible.is_none()
 	}
 
 	pub fn is_layer(&self, node_id: &NodeId, network_path: &[NodeId]) -> bool {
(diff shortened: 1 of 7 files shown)


editor/src/messages/portfolio/document/utility_types/network_interface/resolved_types.rs
```diff
@@ -203,7 +203,7 @@ impl NodeNetworkInterface {
 				concrete!(())
 			}
 		};
-		TaggedValue::from_type_or_none(&guaranteed_type)
+		TaggedValue::from_type(&guaranteed_type).expect("Failed to construct TaggedValue for identity type")
```
adamgerhant: This can crash when used on types from old documents which are no longer handled by the macro. A log::error! with the `from_type_or_none` would be safer

node-graph/graph-craft/src/document.rs
```diff
@@ -29,6 +29,32 @@ fn return_true() -> bool {
...
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct HiddenNodeInput {
+	pub input_index: usize,
+	pub tagged_value: TaggedValue,
+}
+
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub enum Visible {
+	Passthrough,
+	Value(Type), //kept for backward compatibility
```
adamgerhant: backward compatibility with what? This is a new enum.

node-graph/graph-craft/src/document.rs
```diff
@@ -29,6 +29,32 @@ fn return_true() -> bool {
 	true
 }
 
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct HiddenNodeInput {
+	pub input_index: usize,
+	pub tagged_value: TaggedValue,
+}
```
adamgerhant, line 37: I would expect this to include the NodeId so it can be matched with the downstream (toward the export) node

node-graph/graph-craft/src/document.rs
```diff
@@ -50,8 +76,8 @@ pub struct DocumentNode {
 	// A nested document network or a proto-node identifier.
 	pub implementation: DocumentNodeImplementation,
 	/// Represents the eye icon for hiding/showing the node in the graph UI. When hidden, a node gets replaced with an identity node during the graph flattening step.
-	#[serde(default = "return_true")]
-	pub visible: bool,
+	#[serde(default = "return_visible_true")]
+	pub visible: Visible,
```
adamgerhant: This needs to be an option. None means visible, Some means hidden with the appropriate handling.

node-graph/graph-craft/src/document.rs
```diff
@@ -29,6 +29,32 @@ fn return_true() -> bool {
...
 
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct HiddenNodeInput {
+	pub input_index: usize,
+	pub tagged_value: TaggedValue,
+}
+
+#[derive(Clone, Debug, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub enum Visible {
+	Passthrough,
```
adamgerhant, line 41: Passthrough means to use the current hiding logic, where it is replaced with a Passthrough node

node-graph/graph-craft/src/document.rs
```diff
@@ -29,6 +29,32 @@ fn return_true() -> bool {
...
+	Value(Type), //kept for backward compatibility
+	TaggedValues(Vec<HiddenNodeInput>),
```
adamgerhant, line 42: Tagged values only occurs for nodes with no inputs, where valid input values must be chosen for the nodes it is connected to

editor/src/messages/portfolio/document/utility_types/network_interface.rs
```diff
@@ -4481,15 +4484,49 @@ impl NodeNetworkInterface {
 	}
 
 	pub fn set_visibility(&mut self, node_id: &NodeId, network_path: &[NodeId], is_visible: bool) {
-		let Some(network) = self.network_mut(network_path) else {
-			return;
-		};
-		let Some(node) = network.nodes.get_mut(node_id) else {
-			log::error!("Could not get node {node_id} in set_visibility");
+		if is_visible {
```
adamgerhant, line 4485: This logic is generally wrong. For nodes with no inputs, we want to use the downstream traversal and type lookup solution. For nodes with any number of inputs they just become pass through nodes as is the current behavior.

adamgerhant (changes requested):

node-graph/graph-craft/src/document.rs
adamgerhant: Also it might be more clear to rename this to Hidden, since it is just behavior for Hidden nodes

PR #3766, opened 2026-02-15
added feature to snap at midpoint while using path tool

### LINKED ISSUE #3760: Segment molding Shift key X/Y constraint
Webcast explanation:

https://github.com/user-attachments/assets/9d8dd585-55b2-4b1a-bf21-2ab7bbb23afb

This, of course, should reuse the code for showing the X/Y snapping lines.

krVatsal:
fixes #3760 

Now when we use path tool to snap any shape(while holding shift) then it snaps from midpoint irrespective of wherever we drag from.


krVatsal:
One small question, does the axes are meant to be at drag point or at the midpoint?

Annonnymmousss:
Midpoint of the segment.

Keavon:
!build

Keavon:
This is behaving correctly except for the fact that we want the X and Y snapping lines to be centered through the initial position of the midpoint of the Bezier segment being molded, rather than the point at which the pointer began dragging the molded segment.

krVatsal:
Yepp, was working on that part only. Almost done with the work I will push the new commit with that change.

Keavon:
!build ([Run ID 22294708633](https://github.com/GraphiteEditor/Graphite/actions/runs/22294708633))

Keavon:
The expected midpoint (t=0.5, where the X and Y axes should cross through) should be here, roughly where I marked it with the red dot:

image

Keavon:
!build ([Run ID 22639891656](https://github.com/GraphiteEditor/Graphite/actions/runs/22639891656))

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -2371,6 +2399,10 @@ impl Fsm for PathToolFsmState {
 				tool_data.molding_segment = false;
 				tool_data.temporary_adjacent_handles_while_molding = None;
 				tool_data.angle_locked = false;
+				tool_data.snapping_axis = None;
+				tool_data.snap_axis_origin = None;
+				tool_data.snapping_axis = None;
```
Ayush2k02: why are you setting this to NONE twice ??

Ayush2k02 (changes requested):

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -2619,6 +2651,8 @@ impl Fsm for PathToolFsmState {
 				}
 
 				tool_data.snapping_axis = None;
+				tool_data.snap_axis_origin = None;
```
Ayush2k02, line 2653: same here !

Ayush2k02 (changes requested):

editor/src/messages/tool/tool_messages/path_tool.rs
krVatsal: ohh., thanks for pointing out, ig i just pasted it twice while testing.

PR #3768, opened 2026-02-15
Fix inability to navigate up and down with arrow keys in interactive-style menu lists

### LINKED ISSUE #3586: Font selection menu keybinds
Using arrow keys to navigate to a font only works as far as one in either direction of the selected font. Also, using enter to select a font (from the arrow keys, or by searching) doesn't do anything.

https://github.com/user-attachments/assets/7e82b129-b73c-48e1-a54d-3ca11b099a34

JustJ01:
Closes #3586 

While selecting font of a text, there were few issues:

### Before:

- Arrow Navigation handling was buggy, like get stuck at one item away from selected font.
- `Enter` didn't applied to highlighted font correctly.

https://github.com/user-attachments/assets/b75017b5-3afd-4c5a-83f4-0097160105dd

### After:

I removed the line that reset highlighted to activeEntry on every key press. That reset was causing the issues.

https://github.com/user-attachments/assets/e7c4cfde-4d80-41d9-b9aa-8a9f8740c33e

### DIFF
--- a/frontend/src/components/floating-menus/MenuList.svelte
+++ b/frontend/src/components/floating-menus/MenuList.svelte
@@ -252,9 +252,6 @@
 	/// Handles keyboard navigation for the menu.
 	// Returns a boolean indicating whether the entire menu stack should be dismissed.
 	export function keydown(e: KeyboardEvent, submenu = false): boolean {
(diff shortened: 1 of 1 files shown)


JustJ01:
@Keavon review pls!

PR #3770, opened 2026-02-15
Fix Vello canvas on web not being transparent by using premultiplied alpha mode

timon-schelling:
Important note the surface caps doesn't report `PreMultiplied`, wasn't able to find out why. But this seems to work for all browsers I was able to test.

before

image

after

image


timon-schelling:
@TrueDoctor Alternative would be setting a background color to the background color in the render node.

TrueDoctor:
I think I did have issues with safari at some point so we should double-check to make sure it works everywhere. Otherwise, I think this should be fine

Keavon:
Tested as working on Safari 26.2 as well. So we're good to go with this fix.

PR #3772, opened 2026-02-16
Desktop: Auto-fallback to non-accelerated UI when CEF texture fails at startup

### LINKED ISSUE #3667: On desktop, automatically fall back to non-hardware accelerated UI on failure

JustJ01:
Closes #3667

This PR implements an automatic compatibility fallback for desktop startup when accelerated UI fails to render(black screen).

I have changed:
1) Added a startup watchdog that waits for the first CEF UI texture and triggers fallback if none is received within 1 second.
2) Added persisted startup setting, ui_compatibility_mode in startup-settings.ron
3) Updated startup flow:
- If compatibility mode is enabled (persisted or CLI), launch with UI acceleration disabled.
- On watchdog failure, persist compatibility mode and relaunch automatically.
4) Added status bar indicator in compatibility mode ( a warning IconButton appears beside app version in bottom-right status area)

Screenshot 2026-02-16 162405

5) If you press that icon , it relaunches with compatibility mode disabled , if accelerated mode still fails, watchdog fallback re-enables compatibility mode.

- Users with startup blank screen should be auto-recovered into compatibility mode. 
- Compatibility mode state survives relaunches. 
- Screenshots/screencasts clearly show compatibility mode via status bar icon. 
- Users can manually retry normal mode from the icon.

### Manual Setup ( Tested on Windows)
1) Run this command:  `Remove-Item "$env:APPDATA\Graphite\startup-settings.ron" -ErrorAction SilentlyContinue`
2) `cargo run -p graphite-desktop`
- Apps opens normally , no compatibility icon in bottom right status area
Screenshot 2026-02-16 161928

3)  Force compatibility mode UI (to test icon + relaunch button):
`@'
(
    ui_compatibility_mode: true,
)
'@ | Set-Content "$env:APPDATA\Graphite\startup-settings.ron"`

Launch again:

`cargo run -p graphite-desktop`
Expected: Apps runs in compatibility mode , warning icon appears beside version text in status bar 

Screenshot 2026-02-16 162405

4) Test disable compatibility mode button:
- Click the warning icon in status bar. App relaunches 
- startup-settings.ron is updated to `ui-compatibility: false`, to verify it run `Get-Content "$env:APPDATA\Graphite\startup-settings.ron"`
- Icon disappears if accelerated mode works
5) Test auto-fallback path : This only triggers when accelerated UI fails to produce a CEF texture within ~1s.
Expected on affected machine:
- Log: `No CEF texture received within the first second...` 
- App restarts automatically in compatibility mode.
- startup-settings.ron becomes true
- Warning icon is visible after relaunch


JustJ01:
@Keavon pls review!

timon-schelling:
Sorry @JustJ01 we decided against this approach, as we will need restart functionality for other preferences as well and in the end didn't like the UX of what is described in the issue.

For future reference, issues that are already assigned, especially those assigned to the core team, are likely not a good fit for beginners. Assigned issues are generally not available to work on unless otherwise stated.

Closing in favor of #3774

PR #3773, opened 2026-02-16
Line Tool: Only lock angle with Ctrl after drag has started

### LINKED ISSUE #3771: Line tool needs Ctrl to lock angle only after drag has begun
Ctrl is the shortcut for the Line tool to lock the current angle. But currently, it takes effect if Ctrl is held before the click and drag for drawing a new line, which is not desired, because Ctrl is also used to allow drawing a new line snapped to the endpoint of an existing line (rather than dragging around that endpoint). Ctrl should only lock the line's angle if it is pressed after the dragging is already occurring, whether that's for a new line or for dragging the endpoint of an existing line.

JustJ01:
Closes #3771

Implemented the fix so Ctrl only angle locks the Line tool after drag has already started.

Applies to both drawing a new line and dragging an existing line endpoint.

Added per-drag eligibility tracking for lock-angle and reset it at drag lifecycle boundaries.

Also updated tests:

1. `test_line_tool_ctrl_before_drag_start_does_not_lock_angle`
2. `test_line_tool_ctrl_pressed_during_drag_locks_angle`


JustJ01:
@Keavon @timon-schelling @adamgerhant pls review!

timon-schelling:
Thanks for the contribution. Tagging us will not get the PR reviewed quicker and just adds noise. We see open PRs and will review them when we find the time.

JustJ01:
Sorry for unnecessary tagging , won't do it again.

PR #3774, opened 2026-02-16
Desktop: Add "Disable UI Acceleration" to preferences

### LINKED ISSUE #3667: On desktop, automatically fall back to non-hardware accelerated UI on failure

timon-schelling:
Closes #3667


Keavon (approved):

Keavon:
I ended up making it so clicking out of the restart dialog dismisses it just like the "Later" button since that felt like the better behavior.

PR #3783, opened 2026-02-18
Desktop: Make shutdown more robust and fix panic caused by invalid viewport scale

### LINKED ISSUE #3514: On Windows desktop, closing the app takes several seconds
The window remains frozen while all CEF processes are being shut down gracefully before the window unfreezes and closes, maybe we should kill them non-gracefully?

timon-schelling:
fixes

```
thread 'main' (6871) panicked at editor/src/messages/viewport/viewport_message_handler.rs:526:5:
assertion `left == right` failed: Cannot multiply with diffent scale
  left: NaN
 right: NaN
 ```

Re-fixes #3514.

### DIFF
--- a/desktop/src/app.rs
+++ b/desktop/src/app.rs
@@ -1,6 +1,8 @@
 use rand::Rng;
 use rfd::AsyncFileDialog;
 use std::fs;
+use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


Keavon:
This fixes slow shutdown on Windows as well as stuttery window minimization.

PR #3788, opened 2026-02-18
Fix missing crash dialog for panics thrown inside requestAnimationFrame loop

### LINKED ISSUE #3402: Panics originating in a requestAnimationFrame loop silently freeze the editor
We expect that anytime Graphite's Rust code panics, it should display the crash dialog. This can be triggered successfully from the *Help* > *Developer Debug* > *Trigger a Crash* menu, which creates the following call stack:

Image

But some of our code (which is also susceptible to panicking) runs in the recursive `requestAnimationFrame` loop, where its call stack is essentially infinite (there is one per frame since the application was initialized— notice the scrollbar below which is also still truncated):

Image

One good way to reproduce this is to add a `panic!();` line to the body of `fn brightness_contrast_properties()` in `node_properties.rs`, then create a Brightness/Contrast node in the graph, and then select it so that function runs to generate its Properties panel content. The editor will freeze, but won't display a crash dialog. Check the console and you'll see a panic log with a really long call stack due to the `requestAnimationFrame` recursion.

This needs to be solved so our panic handler architecture is able to detect and handle this class of situations just like it does in the regular non-recursive call stack situation.

This is important because freezing without informing the user explicitly that a crash occurred is confusing and frustrating, and means that we won't receive crash reports about those situations either.

JustJ01:
Closes #3402 

Panics triggered from code running in the recursive `requestedAnimationFrame` path could freeze the editor without showing the crash dialog.

This PR makes those panics follow the same crash-reporting flow as other panics.

### The Root Cause: 
The panic handler could clash while trying to lock `EDITOR_HANDLE` during `requestedAnimationFrame` dispatch, so the crash dialog message was not always sent in that path.

Old panic code used a blocking lock (`lock()`) , so panic handling could get stuck waiting for that same lock, as a result app froze.

What I changed:

- `panic_hook()` now uses non blocking strategy:
    - Try send_panic_dialog_via_callback() first.
    - If that fails , try `send_panic_dialog()`, uses `try_lock()`, not lock(),
    - If that also fails, schedule `send_panic_dialog_deferred()` with `setTimeout(0)` and retry next tick.
 - The raw frontend JS callback is stored separately in thread-local state: `PANIC_DIALOG_MESSAGE_CALLBACK`, that callback is set during `EditorHandle::create()`, this path avoids `EDITOR_HANDLE` mutex contention entirely, so it stills works during `requestedAnimationFrame` related panic timing.

### Verification:

Add `panic!()` in `brightness_contrast_properties()` and select Brightness/Contrast node , Crash dialog now appears (no freeze)

https://github.com/user-attachments/assets/fa4a302b-ec32-4d13-8bd3-a5921937ac60

### DIFF
--- a/frontend/wasm/src/lib.rs
+++ b/frontend/wasm/src/lib.rs
@@ -74,11 +74,8 @@ pub fn panic_hook(info: &panic::PanicHookInfo) {
 	log::error!("{info}");
 
 	// Prefer using the raw JS callback to avoid mutex lock contention inside the panic hook.
-	// Fall back to the editor handle path if needed.
 	if let Err(info) = send_panic_dialog_via_callback(info) {
-		if let Err(info) = send_panic_dialog(info) {
-			send_panic_dialog_deferred(info);
-		}
+		send_panic_dialog_deferred(info);
 	}
 }
 
@@ -114,22 +111,10 @@ fn send_panic_dialog_via_callback(panic_info: String) -> Result<(), String> {
 	})
 }
 
-fn send_panic_dialog(panic_info: String) -> Result<(), String> {
-	EDITOR_HANDLE.with(|editor_handle| {
-		let mut guard = editor_handle.try_lock();
-		let Ok(Some(handle)) = guard.as_deref_mut() else {
-			return Err(panic_info);
-		};
-
-		handle.send_frontend_message_to_js_rust_proxy(FrontendMessage::DisplayDialogPanic { panic_info });
-		Ok(())
-	})
-}
-
 #[cfg(not(feature = "native"))]
 fn send_panic_dialog_deferred(panic_info: String) {
 	let callback = Closure::once_into_js(move || {
-		if send_panic_dialog(panic_info).is_err() {
+		if send_panic_dialog_via_callback(panic_info).is_err() {
 			log::error!("Failed to send crash dialog after panic because the editor handle is unavailable");
 		}
 	});


TrueDoctor:
@JustJ01 the suggested allocation removal is bs. Calling `.clone()` on a `&String` will still allocate and basically compiles to the same code as would `&str`. But you could try taking an owned `String` as the argument to the function (instead of the `&String`) to actually avoid the allocation

JustJ01:
> @JustJ01 the suggested allocation removal is bs. Calling `.clone()` on a `&String` will still allocate and basically compiles to the same code as would `&str`. But you could try taking an owned `String` as the argument to the function (instead of the `&String`) to actually avoid the allocation

@TrueDoctor I have made the changes

TrueDoctor:
Have you verified that all of these fallbacks are needed? If not, I'd prefer to only keep the necessary ones to avoid excessive code bloat

JustJ01:
> Have you verified that all of these fallbacks are needed? If not, I'd prefer to only keep the necessary ones to avoid excessive code bloat

Verified. The EDITOR_HANDLE fallback was redundant (same callback path + extra lock dependency), so I removed it and kept only direct callback + deferred callback retry.

TrueDoctor (approved):

TrueDoctor:
!build ([Run ID 22281477614](https://github.com/GraphiteEditor/Graphite/actions/runs/22281477614))

PR #3793, opened 2026-02-20
Generalize the 'Map Vector' node as 'Map' with support for all graphical types

TrueDoctor:
image


Keavon (approved):
Awesome, I am very glad to generalize this, thank you for making it happen!

PR #3795, opened 2026-02-21
Pen/Freehand/Splines tool set its origin to the first drawn point

### LINKED ISSUE #2776: Pen/Freehand/Splines tool should set its origin to the first drawn point location when the new layer is created
Right now, the origin gets set to the parent (usually canvas) origin which is often far away from where the user is drawing. At least we know that, when drawing a shape and its layer gets created, the user likely has an interest in a particular area in space roughly nearby the first point being drawn. So that's the better choice. This means adding a Transform node to provide the offset to the first point location.

jsjgdh:
closes #2776


Keavon:
Superseded by #3872.

PR #3801, opened 2026-02-22
Add Global Eyedropper for Windows  

yush-1018:
Added a global eyedropper tool for the desktop app on Windows. 

Changes include:
- A floating magnifier window that follows the cursor.
- Native Windows GDI integration to sample colors from any application.
- Seamless color updates in the Graphite editor.

### DIFF
--- a/editor/src/messages/portfolio/document/data_panel/data_panel_message_handler.rs
+++ b/editor/src/messages/portfolio/document/data_panel/data_panel_message_handler.rs
@@ -433,10 +433,6 @@ impl TableRowLayout for Vector {
 						TextLabel::new(format_transform_matrix(&stroke.transform)).narrow(true).widget_instance(),
 					]);
 					table_rows.push(vec![
-						TextLabel::new("Stroke Non-Scaling").narrow(true).widget_instance(),
-						TextLabel::new((if stroke.non_scaling { "Yes" } else { "No" }).to_string()).narrow(true).widget_instance(),
-					]);
-					table_rows.push(vec![
 						TextLabel::new("Stroke Paint Order").narrow(true).widget_instance(),
 						TextLabel::new(stroke.paint_order.to_string()).narrow(true).widget_instance(),
 					]);
--- a/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
+++ b/editor/src/messages/portfolio/document/graph_operation/graph_operation_message_handler.rs
@@ -515,7 +515,6 @@ fn apply_usvg_stroke(stroke: &usvg::Stroke, modify_inputs: &mut ModifyInputsCont
 			align: StrokeAlign::Center,
 			paint_order: PaintOrder::StrokeAbove,
 			transform,
-			non_scaling: false,
 		})
 	}
 }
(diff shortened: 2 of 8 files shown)


yush-1018:
Addressed all review feedback. 

Thanks!

timon-schelling:
Looks like this doesn't compile.

The global eyedropper feature will need to be designed in a way that allows for a consistent API for all platforms.

Currently all platforms specific APIs are abstracted away with the native window handle. Ideally we would also do that for the eyedropper or make that more generic, something like a platform abstraction.

Implementation for Wayland/mac needs some research (this PR can still only implement Windows support).
But the eyedropper interface needs to work for all 3 platforms.

desktop/src/global_eyedropper/win.rs
```diff
@@ -0,0 +1,176 @@
...
+
+		let size = PhysicalSize::new(MAGNIFIER_SIZE, MAGNIFIER_SIZE);
+		window.set_outer_position(PhysicalPosition::new(position.x - size.width as f64 / 2., position.y - size.height as f64 / 2.));
+		window.set_min_surface_size(Some(size.into()));
+		window.set_visible(true);
+		window.request_redraw();
+	}
+
+	fn render(&self) {
+		let Some(_window) = &self.window else { return };
```
Ayush2k02, line 74: the window variable is unused, consider using this

`if self.window.is_none() { return; }`

desktop/src/global_eyedropper/win.rs
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			let pixel_size = MAGNIFIER_SIZE / MAGNIFIER_RES;
+			let half = MAGNIFIER_RES as i32 / 2;
+
+			let desktop_dc = GetDC(HWND::default());
+			let window_hwnd = self.window_hwnd();
+			let window_dc = GetDC(window_hwnd);
+
+			let mem_dc = CreateCompatibleDC(desktop_dc);
```
Ayush2k02, line 89: I see that you are doing this create and delete cycle on every render, Consider using the below code for optimization , basically instead do caching in GlobalEyedropperImpl struct . Since render happens very frequently , this should be a good caching optimization

```
pub(crate) struct GlobalEyedropperImpl {
    window: Option<Window>,
    primary: bool,
    mem_dc: HDC,      // Created once in start()
    bitmap: HBITMAP,  // Created once in start()
}

fn start(&mut self, ...) {
    // Create window...
    
    // Allocate GDI resources ONCE
    let desktop_dc = GetDC(HWND::default());
    self.mem_dc = CreateCompatibleDC(desktop_dc);
    self.bitmap = CreateCompatibleBitmap(desktop_dc, 11, 11);
    SelectObject(self.mem_dc, self.bitmap);
    ReleaseDC(HWND::default(), desktop_dc);
}

fn render(&self) {
    // Just use the cached resources - no allocation
    StretchBlt(self.mem_dc, ...);
    StretchBlt(window_dc, ...
}

fn stop(&mut self) {
    // Deallocate ONCE when eyedropper closes
    DeleteObject(self.bitmap);
    DeleteDC(self.mem_dc);
    self.window = None;
}
```

desktop/src/global_eyedropper/mac.rs
Ayush2k02, line 1: can this and the linux.rs be absracted ? Not super sure on this but the code is duplicated here

Keavon:
What Timon mentioned is correct and I am not seeing how this PR is likely to have a path towards being merged because its approach does not follow one that meets our required direction. I think the best option is to close it and discuss any plans towards a proper implementation, if interest arises, in an issue rather than this PR. Thanks for the attempt.

PR #3806, opened 2026-02-22
Implement node output destructuring

JustJ01:
From the discord discussions: https://discord.com/channels/731730685944922173/1210129484255076354/1475229902788628573


Keavon:
Please rebase on the latest upstream master.

node-graph/nodes/gcore/src/extract_xy.rs
```diff
@@ -2,6 +2,14 @@ use core_types::Ctx;
 use dyn_any::DynAny;
 use glam::{DVec2, IVec2, UVec2};
 
+#[derive(Debug, Clone, DynAny, node_macro::Destruct)]
+pub struct SplitVec2Output {
+	#[output(name = "X")]
+	pub x: f64,
+	#[output(name = "Y")]
```
TrueDoctor: we should not require these name annotations

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -1309,62 +1219,16 @@ fn document_node_definitions() -> HashMap<DefinitionIdentifier, DocumentNodeDefi
...
-						]
-						.into_iter()
-						.enumerate()
-						.map(|(id, node)| (NodeId(id as u64), node))
-						.collect(),
-
-						..Default::default()
-					}),
+					document_node: DocumentNode {
+					implementation: DocumentNodeImplementation::ProtoNode(extract_xy::split_vec_2::IDENTIFIER),
```
TrueDoctor: This should be gone completely

TrueDoctor:
Also, please configure your ai / editor / git to automatically run `cargo fmt` on save or pre commit

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
JustJ01: Just wanted to confirm did you meant by removing the manual DocumentNodeDefinition entry , as I have removed the old Network-based implementation and kept the ProtoNode definition.

Keavon: Basically, anything no longer used with this new system can be considered dead code and should be removed, which applies to anything throughout any file.

JustJ01:
@TrueDoctor I have made the requested changes which u asked for

node-graph/nodes/raster/src/adjustments.rs
```diff
@@ -120,6 +129,17 @@ fn extract_channel<T: Adjust<Color>>(
 	input
 }
 
+#[cfg(feature = "std")]
+#[node_macro::node(name("Split Channels"), category("Raster: Channels"), deconstruct_output)]
+fn split_channels(_: impl Ctx, input: Table<Raster<CPU>>) -> SplitChannelsOutput {
```
Ayush2k02, line 134: Each call to extract_channel iterates over the entire raster. This means the image is traversed 4 times. Maybe do a single-pass implementation ?

Maybe do something like this 

```
fn split_channels(_: impl Ctx, input: Table<Raster<CPU>>) -> SplitChannelsOutput {
    let (red, green, blue, alpha) = input.into_four_channels(); // hypothetical single-pass function
    SplitChannelsOutput { red, green, blue, alpha }
}
```

I can commit this function and implimentation if you'd like

Ayush2k02 (changes requested):

node-graph/nodes/raster/src/adjustments.rs
TrueDoctor: I think in this case we should not touch the split channels node at all since it makes the code less efficient, this was done as per Keavons request.

node-graph/node-macro/src/destruct.rs
```diff
@@ -0,0 +1,114 @@
...
+
+	Ok(quote! {
+		#(#node_implementations)*
+
+		impl #impl_generics #core_types::registry::Destruct for #struct_name #ty_generics #where_clause {
+			fn fields() -> &'static [#core_types::registry::StructField] {
+				static FIELDS: std::sync::OnceLock<Vec<#core_types::registry::StructField>> = std::sync::OnceLock::new();
+				FIELDS.get_or_init(|| vec![
+					#(#output_fields,)*
+				]).as_slice()
```
TrueDoctor, line 71: Why do you use heap memory + syncronization here instead of constructing a slice in place?

node-graph/node-macro/src/destruct.rs
```diff
@@ -0,0 +1,114 @@
...
+			}
+		}
+	})
+}
+
+fn generate_extractor_node(core_types: &TokenStream2, fn_name: &syn::Ident, struct_name: &syn::Ident, field_name: &syn::Ident, ty: &Type, output_name: &LitStr) -> TokenStream2 {
+	quote! {
+		#[node_macro::node(category(""), name(#output_name))]
```
TrueDoctor, line 79: Why do we add the `name()` annotation here?

node-graph/node-macro/src/destruct.rs
```diff
@@ -0,0 +1,114 @@
...
+
+		let this_output_name = this_output_name.ok_or_else(|| Error::new_spanned(attr, "Missing output name. Use #[output(name = \"...\")]"))?;
+		if output_name.is_some() {
+			return Err(Error::new_spanned(attr, "Multiple #[output(...)] attributes are not allowed on one field"));
+		}
+		output_name = Some(this_output_name);
+	}
+
+	Ok(output_name)
+}
```
TrueDoctor, line 131: Is this function only used for overriding the field name? I don't think this code is currently used anywhere right? If it is indeed not used, it should be removed to reduce the code complexity

node-graph/preprocessor/src/lib.rs
```diff
@@ -117,16 +135,72 @@ pub fn generate_node_substitutions() -> HashMap<ProtoNodeIdentifier, DocumentNod
...
 
-		nodes.insert(NodeId(input_count as u64), document_node);
+		nodes.insert(source_node_id, document_node);
+
+		let use_memo = !available_output_fields.is_empty()
+			&& node_registry.get(&memo_node).is_some_and(|memo_implementations| {
+				memo_implementations
+					.iter()
+					.any(|(_, node_io)| node_io.call_argument == *input_type && node_io.return_value == first_node_io.return_value)
+			});
```
TrueDoctor, line 157: use_memo will currently always be false since you did not the new structs to the memo node implementations

node-graph/preprocessor/src/lib.rs
```diff
@@ -117,16 +135,72 @@ pub fn generate_node_substitutions() -> HashMap<ProtoNodeIdentifier, DocumentNod
...
+		if use_memo {
+			let memo_node_id = NodeId((input_count + generated_node_count) as u64);
+			generated_node_count += 1;
+			nodes.insert(
+				memo_node_id,
+				DocumentNode {
+					inputs: vec![NodeInput::node(source_node_id, 0)],
+					call_argument: input_type.clone(),
+					implementation: DocumentNodeImplementation::ProtoNode(memo_node.clone()),
+					visible: false,
```
TrueDoctor, line 168: there is no need to set visible to false since the entire network is hidden anyway

node-graph/preprocessor/src/lib.rs
```diff
@@ -117,16 +135,72 @@ pub fn generate_node_substitutions() -> HashMap<ProtoNodeIdentifier, DocumentNod
...
+
+		if use_memo {
+			let memo_node_id = NodeId((input_count + generated_node_count) as u64);
+			generated_node_count += 1;
+			nodes.insert(
+				memo_node_id,
+				DocumentNode {
+					inputs: vec![NodeInput::node(source_node_id, 0)],
+					call_argument: input_type.clone(),
+					implementation: DocumentNodeImplementation::ProtoNode(memo_node.clone()),
```
TrueDoctor, line 167: shouldn't we get a static string for the identifier? why do we need heap allocations here?

TrueDoctor (changes requested):

TrueDoctor:
@cubic-dev-ai

editor/src/node_graph_executor/runtime.rs
```diff
@@ -140,7 +140,14 @@ impl NodeRuntime {
...
 			monitor_nodes: Vec::new(),
 
-			substitutions: preprocessor::generate_node_substitutions(),
+			substitutions: {
+				// Exclude nodes with destructured outputs since their definition networks already contain the complete structure
+				let mut substitutions = preprocessor::generate_node_substitutions();
+				let node_metadata = graphene_std::registry::NODE_METADATA.lock().unwrap();
+				substitutions.retain(|id, _| node_metadata.get(id).is_none_or(|meta| meta.output_fields.is_empty()));
+				drop(node_metadata);
+				substitutions
```
TrueDoctor, line 149: why is this done here instea of in the generate_node_substitutions method?

Keavon: This is all out of my wheelhouse. If you're able to take it from here, I'd appreciate that. I basically just attempted to get it to work from a QA perspective without really understanding what's going on internally, since the alternative would have been to close the PR outright. Sorry I don't know the answer to the question, although it's probably worth fixing if it doesn't match your expectations.

TrueDoctor: I mean you made the changes you should be able to address comments regarding them. the generate substitutions functions is used in other places as well, and it looks to me like the changes should probably apply to all callers so it would likely make sense to integrate the changes in  the function (or even better instead of first generating the substitutions and then filtering them out, we should just not generate them in the first place) I won't have time to take over the pr myself

node-graph/nodes/raster/src/adjustments.rs
Ayush2k02: got it

PR #3808, opened 2026-02-22
Reimplement notice file generation for third-party licenses through Rust, now with CEF credits

### LINKED ISSUE #3730: Desktop: parse and include CEF credits
The CEF binary archive's root `CREDITS.html` contains its license credits, and we use that dependency on desktop. We should parse this HTML ourselves and merge it into our third-party licenses dialog. Doing so could allow us to take advantage of our existing deduplication strategy and cut down on some of the 15 MB of text in that HTML which doesn't do any deduplication. Maybe we should also attempt to compress this text within the bundle so it gets decompressed at runtime to reduce binary size.

timon-schelling:
Closes #3730

### DIFF
--- a/.nix/pkgs/tools/third-party-licenses.nix
+++ b/.nix/pkgs/tools/third-party-licenses.nix
@@ -21,9 +21,11 @@ let
     doCheck = false;
   };
 in
-deps.crane.lib.buildPackage common
-// {
-  inherit cargoVendorDir;
-  cargoArtifacts = deps.crane.lib.buildDepsOnly common;
-  meta.mainProgram = "third-party-licenses";
-}
+deps.crane.lib.buildPackage (
+  common
+  // {
+    inherit cargoVendorDir;
+    cargoArtifacts = deps.crane.lib.buildDepsOnly common;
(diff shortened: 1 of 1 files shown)


desktop/wrapper/src/messages.rs
```diff
@@ -126,6 +127,7 @@ pub enum DesktopWrapperMessage {
 		x: f64,
 		y: f64,
 	},
+	LoadThirdPartyLicenses(String),
```
Keavon:
```suggestion
	LoadThirdPartyLicenses {
		text: String,
	},
```

.github/workflows/build-win-bundle.yml
```diff
@@ -70,6 +70,7 @@ jobs:
           cargo binstall --no-confirm --force "wasm-bindgen-cli@$env:WASM_BINDGEN_CLI_VERSION"
 
       - name: Build Windows Bundle
+        shell: bash # cargo about refuses to run in powershell
```
Keavon:
```suggestion
        shell: bash # `cargo-about` refuses to run in powershell
```

Issue #3807, opened 2026-02-22
Path tool overlays no longer showing with Vello renderer active (regression)


Keavon:
#3722 has broken the ability to see any overlays with the Path tool active.

https://github.com/user-attachments/assets/5b79c757-b664-43c0-a849-f0ef2e8c86b0

PR #3810, opened 2026-02-23
Fix path tool overlays

TrueDoctor:
A change to the document metadata struct in #3727 added a new field which was then not part of the new metadata merge function introduced in render output caching pr and was missed during the rebase process. To avoid issues like this in the future, we are now destructuring the metadata struct so we get compiler warnings if the fields are modified again.
Fixes: #3807


Keavon:
I like this approach. We should also do the destructuring idea for cases like our Hash trait impls, where it's also easy to miss those when adding new fields.

PR #3829, opened 2026-02-25
Fix hardcoded input indices

### LINKED ISSUE #1490: Avoid using magic number indexes in the node system

Tyooughtul:
fix: replace hardcoded input indices with ::INDEX constants
replaces hardcoded numeric indices with predefined ::INDEX constants across various node input operations, improving code maintainability and reducing the risk of index mismatches.

changes include:
for native nodes, replaced hardcoded indices like 1, 2, 3 with module-level constants such as TranslationInput::INDEX, BlendModeInput::INDEX, etc.
in contexts like array access within functions, introduced local const declarations.

during testing, some failing cases were debugged with comments, but these were reverted to keep the code clean in this pr.

Fixes #1490


Keavon:
(you don't have to follow the AI review comments, they are advisory only)

editor/src/messages/tool/common_functionality/shapes/rectangle_shape.rs
```diff
@@ -9,6 +9,8 @@ use crate::messages::tool::tool_messages::tool_prelude::*;
 use glam::DAffine2;
 use graph_craft::document::NodeInput;
 use graph_craft::document::value::TaggedValue;
+use graphene_std::NodeInputDecleration;
```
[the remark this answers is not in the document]
Tyooughtul: this is a module introduced in March 2025

Keavon:
This is a surprisingly small number of changes. Does this comprehensively capture every instance, or is this just progress towards that goal?

Tyooughtul:
> This is a surprisingly small number of changes. Does this comprehensively capture every instance, or is this just progress towards that goal?

Thanks for the review! 😊 The left  hardcoded 0 and 1 indices are primary input for main data flow into any node and index 1 is the secondary input used by layer nodes for chain stacking. As for Artboard's location (2) and dimensions (3), I found that LocationInput::INDEX and DimensionsInput::INDEX simply don't exist anywhere in the codebase.The node macro apparently didn't generate these constants for the artboard node, so those two may not be converted without first adding that infrastructure.

node-graph/libraries/no-std-types/src/color/discrete_srgb.rs
```diff
@@ -162,6 +162,11 @@ mod tests {
 	#[test]
 	fn test_float_to_srgb_u8() {
 		for u in 0..=u8::MAX {
+			// let a = srgb_u8_to_float(u);
+			// let b = srgb_u8_to_float_ref(u);
+			// if a != b {
+			// panic!("Mismatch at u={}: {} != {}", u, a, b);
+			// }
```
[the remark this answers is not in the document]
Ayush2k02: +1

Ayush2k02 (changes requested):

Tyooughtul:
@Ayush2k02 @Keavon Thanks for review! There is a precision test error involved here, and I have now fixed it all at once. Please have a look when you have time. 😊

Keavon:
Thanks for the updates. To partially (but not fully) respond to your previous answer to my question, I will emphasize that the original issue describes a need for all nodes (both proto nodes and document nodes) to only use consts. For the document nodes, those will have to be written as a `const ...: usize = ...;` above its site. For proto nodes, that's `::INDEX` generated by the node macro. The really crucial part is that any PR must actually go through every single one of the search results produced by the regex I provided in the issue description. It's ok to submit a PR that makes some changes but doesn't comprehensively do every one. But that issue cannot be closed until all search results have been manually verified. Can you comment about that? Because I am surprised there are so few changes, unless this is just a partial PR.

Tyooughtul:
> Thanks for the updates. To partially (but not fully) respond to your previous answer to my question, I will emphasize that the original issue describes a need for all nodes (both proto nodes and document nodes) to only use consts. For the document nodes, those will have to be written as a `const ...: usize = ...;` above its site. For proto nodes, that's `::INDEX` generated by the node macro. The really crucial part is that any PR must actually go through every single one of the search results produced by the regex I provided in the issue description. It's ok to submit a PR that makes some changes but doesn't comprehensively do every one. But that issue cannot be closed until all search results have been manually verified. Can you comment about that? Because I am surprised there are so few changes, unless this is just a partial PR.

Sorry, I think I misunderstood you before. I thought it was only necessary to modify the hard-coding of the function signature😞. This time I made more changes, rg verification was used in different ways, the remaining ones are not parameter indexes. 

Also, I have a question:
In node_graph_message_handler.rs:2192, the code uses OutputConnector::node(layer.to_node(), 1) to get the output type for filtering the "New Node" popup. However, Layers (Merge nodes) only have one output at index 0. Accessing index 1 causes a silent failure (returns Error -> None), which disables filtering and shows all nodes. Is this by design (intentionally showing all nodes), or a bug that should use PRIMARY_OUTPUT_INDEX (0) to enable proper type filtering?

Thanks!

node-graph/libraries/no-std-types/src/color/discrete_srgb.rs
```diff
@@ -90,38 +90,70 @@ pub fn float_to_srgb_u8(mut f: f32) -> u8 {
...
-	0.45078585, 0.45641106, 0.46207705, 0.46778384, 0.47353154, 0.47932023, 0.48514998, 0.4910209,
-	0.49693304, 0.5028866, 0.50888145, 0.5149178, 0.5209957, 0.52711535, 0.5332766, 0.5394797,
-	0.5457247, 0.5520116, 0.5583406, 0.5647117, 0.57112503, 0.57758063, 0.5840786, 0.590619, 0.597202,
-	0.60382754, 0.61049575, 0.61720675, 0.62396055, 0.63075733, 0.637597, 0.6444799, 0.6514058,
-	0.65837497, 0.66538745, 0.67244333, 0.6795426, 0.68668544, 0.69387203, 0.70110214, 0.70837605,
-	0.7156938, 0.72305536, 0.730461, 0.7379107, 0.7454045, 0.75294244, 0.76052475, 0.7681514, 0.77582246,
-	0.78353804, 0.79129815, 0.79910296, 0.8069525, 0.8148468, 0.822786, 0.8307701, 0.83879924, 0.84687346,
-	0.8549928, 0.8631574, 0.87136734, 0.8796226, 0.8879232, 0.89626956, 0.90466136, 0.913099, 0.92158204,
-	0.93011117, 0.9386859, 0.9473069, 0.9559735, 0.9646866, 0.9734455, 0.98225087, 0.9911022, 1.,
+	f32::from_bits(0x00000000), f32::from_bits(0x399f22b4), f32::from_bits(0x3a1f22b4), f32::from_bits(0x3a6eb40f),
```
Keavon: Uh, what's with all the changes in this file?

Tyooughtul: > Uh, what's with all the changes in this file?

This resolves the error image I posted under #1490. It offers better accuracy. If constants were used, there might be some errors due to compilation or platform issues. Drawback might be that there is an issue with readability.🤔

Keavon: I see. I'd suggest you stash those changes and maybe later we open a PR for it. But this isn't affected by your PR and I don't think I have ever seen it fail locally or in CI, so it might be something unique to your machine. But it should be unrelated to this entire PR's focus.

Tyooughtul:
@Keavon please review, thanks!

jsjgdh:
@Tyooughtul, could you please resolve the merge conflicts?

Keavon:
Superseded by #4387

PR #3832, opened 2026-02-25
Replace `npm` build script with new `cargo run` tool

timon-schelling:
```
USAGE:
  cargo run [<command>] [<target>] [<profile>] [-- [args]...]

COMMON USAGE:
  cargo run            Run the web app
  cargo run desktop    Run the desktop app

OPTIONS:
<command>:
  [run]        Run the selected target (default)
  build        Build the selected target
  help         Show this message
<target>:
  [web]        Web app (default)
  desktop      Desktop app
  cli          Graphene CLI
<profile>:
  [debug]      Debug symbols enabled,
               optimizations disabled
               (default for run)
  [release]    Debug symbols disabled,
               optimizations enabled
               (default for build)
  profiling    Debug symbols enabled,
               optimizations enabled

MORE EXAMPLES:
  cargo run build desktop
  cargo run desktop profiling
  cargo run cli -- --help
```

### DIFF
--- a/tools/cargo-run/src/main.rs
+++ b/tools/cargo-run/src/main.rs
@@ -39,7 +39,7 @@ fn main() -> ExitCode {
 		Some(run) => run,
 		None => {
 			usage();
-			return ExitCode::FAILURE;
+			return ExitCode::SUCCESS;
 		}
 	};
 


tools/building/src/deps.rs
```diff
@@ -0,0 +1,185 @@
...
+	for msg in &failures {
+		eprintln!("  - {msg}");
+	}
+
+	if !failures.is_empty() {
+		eprintln!();
+		eprintln!("See: https://graphite.art/volunteer/guide/project-setup/");
+	}
+
+	if !installable.is_empty() && std::io::stdout().is_terminal() {
```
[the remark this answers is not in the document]
timon-schelling: that is the behavior we want

timon-schelling:
@cubic-dev-ai

tools/building/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+pub fn run_from(comand: &str, dir: Option<&str>) {
+	let workspace_dir = std::path::PathBuf::from(env!("CARGO_WORKSPACE_DIR"));
+	let dir = if let Some(dir) = dir { workspace_dir.join(dir) } else { workspace_dir };
+	let comand = comand.split_whitespace().collect::<Vec<_>>();
+	let mut cmd = process::Command::new(comand[0]);
+	if comand.len() > 1 {
+		cmd.args(&comand[1..]);
+	}
```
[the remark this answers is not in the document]
timon-schelling: that behavior is ok

timon-schelling:
@cubic-dev-ai

tools/cargo-run/src/requirements.rs
```diff
@@ -0,0 +1,193 @@
...
+use crate::*;
+
+#[derive(Default, Clone)]
+struct Dependency {
+	command: &'static str,
+	args: &'static [&'static str],
+	name: &'static str,
+	version: Option<&'static str>,
+	install: Option<&'static str>,
+	skip: Option<&'static dyn Fn(&Task) -> bool>,
```
TrueDoctor, line 13: Could we use a raw function pointer here?

TrueDoctor: Ah, just saw the closure, so I guess making the change would be slightly less ergonomic

tools/cargo-run/src/requirements.rs
```diff
@@ -0,0 +1,193 @@
+use std::io::IsTerminal;
```
TrueDoctor, line 1: I think the name of the file `deps.rs` is not as useful as it could be. I think something like `build_deps`/ `build_dependenices`/`build_tools` would be more descriptive

desktop/src/lib.rs
```diff
@@ -60,7 +58,7 @@ pub fn start() {
 		}
 		Err(_) => {
 			tracing::error!("Another instance is already running, Exiting.");
-			exit(1);
+			std::process::exit(1);
```
TrueDoctor, line 61: Is there a reason why this was not converted to a panic?

timon-schelling: Yes, this can happen in regular use and doesn't indicate a bug.
A panic would be the wrong feedback.

timon-schelling: I can however make start return result or ExitCode, but that should be separate from this PR.

.github/workflows/comment-!build-commands.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     if: >
       github.event.issue.pull_request &&
       github.event.comment.author_association == 'MEMBER' &&
-      (github.event.comment.body == '!build-dev' || github.event.comment.body == '!build-profiling' || github.event.comment.body == '!build')
+      (github.event.comment.body == '!build-debug' || github.event.comment.body == '!build-profiling' || github.event.comment.body == '!build')
```
Keavon: Remove !build-profiling

.github/workflows/comment-!build-commands.yml
```diff
@@ -1,6 +1,6 @@
 # USAGE:
 # After reviewing the code, core team members may comment on a PR with the exact text:
-# - `!build-dev` to build with debug symbols and optimizations disabled
+# - `!build-debug` to build with debug symbols and optimizations disabled
 # - `!build-profiling` to build with debug symbols and optimizations enabled
```
Keavon: And here, and just Ctrl+F for "profiling" so you don't miss any

timon-schelling: yeah already have that locally

tools/cargo-run/src/requirements.rs
timon-schelling: also includes run deps (cargo watch), `build*` is not fitting.

TrueDoctor: I'd argue these are also part of the build system. But if the tool is called `cargo-run` it is very easy to make the mistake of assuming this deals with rust crate dependencies since that is what cargo usually does

PR #3847, opened 2026-03-03
Add Pixel Preview render mode and fix Vello rendering regression with a tilted viewport

### LINKED ISSUE #320: Implement the pixelated view mode
While in the Pixels view mode, all content should be rendered at the scale of the document's pixels. If zoomed in further, this essentially means that resolution-agnosticism is disabled for the viewport preview and pixels show up. This is useful to see how an image will look when it is exported. This will probably need to wait for the full custom render engine, although it might be possible to achieve by rendering the SVG to an HTML5 canvas and then displaying that.

Keavon:
Closes #320. Fixes tilt behavior broken in #3722 with Vello rendering (the render modes *Normal*, *Outline*, and now this new mode *Pixel Preview*).

### DIFF
--- a/node-graph/libraries/wgpu-executor/src/lib.rs
+++ b/node-graph/libraries/wgpu-executor/src/lib.rs
@@ -154,17 +154,16 @@ impl WgpuExecutor {
 		Ok(())
 	}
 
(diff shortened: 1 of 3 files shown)


Keavon:
Superseded by #3867 and #320.

PR #3852, opened 2026-03-03
Fix node graph export sometimes not sticking to right edge of graph

Annonnymmousss:
Regression caused by commit 5b472a6

Before : 

https://github.com/user-attachments/assets/712a6b14-d28f-4de9-916e-4a8368043890

After :

https://github.com/user-attachments/assets/805addca-fc09-496c-b4f9-30d019e437b3


Keavon:
!build ([Run ID 22639031244](https://github.com/GraphiteEditor/Graphite/actions/runs/22639031244))

PR #3849, opened 2026-03-03
Fix regression where Vello doesn't render new document opened after closing all documents

Keavon:
Fixes #3817, a regression introduced in #3722.

PR #3859, opened 2026-03-05
Fix renderer failing to recover from a type error once it has been resolved

jsjgdh:
Regression of #3783.

https://github.com/user-attachments/assets/f77a1b69-02a2-4625-966d-5d65da7b0b69


Keavon:
Problem was fixed by #3849.

PR #3861, opened 2026-03-05
Images now paste under the cursor

advait87:
Before:

https://github.com/user-attachments/assets/6c547e89-fa33-4556-9391-1320333bdb35

After:

https://github.com/user-attachments/assets/52a9ea36-b828-4214-90a4-132c55f0afac

Tested with normal images as well as svgs.


advait87:
btw this is just for web

timon-schelling:
Implementation in editor code would be better and would also work for desktop.

PR #3864, opened 2026-03-06
Fix vector mesh fills rendering incorrectly by splitting into separate subpaths

### LINKED ISSUE #3378: Render vector mesh fills correctly as separate subpaths

rajesh-puripanda:
## Summary

This PR fixes an issue where vector mesh fills were rendered as a single merged path rather than multiple independent subpaths.

When a mesh contains multiple disconnected regions, the renderer previously treated them as one path, which resulted in incorrect fill behavior.

This change ensures each region is rendered as its own subpath before fill processing.

## Changes

- Extract mesh regions into independent subpaths
- Update renderer path construction logic
- Improve mesh iteration clarity

## Result

Vector meshes containing multiple disconnected regions now render correctly.

## Testing

Tested with:

- multiple independent shapes
- complex mesh geometry
- overlapping regions

All now render with correct fill separation.

## Related Issue

Fixes #3378

### DIFF
--- a/node-graph/libraries/rendering/src/renderer.rs
+++ b/node-graph/libraries/rendering/src/renderer.rs
@@ -825,8 +825,20 @@ impl Render for Table<Vector> {
 				(id, mask_type, vector_row)
 			});
 
-			if vector.is_branching() {
-				for mut face_path in vector.construct_faces().filter(|face| !(face.area() < 0.0)) {
+			if !vector.region_domain.is_empty() || vector.is_branching() {
+				let regions = vector.region_manipulator_groups();
+				let region_count = regions.count();
+				if region_count > 0 {
+					log::trace!("Rendering {} mesh regions", region_count);
+				}
+
+				let faces: Box<dyn Iterator<Item = kurbo::BezPath>> = if region_count > 0 {
+					Box::new(vector.region_manipulator_groups().map(|(_, face)| graphic_types::vector_types::vector::misc::bezpath_from_manipulator_groups(&face, true)))
+				} else {
+					Box::new(vector.construct_faces().filter(|face| !(face.area() < 0.0)))
+				};
+
+				for mut face_path in faces {
 					face_path.apply_affine(Affine::new(applied_stroke_transform.to_cols_array()));
 
 					let face_d = face_path.to_svg();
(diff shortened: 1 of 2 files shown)


rajesh-puripanda:
While testing this change I noticed the mesh path extraction logic could also potentially affect stroke generation.

Would you prefer keeping the separation logic limited to fill generation, or should the same behavior apply to stroke paths as well?

Keavon:
It appears that your PR is written by AI, not yourself, and is not disclosed as such, from what I can tell.

rajesh-puripanda:
> It appears that your PR is written by AI, not yourself, and is not disclosed as such, from what I can tell.


There is no AI disclosure because no AI was used in writing this PR.
The changes, commit messages, and reasoning were authored by me. If you have feedback on the implementation or the approach, I’d welcome that discussion.

Keavon:
Could you please clarify if you used AI to write the PR description? It looks precisely like AI output (and we receive tons that are extremely similar to that). Our AI policy extends to the PR description, and that is a good litmus test for the content of the PR as well. Sorry to accuse, it's just that we are inundated and a quick glance is often enough to be right 99% of the time.

rajesh-puripanda:
Thanks for the clarification. I understand why you'd ask, if you're receiving a lot of AI-generated submissions, I can see how patterns in PR descriptions might raise suspicion.

To clarify, the PR description and the changes themselves were written by me. I didn’t use AI to generate the description or the implementation. I tried to structure it clearly to explain the issue, reasoning, and the fix, which may be why it reads similar to the format you're seeing frequently.

If anything in the description or the implementation could be improved or expanded for clarity, I'm happy to revise it. I’d also appreciate any feedback on the technical approach in the PR itself.

Thanks for taking the time to review it.

Keavon:
Thanks for your understanding and elaboration. Sorry again to accuse, it has been becoming increasingly hard to tell for certain. I'll reopen the PR and hopefully get it reviewed soon, although we are still somewhat behind on our review backlog so no promises about it being as immediate as we'd ideally like it to be. Thanks for the contribution, and feel free to take on other topics while awaiting review in the mean time!

TrueDoctor:
!build ([Run ID 22946817001](https://github.com/GraphiteEditor/Graphite/actions/runs/22946817001))

Keavon:
#3884 supersedes this PR.

PR #3867, opened 2026-03-08
Fix texture cache region size and broken rendering when tilting the viewport

TrueDoctor:
Fixes regression from #3722.

### DIFF
--- a/node-graph/nodes/gstd/src/render_cache.rs
+++ b/node-graph/nodes/gstd/src/render_cache.rs
@@ -291,6 +291,7 @@ fn split_oversized_region(region: RenderRegion, max_region_area: u32) -> Vec<Ren
 	}
 
 	if group1.is_empty() || group2.is_empty() {
+		log::error!("Failed to split oversized region.");
 		return vec![region];
 	}
 


node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -398,15 +387,26 @@ pub async fn render_output_cache<'a: 'n>(
 		ctx.try_pointer_position(),
 	);
 
+	let physical_tile_size = (TILE_SIZE as f64 * device_scale).round() as u32;
```
timon-schelling: I assume this means vello tiles are assumed to be the incorrect size when device scale is not 1? vello always renders in tiles of 256 in device pixels right? The preference `Render Tile Resolution` should be in device pixels right? scale seem to have influence on that currently?

TrueDoctor: This tile size is in no way coupled to vello internals but thevtile size changes based on the device scale. I ccould try to chane this, but that would require more changes to the tile composition logic

TrueDoctor: Also the name of the preference is just wrong since it has nothing to do with the tile size. It was choosen by keavon to make it less confusing for users

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -275,7 +267,7 @@ fn split_oversized_region(region: RenderRegion, max_region_area: u32) -> Vec<Ren
 	if split_horizontally {
 		let min_x = region.tiles.iter().map(|t| t.x).min().unwrap();
 		let max_x = region.tiles.iter().map(|t| t.x).max().unwrap();
-		let mid_x = (min_x + max_x) / 2;
+		let mid_x = min_x + (max_x - min_x) / 2;
```
TrueDoctor, line 270: This should be exactly the same mathematically

timon-schelling: without this I get regions that are not split correctly (both tiles end up in the same group)
image

timon-schelling: test with
```diff
diff --git a/node-graph/nodes/gstd/src/render_cache.rs b/node-graph/nodes/gstd/src/render_cache.rs
index 67ead5f13..b23b1fce4 100644
--- a/node-graph/nodes/gstd/src/render_cache.rs
+++ b/node-graph/nodes/gstd/src/render_cache.rs
@@ -267,7 +267,7 @@ fn split_oversized_region(region: RenderRegion, max_region_area: u32) -> Vec<Ren
 	if split_horizontally {
 		let min_x = region.tiles.iter().map(|t| t.x).min().unwrap();
 		let max_x = region.tiles.iter().map(|t| t.x).max().unwrap();
-		let mid_x = min_x + (max_x - min_x) / 2;
+		let mid_x = (min_x + max_x) / 2;
 
 		for &tile in &region.tiles {
 			if tile.x <= mid_x {
@@ -279,7 +279,7 @@ fn split_oversized_region(region: RenderRegion, max_region_area: u32) -> Vec<Ren
 	} else {
 		let min_y = region.tiles.iter().map(|t| t.y).min().unwrap();
 		let max_y = region.tiles.iter().map(|t| t.y).max().unwrap();
-		let mid_y = min_y + (max_y - min_y) / 2;
+		let mid_y = (min_y + max_y) / 2;
 
 		for &tile in &region.tiles {
 			if tile.y <= mid_y {
@@ -291,6 +291,7 @@ fn split_oversized_region(region: RenderRegion, max_region_area: u32) -> Vec<Ren
 	}
 
 	if group1.is_empty() || group2.is_empty() {
+		log::error!("Failed to split region even though it exceeds max area.");
 		return vec![region];
 	}
 
@@ -534,8 +535,8 @@ fn composite_cached_regions(
 					aspect: wgpu::TextureAspect::All,
 				},
 				wgpu::Extent3d {
-					width,
-					height,
+					width: width - 1,
+					height: height - 1,
 					depth_or_array_layers: 1,
 				},
 			);

```

timon-schelling: maybe we should actually include that log line

TrueDoctor: I see, this is caused by the different rounding directions then

PR #3868, opened 2026-03-08
Fix issue with license generation using new `cargo run` tool when using PowerShell

krVatsal:
Based on the discussion in discord https://discord.com/channels/731730685944922173/731738914812854303/1479960786871779459

On Windows, if cargo-about fails (often due to PowerShell detection in process ancestry), fall back to using a temporary file to work around the issue.


krVatsal:
Do we need any changes here @timon-schelling ?

timon-schelling:
Thanks for the contribution.

1. The tmp file should not be placed in the workspace dir but instead in the standard tmp dir on windows.
2. Move windows specific code to a function.
3. See if you can find a simpler solution than introducing the guard struct.

PR #2670, opened 2025-05-25
Port to linesweeper

jneem:
@Keavon

PR #3882, opened 2026-03-11
Remove the path-bool library

Keavon:
Follow-up to the PR #2670 which replaced path-bool usage with the Linesweeper library.


Keavon:
We updated the commit in master to one authored by @TrueDoctor for Git statistical purposes in a force push.

PR #3901, opened 2026-03-14
Make path editable for grouped graphic inputs

krVatsal:
Continues work from #3735 since @jsjgdh cannot continue.

Refactors path editability checks into a shared helper and applies it consistently in both menu eligibility and Path Tool selection handling. This enables grouped layers `[Table<Graphic>]` to be treated as path editable while preserving existing behavior for vector inputs.


krVatsal:
Just a doubt @Keavon @jsjgdh, the current fix makes it behave as you told but a slight diversion is here. Asking just to confirm if the current behaviour looks good or should i make exactly as you told. 
The current behavior is that when you select a grouped layer using path tool then the button can be seen active(as you told) but the vectors aren't getting auto-selected, so when we click on any one of the shape(without clicking on `make path editable`) then we will be only modifying that vector and not entire group. Further when we click on `make path editable` then we get both the path and flatten path nodes as expected.

https://github.com/user-attachments/assets/037b7bcf-5b73-49f2-82da-81b88547b887

jsjgdh:
> Just a doubt @Keavon @jsjgdh, the current fix makes it behave as you told but a slight diversion is here. Asking just to confirm if the current behaviour looks good or should i make exactly as you told. The current behavior is that when you select a grouped layer using path tool then the button can be seen active(as you told) but the vectors aren't getting auto-selected, so when we click on any one of the shape(without clicking on `make path editable`) then we will be only modifying that vector and not entire group. Further when we click on `make path editable` then we get both the path and flatten path nodes as expected.
> 
>  Recording.2026-03-17.100620.1.mp4


@Keavon will be the only one who will be able to answer this question .

PR #3906, opened 2026-03-17
Fix undo not fully reverting image and SVG imports

### LINKED ISSUE #3904: After importing an image and then doing cmd + z , the layer isn't deleted.
https://github.com/user-attachments/assets/b273a128-b37a-46dc-8d9a-8932a3d3c142

ziad-ashraf7:
Fixes: #3904

### Before:
(See the video in the mentioned issue)

### After:

https://github.com/user-attachments/assets/13b4a206-42f9-490b-ac96-355227ad95f3

### Files Changed:
`editor/src/messages/portfolio/document/document_message_handler.rs`

### Description:

When importing an image or SVG, the SetDisplayName call was creating a separate history step because 
skip_adding_history_step was set to false. 

This meant a single undo would only revert the rename, requiring a second undo to actually delete the imported layer.

Setting skip_adding_history_step to true keeps the rename as part of the same transaction as the layer creation, so one undo fully reverts the import.


Ayush2k02:
yeah LGTM, thanks @0HyperCube for providing details and resolution for this

Ayush2k02 (approved):

ziad-ashraf7:
Thank you Mr. @Ayush2k02 for approving the PR, so happy for that !

Also Thank you Mr. @0HyperCube for putting me on the right way to solve the issue and have a good exposure on the solution.

Is there anything else that i should do now to make the PR merged ? so I can now look for more issues to resolve.

PR #3910, opened 2026-03-18
fix Firefox color rendering by respecting surface texture format 

### LINKED ISSUE #3880: Vello rendering incorrect colors in Firefox on Mac
The colour rendering is off in dev version but works perfectly fine in editor version in firefox. It works perfectly okay bothways in chrome.
OS - MacOS

Reproduction clip:

https://github.com/user-attachments/assets/a9421fae-06a3-4878-ac9d-a1e218125ec4

Ayush2k02:
Replace direct copy_texture_to_texture() with a shader-based blitter that handles format conversion between Rgba8Unorm (Chrome) and Bgra8Unorm (Firefox)

https://github.com/user-attachments/assets/b8e1b989-ee4f-4e30-8e80-2971a6991cc4

Fixes #3880

### DIFF
--- a/node-graph/libraries/canvas-utils/src/wasm.rs
+++ b/node-graph/libraries/canvas-utils/src/wasm.rs
@@ -52,13 +52,20 @@ impl Canvas for CanvasHandle {
 }
 
 #[cfg(feature = "wgpu")]
-pub struct CanvasSurfaceHandle(CanvasHandle, Option<Arc<wgpu::Surface<'static>>>);
+struct SurfaceState {
+	surface: Arc<wgpu::Surface<'static>>,
+	format: wgpu::TextureFormat,
+	blitter: wgpu_executor::cached_blitter::CachedBlitter,
+}
+
+#[cfg(feature = "wgpu")]
+pub struct CanvasSurfaceHandle(CanvasHandle, Option<SurfaceState>);
 #[cfg(feature = "wgpu")]
 impl CanvasSurfaceHandle {
 	pub fn new() -> Self {
 		Self(CanvasHandle::new(), None)
 	}
-	fn surface(&mut self, executor: &WgpuExecutor) -> &wgpu::Surface<'_> {
+	fn state(&mut self, executor: &WgpuExecutor) -> &SurfaceState {
 		if self.1.is_none() {
 			let canvas = self.0.get().canvas.clone();
 			let surface = executor
@@ -66,7 +73,17 @@ impl CanvasSurfaceHandle {
 				.instance
 				.create_surface(wgpu::SurfaceTarget::Canvas(canvas))
 				.expect("Failed to create surface from canvas");
-			self.1 = Some(Arc::new(surface));
+
+			// Use the surface's preferred format (Firefox WebGL prefers Bgra8Unorm, Chrome prefers Rgba8Unorm)
+			let surface_caps = surface.get_capabilities(&executor.context.adapter);
+			let surface_format = surface_caps.formats.iter().copied().find(|f| f.is_srgb()).unwrap_or(surface_caps.formats[0]);
+			let blitter = wgpu_executor::cached_blitter::CachedBlitter::new(&executor.context.device, surface_format);
+
+			self.1 = Some(SurfaceState {
+				surface: Arc::new(surface),
+				format: surface_format,
+				blitter,
+			});
 		}
 		self.1.as_ref().unwrap()
 	}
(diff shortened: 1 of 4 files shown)


TrueDoctor:
This change will degrade performance. (I recently changed it from a blitter to the direct texture copy) Can you give more details for when this workaround is needed?

Ayush2k02:
Oh, this is only needed for Firefox based browsers , while importing an image in Firefox, the colours are inverted
@TrueDoctor

Ayush2k02:
can you tell me which exact test would you use here to benchmark the performance here, that way I might be able to come up with a better solution

TrueDoctor:
> can you tell me which exact test would you use here to benchmark the performance here, that way I might be able to come up with a better solution

I previously used the wgpu inspector extension to debug this. The issue with using a blitter is, that it creates bind groups which we need to wait for javascript to garbage collect (and it is also just less efficient). So it would be could, if we could avoid it.
So this is only an issue on Firefox + MacOS? Because for me, it works fine in Firefox + Linux.
I'm basically wondering if we could check if the formats are different and only in that case use the blitter?

Ayush2k02:
@TrueDoctor , Done, have updated the branch to check for format and then only use blitter if the formats differ

Ayush2k02:
Here is the updated recording for both chrome and firefox after the recent commit 



https://github.com/user-attachments/assets/77152528-53c3-46d1-930b-21f6f1a1d58b

TrueDoctor:
!build ([Run ID 23978166532](https://github.com/GraphiteEditor/Graphite/actions/runs/23978166532))

TrueDoctor:
image
image

It still seems to allocate a bunch of bind groups for me which does not happen in dev.graphite.rs

TrueDoctor:
I unfortunately don't have much time at the moment but if you want to test something for you (e.g. running your branch with debug statements etc.) I can do that

Ayush2k02:
Yeah, that would be awesome . Honestly I can do this and share with you as
well, like by the end of today , totally upto you, feel like it would be
better if I do the testing myself though

Ayush2k02:
Actually, nvm . M taking a look at this right now :) . Also, thanks for sharing the webGpu inspector console, didn't know you could check that way

Ayush2k02:
Hey, sorry for the delayed response but can you tell me how are you importing that image or from where I can get that. The image which I use for my testing is working correctly, like the bind groups used are consistent across chrome and firefox

Also , one thing I noticed was that the bind groups would reset to zero after a certain time on firefox ( last screenshot ) , most likely due to some kind of garbage collection happening , but not sure

dev.graphite.art ( firefox )

image

dev.graphite.art ( chrome )

image

firefox on this branch

Screenshot 2026-04-09 at 7 50 12 AM



chrome on this branch

Screenshot 2026-04-09 at 7 49 54 AM

firefox reseting bindgroups

image

Ayush2k02:
Update, found the image in the codebase itself , testing with this image now

TrueDoctor:
I have not used an image for testing. I just opend the painted dreams demo artwork and zoom in and out, on dev, we don't create more bind groups, and in your branch, it rapidly allocates bind groups when zooming

Ayush2k02:
yeah, noted. Taking a look

Ayush2k02:
This is actually happening on main as well @TrueDoctor  , Like I don't think the current latest is pushed to prod . The bind groups increase decrease on zooming in and out. Taking a look at which commit might have caused it and what are the fixes

TrueDoctor:
@Ayush2k02 I'm really sorry, I think, I gave you the wrong reproduction, instead of zooming you have to pan around for the difference to show up. When zooming an unrelated bug causes the cache to be invalidated creating bind groups. But the one upside of this is that we can fix that issue as well now that we know of it

TrueDoctor:
Here is a video of the difference

https://github.com/user-attachments/assets/77c89d6b-e642-4b95-9bb8-c4bc5658cf74

TrueDoctor:
The last commit is weird, from what I can tell it could not possibly fix the bindgroup creation issue I reported since that happens on web. And it looks like your changes might introduce new bugs on desktop

Ayush2k02:
Hey, converted it to draft for now. I'll let you know once its ready . Was getting confused with all the merge conflicts

Ayush2k02:
https://github.com/user-attachments/assets/60105920-8ee1-4a29-9b3e-0cb8d48685f8

Seems to be working fine now, but yeah Don't review the code yet, I'll let you know when its ready

Ayush2k02:
@TrueDoctor  , have verified the code here. Its not producing any new bind groups on panning around and the colours are also correct for firefox. Have given the code changes a review as well. Let me know if the approach on this branch seems good to you

PR #3916, opened 2026-03-19
Consolidate CI web builds into the build workflow and restore deployments to GitHub environments

Keavon:


.github/workflows/build.yml
```diff
@@ -376,7 +440,7 @@ jobs:
           path: target/artifacts
 
   mac:
-    if: github.event_name == 'push' || inputs.mac
+    if: (github.event_name == 'push' && github.ref != 'refs/tags/latest-stable') || inputs.mac
```
timon-schelling, line 446: why, I would expect it to also build on pushing the tag.

Keavon: That build already occurs when we originally push the commit. We haven't yet set up GitHub Releases deployments, but this clause will be removed when we do that.

PR #3921, opened 2026-03-20
Fix artboards not exporting with transparency using Vello

timon-schelling:


timon-schelling:
!build desktop ([Run ID 23353978912](https://github.com/GraphiteEditor/Graphite/actions/runs/23353978912))

PR #3928, opened 2026-03-21
Fix Mac text selection keyboard shortcuts in CEF text fields

### LINKED ISSUE #3507: Desktop release candidate refinements

Mayank7Bisnewar:
Fixes #3507

This fixes the issue where standard macOS keyboard selection shortcuts (like `Cmd+A` and `Shift+Option+Up`) weren't working properly inside CEF text fields on desktop.

While looking into this, I found that the `unmodified_character = 0` workaround in [input.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/cef/input.rs:0:0-0:0) was preventing CEF from properly registering those text shortcuts on Mac. However, removing it caused an infinite loop where CEF would push unhandled synthetic `NSEvents` back to [app.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/app.rs:0:0-0:0) which routed them straight back into Winit (causing duplicate menu events and hanging).

To fix the whole flow:
- I removed the `unmodified_character = 0` mitigation in [input.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/cef/input.rs:0:0-0:0) so Cmd+A registers.
- I added a `IS_CEF_WORK` thread-local flag in [app.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/app.rs:0:0-0:0) and [singlethreaded.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/cef/context/singlethreaded.rs:0:0-0:0) to detect and safely drop those synthetic CEF events before they loop back into Winit.
- I mapped the mac unichars (like `0xF700`) for the navigational named keys in [keymap.rs](cci:7://file:///Users/mayank/projects/Graphite/desktop/src/cef/input/keymap.rs:0:0-0:0) so Chromium's native `interpretKeyEvents:` can correctly process complex multi-modifier shortcuts like `Shift+Option+Up`.

Tested locally and everything works just like it does on the web version now!

*(Note: I tracked down the CEF/Winit infinite loop architecture and developed this fix with the help of an AI coding assistant.)*


timon-schelling:
Clearly not tested (well), doesn't compile:

```
error[E0603]: module `mac` is private
  --> desktop/src/cef/context/singlethreaded.rs:25:18
   |
25 |         crate::window::mac::app::IS_CEF_WORK.with(|c| c.set(false));
   |                        ^^^       ----------- constant `IS_CEF_WORK` is not publicly re-exported
```
 
And this will lead to the menu bar lighting up when pressing `g` for example.
You proposed one of the obvious fixes that break other things and are therefore not acceptable.

Also linked the wrong issue.

PR #3929, opened 2026-03-21
Fix/text layer shake

### LINKED ISSUE #3896: Shaking a Text layer breaks its type signature
https://github.com/user-attachments/assets/c4b735df-933f-4af4-8101-9372dcc6b9a3

Mayank7Bisnewar:
Fixes #3896 

This fixes a bug where shaking a Text layer in the node graph would break its type signature and put the node into an error state.

While looking into this, I noticed that `ShakeNode` in [node_graph_message_handler.rs](cci:7://file:///Users/mayank/projects/Graphite/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs:0:0-0:0) blindly disconnects *all* inputs and outputs. Since the [Text](cci:2://file:///Users/mayank/projects/Graphite/node-graph/nodes/text/src/text_context.rs:18:0-23:1) node relies on an internal unexposed connection (`WasmEditorApi` provided via the `editor-api` scope) for font resolution, disconnecting it caused the entire node graph pipeline to break.

To fix it:
- I updated the input loop in `ShakeNode` to filter using `input.is_exposed()`. Unexposed internal inputs are now safely skipped and remain connected.
- I applied the same check for downstream inputs during the output disconnection loop, ensuring we only detach user-facing wires.

Tested locally, and Text layers now disconnect securely when shaken without breaking their parameters!

*(Note: I tracked down the node graph disconnection flow and developed this fix with the help of an AI coding assistant.)*

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -1520,6 +1520,11 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
 					// Handle inputs of selected node
 					for input_index in 0..network_interface.number_of_inputs(selected_node, selection_network_path) {
 						let input_connector = InputConnector::node(*selected_node, input_index);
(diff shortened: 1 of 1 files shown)


Keavon:
AI slop

PR #3941, opened 2026-03-22
Remove surface and window from ApplicationIo

timon-schelling:
Preparation for asset database integration.


TrueDoctor:
The perf benchmarks look very suspicious

timon-schelling:
I think this is nothing related to this PR but the result of a failure during the last run.

timon-schelling:
@TrueDoctor I think I addressed all of your comments, merging :)

Issue #3956, opened 2026-03-26
Smooth brush caps for the freehand and line tools


shioprince:
I'd like smooth caps for tools like the freehand and line tools. For reference, here is an example of smooth caps (notice the ends of the line):

<img width="855" height="117" alt="Image" src="https://github.com/user-attachments/assets/9aef0b1b-8d5f-414d-842a-bb247b35ddff" />

Veercodeprog:
please assign me?

Keavon:
@Veercodeprog you can add this to the control bar for both tools. We don't assign issues, please read our contributor guide which explains that.

Veercodeprog:
> @Veercodeprog you can add this to the control bar for both tools. We don't assign issues, please read our contributor guide which explains that.

sure ,i am on it.

Issue #3939, opened 2026-03-22
`cargo run` fails on WebAssembly target due to `CARGO_BUILD_TARGET` leaking into `rustc_codegen_spirv` build


ghchinoy:
### Summary
When running the standard `cargo run` development workflow, the build fails while compiling the web frontend (`wasm-pack build ./wrapper --dev --target=web`). 

The failure occurs in the `raster-nodes-shaders` crate's `build.rs` script. The build script attempts to compile the `rustc_codegen_spirv` compiler backend, but the parent `CARGO_BUILD_TARGET=wasm32-unknown-unknown` environment variable is inherited by the inner `cargo` build. Because WebAssembly does not support dynamic libraries (`dylib`), the inner build fails.

### Error Output
```text
[INFO]: 🎯  Checking for the Wasm target...
[INFO]: 🌀  Compiling to Wasm...
   Compiling raster-nodes-shaders v0.1.0 (../Graphite/node-graph/nodes/raster/shaders)
error: failed to run custom build command for `raster-nodes-shaders v0.1.0 (../Graphite/node-graph/nodes/raster/shaders)`

Caused by:
  ...
  error: cannot produce dylib for `rustc_codegen_spirv v0.9.0 (https://github.com/Firestar99/rust-gpu-new?rev=c12f2161...#c12f2161)` as the target `wasm32-unknown-unknown` does not support these crate types
```

### Steps to Reproduce
1. Clone the repository.
2. Ensure Node.js and Rust prerequisites are installed.
3. Run `cargo run`.
4. Observe the build crash when `wasm-pack` reaches `raster-nodes-shaders`.

### Proposed Solution
The inner build initiated by `cargo_gpu::Install` inside `node-graph/nodes/raster/shaders/build.rs` needs to compile for the host OS, not the WebAssembly target.

Clearing the `CARGO_BUILD_TARGET` environment variable inside `build.rs` before it kicks off the `cargo_gpu` build resolves the issue and allows the compilation to succeed:

```rust
// node-graph/nodes/raster/shaders/build.rs

// Inside main():
unsafe { std::env::remove_var("CARGO_BUILD_TARGET"); }

// ... followed by the existing cargo_gpu::Install::from_shader_crate(...) logic
```

### Environment
* **OS:** macOS 26.3.1 (Build 25D2128)
* **Rust Version:** rustc 1.92.0 (ded5c06cf 2025-12-08)
* **Node.js:** v22.21.1

[...]

PR #3962, opened 2026-03-28
Make round stroke caps for tool strokes

Veercodeprog:
#3956
Screenshot 2026-03-29 at 8 21 31 PM

- New strokes created via tools that use `ToolColorOptions::apply_stroke` (freehand, spline, pen, line/shape tools, etc.) now default to **round** line caps instead of butt, so open paths get smooth end caps without editing the Stroke node.
- Viewport **tile cache** keys now include the document **network hash**, so changing stroke properties (e.g. cap) or any other node input invalidates cached tiles and the canvas updates correctly instead of showing stale pixels.

## Testing
- Drew short open paths with freehand/line; confirmed rounded ends by default.
- Changed stroke cap in properties and confirmed the viewport updates (no stale butt-cap tiles).
- `cargo fmt`, `cargo clippy`, `cargo test --all-features` (as per contributor guide).


node-graph/nodes/raster/shaders/build.rs
```diff
@@ -5,6 +5,14 @@ use std::path::PathBuf;
 pub fn main() -> Result<(), Box<dyn std::error::Error>> {
 	env_logger::builder().filter_level(log::LevelFilter::Debug).init();
 
+	// This crate is often built for wasm32 via wasm-pack, which sets `CARGO_BUILD_TARGET`.
+	// cargo-gpu then spawns nested `cargo` to build `rustc_codegen_spirv` for the host; if that
+	// variable is inherited, the nested build targets wasm32 and fails with "cannot produce dylib".
+	// SAFETY: Build script is single-threaded.
+	unsafe {
```
Veercodeprog: this is n't necessary on your system,not sure how exactly you run the app in your system

Veercodeprog: @Keavon

Veercodeprog:
@Keavon i am GSOC applicant , could you please guide a bit for the application?

Keavon:
@Veercodeprog GitHub issues would not be the appropriate venue for that, see our documented instructions for advice on your next steps.

Veercodeprog:
@Keavon please review this, i could use this pr in my PROPOSAL

node-graph/nodes/raster/shaders/build.rs
Keavon: There is no circumstance in which you should be committing changes to a build script when developing an editor feature.

Veercodeprog: @Keavon 
this is related to the issue I was referring to in #3939 
when running the project locally, the CARGO_BUILD_TARGET environment variable seems to leak from somewhere, which causes cargo run to fail while compiling the web frontend. The failure happens in the raster-nodes-shaders crate’s build.rs

as a temporary workaround, clearing the CARGO_BUILD_TARGET environment variable inside build.rs before it starts cargo_gpu fixes the issue and allows cargo run to work on our system.

This particular change can be skipped when merging the PR:
unsafe {
    std::env::remove_var("CARGO_BUILD_TARGET");
}

infact we can skip the changes made in [node-graph/nodes/raster/shaders/build.rs] for this pr

Veercodeprog: updating this change now

PR #3974, opened 2026-03-29
Finalize and unify the design of the 'Morph' and 'Blend' nodes

Keavon:
image

## Morph node

| Before | After |
| - | - |
| image | image |

## Blend node

Renamed from Blend Shapes. The existing Blend node has also been renamed to Mix.

| Before | After |
| - | - |
| image | image |
| image | image |

## New features

The new *Path* parameter is added to the Morph node so it acts similar to the old Blend Shapes node implementation, and for consistency with the Position on Path node. It also adds a *Reverse* parameter to Morph for Position on Path consistency. Blend and Morph also gain a *Distribution* parameter with the choices: "Objects", "Distances", "Angles", "Sizes", and "Slants" for controlling what parameter of change is kept at equal speed.

This also adds the *Layer* > *Blend* and *Layer* > *Morph* menu entries:

image

### DIFF
--- a/node-graph/nodes/vector/src/vector_nodes.rs
+++ b/node-graph/nodes/vector/src/vector_nodes.rs
@@ -2226,68 +2226,72 @@ async fn morph<I: IntoGraphicTable + 'n + Send + Clone>(
 	let subpath_anchors = anchor_count(control_bezpath);
 	let max_content_index = content.len().saturating_sub(1);
 
-	// Compute per-segment arc lengths for spatial positioning along the control path
-	let segment_lengths: Vec<f64> = control_bezpath.segments().map(|seg| seg.perimeter(DEFAULT_ACCURACY)).collect();
-
-	// Compute segment weights based on the user's chosen spacing metric
-	let segment_weights: Vec<f64> = match distribution {
-		InterpolationDistribution::Objects => vec![1.; segment_count],
-		InterpolationDistribution::Distances => segment_lengths.clone(),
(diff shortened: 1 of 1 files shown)


node-graph/libraries/vector-types/src/vector/vector_attributes.rs
```diff
@@ -509,6 +530,12 @@ impl RegionDomain {
 		self.fill.clear();
 	}
 
+	pub fn reserve(&mut self, additional: usize) {
```
timon-schelling, line 623: `#[inline(always)]`

node-graph/libraries/vector-types/src/vector/vector_attributes.rs
```diff
@@ -234,6 +241,14 @@ impl SegmentDomain {
 		self.stroke.clear();
 	}
 
+	pub fn reserve(&mut self, additional: usize) {
```
timon-schelling, line 246: `#[inline(always)]`

node-graph/libraries/vector-types/src/vector/vector_attributes.rs
```diff
@@ -104,6 +104,11 @@ impl PointDomain {
 		self.position.clear();
 	}
 
+	pub fn reserve(&mut self, additional: usize) {
```
timon-schelling, line 108: `#[inline(always)]`

Keavon: Why? This isn't called in a hot loop.

PR #4008, opened 2026-04-05
Fix Layer Transformation

### LINKED ISSUE #1529: Dragging layers sharing a transform node applies multiple translations

jsjgdh:
closes #1529


jsjgdh:
@cubic-dev

PR #4010, opened 2026-04-05
Add a family of 14 new string processing nodes

Keavon:
## Category - Text

- **Format Number**: Formats a number as a string with control over decimal places, separators, fixed decimals, and thousands grouping.
- **String to Number**: Parses a string into a number, with a configurable fallback value.
- **String Truncate**: Clips a string to a max character length, optionally appending a suffix (e.g. `…`) on truncation.
- **String Trim**: Removes leading and/or trailing whitespace, independently toggleable for start and end.
- **String Escape**: Converts between literal escape sequences (`\n`, `\t`, etc.) and their actual control characters, in either direction.
- **String Reverse**: Reverses the character sequence of a string.
- **String Repeat**: Repeats a string N times, with an optional separator and a "Separator Escaping" option.
- **String Pad**: Pads a string to a target length with a fill substring, with start/end targeting and an "Up To" parameter.
- **String Contains**: Checks if a string contains a substring, optionally anchored to the start and/or end.
- **String Find Index**: Returns the index of the first (or last) occurrence of a substring, or -1 if not found.
- **String Occurrences**: Counts occurrences of a substring within a string.
- **String Join**: Joins a list of strings with a separator; the inverse of String Split.

## Category - Context

- **Map String**: Iterates over a list of strings, evaluating a mapped node operation for each one (loop construct).
- **Read String**: Reads the current string from inside a Map String loop.


Keavon:
@cubic-dev-ai

PR #4019, opened 2026-04-08
Fix Very long text does not forcibly wrap at its max width

Annonnymmousss:
closes #4009 

https://github.com/user-attachments/assets/b6beac10-1396-49ac-b045-345dba66d12e


Keavon:
!build ([Run ID 24163097135](https://github.com/GraphiteEditor/Graphite/actions/runs/24163097135))

PR #4025, opened 2026-04-12
FIX: Made text alignment work when max width is disabled

Chetansahney:
Fixes #4011

This PR fixes the issue where text alignment settings (Center/Right/Justify) were ignored unless a Max Width was set. I've updated the logic so that the text block now calculates its own width based on the longest line, allowing alignment to work correctly relative to the text's own bounding box.

Key Changes

Enabled alignment support for text nodes without a defined Max Width.
Updated the layout engine to use the dynamic width of the text rows as the alignment basis.


Annonnymmousss:
@Chetansahney It would be helpful if you could attach the before after video clips. Thankyou.

PR #4039, opened 2026-04-23
Add G/S shortcut handling to Artboard tool

### LINKED ISSUE #4026: The Artboard tool needs to support G/S (grab and scale) transformation
When the Artboard tool is selected, G/S (like the G/R/S of the Select and Path tools) needs to allow the user to grab and scale the selected artboard. Code should be reused in the implementation.

devviktoria:
Adds support G/S (grab and scale) keyboard shortcut handling to the Artboard tool.

Closes https://github.com/GraphiteEditor/Graphite/issues/4026


devviktoria:
So I checked this: the grab with the G key works but unfortunately the scaling has some issues. The artboard scaling functionality can only work with the absolute position of the mouse, but cannot handle relative change.
So in this version the scaling works, but it follows the mouse position so at the start of the scaling there is a jump and then follows the mouse position and it only scales as if the bottom right corner was dragged.
Here is the video of the behavior.

https://github.com/user-attachments/assets/e14c9c1c-3819-45e7-a627-af0dd0732d37

And also there are no tests created for this functionality.

PR #4044, opened 2026-04-24
New nodes: 'Format JSON', 'Query JSON', and 'Query JSON All', replacing the 'JSON Get' node

Keavon:
## Category - Text: JSON

- **Format JSON**: Pretty prints (or compacts) a valid JSON string.
- **Query JSON**: Obtain a value accessed with JS array/object accessor syntax from a valid JSON string.
- **Query JSON All**: Obtain all values (as a list) accessed with JS array/object accessor syntax from a valid JSON string, where `[]` accesses all.

Broken out from #4010.


Keavon:
@cubic-dev-ai

PR #4053, opened 2026-04-26
Add Artboard Grab/Scale

### LINKED ISSUE #4026: The Artboard tool needs to support G/S (grab and scale) transformation

atchamwa:
### Description
Added artboard tool support to the transform system so you can actually use the G/S (grand and scale) key binds for the artboard. Changing the artboard wont change the other things inside of it. Closes #4026 

### Changes Made:
transformation.rs - Added Artboard to the tool type checks so it skips the normal transform logic when using the Artboard tool
transform_layer_message_handler.rs - Made artboards filterable and gave them special handling. Also clears everything on cancel/finish.

https://github.com/user-attachments/assets/a3073c2e-7a48-4a63-9ea5-2421ec60ec9a


editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -95,6 +94,8 @@ pub struct TransformLayerMessageHandler {
 
 	// Path tool (ghost outlines showing pre-transform geometry)
 	ghost_outline: Vec<(Vec<ClickTargetType>, DAffine2)>,
+
+	original_artboard_bounds: HashMap<LayerNodeIdentifier, [DVec2; 2]>,
```
0HyperCube, line 98: I think this should probably be stored inside `OrigionalTransforms::Artboard`? This would simplify much of the code here.

0HyperCube (commented):
Seems to work well. I'm sorry that nobody bothered to review for many months.

PR #4056, opened 2026-04-26
Migrate iai-callgrind to the renamed successor gungraun

gamma0987:
Migrate from iai-callgrind to gungraun

### Summary

- Replace `iai-callgrind` with `gungraun` and update it from v0.16 to v0.18
- Rename all iai-callgrind benchmark files (`*_iai.rs` -> `*_gungraun.rs`) and their corresponding `[[bench]]` entries
- All imports updated: `use iai_callgrind::{…}` -> `use gungraun::prelude::*`
- Simplify CI workflow by replacing manual Valgrind installation and `iai-callgrind-runner` installation with the [`gungraun/setup-gungraun@v1`](https://github.com/gungraun/setup-gungraun) action.
- Update benchmark cache path from `target/iai` to `target/gungraun`

Other minor changes:

- Add `.nvim.lua` to `.gitignore`
- Update `deny.toml` advisory comment to reference `gungraun` instead of `iai-callgrind`

Advantages of using the `setup-gungraun` action in the CI: 

The action always tries to install the latest `valgrind` version. The action automatically installs the gungraun version which is used in the project, so the CI workflow no longer needs updates when bumping the gungraun version. Note that gungraun outputs the same json format as iai-callgrind, so no changes required here. 

I changed the cache key since just changing the path from `target/iai` to `target/gungraun` might not change the cache hit output from the `actions/cache` action. So, Baselines and cache will be rebuilt from scratch on the next CI run.


TrueDoctor:
The code looks good, a bit of a shame we can't test the full ci workflow before merging the pr

gamma0987:
I'm sorry I can't help with that.

gamma0987:
Hi. With the current approach the CI is failing because this PR would need to be merged into the master branch first to pass. Would you prefer a 2-step approach? It would be a little bit more work and the CI wouldn't fail in this PR. A follow-up PR could then cleanup the migration code.

TrueDoctor (approved):

PR #4058, opened 2026-04-27
fix: prevent text editor flash at origin

### LINKED ISSUE #4028: Text tool layer flashes at origin (0,0) before snapping to cursor position
When creating a new text layer, the text renders at document origin (0,0) for one frame and then places the cursor at the placed location.

anmolxlight:
## Summary
- Initialize the editable text input transform matrix with an identity fallback.
- Apply the backend-provided transform before rendering the contenteditable text input.
- Prevents a newly created text input from briefly appearing at viewport origin before moving to the cursor location.

Fixes #4028

## Test Plan
- `python3` targeted regression check confirming `textInputMatrix = data.transform` occurs before `showTextInput = true`
- `npx eslint src/components/panels/Document.svelte`

Note: `npm run check -- --quiet` was attempted after `npm ci`, but it requires generated `/wrapper/pkg/graphite_wasm_wrapper` bindings. This environment has no `cargo`, `rustc`, or `wasm-pack`, so the wrapper package could not be generated locally.

### DIFF
--- a/frontend/src/components/panels/Document.svelte
+++ b/frontend/src/components/panels/Document.svelte
@@ -32,7 +32,7 @@
 	// Interactive text editing
 	let textInput: undefined | HTMLDivElement = undefined;
 	let showTextInput: boolean;
-	let textInputMatrix: [number, number, number, number, number, number];
+	let textInputMatrix: [number, number, number, number, number, number] = [1, 0, 0, 1, 0, 0];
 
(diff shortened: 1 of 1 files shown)


Keavon:
Your PR description has the tells of being AI generated which is against the contribution policy, and likely implies your PR code is as well. It is our policy to close such PRs.

PR #4078, opened 2026-04-29
Reduce along-path snap influence near colinear anchor point

### LINKED ISSUE #2451: Dragging handle back into its anchor makes its opposite handle move around randomly
We no longer can snap a handle to its anchor point, thereby hiding it (giving it a zero length status where it's treated as gone). When doing this, the other handle shouldn't be affected from its original position.

devviktoria:
If I was editing a path and dragging a handle when the handle moves close to the anchor point then there were some weird jittering issues with the other handle. This occurred only  when the anchor is in collinear mode nad the Alpng Path snapping is tuned on. You can see my reproduction video here:

https://github.com/user-attachments/assets/aeadc965-8cbc-4990-8ab1-402c999c921f

I corrected this by reducing the snap weight when the handle is close to the anchor point (less then 30% of tolerance distance) (and of course we are in collinear mode and the snapping is Along Path) in this way the behavior is similar when there is no snapping.
Here is the video with my fix:

https://github.com/user-attachments/assets/b2462514-117e-484d-85be-42622b57409c

Closes https://github.com/GraphiteEditor/Graphite/issues/2451


devviktoria:
Please note that this does not contain test code, I will look into if I can create some tests for this next.

Keavon:
I force pushed to this, so please pull/reset to the version here on GitHub before you do any further commits or pushes. (Do NOT force push.)

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1062,6 +1063,9 @@ impl PathToolData {
 		snap_angle: bool,
 		tangent_to_neighboring_tangents: bool,
 	) -> f64 {
+		if handle_vector.length_squared() < f64::EPSILON {
+			return self.angle;
```
devviktoria: AI helped code: The drag function calls this and the result is used to calculate the new_handle_position, but in case the handle is at the anchor this returned NaN, which made the new_handle_position NaN. That caused problems. Not it is returning with self.angle wich is better.

devviktoria:
When I was working on the tests I was able to reproduce exactly the reported problem. I think this is not in my video but this screenshot says it all. 
2451_0_0snapping
The root cause of the problem was in the self.calculate_handle_angle. This was called by the drag function and the returned value was used to calculate the new_handle_position. But the .calculate_handle_angle function returned NaN if the handle was on the anchor's position. Now I had corrected this too and one of the test cases it covering it too.

PR #4100, opened 2026-05-03
New nodes: 'Attach Attribute' and 'Read Attribute *'

Keavon:
Re-attempt at the rolled-back #4099 due to the 1.77 MB increase in binary size, way more than I expected.

Now we're using the `Convert` trait and introducing +0.26 MB, more manageable.


Keavon:
!build ([Run ID 25306544541](https://github.com/GraphiteEditor/Graphite/actions/runs/25306544541))

PR #4101, opened 2026-05-03
Clean up document message wrappers around proto nodes so they're now used directly

Keavon:
Fixes the `skip_impl` bugs only showing up on web but not desktop, and simplifies the code by removing the wrapper document node definitions when the auto-generated proto nodes can now be sufficient.


editor/src/messages/portfolio/document/node_graph/document_node_definitions/document_node_derive.rs
```diff
@@ -39,12 +44,18 @@ pub(super) fn post_process_nodes(custom: Vec<DocumentNodeDefinition>) -> HashMap
...
 
 		let valid_inputs: HashSet<_> = implementations.iter().map(|(_, node_io)| node_io.call_argument.clone()).collect();
-		let input_type = if valid_inputs.len() > 1 { &const { generic!(D) } } else { &first_node_io.call_argument };
+		let input_type = if valid_inputs.is_empty() {
+			&context_type
+		} else if valid_inputs.len() > 1 {
+			&const { generic!(D) }
+		} else {
+			&first_node_io.call_argument
+		};
```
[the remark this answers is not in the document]
TrueDoctor: This is valid in theory but won't currently cause any test failures so this is something to keep in mind

TrueDoctor (changes requested):
This breaks basically all tool interactions in the editor

node-graph/node-macro/src/codegen.rs
```diff
@@ -623,7 +623,17 @@ fn generate_phantom_data<'a>(fn_generics: impl Iterator<Item = &'a crate::Generi
 
 fn generate_register_node_impl(parsed: &ParsedNodeFn, field_names: &[&Ident], struct_name: &Ident, identifier: &Ident) -> Result<TokenStream2, Error> {
 	if parsed.attributes.skip_impl {
-		return Ok(quote!());
+		// Skip generating a `register_node` implementation, but still emit a Wasm-side shim that registers metadata.
+		// On native, `register_metadata` runs automatically via `#[ctor]`. On Wasm it needs an explicit shim, otherwise
+		// NODE_METADATA stays empty for `skip_impl` nodes on the web target.
+		let registry_name = format_ident!("__node_registry_{}_{}", NODE_ID.fetch_add(1, std::sync::atomic::Ordering::SeqCst), struct_name);
+		return Ok(quote! {
```
TrueDoctor: This is duplicate code which should be shared with the other code path

node-graph/nodes/gcore/src/memo.rs
```diff
@@ -6,15 +6,19 @@ use std::hash::Hasher;
 use std::sync::Arc;
 use std::sync::Mutex;
 
-/// Caches the output of a given node called with a specific input.
+/// Improves rendering performance if used in rare circumstances where automatic caching is not yet advanced enough to handle the situation.
```
TrueDoctor: This is just bs we don't really have an automatic caching system, our current system only adds cache nodes when context deps change which is rarely the case and basically never when you don't use animation / read* nodes.
Inserting memo nodes is especially useful when working on large documents. If you want to change the comment you can add a non doc comment which could mention that the node macro can now be used to post append a memo node automatically

node-graph/nodes/gcore/src/memo.rs
```diff
@@ -23,23 +27,23 @@ async fn memo<I: CacheHash + Send + 'n, T: Clone + WasmNotSend>(input: I, #[data
...
+	let value = content.eval(input).await;
 	*cache.lock().unwrap() = Some((hash, value.clone()));
 	value
 }
 
 type MonitorValue<I, T> = Arc<Mutex<Option<Arc<IORecord<I, T>>>>>;
 
-/// Caches the output of the last graph evaluation for introspection.
-#[node_macro::node(category(""), path(graphene_core::memo), serialize(serialize_monitor), skip_impl)]
+/// The Monitor node is used by the editor to access the data flowing through it.
```
TrueDoctor, line 37: > The Monitor node ~is~ *is no longer* used by the editor to access the data flowing through it.

To be consistent with the transform changes

TrueDoctor (dismissed):

node-graph/nodes/gcore/src/memo.rs
Keavon: This is just transferring the existing-in-master description but you're right that the user-facing description could be made more precise without overwhelming the user with detail. I'll give it a shot to reword that.

node-graph/nodes/gcore/src/memo.rs
TrueDoctor: Feels free to ignore this one, I was just annoyed and made a snarky comment

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -133,46 +132,6 @@ static DOCUMENT_NODE_TYPES: once_cell::sync::Lazy<HashMap<DefinitionIdentifier,
 /// Only the position can be set for protonodes within a definition. The rest of the metadata comes from the node macro in NODE_METADATA
```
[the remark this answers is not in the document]
Keavon: That's expected and is a minor drawback of this approach but each usage still retains it, so it should be safe as long as we continue to keep that set for each usage.

editor/src/messages/portfolio/document/node_graph/document_node_definitions.rs
```diff
@@ -1742,132 +1520,6 @@ fn document_node_definitions() -> HashMap<DefinitionIdentifier, DocumentNodeDefi
...
-					output_names: vec!["Data".to_string()],
-					..Default::default()
-				},
-			},
-			description: Cow::Borrowed("TODO"),
-			properties: None,
-		},
 	];
 
 	document_node_derive::post_process_nodes(custom)
```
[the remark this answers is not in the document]
Keavon: Fixed.

Keavon:
> This breaks basically all tool interactions in the editor

Solved. Forgot to update a few tooling references to the old document node. Should have tested a bit more than a few seconds but I was really late to going to bed and thought I'd test more in the morning.

PR #4106, opened 2026-05-05
Desktop: Register the Graphite app as a file handler on Mac

### LINKED ISSUE #3575: Graphite should support importing image files by opening them as a launch argument to the desktop executable
blender lets you set any painting app as an external editor but currently gets this error

Image

Image

maybe if default reaction to graphite trying to open an image file it would would open a new graphite file and import the image file that is being tried to be opened

timon-schelling:
closes #3575


timon-schelling:
!build desktop ([Run ID 25389980031](https://github.com/GraphiteEditor/Graphite/actions/runs/25389980031))

Issue #4103, opened 2026-05-04
Add other boolean modes to the Boolean Operation node


devviktoria:
On [discord](https://discord.com/channels/731730685944922173/731738914812854303/1499412005428396122) boldline suggested to implement the following feature: 
Would you consider adding a "merge" option to the boolean tools? Specifically one that functions like in Illustrator and in Vectorstyler? Essentially, if the two objects are the same color, it will merge them together, if they are different colors, it will cookie-cutter them but leave both pieces there. And also @Keavon suggested to add the other boolean modes that can be constructed out of the existing boolean ops algorithm from the Linesweeper library that would be expected in other vector graphics apps.

So far Graphite implemented: Union, Subtract front, Subtract back, Intersect, Difference (this one is equivalent to Illustrator's Exclude, VectorStyler's Remove Overlaps, Inkscape's Exclusion. But maybe a better name can be used for this?)

My proposal is to add the following operators to the Boolean Operation Node:

- Merge: Apply union for all the objects that share the same fill property, removes strokes from all objects  (Equivalent of Illustrator's Merge, VectorStyler's Merge functionality)
- Divide/Fracture: Calculates all the overlapping and non overlapping parts (Equivalent of Illustrator's Divide, VectorStyler's Divide, Inkscape's Fracture)
- Trim: Subtracts top objects from the objects below them and keeps the remainder of the operation and also the top objects, removes stroke from all objects (Equivalent of Illustrator's Trim, VectorStyler's Trim functionality)

Reference material:
How to Use the Pathfinder Panel in Adobe Illustrator (Full Overview)
https://www.youtube.com/watch?v=mkNnyDEVKUY
The Pathfinder Tool Explained // Adobe Illustrator Tutorial 
https://www.youtube.com/watch?v=tJLMnaP3VyM
https://helpx.adobe.com/illustrator/desktop/manage-objects/reshape-transform-objects/pathfinder-panel-overview.html

Boolean Operations Reference Guide
https://vectorstyler.com/forum/topic/1758/boolean-operations-reference-guide/2

All Boolean Path Operations in Inkscape Explained (with Examples) 
https://www.youtube.com/watch?v=Xr8HqSwqn2E
https://inkscape-manuals.readthedocs.io/en/latest/boolean-operations.html

devviktoria:
I think I am going to start with the Merge functionality since this was requested on discord.
[...]

PR #4122, opened 2026-05-07
Add additional boolean operations

devviktoria:
This pull requests adds the Crop and Trim boolean operations to Grahpite. 

The Trim operation is best described in the following video:
Adobe Illustrator Pathfinder Panel Explained! Understand Your Shape-Merging Options
https://youtu.be/Me5oOw8r0Is?t=370

The Crop operation is best described in this part of the video:
Adobe Illustrator Pathfinder Panel Explained! Understand Your Shape-Merging Options
https://youtu.be/Me5oOw8r0Is?t=428

Part of https://github.com/GraphiteEditor/Graphite/issues/4103

### DIFF
--- a/node-graph/nodes/path-bool/src/lib.rs
+++ b/node-graph/nodes/path-bool/src/lib.rs
@@ -18,8 +18,6 @@ use vector_types::kurbo::{Affine, BezPath, CubicBez, Line, ParamCurve, PathSeg, 
 pub use vector_types::vector::misc::BooleanOperation;
 
 // TODO: Fix boolean ops to work by removing .transform() and .one_instance_*() calls,
-// TODO: since before we used a Vec of single-item `List`s and now we use a single `List`
-// TODO: with multiple items while still assuming a single item for the boolean operations.
 
 /// Combines the geometric forms of one or more closed paths into a new vector path that results from cutting or joining the paths by the chosen method.
 #[node_macro::node(category("Vector: Modifier"), memoize)]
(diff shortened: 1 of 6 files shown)


node-graph/nodes/path-bool/src/lib.rs
```diff
@@ -36,24 +34,29 @@ async fn boolean_operation<I: graphic_types::IntoGraphicTable + 'n + Send + Clon
 
 	// The first index is the bottom of the stack
 	let flattened = flatten_vector(&content);
-	let mut result_vector_table = boolean_operation_on_vector_table(&flattened, operation);
+	let mut result_vector_table = match operation {
```
devviktoria: This code is created with AI help. For the existing Union, SubtractFront, SubstrackBack, Intersect and Difference we should use the boolean_operation_on_vector_table, but for Trim and Crop now we should use the cascading_subtract function.

devviktoria:
So far I had tested only the basic functionality so I would like to test it with edge cases to make sure everything is working fine.  And also I will check Geminis code review and creating tests for the Crop and Trim.

Keavon:
Happy to see this coming along, nice work. Beware of the merge conflict, you'll want to sync and resolve that when you get a chance. When you are ready to mark this as no longer a draft, you'll also get a round of reviews from the Cubic bot which will be helpful (and its reviews tend to have fewer false positives than Gemini which is more hit-or-miss).

devviktoria:
Please note that I renamed BooleanOperation::Difference to BooleanOperation::Exclude. I did renamed the icon name in icons.ts however I could not rename the boolean-difference.svg because I noticed that is in a different repository. Please let me know if I should rename it in that other repository.

devviktoria:
Basically I finished coding the Trim and the Crop boolean operations with some tests included too.

However I cannot finalize this without some feedback from the maintainers.

I have a question about the node's UI. Currently only the operation can be selected and the operation keeps the fill and the stroke of the shapes. The question is shall we give the user the option which appearance shall we keep for the end result, fill or the stroke or both? Or shall we keep this simple UI we currently have. Once I got the answer to this question I can finish the test code to test if the look is what it is expected.
extrabool_op_idea


There is a boolean-difference.svg in branding/assets/icon-16px-solid/. Since I had renamed the difference operation to Exclude (because the difference name was too confusing, for this operation the Exclude/Exclusion is the typical name in other applications.) This icon should be renamed to boolean-exclude.svg. But this icon is in another repository so I cannot rename it. Or this renaming is going to occur after this has been merged?

devviktoria:
And also there are 2 new icons needed for this 2 boolean operation. (I forgot to mention this in my previous comment.)

Issue #3779, opened 2026-02-17
Data trees


Keavon:
This is a spec for bringing clarity to the plans for the next evolution of our visual programming language in Graphite as we move towards better support for pure functional programming paradigms without the limitations imposed by today's incomplete architecture.

This spec is also related to:
- #2988
- #1832
- #1173

This is described from a user-facing perspective, from which the technical requirements arise in order to produce the necessary solution to the problem space we uniquely face. A technical implementation plan will still need to designed meeting this spec which will supersede #2522, but take inspiration from it for the dynamic handling of attributes.

## Key ideas

In the past, Graphite passed only singleton data through the data flow, although the old `GraphicGroup` type allowed nesting. In #1834, we transitioned to the current system where graphical types use `Table<T>` in order to make everything work with lists of data, for all graphical types `T` such as `Vector`, `Raster`, `Color`, and `Graphic`, but all other types like `String`, `f64`, `DVec2`, and `bool` all still get passed around as singleton types. And really, `Table<T>` is itself just a singleton that internally stores multiple child elements. This spec is the answer to how to progress further towards a world where everything (all types, even `f64`, etc.) can be treated as a singleton but become part of a list of multiple elements, and have attributes even as a singleton.

Instead of currently passing `f64` from a node, we pass an object/record like `{ element: f64 }` where `element` is its primary key and thus determines its type in our type system, which would be `f64` in this case. This is not a list, just a singleton object. However, if this is passed through a node which adds an attribute, this could become `{ element: f64, enabled: bool }` by adding some string-indexed value to the object. The Data panel visualizes this as a table with the headers "element" and "enabled" and a single row with the values, but no column for the row index since this is not part of a table. If this singleton object eventually reaches a layer that combines it with other `f64`-type objects such as a different `{ element: f64, name: String }`, they would be combined into a list (visualized in the Data panel as a table with columns for the row index, `element`, `enabled`, and `name`) that looks like:

```
[
	0:{ element: f64, enabled: bool, name: String }
	1:{ element: f64, enabled: bool, name: String }
]
```

Between the two merged `f64` objects, as long as their primary key (`element`) is of the same type, their other attributes do not have to match since they can be merged while filling in the missing attributes with each type's default value.

[...]

PR #4129, opened 2026-05-08
Remove serialization from Table<T> and make TaggedValue only store tooling/widget node inputs

Keavon:
Split commit by commit.

Partly closes #3779.


Keavon:
22.92 MB -> 22.24 MB Wasm bundle size, nice!

PR #4134, opened 2026-05-09
Fix panic when hiding nodes with zero inputs

### LINKED ISSUE #3629: Crash when hiding an 'Instance Position' or 'Pointer Position' node feeding a value

kimjune01:
Fixes #3629.

Nodes like `PointerPositionNode` have no value inputs — they only receive context. `drain(1..)` panics on an empty `Vec`. Guard with an `is_empty` check: push a synthetic `None` input for empty-input nodes, `truncate(1)` otherwise.

Adds regression tests for zero-input and multi-input hidden nodes.

### DIFF
--- a/node-graph/graph-craft/src/document.rs
+++ b/node-graph/graph-craft/src/document.rs
@@ -813,8 +813,12 @@ impl NodeNetwork {
 		if !node.visible && node.implementation != passthrough_node {
 			node.implementation = passthrough_node;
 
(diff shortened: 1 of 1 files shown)


timon-schelling:
This doesn't really fix the problem, just removes the symptom. Also seems LLM generated.

PR #4166, opened 2026-05-24
GPU: add support for N-many input images and not just exactly one

Firestar99:
I'm sitting in the Eurostar from rustweek to Paris, running down the track at 300km/h, with a 30min delay of course... and finally found some time for something I wanted to add right after GSoC but never came around to it :D 

**GPU blend nodes**
image

* add support for N-many image inputs for nodes, not just one
  * enables `blend GPU` node with 2 inputs
    * I could not get the CPU-based blend node to work, it'll always error out, so can't compare it's outputs
  * Can you think of other nodes this could enable?
* TODO support for 0 input textures
  * the output texture and format is determined from the first input image and format, matching behavior of blend node writing it's result into the first input texture
  * with no first input texture, there's no way to determine the output texture resolution
  * mandelbrot node's output image is created based on viewport and some weird code computing something?
* Makes multiple assumptions on how raster images should behave. I feel like a universal definition on how raster textures should handle things would make this a lot easier:
  * size of the output texture? Assumption: based on first input texture
  * format of the output texture? Assumption: currently first input texture, and upload uses RGBA8 so every texture is that
    * changing away from RGBA8 requires a pass to convert ton RGBA8 before using it with vello
  * 2 input textures with mismatched input sizes: **couldn't verify CPU impl**, currently just places image in the top left and matches pixel by pixel, no scaling

### DIFF
--- a/node-graph/libraries/wgpu-executor/src/shader_runtime/per_pixel_adjust_runtime.rs
+++ b/node-graph/libraries/wgpu-executor/src/shader_runtime/per_pixel_adjust_runtime.rs
@@ -33,7 +33,8 @@ impl PerPixelAdjustShaderRuntime {
 }
 
 impl ShaderRuntime {
-	pub async fn run_per_pixel_adjust<T: BufferStruct>(&self, shaders: &Shaders<'_>, textures: List<Raster<GPU>>, args: Option<&T>) -> List<Raster<GPU>> {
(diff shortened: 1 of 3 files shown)


TrueDoctor:
The mandelbrot node computes only the which is going to be visible in the viewport and at the correct resolution, that is what that math is for

PR #4189, opened 2026-06-02
Add smoothstep node

benstockil:
Add smoothstep node, as per [OpenGL reference](https://registry.khronos.org/OpenGL-Refpages/gl4/html/smoothstep.xhtml).

### DIFF
--- a/node-graph/nodes/math/src/lib.rs
+++ b/node-graph/nodes/math/src/lib.rs
@@ -568,29 +568,26 @@ fn clamp<T: std::cmp::PartialOrd>(
 
 /// Performs smooth Hermite interpolation between two values.
 #[node_macro::node(category("Math: Numeric"))]
-fn smoothstep<T: num_traits::float::Float>(
+fn smoothstep(
 	_: impl Ctx,
 	/// The value to be mapped, which is restricted to the range between the minimum and maximum values.
-	#[implementations(f64, f32)]
-	value: T,
+	value: f64,
 	/// The lower bound of the input range. Input values below this edge are mapped to 0.
-	#[implementations(f64, f32)]
 	#[default(0.)]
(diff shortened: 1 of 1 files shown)


node-graph/nodes/math/src/lib.rs
```diff
@@ -566,6 +566,33 @@ fn clamp<T: std::cmp::PartialOrd>(
...
+	/// The value to be mapped, which is restricted to the range between the minimum and maximum values.
+	#[implementations(f64, f32)]
+	value: T,
+	/// The lower bound of the input range. Input values below this edge are mapped to 0.
+	#[implementations(f64, f32)]
+	#[default(0.)]
+	edge_min: T,
+	/// The upper bound of the input range. Input values above this edge are mapped to 1.
+	#[implementations(f64, f32)]
+	#[default(1.)]
```
TrueDoctor: I think it might be best to just make this node work on f64 and don't have it be generic over f32 and f64. This was previously required, but nowadays, we have automatic conversions to f64 values, so this should no longer be necessary. @Keavon thoughs?

Keavon: Yes, just f64

Keavon: @benstockil

Keavon:
I will mark this as a draft while awaiting the requested change. Please be sure to un-mark it as draft when you've completed it.

benstockil:
Apologies for the delay! I've replaced the generic floats with f64s.

Keavon:
Thanks. Before we merge this, can you please leave the requested comment in #4208?

benstockil:
That's done now :+1:

TrueDoctor (approved):

PR #4197, opened 2026-06-04
Add Frame Number node

benstockil:
Add `Frame Number` node to get the current frame index, complementing the existing `Animation Time` node. 

New context traits `ExtractFrameNumber`  and `InjectFrameNumber` have been added to support this.


TrueDoctor:
I think the frame number should be derived from the animation time instead. But you can build an exact frame number node which reads the animation time and the frame rate

benstockil:
> I think the frame number should be derived from the animation time instead.

@TrueDoctor Is this true for both animation modes, or just `TimeBased`? If acceptable I'd prefer to maintain some way to read the exact frame counter for `FrameBased` animations, i.e.:

```rs
// in AnimationMessageHandler::timing_information:

let frame_number = match self.animation_time_mode {
	AnimationTimeMode::TimeBased => self.fps * (animation_time_millis / 1000.0),
	AnimationTimeMode::FrameBased => self.frame_index,
};
```

TrueDoctor:
When in frame based animation mode, the timing will always be an increment of `1/fps`. If we actually needed to pass the frame number through the context, we would most likely store this as a tuple with the animation time instead of as its own field / extract traits since they semantically express the same thing. Keeping them separate would introduce a different dependency flow analysis / different cache behavior. And we try to keep the context as light weight as possible so adding and extra field is not ideal if we can avoid it.

Keavon:
I'm sort of surprised there so *so many* merge conflicts, but I'll just point that out for you to solve when you get the chance. To help my sanity as I catch up on code reviews, I'm going to mark this as a draft until you make this merge-ready again with the conflict resolution (plus anything else you may think is necessary to make it ready to land, if you're aware of anything else before we review it). Please mark it as ready for review when it is (I won't do that for you). Thanks :)

PR #4208, opened 2026-06-07
Relicense Graphite under the dual MIT/Apache-2.0 license

Keavon:
As of merging this PR, Graphite has moved from an Apache 2.0 license (with certain MIT dual-licensed components) to a dual-licensed pair of the Apache 2.0 and MIT licenses throughout all code in the repository (unless noted otherwise). It does not apply retroactively to all parts (the non-respondents listed below are blocking specific small selections of code). Upon this PR landing, this becomes the first commit of the dual-licensed code base.

## Explanation

Hi there! Thanks for your past contributions to Graphite, the open source 2D graphics editor. We need your consent to add the MIT open source license to the project, so please take a few seconds to respond right now. This will make it possible to integrate Graphite code into Blender and fix some technicalities to make Graphite more usable across the Rust + open source ecosystems. What we need from you: please reply by copy-pasting this exact sentence if you agree:

```
I license my past and future contributions to Graphite under the dual MIT/Apache-2.0 license, allowing licensees to choose either at their option.
```

We are hoping for 100% of past contributors to reply with this comment. Unanimous consent is needed, so we urge you to help us out with a timely reply.

What this means: Graphite was started with one of the two popular permissive open source licenses, Apache 2.0, but (somewhat accidentally) without the other, MIT. The Rust ecosystem typically dual-licenses code under both of these similar licenses, and we're now aiming to rectify that split. This does not change the existing Apache 2.0 license, but adds MIT (used by the Godot game engine) as an additional option, which is needed for compatibility with older licenses like GPL-2 (used by Blender). This should be a noncontroversial change, but it's a technicality we legally need your consent for in order to standardize Graphite's open source licensing situation to open it up for integration by more open source projects. Thank you for your cooperation and prior involvement with the project!

## If you are one of the names below, we'd still appreciate a response with your explicit agreement (late is better than never)

- @akshay1992kalbhor
- `AlexiWolf`
- @AustinHen
- @aybdee
- @Daniel53245
- @nat-rix
- @renshuncui
- @Sahra-Zhou
- @Sambhram1
- @theaniketgiri
- @tlaplaca
- @zhiyuang

## Checklist

- [x] @0HyperCube
- [x] @0SlowPoke0
- [x] @4adex
- [x] @adamgerhant
- [x] @afrdbaig7
- [x] @ajjahn
- [x] @ajweeks
- [x] @akozlev
[...]

PR #4222, opened 2026-06-10
Fix text migration failure caused by outdated inputs_count

### LINKED ISSUE #4212: Text migration fail from Jan 16 2026
Text nodes created before #3643 (Jan 16 2026) such as [text_old.graphite.json](https://github.com/user-attachments/files/28721172/hello_text_old.graphite.json) initially import with error. If you import the modified file again (e.g. after autosave), it then works fine.
<img width="261" height="265" alt="These input types are incompatible:
• Input 1:
…found:       EditorApi<PlatformApplicationIo>
…expected: ()
• Input 3:
…found:       Font
…expected: Resource" src="https://github.com/user-attachments/assets/3afb7bfb-cbdd-4e42-b407-7fe1237c489d" />

This is because the first migration adding two parameters for max width doesn't update `inputs_count`
https://github.com/GraphiteEditor/Graphite/blob/2835c3b53669f6477473da2bb4fb6fb408b005e0/editor/src/messages/portfolio/document_migration.rs#L1537-L1538
Preventing the second migration from running as it requires `inputs_count == 13`:
https://github.com/GraphiteEditor/Graphite/blob/2835c3b53669f6477473da2bb4fb6fb408b005e0/editor/src/messages/portfolio/document_migration.rs#L1688-L1690
It seems like the other migrations manually update the input count. Doing the same here resolves the issue:
```diff
diff --git a/editor/src/messages/portfolio/document_migration.rs b/editor/src/messages/portfolio/document_migration.rs
index 8f7915b7a..5f27c293e 100644
--- a/editor/src/messages/portfolio/document_migration.rs
+++ b/editor/src/messages/portfolio/document_migration.rs
@@ -1574,6 +1574,7 @@ fn migrate_node(node_id: &NodeId, node: &DocumentNode, network_path: &[NodeId],
 		for i in 10..=12 {
 			document.network_interface.set_input(&InputConnector::node(*node_id, i), old_inputs[i - 2].clone(), network_path);
 		}
+		inputs_count = 13;
 	}
 
 	// Upgrade Sine, Cosine, and Tangent nodes to include a boolean input for whether the output should be in radians, which was previously the only option but is now not the default
```
However this seems a bit error prone (there are other places in the file that do not do this and rely on being the last migration). Perhaps retrieving the actual value each time would be better (it shouldn't have particular performance degradation I don't think)
```rs
document.document_network().nested_network(network_path).expect("network").nodes.get(node_id).expect("node").inputs.len()
```

sanket-jadhav-cse:
Fixes #4212

I found a problem, with text migration. When we upgrade nodes from 11 inputs to 13 inputs two new inputs are added. The inputs_count variable is not updated.

Later another text migration checks if inputs_count is 13. Since the value is not updated this migration is skipped when we import text documents.

This change makes sure inputs_count is updated after migration. So the next text migration can work correctly.


timon-schelling:
!build ([Run ID 27442637414](https://github.com/GraphiteEditor/Graphite/actions/runs/27442637414))

Keavon:
Thanks! Before we merge this, can you please submit your relicensing agreement in #4208? Context in that thread. Thanks.

timon-schelling:
This patch is originally from hypercube https://github.com/GraphiteEditor/Graphite/issues/4212#issue-4615461917

Keavon:
We actually independently fixed this earlier so now that I've updated the branch, this is a +0/-0 change, so I'll close it. But thanks for the attempt!

PR #4232, opened 2026-06-12
Move `TaggedValue::ResourceHash` to macro definition site

timon-schelling:

### DIFF
--- a/node-graph/graph-craft/src/document/value.rs
+++ b/node-graph/graph-craft/src/document/value.rs
@@ -10,6 +10,7 @@ use core_types::{CacheHash, Color, ContextFeatures, MemoHash, Node, Type, TypeDe
 use dyn_any::DynAny;
 pub use dyn_any::StaticType;
 pub use glam::{DAffine2, DVec2, IVec2, UVec2};
+use graphene_application_io::resource::ResourceHash;
 use graphic_types::raster_types::{CPU, Image, Raster};
 use graphic_types::vector_types::vector::style::{Fill, Gradient, GradientStops};
 use graphic_types::vector_types::vector::{self, ReferencePoint};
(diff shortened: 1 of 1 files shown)


timon-schelling:
@TrueDoctor any reason not to do this?

0HyperCube (changes requested):
Why move to the variant macro definition site?

The `tagged_value` macro will accept the `#[serde(skip)]` and use it in the definition of the enum variant (as you can see for other variants such as `DAffine2`).

This approach seems to just add more duplicated code. If you do not care about the duplicated code, you might as well get rid of the `tagged_value` macro.

timon-schelling:
> Why move to the variant macro definition site?

Not 100% sure, @Keavon said I should have done this when originally introducing this variant.

TrueDoctor:
I agree with @0HyperCube, I don't see a technical reason for this change and if this is purely about code organization, there are other way of achieving that

Keavon (approved):

Keavon:
This is a runtime-only, ephemeral variant that is not serialized, so this fixes its incorrect placement amongst the serialized variants to be now correctly amongst its runtime-only variants. That was the original purpose of splitting these up between the ones that describe value node input widget persisted values in a .graphite document versus those that are internal machinery.

TrueDoctor:
> This is a runtime-only, ephemeral variant that is not serialized, so this fixes its incorrect placement amongst the serialized variants to be now correctly amongst its runtime-only variants. That was the original purpose of splitting these up between the ones that describe value node input widget persisted values in a .graphite document versus those that are internal machinery.

Then the original split was not the right way to enforce this. The current solution seems pretty fragile with a non compiler enforced invariant upheld by you. If this is something you have strong opinions on, we should either add nesting to the tagged value, or enforce this in the generation macro using two input lists

Keavon:
I like the latter option.

PR #4235, opened 2026-06-14
Resolve scopes before flattening network

timon-schelling:
Followup for #4221 that resolves issues with scope resolution when using it #4210
Also clean up accidentally committed test nodes

### DIFF
--- a/node-graph/graph-craft/src/document.rs
+++ b/node-graph/graph-craft/src/document.rs
@@ -659,39 +659,40 @@ impl NodeNetwork {
 impl NodeNetwork {
 	pub fn resolve_scope_inputs(&mut self) {
 		let mut leftover = Vec::new();
-		self.resolve_scope_inputs_with(None, &mut leftover);
+		self.resolve_scope_inputs_impl(None, &mut leftover);
 		assert!(leftover.is_empty(), "Unresolved scope keys at top level: {leftover:?}");
 	}
 
-	fn resolve_scope_inputs_with(&mut self, parent: Option<&ScopeChain<'_>>, parent_inputs: &mut Vec<NodeInput>) {
-		let chain = ScopeChain {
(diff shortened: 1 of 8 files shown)


node-graph/graph-craft/src/document.rs
```diff
@@ -655,6 +655,59 @@ impl NodeNetwork {
...
 
+// Functions for resolving scope inputs
+impl NodeNetwork {
+	pub fn resolve_scope_inputs(&mut self) {
+		let mut leftover = Vec::new();
+		self.resolve_scope_inputs_with(None, &mut leftover);
+		assert!(leftover.is_empty(), "Unresolved scope keys at top level: {leftover:?}");
+	}
+
+	fn resolve_scope_inputs_with(&mut self, parent: Option<&ScopeChain<'_>>, parent_inputs: &mut Vec<NodeInput>) {
```
TrueDoctor: I think we usually call this `resolve_scope_inputs_impl`

node-graph/graph-craft/src/document.rs
```diff
@@ -655,6 +655,59 @@ impl NodeNetwork {
...
+		}
+
+		let mut key_to_idx: FxHashMap<Cow<'static, str>, usize> = FxHashMap::default();
+		for node in self.nodes.values_mut() {
+			for input in node.inputs.iter_mut() {
+				let NodeInput::Scope(key) = input else { continue };
+				if let Some((producer_id, _)) = self.scope_injections.get(key.as_ref()) {
+					*input = NodeInput::node(*producer_id, 0);
+					continue;
+				}
```
TrueDoctor: why do we need this special case handling? wouln't this be resolved through the chain as well?

TrueDoctor (approved):

node-graph/graph-craft/src/document.rs
timon-schelling: should be more clear now

TrueDoctor (approved):

PR #4236, opened 2026-06-14
Replace the 'Vec2 to Point' node with automatic conversion machinery

Keavon:
This removes the 'Vec2 to Point' node and it now auto-converts where possible (when it feeds a concrete, single-type input connector). If users need to plug a Vec2 into a generic (multi-type) input, they can also now pass it through the 'As Vector' type assertion node. Also, the generic (multi-type) 'Wrap Graphic' node, which is used by the Merge layer node subgraph, also now takes Vec2 (Rust type `DVec2`) since it's a common case. And this PR also renames 'To String' -> 'As String', 'To f64' -> 'As f64', 'To u32' -> 'As u32', and 'To u64' -> 'As u64' type assertion nodes for consistency with the new 'As Vector'.

### DIFF
--- a/node-graph/libraries/vector-types/src/vector/vector_types.rs
+++ b/node-graph/libraries/vector-types/src/vector/vector_types.rs
@@ -55,6 +55,23 @@ impl graphene_hash::CacheHash for Vector {
 	}
 }
 
+impl core_types::ops::FromAnchorPosition for Vector {
+	fn from_anchor_position(position: DVec2) -> Self {
+		let mut point_domain = PointDomain::new();
+		point_domain.push(PointId::generate(), position);
+
+		Self { point_domain, ..Default::default() }
+	}
+}
+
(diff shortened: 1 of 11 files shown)


node-graph/libraries/vector-types/src/vector/vector_types.rs
```diff
@@ -55,6 +55,23 @@ impl graphene_hash::CacheHash for Vector {
...
+impl core_types::ops::FromAnchorPosition for Vector {
+	fn from_anchor_position(position: DVec2) -> Self {
+		let mut point_domain = PointDomain::new();
+		point_domain.push(PointId::generate(), position);
+
+		Self { point_domain, ..Default::default() }
+	}
+}
+
+// Identity item conversion so `List<Vector>` satisfies the blanket `Convert<List<U>, ()> for List<T>`, letting its
```
timon-schelling, line 67: Could we make this less verbose?

timon-schelling (approved):

PR #4237, opened 2026-06-15
fix: avoid invalid transforms in repeat_array when count is 1

sanket-jadhav-cse:
## Summary

While looking through the repeat nodes, I noticed that `repeat_array` allows `count = 1`, but the calculation uses `(count - 1)` as a denominator.

When `count` is 1, this results in a zero denominator and can produce invalid transform values during the angle and translation calculations.

This change guards against that edge case by ensuring the denominator is at least 1.

## Changes

* Prevent division by zero when `count = 1`
* Keep existing behavior unchanged for `count >= 2`

## Testing

Reasoned through the `count = 1` execution path and verified that the denominator can no longer become zero.

### DIFF
--- a/node-graph/nodes/repeat/src/repeat_nodes.rs
+++ b/node-graph/nodes/repeat/src/repeat_nodes.rs
@@ -65,7 +65,7 @@ pub async fn repeat_array<T: Into<Graphic> + Default + Send + Clone + 'static>(
 ) -> List<T> {
 	let angle = angle.to_radians();
 	let count = count.max(1);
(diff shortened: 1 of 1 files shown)


0HyperCube (commented):
Thanks for the contribution. This works well.

This exact fix was implemented by @Keavon in #4307 last week https://github.com/GraphiteEditor/Graphite/pull/4307/changes/cef6fefa332a8b77f2e34020f20016e1855895cc.

Please close PR. Thanks.

PR #4243, opened 2026-06-16
Fix repeated randomization patterns across paths in Copy to Points

sanket-jadhav-cse:
## Summary

This fixes an issue in `copy_to_points()` where random scale and rotation values could repeat across different paths when using the same seed.

The RNGs were being initialized inside the outer row loop, which caused the random sequence to restart for each path. As a result, corresponding points on different paths could receive identical random values.

The fix moves RNG initialization outside the loop so the RNG state is preserved across all paths while still remaining deterministic for a given seed.

## Testing

Manual testing:

* Created multiple paths and used Copy to Points with random scale enabled
* Verified copies were generated correctly across paths
* Tested random rotation with non-symmetrical shapes
* Confirmed no crashes or rendering issues

Automated testing:

* Added a regression test covering multiple paths
* Verified the test fails with the buggy implementation
* Verified the test passes with the fix
* Ran the vector-nodes test suite successfully

### DIFF
--- a/node-graph/nodes/vector/src/vector_nodes.rs
+++ b/node-graph/nodes/vector/src/vector_nodes.rs
@@ -286,12 +286,14 @@ async fn copy_to_points<I: 'n + Send + Clone>(
 
 	let random_scale_difference = random_scale_max - random_scale_min;
 
-	for row in points.into_iter() {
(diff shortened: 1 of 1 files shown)


Keavon:
This PR appears to be entirely AI generated based on the looks of it, in violation of our AI contribution policy. I also do not see an issue mentioned, so as far as I know, the issue being addressed here may be entirely hallucinated.

PR #4244, opened 2026-06-16
Fix #4239: Support more types in Extract Element node

### LINKED ISSUE #4239: Support more types in the «Extract Element» node
Currently the «Extract Element» node is missing some types, making them difficult to use. For example, you cannot do anything with a `List<DAffine2>` or a `List<BlendMode>` that you can get from the «Read attribute [...]» node.

sanket-jadhav-cse:
Fixes #4239

The `Extract Element` node was missing support for a few list types that are already used elsewhere in the attribute system.

This adds support for:

* `List<bool>`
* `List<DAffine2>`
* `List<BlendMode>`
* `List<GradientType>`
* `List<GradientSpreadMethod>`

These types already have existing support through attribute readers/writers and related node infrastructure, but were not included in the `Extract Element` node's implementation list.

I verified that the added types satisfy the required trait bounds and that `cargo check --package graphic-nodes` completes successfully after the change.

This is a small registration-only change and does not modify the node's runtime behavior.


0HyperCube (changes requested):
Perhaps the same types could also be added to the [« Extend »](https://github.com/GraphiteEditor/Graphite/blob/b1d8ac0e64d2bf3f4730b9b147d0ef3c7f6d1e20/node-graph/nodes/graphic/src/graphic.rs#L495), [« Map »](https://github.com/GraphiteEditor/Graphite/blob/b1d8ac0e64d2bf3f4730b9b147d0ef3c7f6d1e20/node-graph/nodes/graphic/src/graphic.rs#L111) and [« Index Element »](https://github.com/GraphiteEditor/Graphite/blob/b1d8ac0e64d2bf3f4730b9b147d0ef3c7f6d1e20/node-graph/nodes/graphic/src/graphic.rs#L16) node? Ideally they should all support all the possible `List<T>` that can be found in the editor. This would significantly improve the utility of the nodes.

sanket-jadhav-cse:
Thanks! I've added support to those nodes as well

0HyperCube:
This is implemented by Keavon in #4335. Please close PR. Thanks and sorry that Keavon has wasted your time.

Keavon:
Superseded by #4335.

Issue #4213, opened 2026-06-08
Pen tool offset when merging


0HyperCube:
# Reproduction
- New document
- Select pen tool
- Draw a line not at the origin
- End drawing
- Start new layer somewhere else
- Place anchor on the end point of the original line
- Observe offset

# Video
https://github.com/user-attachments/assets/8a0009f0-947b-4fd4-99f8-23c0872f7188

# Code
The pen tool appears to refer to a "buffer" system e.g. `buffering_merged_vector`. However buffering system is thankfully removed. I have literally no clue what is meant to be going on, however I fixed the issue with the following diff. It probably breaks other things though.

```diff
diff --git a/editor/src/messages/tool/tool_messages/pen_tool.rs b/editor/src/messages/tool/tool_messages/pen_tool.rs
index 529404b12..89690ee68 100644
--- a/editor/src/messages/tool/tool_messages/pen_tool.rs
+++ b/editor/src/messages/tool/tool_messages/pen_tool.rs
@@ -401,8 +401,6 @@ struct PenToolData {
 	auto_panning: AutoPanning,
 	modifiers: ModifierState,
 
-	buffering_merged_vector: bool,
-
 	previous_handle_start_pos: DVec2,
 	previous_handle_end_pos: Option<DVec2>,
 	toggle_colinear_debounce: bool,
@@ -1851,26 +1849,30 @@ impl Fsm for PenToolFsmState {
 
 				self
 			}
+			(PenToolFsmState::PlacingAnchor, PenToolMessage::RecalculateLatestPointsPosition) => {
+				tool_data.recalculate_latest_points_position(document);
+
+				// If we were placing anchors then it would be a good idea to update the anchor if possible
+				if let Some(layer) = layer {
+					tool_data.handle_mode = HandleMode::ColinearLocked;
+					tool_data.bend_from_previous_point(SnapData::new(document, input, viewport), transform, layer, shape_editor, responses);
+					tool_data.place_anchor(SnapData::new(document, input, viewport), transform, input.mouse.position, responses);
+					PenToolFsmState::DraggingHandle(tool_data.handle_mode)
+				} else {
+					PenToolFsmState::Ready
+				}
+			}
 			(state, PenToolMessage::RecalculateLatestPointsPosition) => {
 				tool_data.recalculate_latest_points_position(document);
 				state
 			}
-			(PenToolFsmState::PlacingAnchor, PenToolMessage::DragStart { append_to_selected }) => {
[...]

PR #4263, opened 2026-06-21
Add option for disabling merging on shared pen tool anchors

### LINKED ISSUE #4214: Allow avoid merging layers with pen tool
# Reproduction
When drawing with the pen tool, a user may wish to keep two layers separate. This could be useful for e.g. applying different fills or animations. However if the anchor shares a position with an anchor from another layer, it will always automatically merge.

https://github.com/user-attachments/assets/78a04e27-e7d0-4303-8882-aa034ef2bfe9

# Fix
- Indication when placing the anchor will automatically merge layers (e.g. in the tooltip). It is currently very hard to tell.
- Option to avoid merging the layers even when they have the same anchor position.

# Code
https://github.com/GraphiteEditor/Graphite/blob/cc7d700bfda6961fece0b6125289ef39c9caa0a5/editor/src/messages/tool/tool_messages/pen_tool.rs#L1883-L1900

FadedBronze:
Resolves #4214

In that issue you can also see that the incorrect transform seems to be applied to the point that is being merged on. I'd like to fix that, but I wonder if It needs a refactor in order to fix the seeming current jank in the pen tool (like the deferred addition of the point). The spline tool seems to hold a list of layers for this but the pen tool doesn't store this list. 

Additionally, the issue outlines adding in a message in the tooltip, it seems this message needs to go through the snap manager in order to be displayed on the same tooltip and any solution here should work if in the future different messages are required to indicate extra behavior on snapping. 

It likely makes sense to make separate issues for these but holding off in case this isn't important.

### DIFF
--- a/editor/src/messages/tool/tool_messages/pen_tool.rs
+++ b/editor/src/messages/tool/tool_messages/pen_tool.rs
@@ -1595,6 +1595,9 @@ impl Fsm for PenToolFsmState {
 				let vector = document.network_interface.compute_modified_vector(layer);
 				let Some(vector) = vector else { return PenToolFsmState::GRSHandle };
 
+				// Keep the preview segment's endpoint tracking the cursor while the outgoing handle is transformed by the GRS operation
+				tool_data.place_anchor(SnapData::new(document, input, viewport), transform, input.mouse.position, responses);
+
 				if let Some(latest_pt) = tool_data.latest_point_mut() {
 					let layer_space_to_viewport = document.metadata().transform_to_viewport(layer);
 					let final_pos = layer_space_to_viewport.inverse().transform_point2(final_position);


0HyperCube:
Hi @FadedBronze, thanks for your interest.

This solution adds more state for users to keep track of. They won't be familiar with this from other tools and so new users will be confused when they accidentally enable/disable this.

Other tools such as Figma or Inkscape do not merge layers at all without explicit action. For example <kbd>Ctrl</kbd>+<kbd>E</kbd> in Figma or pressing the « Join nodes » button in inkscape (nodes=anchors)


> In that issue you can also see that the incorrect transform seems to be applied to the point that is being merged on.

This concerns issue #4213. I have found out roughly the problem and given a potential solution (although it is not checked for correctness).

> seeming current jank in the pen tool

The pen/path/select tools have too much state and are very confusing to work with the giant state machine. See #2740.

> tooltip, it seems this message needs to go through the snap manager

It is possible to disable the snapping by pressing on the magnet icon in the toolbar. For this reason the end path is handled separately:
https://github.com/GraphiteEditor/Graphite/blob/f184148a6cb87345c5f3a33be006930454130b12/editor/src/messages/tool/tool_messages/pen_tool.rs#L756-L772
Your solution would need to function without snapping enabled, however I would consider a similar tooltip to be reasonable (although @Keavon probably has other ideas)
 https://github.com/GraphiteEditor/Graphite/blob/f184148a6cb87345c5f3a33be006930454130b12/editor/src/messages/tool/common_functionality/snapping.rs#L509-L512

FadedBronze:
@0HyperCube 

> Your solution would need to function without snapping enabled, however I would consider a similar tooltip to be reasonable 

to clarify: are you saying when snapping is disabled the message should look similar to how it does now despite not going through the snap manager?

> This solution adds more state for users to keep track of. They won't be familiar with this from other tools and so new users will be confused when they accidentally enable/disable this.

the exact way this should be done seems to require more planning then

> The pen/path/select tools have too much state and are very confusing to work with the giant state machine.

I am looking through the pen tool trying to understand all the features it has. I have opened an issue documenting some issues with the GRSHandle state. I think it makes sense to open a tracking issue for all the problems with the pen, spline, and path tools especially if they are to be merged/refactored/rewritten. 

https://github.com/GraphiteEditor/Graphite/blob/d753ce26f7b9a9cf8ce83c0141978d3254ff1318/editor/src/messages/tool/tool_messages/pen_tool.rs#L104-L112

0HyperCube:
> to clarify: are you saying when snapping is disabled the message should look similar to how it does now despite not going through the snap manager?

In my view, it should be possible to close a path without the snapping enabled. Currently we just magically do this when you click within a certain distance with no indication. I don't have strong opinions about how this should be done, but I think that some indication of the region being closed / connected would be beneficial.

> the exact way this should be done seems to require more planning then

I don't think that merging the layers is a super frequent operation. It can also be quite frustrating to do it accidentally and then not realise what has happened. Potentially it could be dropped entirely from the pen tool. Users are still able to use the path tool's <kbd>Ctrl</kbd>+<kbd>J</kbd> to merge paths.

PR #4273, opened 2026-06-22
Bump linesweeper to 0.4

jneem:
As linesweeper 0.4 reversed the path direction, this removes the temporary reversal.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3169,9 +3169,9 @@ checksum = "d4a5ff6bcca6c4867b1c4fd4ef63e4db7436ef363e0ad7531d1558856bae64f4"
 
 [[package]]
 name = "linesweeper"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8421b276e96af0ace5f3d8d2d165d0dea07fe764d2fe94ec06bb1acaf8a1e759"
+checksum = "9c19728333c060c6569a53c9a0e56c4be0df52cb4e6e07a8fbe16084cecce769"
 dependencies = [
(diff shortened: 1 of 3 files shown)


Keavon:
Thanks! What's the `of_tag` function?

jneem:
It's an annoying consequence of the fact that linesweeper abstracts over the definition of winding numbers. So there's a `WindingNumber` trait that implements a collection of winding numbers, one for each set that you care about, and the type `WindingNumber::Tag` identifies a single set. For example, for binary ops linesweeper provides [BinaryWindingNumber](https://docs.rs/linesweeper/latest/linesweeper/topology/struct.BinaryWindingNumber.html#impl-WindingNumber-for-BinaryWindingNumber) and its tag type takes 2 values: "set a" or "set b".

`WindingNumber::of_tag` just restricts the winding number to a single set. So if `w` is a binary winding number representing `-5` for "set a" and `3` for "set b" then `w.of_tag(SetA)` is `(-5, 0)` and `w.of_tag(SetB)` is `(0, 3)`.

Since Graphite supports n-ary set ops, the winding number type has dynamic length and `of_tag` uses dynamic indexing. If the tag is `(tag, out_of)` then `out_of` is always the length of the array and `tag` is the index of an element. It's definitely a bug if the array access is out of bounds, and I'd be inclined to just panic in that case. But if you want to be defensive about it, I'd say that returning the all-zero array is acceptable.

Keavon:
I see, thanks for the explanation. And yes, we try to avoid crashing the app if a bug occurs, since broken behavior with a console error (`log::error!("...");`) is better than making the user potentially lose their work.

Keavon (approved):

PR #4280, opened 2026-06-24
Change footprint to always be in physical instead of logical space

timon-schelling:
The render node(s) pass along the logical instead of the physical scale as part of the footprint, leading to upstream nodes rendering at the wrong scale and producing textures of the wrong size.

wrong at 2.8x scale
image

correct at 2.8x scale
image

The first Commit is a minimal fix.
The second makes the create context node responsible for converting the render configs footprint to physical scale and then passing that along. Things that explicitly require logical scale like metadata and svg rendering convert back to logical.

Both approaches work, @TrueDoctor any opinions?
I'm slightly in favor of the second.

### DIFF
--- a/node-graph/nodes/gstd/src/render_cache.rs
+++ b/node-graph/nodes/gstd/src/render_cache.rs
@@ -379,16 +379,7 @@ pub async fn render_output_cache<'a: 'n>(
 		if missing_region.tiles.is_empty() {
 			continue;
 		}
-		let region = render_missing_region(
-			missing_region,
-			|ctx| data.eval(ctx),
-			ctx.clone(),
-			render_params,
-			&footprint.transform,
-			&device_origin_offset,
-			render_params.scale,
-		)
-		.await;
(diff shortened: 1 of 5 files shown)


TrueDoctor:
Why is there the Vello vs SVG asymmetry in the first commit? Would this lead to blurry results when embedding raster content using the svg rendering path?

timon-schelling:
> Why is there the Vello vs SVG asymmetry in the first commit? Would this lead to blurry results when embedding raster content using the svg rendering path?

I think yes, I just accepted the current behavior for SVG but you are likely right that we should also change that.

node-graph/nodes/gstd/src/render_cache.rs
```diff
@@ -345,12 +345,10 @@ pub async fn render_output_cache<'a: 'n>(
 		return data.eval(context.into_context()).await;
 	}
 
-	let device_scale = render_params.scale;
-	let zoom = footprint.scale_magnitudes().x;
+	let zoom = footprint.scale_magnitudes().x / render_params.scale;
```
TrueDoctor: could the scale be just removed here?

TrueDoctor (approved):

PR #4285, opened 2026-06-26
Implement pasting nodes at cursor position in graph

### LINKED ISSUE #3798: Pasted node should spawn at cursor's position in the graph
As the title suggests.

kevlu93:
Closes #3798 

This PR implements cursor-based paste by calculating the offset of the first selected node to be copied from the location of the most recent click. The following changes were added:

1. The copied nodes are sorted by new id, so that we know the selection order of the nodes.
2. Calculate the difference between the mouse position (`ipp.mouse.position`) and the location of the first selected node.
3. Add the nodes to the graph (`NodeGraphMessage::AddNodes`), and then shift by the offset  (`NodeGraphMessage::ShiftSelectedNodesByAmount`)

One thing I'm not sure about is that one needs to actually click the desired paste location in order for `ipp.mouse.position` to update. Is there another something else tracking the cursor location?

Here's two examples of copy and paste

First we select the Transform node at (-15,3) as well as the Path node.

initial copy

Next, I click (-16,4) based on eyeballing the grid, then paste the nodes with Ctrl+v. The Transform node is pasted to (-16,4), while the pasted Path node maintains the relative distance to the Transform node, without any wires to the Transform node.

paste1

Next, I select Stroke -> Transform -> Path and then copy the nodes. Stroke is at (-22, 3).

second copy

Finally, I click at (-16, 8) then paste with Ctrl+v. We paste Stroke node at (-16, 8), and the other selected nodes are copied as well with wires maintained.

paste2

### DIFF
--- a/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
+++ b/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
@@ -12,7 +12,9 @@ use crate::messages::portfolio::document::node_graph::document_node_definitions:
 use crate::messages::portfolio::document::node_graph::utility_types::{ContextMenuData, Direction, FrontendGraphDataType, NodeGraphErrorDiagnostic};
 use crate::messages::portfolio::document::utility_types::document_metadata::LayerNodeIdentifier;
 use crate::messages::portfolio::document::utility_types::misc::GroupFolderType;
-use crate::messages::portfolio::document::utility_types::network_interface::{self, FlowType, InputConnector, NodeNetworkInterface, NodeTypePersistentMetadata, OutputConnector, Previewing};
+use crate::messages::portfolio::document::utility_types::network_interface::{
+	self, FlowType, InputConnector, LayerPosition, NodeNetworkInterface, NodePosition, NodeTemplate, NodeTypePersistentMetadata, OutputConnector, Previewing,
+};
 use crate::messages::portfolio::document::utility_types::nodes::{CollapsedLayers, LayerPanelEntry};
 use crate::messages::portfolio::document::utility_types::wires::{GraphWireStyle, WirePath, WirePathUpdate, build_vector_wire};
 use crate::messages::prelude::*;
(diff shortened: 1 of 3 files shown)


timon-schelling:
!build ([Run ID 28262192850](https://github.com/GraphiteEditor/Graphite/actions/runs/28262192850))

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -774,6 +774,48 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
...
+				let cursor_viewport_location = ipp.mouse.position;
+				let cursor_to_node_graph = network_metadata
+					.persistent_metadata
+					.navigation_metadata
+					.node_graph_to_viewport
+					.inverse()
+					.transform_point2(cursor_viewport_location);
+
+				// Sort the selected nodes by the new id so that we know which node was selected first
+				data.sort_by_key(|a| a.0);
```
0HyperCube: The nodes are already in order of the ascending node ids:

https://github.com/GraphiteEditor/Graphite/blob/97f8113fe43d41b15ca3a56c300c48b7c8fda5c4/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs#L255

0HyperCube (changes requested):
Please rebase. #4296 rewrote the clipboard pasting system so the functionality goes inside `NodeGraphMessage::InsertNodes`:

https://github.com/GraphiteEditor/Graphite/blob/97f8113fe43d41b15ca3a56c300c48b7c8fda5c4/editor/src/messages/clipboard/clipboard_message_handler.rs#L292-L294

Also I think that this could benefit from a test.

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
kevlu93: > The nodes are already in order of the ascending node ids:
> 
> https://github.com/GraphiteEditor/Graphite/blob/97f8113fe43d41b15ca3a56c300c48b7c8fda5c4/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs#L255

I don't think new_ids are in order since it's a hash map. ```new_ids``` is keyed by the old node id value, but since it's a hashmap you can't guarantee that elements are sorted by key. To guarantee order you would need to use a BTreeMap instead, or sort the nodes before creating the iterator returned by `copy_nodes`, or what I did where you sort the copied nodes when the original order is relevant. 

https://github.com/GraphiteEditor/Graphite/blob/97f8113fe43d41b15ca3a56c300c48b7c8fda5c4/editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs#L256

If we want to preserve the original order when copying nodes, then I think we sort by key [here](https://github.com/GraphiteEditor/Graphite/blob/97f8113fe43d41b15ca3a56c300c48b7c8fda5c4/editor/src/messages/portfolio/document/utility_types/network_interface.rs#L445) ie. 
```
new_nodes.sort_by_key(|a| a.0);
new_nodes.into_iter()...
```

This preserves the order whenever we copy nodes, with minimal code changes. We don't need the new nodes to be sorted until this step, so we can continue using HashMap in ```copy_nodes``` instead of changing to BTreeMap.

0HyperCube: @kevlu93 You misunderstand. I am not claiming that the `HashMap` will be ordered as it obviously will not. I am saying that the `all_selected_nodes[0]` will get `NodeId(0)` since that is how the enumeration works.

PR #4164, opened 2026-05-23
Add animation curves support

Gavin-Niederman:
A PR adding support for f-curves and a timeline `NodeInput` variant.

TODO:

- [X] `AnimationCurve`
- [X] Timeline `NodeInput` variant.
- [X] Store timeline curves in document

## Demo

https://github.com/user-attachments/assets/8b62f555-af24-4873-bc12-0113e4a7b540

PR #4290, opened 2026-06-28
Replace AnimationCurve curve value inputs during preprocessing

Gavin-Niederman:
This PR adds a `replace_animation_curve_inputs` method to the preprocessor that replaces any `TaggedValue::AnimationCurve(curve)` value inputs in the network new `eval_curve` nodes. It also refactors out some of the logic in `replace_resource_inputs` into `visit_*` functions to reduce code duplication.

In the future, this will be used in the expose dialog to allow exposing inputs to the timeline.

This PR is a follow-up to #4164. Until that PR is merged, this one will include its commits.

# Demo

https://github.com/user-attachments/assets/e2c49d6b-1a8b-4ca0-a8b5-19ec17df7c4d


node-graph/preprocessor/src/lib.rs
```diff
@@ -23,6 +23,7 @@ impl Preprocessor {
 	pub fn preprocess(&self, network: &mut NodeNetwork, resolve_resource: &dyn Fn(ResourceId) -> Option<ResourceHash>) -> Result<(), PreprocessorError> {
 		self.insert_inject_scopes(network);
 		self.replace_resource_inputs(network, resolve_resource)?;
```
timon-schelling, line 25: Maybe we should consolidate both input replace passes into one.

Having one replace_inputs_with_producer_nodes (or similar) that has a match for the node input types for node in the network.

```rust
match input {
  <resource> ...
  <animation_curve> ...
  _ => {}
}
```

node-graph/preprocessor/src/lib.rs
```diff
@@ -41,47 +42,103 @@ impl Preprocessor {
...
-				self.replace_resource_inputs(nested, resolve_resource)?;
-				continue;
+			for node in network.nodes.values_mut() {
+				if let DocumentNodeImplementation::Network(nested) = &mut node.implementation {
+					inner(nested, visitor)?;
+					continue;
+				}
+
+				visitor(node, &mut inserts)?;
 			}
```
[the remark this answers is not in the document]
Gavin-Niederman: This pr doesn't not introduce or modify this behavior.  This is the same as the previous implementation.

node-graph/libraries/animation/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+		}
+
+		let segment_start = &self.keyframes[index - 1];
+		let segment_end = &self.keyframes[index];
+
+		match segment_start.interp_behavior {
+			InterpolationBehavior::Bezier { right_handle } => {
+				let start = segment_start.knot;
+				let end = segment_end.knot;
+				let left_handle = segment_end.left_handle.unwrap_or(end);
```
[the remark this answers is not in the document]
Gavin-Niederman: This diff is from the first pr.

node-graph/preprocessor/src/lib.rs
Gavin-Niederman: > Maybe we should consolidate both input replace passes into one.

Would you also want to remove the visitor abstractions in this case since there will only be one function that uses it?

node-graph/libraries/animation/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "graphene-animation"
+version = "0.1.0"
+edition = "2024"
```
timon-schelling, line 4: use *.workspace = true like in other crates for version etc. (copy from other crates to keep order)

node-graph/preprocessor/src/lib.rs
```diff
@@ -41,48 +41,67 @@ impl Preprocessor {
...
+					continue;
+				}
+				DocumentNodeImplementation::ProtoNode(identifier) => Some(&*identifier),
+				DocumentNodeImplementation::Extract => None,
+			};
 
-			if matches!(&node.implementation, DocumentNodeImplementation::ProtoNode(identifier) if *identifier == platform_application_io::resource::IDENTIFIER) {
-				continue;
-			}
+			let is_resource_node = identifier == Some(&platform_application_io::resource::IDENTIFIER);
```
timon-schelling, line 59: Not sure if this is needed.

The function should never run on those nodes, right?

node-graph/libraries/animation/src/lib.rs
```diff
@@ -0,0 +1,267 @@
+//! Animation Curves
```
timon-schelling, line 1: I was not that involved in the discussion about this but it seems like this fits the usecase.
personally I would changes some names but I assume the naming is common for similar keyframe based animation systems (like blender)?

timon-schelling (changes requested):
Also needs a rebase

PR #4292, opened 2026-06-29
Box big enum variants to satisfy Clippy's lint warnings

Keavon:


editor/src/messages/portfolio/document/node_graph/node_properties.rs
```diff
@@ -2112,13 +2115,13 @@ pub(crate) fn string_capitalization_properties(node_id: NodeId, context: &mut No
 						NodeGraphMessage::SetInputValue {
 							node_id,
 							input_index: UseJoinerInput::INDEX,
-							value: TaggedValue::Bool(true),
+							value: Box::new(TaggedValue::Bool(true)),
```
TrueDoctor: For these I would generally prefer to write this using `.into()` instead of the more verbose `Box::new` but this is personal preference

editor/src/messages/portfolio/document/node_graph/node_graph_message_handler.rs
```diff
@@ -1764,6 +1764,7 @@ impl<'a> MessageHandler<NodeGraphMessage, NodeGraphMessageContext<'a>> for NodeG
 			NodeGraphMessage::SetInputValue { node_id, input_index, value } => {
 				use graphene_std::vector::generator_nodes::*;
 
+				let value = *value;
```
TrueDoctor: This is a bit ugly since without inlay hints, it is a bit hard to understand what this does. We could change the name of one of the variables to make this clearer, but this is not a big issue

TrueDoctor (approved):

PR #4294, opened 2026-06-30
Copy paste with resources

Keavon:
Fixes the bug that copy-pasting a Text layer into a different document results in intermittent render errors and console errors.


timon-schelling:
This would still miss the content making for example images not work.

I suggest that we instead make the content of a clipboard:

```rust
// string format "graphite: <serialized GraphiteClipboardContent>"

type GraphiteClipboardContent = Vec<GraphiteClipboardItem>;

enum GraphiteClipboardItem {
    Layers(...),
    ...
    Resource {
      id: ResourceId,
      hash: Option<ResourceHash>,
      sources: Vec<DataSource>,
      data: Option<ResourceData>,
    }
}

type ResourceData = Vec<u8>; // serialized as base64 string, maybe with compression
```

timon-schelling:
We should also take this as a opportunity to move all the clipboard code into a separate message handler.

PR #4305, opened 2026-07-02
Desktop: Use dirty rects for UI texture uploads

timon-schelling:
Slightly improves latency for large windows. Most noticeable on low end hardware with a "slow" CPU to GPU bus. Almost no effect on faster systems or systems with unified memory. 
Mostly useful for Linux, we have ui acceleration disabled by default currently.

Wanted to do this for quite some time now.
Not as effective as I hoped and dose nothing about the still not optimal resize latency.
Still consider it a good idea to merge this and if only for the cleanup.


timon-schelling:
@TrueDoctor Would likely self merge soon. Not sure if this needs your review.

PR #4307, opened 2026-07-04
Revamp the syntax to #[hard(a..b)] and #[soft(a..b)] bounds on node definitions

Keavon:


timon-schelling (commented):
This is not optimal. Why not `#[slider(0..100)]` for the slider range and `#[clamp(-500..500)]` for optional clamping also allowed for non slider inputs. Way more descriptive.

Issue #2458, opened 2025-03-18
Add node that generates geometry using parametric equations


optozorax:
I wanted to draw this illustration using graphite, but ended up using desmos and screenshotting it, because they support parametric equations. It will be nice to have such a node.

<details>
  <summary>Desmos illustration</summary>
  
<img width="1920" alt="Image" src="https://github.com/user-attachments/assets/e1311998-71dc-4bfe-8d05-d87e44e473ed" />

  
</details>

I think that in this node we should be able to specify density of the geometry (amount of points), because different equations require different density, and it can affect performance.

I think there also should be density multiplier for the export rendering (if there is such a regime). So that at the editing stage we may see segmented geometry, and for the final render it will be smooth.

Also, unsolved problem: how to make this node accept any amount of variables (so that you can tweak behavior of your equation, like `l` in my example). This is also the problem for the Math node. Without this, this node will be not as useful is it might be.

Keavon:
I was overthinking it by trying to consider how a Bezier path might be generated, but you're right that a polyline should be sufficient which would drastically improve the simplicity of this.

> Also, unsolved problem: how to make this node accept any amount of variables (so that you can tweak behavior of your equation, like `l` in my example). This is also the problem for the Math node. Without this, this node will be not as useful is it might be.

I just filed #2462 for this.

laundmo:
If i understand the feature request correctly, a node like this would be easiest to accomplish using something like https://crates.io/crates/fasteval which parses and evaluates math expressions.

(i'm just passing by, i just noticed this issue and remebered the crate)

Chen-Hao-WANG:
are this issue be assigned ? i would like to solve this issue as a start for the GSoC

Keavon:
Feel free to get started with it but you'll probably have to first work with us in Discord to figure out the appropriate design for how it's used and implemented. Like do we use the math expression library we've written for our *Math* node? Does that require adaption to support equations (not just expressions)?

Keavon:
Insights on the Desmos approach: https://www.reddit.com/r/desmos/comments/wak098/rendering_algorithm_used_for_implicit_functions/

PR #4313, opened 2026-07-06
Parametric curve plot

devviktoria:
This adds a simple parametric curve plot to Graphite.
This algorithm is designed to be simple it allows the user to add simple function plots to the document. It has no intention to compete with other more professional plotting software, like Desmos.
It uses an adaptive sampling (instead of fixed one), because in this way the plot looks better with less points and also the discontinuity detection works better too.

Known limitations:
- at most one variable is currently supported in the expressions
- there are no symbolic analysis for precise function domain detection, we are figuring out the domain by evaluating the given expressions
- the discontinuity detection is also happening by analysing the sampled data; it may detect false positives/false negatives
- there are no periodicity detection, since we are doing a simple adaptive sampling

Part of: [#2458](https://github.com/GraphiteEditor/Graphite/issues/2458)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6476,6 +6476,7 @@ dependencies = [
  "graphic-types",
  "kurbo",
  "log",
+ "math-parser",
  "node-macro",
  "qrcodegen",
  "rand",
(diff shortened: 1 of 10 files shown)


devviktoria:
Here is a video about the current functionality:

https://github.com/user-attachments/assets/e1b5569b-b779-4c28-bd9b-b83dee0c303a

PR #4319, opened 2026-07-09
Fix node catalog preserving type search on autofocus

### LINKED ISSUE #4311: Type hint is selected in node graph view
## Reproduction
- Drag a wire from an output with a resolved type
- Type something to narrow down your search
- Note how the type information is lost from the search since it is automatically selected

This is really annoying UX.

## The node catalogue is only temporary

Is it really? It is basically unchanged from its initial implementation for 2 years. See #2072.

## Code
The node catalogue automatically focuses the input (this is desired):
https://github.com/GraphiteEditor/Graphite/blob/9f9899cfd0c2ddea4a8664358b6b1834f30adc66/frontend/src/components/floating-menus/NodeCatalog.svelte#L110-L111
When the focus is changed, the text input automatically selects all the text (not desired)
https://github.com/GraphiteEditor/Graphite/blob/9f9899cfd0c2ddea4a8664358b6b1834f30adc66/frontend/src/components/widgets/inputs/TextInput.svelte#L31-L35

sahilsahni18:
Fixes #4311 
Prevent Node Catalog autofocus from selecting all text so prefilled type search context is preserved while refining results.

### DIFF
--- a/frontend/src/components/widgets/inputs/TextInput.svelte
+++ b/frontend/src/components/widgets/inputs/TextInput.svelte
@@ -16,10 +16,12 @@
 	// Sizing
 	export let minWidth = 0;
 	export let maxWidth = 0;
-	// Tooltips
+		// Tooltips
 	export let tooltipLabel: string | undefined = undefined;
 	export let tooltipDescription: string | undefined = undefined;
 	export let tooltipShortcut: ActionShortcut | undefined = undefined;
+	// Behavior
+	export let selectAllOnFocus = true;
 
 	let className = "";
 	export { className as class };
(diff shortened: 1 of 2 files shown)


frontend/src/components/widgets/inputs/TextInput.svelte
```diff
@@ -16,10 +16,12 @@
 	// Sizing
 	export let minWidth = 0;
 	export let maxWidth = 0;
-	// Tooltips
+		// Tooltips
```
0HyperCube, line 19: Please format the file

0HyperCube (changes requested):
It would be more functional if the space was added after the type hint. Currently you either have to manually add the space or you will create an invalid type such as "type:Vector[]jitter".

PR #4320, opened 2026-07-09
Add SVG Gaussian blur filter effect

### LINKED ISSUE #4308: Support for SVG filter effects
SVG has support for many filter effects (elements prefixed with `fe`). Documentation is available on [MDN list of SVG elements](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/feDropShadow). A filter can be made by composing one or more of these filter effects. For example a glow could be made by combining the [`feMorphology`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/feMorphology) to make the element thicker and then the [`feGaussianBlur`](https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/feGaussianBlur) to blur the result. This can allow for powerful graphic effects whilst remaining a vector (useful for exporting and performance). Examples from the Inkscape filter gallery:

Inkscape filter gallery featuring a grid of different filter effects such as oil painting

Filters can also be composed by hand using a filter editor view:

a filter editor, featuring a list of filter effects and the details for a selected effect

This style of approach lends itself well the the existing graphite node graph.

Currently there is implemented a raster based blur and the ability to change the blend mode (this is implemented with the blend mode not a filter effect). However this significantly limits the possibilities compared to Inkscape.

The filter effect must also be implemented in the vello rendering pipeline, attempting to match the SVG rendering in a browser. For efficient SVG output, it should be possible to combine filter effects applied to an element into a single filter that can be reused. However this is not a particular priority given the rest of the SVG output is horrible.

sahilsahni18:
## Summary

Adds an SVG-compatible Gaussian blur filter effect for vector content.

## Changes

- Adds `ATTR_FILTER_EFFECTS` as a vector render attribute
- Adds `SvgFilterEffect::GaussianBlur`
- Adds a Vector Style node for SVG Gaussian blur
- Emits `<filter>` and `<feGaussianBlur>` during SVG export from `List<Vector>::render_svg`

## Notes

This is an MVP for #4308 focused on SVG export. Vello renderer parity and additional filter primitives can be added in follow-up work.

Fixes #4308 partially.

### DIFF
--- a/frontend/src/components/floating-menus/NodeCatalog.svelte
+++ b/frontend/src/components/floating-menus/NodeCatalog.svelte
@@ -113,7 +113,7 @@
 </script>
 
 <LayoutCol class="node-catalog">
-	<TextInput placeholder="Search Nodes…" value={searchTerm} on:value={({ detail }) => (searchTerm = detail)} bind:this={nodeSearchInput} />
+	<TextInput placeholder="Search Nodes…" value={searchTerm} on:value={({ detail }) => (searchTerm = detail)} selectAllOnFocus={false} bind:this={nodeSearchInput} />
 	<div class="list-results" on:wheel|passive|stopPropagation>
 		{#each nodeCategories as nodeCategory}
 			<details open={nodeCategory[1].open}>
--- a/node-graph/libraries/graphic-types/src/filter.rs
+++ b/node-graph/libraries/graphic-types/src/filter.rs
@@ -1,0 +1,17 @@
+#[derive(Debug, Clone, PartialEq, graphene_hash::CacheHash, dyn_any::DynAny)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum SvgFilterEffect {
+	GaussianBlur {
+		std_deviation_x: f64,
+		std_deviation_y: f64,
+	},
+}
+
+impl Default for SvgFilterEffect {
+	fn default() -> Self {
+		Self::GaussianBlur {
+			std_deviation_x: 0.,
+			std_deviation_y: 0.,
+		}
+	}
+}
--- a/node-graph/libraries/graphic-types/src/lib.rs
+++ b/node-graph/libraries/graphic-types/src/lib.rs
@@ -1,10 +1,12 @@
 pub mod artboard;
 pub mod graphic;
+pub mod filter;
 
 // Re-export all transitive dependencies so downstream crates only need to depend on graphic-types
 pub use core_types;
 pub use raster_types;
 pub use vector_types;
+pub use filter::SvgFilterEffect;
 
 // Re-export commonly used types at the crate root
 pub use artboard::Artboard;
--- a/node-graph/libraries/rendering/src/renderer.rs
+++ b/node-graph/libraries/rendering/src/renderer.rs
@@ -7,7 +7,7 @@ use core_types::bounds::RenderBoundingBox;
 use core_types::color::Color;
 use core_types::color::SRGBA8;
 use core_types::consts::DEFAULT_FONT_SIZE;
-use core_types::list::{ATTR_FILL, ATTR_STROKE, Item, List};
+use core_types::list::{ATTR_FILL, ATTR_FILTER_EFFECTS, ATTR_STROKE, Item, List};
 use core_types::math::quad::Quad;
 use core_types::render_complexity::RenderComplexity;
 use core_types::transform::Footprint;
@@ -21,6 +21,7 @@ use dyn_any::DynAny;
 use glam::{DAffine2, DMat2, DVec2};
 use graphene_hash::CacheHashWrapper;
 use graphene_resource::Resource;
+use graphic_types::SvgFilterEffect;
 use graphic_types::graphic::{graphic_list_at, has_paint_at, is_paint_present, set_paint_attribute};
 use graphic_types::raster_types::{BitmapMut, CPU, GPU, Image, Raster};
 use graphic_types::vector_types::gradient::{GradientStops, GradientType};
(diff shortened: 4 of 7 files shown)


frontend/src/components/floating-menus/NodeCatalog.svelte
```diff
@@ -113,7 +113,7 @@
 </script>
 
 <LayoutCol class="node-catalog">
-	<TextInput placeholder="Search Nodes…" value={searchTerm} on:value={({ detail }) => (searchTerm = detail)} bind:this={nodeSearchInput} />
+	<TextInput placeholder="Search Nodes…" value={searchTerm} on:value={({ detail }) => (searchTerm = detail)} selectAllOnFocus={false} bind:this={nodeSearchInput} />
```
0HyperCube, line 116: This is not relevant to this PR (probably #4319). Please keep your branches separate to allow for reviewing small changes…

node-graph/libraries/graphic-types/src/filter.rs
```diff
@@ -0,0 +1,17 @@
...
+pub enum SvgFilterEffect {
+	GaussianBlur {
+		std_deviation_x: f64,
+		std_deviation_y: f64,
+	},
+}
+
+impl Default for SvgFilterEffect {
+	fn default() -> Self {
+		Self::GaussianBlur {
```
0HyperCube, line 12: If more `SvgFilterEffect`s are added, it doesn't make sense to have the default be a gaussian blur. Perhaps the default trait should not be implemented?

node-graph/libraries/graphic-types/src/lib.rs
```diff
@@ -1,10 +1,12 @@
 pub mod artboard;
 pub mod graphic;
+pub mod filter;
 
 // Re-export all transitive dependencies so downstream crates only need to depend on graphic-types
 pub use core_types;
 pub use raster_types;
 pub use vector_types;
+pub use filter::SvgFilterEffect;
```
0HyperCube, line 9: Per code organisation, please put in the section below. This section is meant to "Re-export all transitive dependencies".

node-graph/libraries/graphic-types/src/lib.rs
```diff
@@ -1,10 +1,12 @@
 pub mod artboard;
 pub mod graphic;
+pub mod filter;
```
0HyperCube, line 3: Would this be better in the `vector-types` crate rather than in `graphic-types`? This matches other vector graphics rendering related features such as gradients.

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -268,6 +269,29 @@ pub fn format_transform_matrix(transform: DAffine2) -> String {
 	}) + ")"
 }
 
+fn write_svg_filter_def(svg_defs: &mut String, effects: &[SvgFilterEffect]) -> Option<String> {
+	if effects.is_empty() {
+		return None;
+	}
+
+	let id = format!("filter-{}", generate_uuid());
+	write!(svg_defs, r#"<filter id="{id}" x="-50%" y="-50%" width="200%" height="200%" color-interpolation-filters="sRGB">"#).unwrap();
```
0HyperCube, line 278: Perhaps add comment to explain why the default expansion was not enough to avoid clipping.

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -268,6 +269,29 @@ pub fn format_transform_matrix(transform: DAffine2) -> String {
...
+
+	let mut input = "SourceGraphic".to_string();
+	for (index, effect) in effects.iter().enumerate() {
+		let result = format!("effect{index}");
+		match effect {
+			SvgFilterEffect::GaussianBlur { std_deviation_x, std_deviation_y } => {
+				write!(svg_defs, r#"<feGaussianBlur in="{input}" stdDeviation="{std_deviation_x} {std_deviation_y}" result="{result}"/>"#).unwrap();
```
0HyperCube, line 285: There is no need to specify the `in` or `result` attributes if you just want to apply filter effects from top to bottom (which is what you are doing here anyway).

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -835,6 +859,7 @@ impl Render for List<Graphic> {
 		for index in 0..self.len() {
 			let transform: DAffine2 = self.attribute_cloned_or_default(ATTR_TRANSFORM, index);
 			let blend_mode: BlendMode = self.attribute_cloned_or_default(ATTR_BLEND_MODE, index);
+			let filter_effects: Vec<SvgFilterEffect> = self.attribute_cloned_or_default(ATTR_FILTER_EFFECTS, index);
```
0HyperCube, line 862: As per #3779, all things must now be a `List<T>` for unknown reasons. Consult with Keavon if unsure.

node-graph/nodes/vector/src/vector_nodes.rs
```diff
@@ -230,6 +231,48 @@ async fn fill<V: VectorListIterMut + 'n + Send, F: IntoGraphicList + 'n + Send +
...
+	#[hard(0..)]
+	#[soft(..100)]
+	std_deviation: PixelLength,
+) -> T
+where
+	T: FilterListIterMut + 'n + Send,
+{
+	content.push_filter_effect(SvgFilterEffect::GaussianBlur {
+		std_deviation_x: std_deviation,
+		std_deviation_y: std_deviation,
```
0HyperCube, line 270: Why not allow users to set the x and y to different values?

node-graph/libraries/graphic-types/src/filter.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Debug, Clone, PartialEq, graphene_hash::CacheHash, dyn_any::DynAny)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum SvgFilterEffect {
```
0HyperCube, line 3: I'm not sure if this name should include SVG since gaussian blur is a standard filter that exists outside of the SVG specification. Other graphic types do not include SVG specific names.

0HyperCube, line 3: The guassian blur has only one input, so can get it implicity from the filter effect before. Other effects take several inputs, so you need to expand the data structure to take this into account.

0HyperCube (changes requested):
An implementation must be added to vello before this can be merged. Otherwise it looks broken by default.

Please mark as draft until this is complete.

node-graph/libraries/rendering/src/renderer.rs
```diff
@@ -1161,6 +1186,10 @@ impl Render for List<Vector> {
 					attributes.push(ATTR_TRANSFORM, matrix);
 				}
 
+				if let Some(filter_id) = write_svg_filter_def(&mut attributes.0.svg_defs, &filter_effects) {
+					attributes.push("filter", format!("url(#{filter_id})"));
```
0HyperCube, line 1190: When zooming the viewport, the amount of blur changes. This is not desirable.

Keavon:
Hey, thanks for submitting the PR, but unfortunately this is not going to be possible to merge at the present time. We have to support both our Vello and SVG renderer features simultaneously, and additionally we are in the process of reworking the internal graphical data imaging model and amongst the upcoming parts of that design process will be the inclusion of SVG-compatible features such as this, however right now we don't know exactly how that will fit in, and merging this today would mean we have an incomplete feature requiring future maintenance and conversion to the new system, plus migrations overhead. But thank you for the attempt!

PR #4325, opened 2026-07-10
Lock segment molding to an axis with Shift

### LINKED ISSUE #3760: Segment molding Shift key X/Y constraint

Chetansahney:
Closes #3760

While molding a path segment with the Path tool, holding <kbd>Shift</kbd> now constrains the drag to a single axis (horizontal or vertical), matching the direction-locking behavior of other drag operations in the tool.

### Behavior
- Hold <kbd>Shift</kbd> during a segment mold to lock the movement to the dominant screen axis (whichever of X/Y the drag has travelled further along).
- Release <kbd>Shift</kbd> to resume free movement.
- A "Lock Direction" hint is shown in the status bar while molding.

### Implementation
- `mold_handle_positions` gains a `lock_direction` flag; when set, the incoming pointer position is snapped to the grab point's x or y before the drag delta is computed, so the segment moves purely along one axis.
- The Path tool forwards the existing `snap_angle` (<kbd>Shift</kbd>) key state into the molding branch, where it was previously unused, and adds the corresponding hint.

### DIFF
--- a/editor/src/messages/tool/common_functionality/shape_editor.rs
+++ b/editor/src/messages/tool/common_functionality/shape_editor.rs
@@ -354,7 +354,6 @@ impl ClosestSegment {
 		(c1, c2): (DVec2, DVec2),
 		new_b: DVec2,
 		break_colinear_molding: bool,
-		lock_direction: bool,
 		temporary_adjacent_handles_while_molding: Option<[Option<HandleId>; 2]>,
 	) -> Option<[Option<HandleId>; 2]> {
 		let transform = document.metadata().transform_to_viewport_if_feeds(self.layer, &document.network_interface);
@@ -364,13 +363,6 @@ impl ClosestSegment {
 
 		// Apply the drag delta to the segment's handles
 		let b = self.bezier_point_to_viewport;
-		// When the lock direction modifier is held, constrain the drag to the dominant screen axis (purely horizontal or vertical).
-		let new_b = if lock_direction {
-			let drag = new_b - b;
-			if drag.x.abs() >= drag.y.abs() { DVec2::new(new_b.x, b.y) } else { DVec2::new(b.x, new_b.y) }
-		} else {
-			new_b
-		};
 		let delta = transform.inverse().transform_vector2(new_b - b);
 		let (nc1, nc2) = (c1 + delta, c2 + delta);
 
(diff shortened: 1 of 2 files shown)


0HyperCube:
The implementation in #3766 is far more complete. It features overlays so you can see where you are aligning to (in your implementation the user just has to guess). It also does not offset the cursor position in such a strange way.

Chetansahney:
@0HyperCube 
Thanks for the review, and fair points on both counts. Pushed 01993397.

On the overlays: you're right that there was nothing to see. I'd missed that molding already runs inside the Dragging FSM state, which means the existing X/Y guide-line overlay just needs snapping_axis set to appear. Setting it during a mold gets the same red/green lines the tool already draws when dragging points, with no new drawing code or state. That's also what the issue asked for, so I should have gone there first.

On the cursor offset: that was a genuine bug, not just an odd choice. The constraint was snapping against bezier_point_to_viewport rather than the grab point, so pressing Shift shifted the segment by the gap between the two instead of only locking the axis. It now constrains relative to drag_start_pos, consistent with point dragging.

Also moved the constraint out of mold_handle_positions into the path tool, since the drag start isn't available in the shape editor, which drops the lock_direction parameter from that API. And cleared snapping_axis on the escape/abort path and on the early return in drag stop, both of which could otherwise leak a locked axis into the next drag.

That said, #3766 predates this by several months and covers the same issue. I'm not looking to compete with it, and I arrived at the same conclusion you'd already pointed at. If you'd rather land that one, I'm happy to close this in its favour. The two cleanup items above are independent of the feature, so I can split them into a separate PR if they're useful on their own.

cargo check, clippy, and fmt are clean. I haven't verified the guide lines visually yet.

0HyperCube:
> On the overlays: you're right that there was nothing to see.

> that was a genuine bug, not just an odd choice

Sorry what? You perhaps mean that you didn't implement them? This sounds LLM generated. Note that you *must not* use LLM generated text for PR comments [per the LLM contribution policy](https://graphite.art/volunteer/guide/starting-a-task/ai-contribution-policy/).

> I haven't verified the guide lines visually yet.

Well it is not worth anybody reviewing the code if you haven't checked if it worked. Please set state to draft PR until you have ensured you submit code that works to the best of your knowledge.

Thanks for your contributions so far!

PR #4326, opened 2026-07-10
FIX: Show local-axis grab measurement as a single tilted component

### LINKED ISSUE #3341: Local-axis grab (G key) dashed line should show only the X or Y component, tilted

Chetansahney:
Closes #3341

When grabbing (<kbd>G</kbd>) and pressing <kbd>X</kbd>/<kbd>Y</kbd> a second time to switch from a global-axis to a **local-axis** constraint, the on-canvas measurement overlay was drawn incorrectly. It showed an axis-aligned box with both the X and Y screen-space components, and when typing a value it printed that number on *both* axes, instead of a single line along the tilted local axis.

This implements the change requested in https://github.com/GraphiteEditor/Graphite/pull/3527#issuecomment-3910935758.

### Change
In the `Grabbing` overlay branch, when the transform is constrained to a single axis in local space, draw one dashed line from the pivot along the (already correctly tilted) `translation_viewport` vector, with a single measurement label — instead of the screen-aligned `translation_box`.

The unconstrained and global-axis cases are untouched and still use `translation_box`. The equivalent overlay for Scale (<kbd>S</kbd>) already behaved this way, so this brings Grab in line with it.


editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -234,15 +234,34 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
 
 						let typed_string = (!self.typing.digits.is_empty() && self.transform_operation.can_begin_typing()).then(|| self.typing.string.clone());
-						overlay_context.translation_box(translation_viewport / document_to_viewport.matrix2.y_axis.length(), quad, typed_string);
+
+						// When constrained to a single axis in local space, the translation vector is tilted along the layer's
+						// local axis. Draw one dashed line (with a single measurement label) along that axis, instead of the
+						// axis-aligned box, which would incorrectly show both the X and Y screen-space components.
+						if self.state.is_transforming_in_local_space && axis_constraint != Axis::Both {
+							let end_point = pivot + translation_viewport;
+							overlay_context.dashed_line(pivot, end_point, None, None, Some(2.), Some(2.), Some(0.5));
```
0HyperCube: I would expect the drawing of the overlays to be implemented in the `fn translation_box` code.

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -234,15 +234,34 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
+
+							let axis_direction = self.state.constraint_axis(axis_constraint).unwrap_or(DVec2::X).normalize_or(DVec2::X);
```
0HyperCube: You can just make this the overall condition e.g.

```rs
if let Some(axis_contraint) = self.state.constraint_axis(axis_constraint)
         && self.state.is_transforming_in_local_space {
    // Logic for constraint in local space
}
```

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -234,15 +234,34 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
+							let length = translation_viewport.dot(axis_direction) / document_to_viewport.matrix2.y_axis.length();
```
0HyperCube: This doesn't really make any sense to me? Why take the length of a column of the document to viewport matrix?

I guess it works given that you only rotate and scale in a uniform fashion the viewport.

I would suggest you do something like `document_to_viewport.transform_vector2(translation_viewport)`

editor/src/messages/tool/transform_layer/transform_layer_message_handler.rs
```diff
@@ -234,15 +234,34 @@ impl MessageHandler<TransformLayerMessage, TransformLayerMessageContext<'_>> for
...
+							let label = match typed_string {
+								Some(typed_string) => typed_string,
+								None => format!("{length:.2}").trim_end_matches('0').trim_end_matches('.').to_string(),
+							};
+							let text_position = pivot.midpoint(end_point) + translation_viewport.perp().normalize_or(DVec2::X) * 24.;
+							let transform = DAffine2::from_translation(text_position);
+							overlay_context.text(&label, COLOR_OVERLAY_BLUE, None, transform, 16., [Pivot::Middle, Pivot::Middle]);
```
0HyperCube: Magic constants 16 and 24 should be given names since I don't know what they do.

0HyperCube (changes requested):

Chetansahney:
@0HyperCube 
Thanks for the review! Pushed e51e5978 addressing all the feedback.

The main change: instead of hand-drawing the dashed line and label in the message handler, the measurement quad is now built in the layer's local (tilted) frame and handed to the existing translation_box. Because Quad holds four arbitrary corners and translation_box draws its X edge from top_left to top_right, this renders the single tilted line and one label with no new drawing code. The handler branch went from 13 lines to 3.

That folds in the rest of the feedback as a side effect:

Drawing now lives in translation_box rather than the handler.

The magic constants 16 and 24 are gone, along with the duplicated label formatting, since translation_box already handles both including the typed-string case. This also covers the format_rounded suggestion from gemini-code-assist, as the manual formatting no longer exists.

The branch condition is now constraint_axis(axis_constraint).is_some() && is_transforming_in_local_space, dropping the != Axis::Both comparison and the unwrap_or. I used .is_some() rather than binding the axis, since the vector itself is no longer needed once the quad is built in local space.

The matrix-column division is gone. The label length now comes from document_to_viewport.inverse().transform_vector2(translation_viewport).length(), which is the proper viewport-to-document conversion and stays correct under a non-uniform viewport transform.

Verified locally: cargo check -p graphite-editor passes, cargo clippy reports no warnings in the changed file, and cargo fmt --check is clean.

0HyperCube (approved):
Much cleaner solution; thanks for your continued work on this.

Consider merging @timon-schelling.

PR #4333, opened 2026-07-12
Fix lifetime of cached textures

timon-schelling:
Fixing

```
wgpu error: Validation Error

Caused by:
  In Queue::submit
    Texture with 'cached_texture_1954x1712' label has been destroyed
```

### DIFF
--- a/node-graph/libraries/wgpu-executor/src/texture_conversion.rs
+++ b/node-graph/libraries/wgpu-executor/src/texture_conversion.rs
@@ -50,6 +50,7 @@ struct RasterGpuToRasterCpuConverter {
 	height: u32,
 	unpadded_bytes_per_row: u32,
 	padded_bytes_per_row: u32,
+	_source: raster_types::Texture,
 }
 impl RasterGpuToRasterCpuConverter {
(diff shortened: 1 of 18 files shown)


node-graph/libraries/wgpu-executor/src/texture_conversion.rs
```diff
@@ -97,6 +98,8 @@ impl RasterGpuToRasterCpuConverter {
 			height,
 			unpadded_bytes_per_row,
 			padded_bytes_per_row,
+			// Keep source texture alive
+			_source: data_gpu.texture.clone(),
```
TrueDoctor, line 102: I would expect this to be the only actual fix right? Why is the diff so large?

timon-schelling: No, the Arc inside Raster<GPU> is important, this case here is more theoretical. The rest is changes needed because of the different types and removal of the duplicated ImageTexture struct. No need to have multiple `struct Type(Arc<wgpu::Texture>)`.

TrueDoctor (approved):

PR #4335, opened 2026-07-15
Make the data model use Item and List types universally, with nodes authored as rank-polymorphic kernels

### LINKED ISSUE #2047: Identity node has no compiled type
https://github.com/user-attachments/assets/36009c4e-07d4-48b4-8c41-a2180d8f63c9

There may be some optimization to remove the identity nodes when compiling the network, which means the type is not added to the `ResolvedDocumentNodeTypes` hashmap

### LINKED ISSUE #4239: Support more types in the «Extract Element» node

Keavon:
Closes #3779.

Bonus: Closes #2047. Closes #4239.

Here is a thematically split-up version of events with the same end state: https://github.com/GraphiteEditor/Graphite/compare/a040362d4a56a7ffac08c57697e409dc427a2694...876b49d4f4c5008bfa9d1ac1c81f33879b4a30b2


Keavon:
!build ([Run ID 29391783747](https://github.com/GraphiteEditor/Graphite/actions/runs/29391783747))

Keavon:
The expected 6.16 MB build size increase from doubling the monomorphizations should be temporary once we can move the dual List-and-Item implementations from the macro to deeper in the compiler in the future.

Instruction counting based metrics look worse than expected, however my real-world benchmarks show a frame rate drop from a consistent 28.8 to 27.6 FPS, which is an acceptable 4%. Tested using Chrome and this file in [master](https://16ebcd75.graphite.pages.dev/) and [this branch](https://808bd355.graphite.pages.dev/).
[Benchmark.graphite.txt](https://github.com/user-attachments/files/30035910/Benchmark.graphite.txt)

Since this PR must meet its deadline, the above can be recovered through future optimizations subsequent to this landing.

editor/src/messages/color_picker/color_picker_message_handler.rs
```diff
@@ -28,7 +28,7 @@ pub struct ColorPickerMessageHandler {
 	old_is_none: bool,
 
 	// When set, the picker is editing a gradient: the visual pickers and inputs target the active stop's color.
-	gradient: Option<GradientStops>,
+	gradient: Option<Gradient>,
```
0HyperCube, line 31: Perhaps refrain from the unrelated renames in the 8000 line PR?

editor/src/messages/portfolio/document/data_panel/data_panel_message_handler.rs
0HyperCube: I don't think it is a good idea to have implementations for both `f64` and `Item<f64>`. This is likely the cause of the bloated binary size and instruction count in the benches.

Why not add some automatic type casting from `T` to `Item<T>`? Then you don't have to have confusion around having `f64` and `Item<f64>` in the graph.

node-graph/interpreted-executor/src/util.rs
```diff
@@ -84,12 +84,12 @@ pub fn wrap_network_in_scope(network: NodeNetwork, editor_api: Arc<PlatformEdito
 					..Default::default()
 				},
 				DocumentNode {
-					call_argument: concrete!(graphene_std::application_io::RenderConfig),
+					call_argument: concrete!(Context),
```
0HyperCube, line 87: For what purpose is this changed?

node-graph/interpreted-executor/src/dynamic_executor.rs
```diff
@@ -430,19 +442,4 @@ impl BorrowTree {
...
-		let val_1_protonode = ProtoNode::value(ConstructionArgs::Value(TaggedValue::U32(2u32).into()), vec![]);
-		let context = TypingContext::default();
-		let future = tree.push_node(NodeId(0), val_1_protonode, &context);
-		futures::executor::block_on(future).unwrap();
-		let _node = tree.get(NodeId(0)).unwrap();
-		let result = futures::executor::block_on(tree.eval(NodeId(0), ()));
-		assert_eq!(result, Some(2u32));
-	}
-}
+mod test;
```
0HyperCube, line 445: There is no reason to have an empty test module?

Also why delete the old test?

node-graph/interpreted-executor/src/dynamic_executor.rs
```diff
@@ -413,7 +413,19 @@ impl BorrowTree {
 			ConstructionArgs::Inline(_) => unimplemented!("Inline nodes are not supported yet"),
 			ConstructionArgs::Nodes(ids) => {
 				let ids = ids.to_vec();
-				let construction_nodes = self.node_deps(&ids);
+				let mut construction_nodes = self.node_deps(&ids);
+
+				// Wrap arguments the typing pass marked for rank promotion with their adapter node
```
0HyperCube, line 418: Why does this happen in the borrow tree?

I would have thought this should occur before, however I'll defer to @TrueDoctor.

node-graph/libraries/core-types/src/list.rs
```diff
@@ -79,6 +80,54 @@ pub const ATTR_LETTER_TILT: &str = "letter_tilt";
...
+
+// ================
+// TYPE: Bundle
+// ================
+
+/// A whole `List<T>` treated as one rank-0 value (`Item<Bundle<T>>`) rather than a rank-1 `List<T>`.
+/// Bundling a collection lets it pass through a connector that selects or carries the entire collection as one opaque
+/// cell (such as a Switch branch), instead of the element-wise machinery zipping over it per element.
+#[derive(Clone, Debug, PartialEq)]
+pub struct Bundle<T>(pub List<T>);
```
0HyperCube, line 107: Why? You could do a `List<List<T>>`, avoiding introduction of more terminology (which I'm unfortunately struggling to keep track of).

Alternatively you could do the approach previously discussed whereby the `List` type stores the shape of the array. This would allow a single List type to represent all things including an `Item`.

node-graph/libraries/core-types/src/list.rs
```diff
@@ -1292,9 +1296,22 @@ impl<T: Default> Default for Item<T> {
...
+		}
+	}
+}
+
 impl<T: PartialEq> PartialEq for Item<T> {
 	fn eq(&self, other: &Self) -> bool {
-		self.element == other.element
+		// Attributes participate in equality for the same `cache_hash` contract reason as `List`
+		self.element == other.element && self.attributes == other.attributes
 	}
```
0HyperCube: This is the same as `#[derive(PartialEq)]`?

node-graph/libraries/no-std-types/src/list.rs
```diff
@@ -0,0 +1,41 @@
+//! A zero-cost stand-in for `core_types::list::Item` used when node kernels are compiled for the GPU.
+//!
+//! Shader node kernels compile twice: under `std` against the real attribute-carrying `Item`, and under
+//! `no_std` (SPIR-V) against this transparent wrapper, imported as `Item`. Only the element-access surface is
+//! provided, since rust-gpu cannot allocate and attributes have no per-pixel meaning; attribute use fails the
```
0HyperCube, line 5: If you plan to make a lot of things into attributes, then how will this work given the shaders cannot read attributes?

For example if the blend node is given a gpu raster item with opacity and blend node attributes. If the blend node is a shader then it will not be able to extract such information

Perhaps it is necessary to have a preprocess step to extract and type the attributes? If this is the case, would it not be advantageous to also do this for CPU nodes?

node-graph/nodes/repeat/src/repeat_nodes.rs
```diff
@@ -137,19 +171,31 @@ async fn repeat_radial<T: Into<Graphic> + Default + Send + Clone + 'static>(
...
 		Context -> List<Raster<CPU>>,
+		Context -> List<Raster<GPU>>,
 		Context -> List<Color>,
-		Context -> List<GradientStops>,
+		Context -> List<Gradient>,
+		Context -> List<Artboard>,
 	)]
 	content: impl Node<'n, Context<'static>, Output = List<T>>,
-	reverse: bool,
+	reverse: Item<bool>,
```
0HyperCube, line 195: What is the point of this change? There is never going to be a need to add arbitrary data to a boolean for a reverse flag.

node-graph/nodes/repeat/src/repeat_nodes.rs
```diff
@@ -137,19 +171,31 @@ async fn repeat_radial<T: Into<Graphic> + Default + Send + Clone + 'static>(
...
 
 #[node_macro::node(category("Repeat"), name("Repeat on Points"))]
-async fn repeat_on_points<T: Into<Graphic> + Default + Send + Clone + 'static>(
+async fn repeat_on_points<T: Send + Clone + 'static>(
 	ctx: impl ExtractAll + CloneVarArgs + Sync + Ctx + InjectVarArgs,
 	points: List<Vector>,
 	#[implementations(
-		Context -> List<Graphic>,
+		Context -> List<String>,
+		Context -> List<bool>,
```
0HyperCube, line 179: What is the point of adding a transform attribute to booleans?

0HyperCube (commented):
I think that this is in principle a great improvement. Having the rectangle node return a list of vectors has baffled me ever since you first changed it.

I have left a couple of considerations here, however I was unable to look at most of the files in this PR (GitHub apparently can't render long diffs and I don't have the attention span).

> Since this PR must meet its deadline, the above can be recovered through future optimisations subsequent to this landing.

> +8,774 -5,043

I'm a bit confused as to why you are rushing this large change

>  Instructions: 59,385,391 (master) → 93,004,444 (HEAD) : + 56.61 % 

> 6.16 MB build size increase

This is pretty dramatic +25% binary size for a web file. I guess you did add a lot of types for the monomorphisation in the dynamic executor and we enumerate all combinations. I feel there are some easy wins in this regard which I have pointed out in my comments.

node-graph/interpreted-executor/src/dynamic_executor.rs
TrueDoctor: This is a concern I have raised as well

node-graph/interpreted-executor/src/util.rs
Keavon: This standardizes how the RenderConfig is fed through all the nodes. It's now consistent, which is also required by the infrastructure now.

editor/src/messages/portfolio/document/data_panel/data_panel_message_handler.rs
Keavon: Good find, that was leftover code.

node-graph/interpreted-executor/src/dynamic_executor.rs
Keavon: This was broken out into `test.rs` which this now references.

editor/src/messages/color_picker/color_picker_message_handler.rs
Keavon: Technically separable but it was part of this change set and is not worth the effort of extracting.

node-graph/nodes/repeat/src/repeat_nodes.rs
Keavon: All types are now ranked by this PR, which is the new consistent data model.

node-graph/nodes/repeat/src/repeat_nodes.rs
Keavon: There is technically no reason against it, so no reason to surprise an adventurous user by disallowing it.

node-graph/interpreted-executor/src/dynamic_executor.rs
Keavon: Valid opportunity for future cleanup.

node-graph/libraries/no-std-types/src/list.rs
Keavon: Currently no shaders need to read attributes. Further down the road in this design evolution, we'll have the concept of wires being compiler-routed bundled scalar data flows, one wire per attribute/column, and those will probably all be typed and thus the shaders will be able to access these as uniforms.

node-graph/libraries/core-types/src/list.rs
Keavon: `List<T>` (and `Item<T>`) is type infrastructure while the inner `T` may be anything else: `f64`, `Raster<X>`, or in this case `Bundle<X>`. A `List<List<T>>` would be type infrastructure, a rank-2 type, which this does not represent. It is a workaround for limitations hit by the Switch node, and that's all.

node-graph/interpreted-executor/src/util.rs
TrueDoctor: And it also makes the code unsound. Please don't change such core mechanims without thinking them through. For many things there is a reason why I designed them that way

TrueDoctor (changes requested):

Keavon:
For the record, here is a thematically split-up version with the same end state: https://github.com/GraphiteEditor/Graphite/compare/a040362d4a56a7ffac08c57697e409dc427a2694...876b49d4f4c5008bfa9d1ac1c81f33879b4a30b2

PR #4343, opened 2026-07-15
made luminance work as a mask

NicTanghe:
## Summary

Adds typed support for distinguishing alpha and luminance clipping masks throughout
Graphite’s document and rendering pipeline.

Previously, all clipping masks were treated as alpha masks. Consequently, opaque black
and white SVG mask artwork produced identical results, even though luminance masks
should use color luminance multiplied by alpha.

When merging take special attention to how it interacts with goupmasks becouse it should transfer extra fields ecs.


NicTanghe:
Was only able to get the luminance masks working after merging the 2 branches.

PR #4354, opened 2026-07-19
Desktop: Support input from drawing tablets

### LINKED ISSUE #4312: Update winit
Winit is causing some users to crash at startup when wacom tablet connected https://github.com/rust-windowing/winit/issues/4493. [Reported by myhouse](https://discord.com/channels/731730685944922173/731738914812854303/1521886300741963806).

This was fixed in February. However graphite still uses a winit commit from October 2025:

https://github.com/GraphiteEditor/Graphite/blob/6a54dcb5dabe34a511a91654e08839a44d7e1062/Cargo.lock#L7477

Updating the winit version to the latest release would likely resolve this issue.

timon-schelling:
winit bump not 100% necessary. Some more drawing tablet edge cases have been addressed that I never saw. But might as well do the update now that i touch that code anyway. Could be separated out if that is wanted.

Everything else is pretty well described with the commit messages.


timon-schelling:
!build desktop ([Run ID 29704545193](https://github.com/GraphiteEditor/Graphite/actions/runs/29704545193))

timon-schelling:
PoC with future PRs using the input attributes that are passed through.
Just taking every sample here as input for a stamp with width and opacity mapped to pressure.

https://github.com/user-attachments/assets/ca60e61a-d411-49eb-8ef2-b4b26ab0646e

desktop/src/app.rs
```diff
@@ -33,8 +35,8 @@ pub(crate) struct App {
 	window_maximized: bool,
 	window_fullscreen: bool,
 	window_pending_drag: bool,
-	pointer_position: PhysicalPosition<f64>,
 	pointer_lock_position: Option<PhysicalPosition<f64>>,
```
0HyperCube: Should the `pointer_lock_position` not also be part of `InputState`?

editor/src/messages/input_mapper/utility_types/pointer.rs
```diff
@@ -50,47 +50,70 @@ impl ScrollDelta {
 	}
 }
 
-// TODO: Document the difference between this and EditorMouseState
+// TODO: Document the difference between this and EditorPointerState
```
0HyperCube, line 53: Perhaps this todo comment could be addressed instead of just updated? It seems like the only difference is one has `ViewportPosition` and one has `EditorPosition`.

I think that it should be reasonably easy to remove the `PointerState` entirely and just use the `EditorPointerState` (computing the other position using the viewport data which is usually available). Thoughts?

0HyperCube (approved):
Looks good; mostly just renaming some things.

editor/src/messages/input_mapper/utility_types/pointer.rs
timon-schelling: I would prefer to do that in a followup but yes we should do that

timon-schelling:
!build desktop ([Run ID 31918893822](https://github.com/GraphiteEditor/Graphite/actions/runs/31918893822))

TrueDoctor (approved):

PR #4406, opened 2026-08-04
Add RFC for the attribute system

TrueDoctor:

### DIFF
--- a/node-graph/rfcs/attribute-model.md
+++ b/node-graph/rfcs/attribute-model.md
@@ -7,40 +7,50 @@ declare their attribute reads and writes in their signatures, and the
 compiler resolves every access to a byte offset during wiring, so there is
 no name lookup at runtime. Storage and batch results are per-attribute
 columns. The contiguous record only exists as a per-lane view, assembled
-into buffers the compiler assigns. All of the machinery that could
+into activation frames on a per-thread stack. All of the machinery that could
 corrupt a layout is generated code, so getting it wrong is a type error or
 a graph compile error rather than undefined behavior.
 
 # Motivation
 
-Attributes currently exist as string-keyed pairs of boxed trait objects
-carried inside `List<T>`:
+Attributes currently exist as a string-keyed store of type-erased
+columns beside the elements:
 
 ```rs
 pub struct List<T> {
 	element: Vec<T>,
-	attributes: Vec<(String, Box<dyn AnyAttributeValue>)>,
+	attributes: Attributes,
+}
+
+struct Attributes {
+	attributes: Vec<(String, Box<dyn AnyAttribute>)>,
+	len: usize,
 }
 ```
 
-Every access does a string comparison and a downcast, every value is
-boxed, and merging eagerly pads missing attributes with materialized
-defaults. On a ten-node chain with eight attributes over 64k items this
-costs us around 500ns per item. The design described here measures
-between 3.5 and 47ns on the same workload, depending on the execution
-mode, and the cost is mostly independent of the attribute count.
+So the storage is already columnar: one boxed column per name, each
+holding `len` values. What it is not is resolved. Every access searches
+the key list by string comparison and downcasts the column, and merging
+eagerly pads missing attributes with materialized defaults. Scalar
+access boxes per value at the item boundary, where the same store
+appears as `ItemAttributeValues(Vec<(String, Box<dyn AnyAttributeValue>)>)`.
+On a ten-node chain with eight attributes over 64k items this costs us
+around 500ns per item. The design described here measures between 3.5
+and 47ns on the same workload, depending on the execution mode, and the
+cost is mostly independent of the attribute count.
 
 There is also a cost at compile time and in the node catalog. Because
-attributes ride inside `List<T>`, a node that touches a property needs
-per-type traits (`MultiplyAlpha` and kin) and an implementations list
-enumerating every carrier type. Each row monomorphizes, adding a new
+attributes ride inside `List<T>`, a node that touches a property carries
+an implementations list enumerating every carrier type: the opacity and
+blend-mode nodes each name seven. Each row monomorphizes, adding a new
 carrier type means editing every one of these lists, and the duplicated
-instantiations show up in the build size. The blending nodes also carry
-a TODO ("find a way to make this apply once to the list's parent rather
-than applying to each item") that the current representation cannot
-express at all: opacity on a group and opacity on each member composite
-differently once members overlap, so the difference is semantic, and
-there is currently nowhere to put it.
+instantiations show up in the build size, measured at 6.16 MB and about
+4% of frame rate when the rank work landed. An escape hatch exists for
+the narrow case of reading an attribute without the element (`ListDyn`,
+which erases the carrier), but it does not extend to nodes that write.
+Separately, opacity on a group and opacity on each member composite
+differently once members overlap, so the two are semantically distinct
+and the representation has nowhere to record which one a node meant.
 
 The requirements, briefly. Attributes are named with strings and work
 for all types, and users can author read/write nodes with custom names.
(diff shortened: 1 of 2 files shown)


timon-schelling (approved):

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
+# Summary
+
+Give every item flowing through the graph a set of named, typed attributes
+next to its primary `element` value. Attributes are stored as a packed
```
0HyperCube, line 4: I'm not sure it is necessary or desirable to have a primary element. For example you might wish to store a list containing `x: f64` and `y: f64` as attributes. It would not make sense to set one of these as the primary « element » ?

I think your approach would work just as well with the primary element removed. It would perhaps even simplify it somewhat.

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+	(_, attr): ((), Attr<Custom<A, Y>>),
+) -> (T, Attr<Custom<A, Y>>) {
+	(element, attr)
+}
+```
+
+A unit value component means the edge exists and carries only its
+attributes (`_: ()` still means no edge at all). The name enters the
+graph at a source node holding the constant text input, whose output
+type is filled at graph compile time, where user-supplied names join
```
0HyperCube, line 124: A user may currently choose to give a dynamic name (such as the current animation time as a string). Would this be forbidden under this proposal?

A node writing an attribute who's name is the current time

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+attribute then arrives on its own wire, as an input whose element is
+`()`:
+
+```rs
+/// Attaches the attribute to the content.
+#[node_macro::node(category("Attributes"))]
+fn set_attribute<T, A, Y>(
+	_: impl Ctx,
+	element: T,
+	(_, attr): ((), Attr<Custom<A, Y>>),
```
0HyperCube, line 115: What is `A` and `Y`? What is `Custom`?

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+
+Where a node sits in the chain decides which nesting level it affects.
+Applying the opacity node to a shape and then repeating it gives every
+copy its own opacity. Repeating first and then applying opacity sets one
+value on the whole group, which composites differently where copies
+overlap. The node's code is identical in both cases. Reads and writes
+bind to the top level of the wire at the node's position in the chain,
+and Repeat pushed a level in one of the two arrangements. Reaching an
+inner level from outside is an explicit map/enter construct, so "set on
+the parent" and "map over the children" are visibly different graphs.
```
0HyperCube: It is surely useful to allow mapping over children after constructing the list? For example, I might have a « Trace Raster » node which converts a raster image into several shapes. I would then like to give each shape its own opacity.

How could this be accomplished within this system?

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+	emit(element, Attr(DAffine2::from_translation(offset) * *transform))
+}
+```
+
+The body is one lane of the declared list: the kernel reads its own copy
+index and produces that copy's values. `level_extent = count` declares
+the size of the new level, and the compiler derives the structural parts
+from that one declaration: the multiplication with the carrier's extent,
+and the index decomposition that routes an output lane to the right copy
+and content item. The `emit(...)` tail marks the one-lane form and
```
0HyperCube: I find this syntax quite hard to read and reason about. What is the motivation over just explicitly writing a `for index in 0..count` ?

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+otherwise the pushed value fills the items that never wrote the name
+and the inner value wins where they did, resolved from the write sets
+at graph compile time), and the merged list starts with an empty
+top level. If the user wants to keep the groups as groups, they wrap
+explicitly instead.
+
+## Selecting
+
+Switch takes two lazy inputs and returns one of them, for any carrier,
+without an implementations list:
```
0HyperCube, line 280: Quite neat.

node-graph/rfcs/attribute-model.md
```diff
@@ -0,0 +1,675 @@
...
+  push-down's fallback and the Data panel, which presents the layout:
+  column presence is a pure function of the graph, stable across frames
+  and across the branches a selector takes.
+- Writes are unconditional: presence never depends on a value, so a
+  conditionally relevant attribute is written at its default, and a
+  runtime `Option` around a value buys nothing (`None` could only mean
+  the default). A name that wants a distinguished unset declares an
+  `Option` value type on its marker.
+
+Fields are `Copy`, and larger payloads go behind a pointer-sized field.
```
0HyperCube, line 378: What would own the larger payloads? Is it necessary to clone them when cloning a table?

node-graph/rfcs/attribute-model.md
timon-schelling: I agree with this somewhat, the asymmetry is definitely weird, I think mainly inherited from the "Old" Table<T> type.
If we go all in on "dynamic" types I think we could get inspiration from [NuShell's
structured data types](https://www.nushell.sh/book/types_of_data.html#structured-data-types).

node-graph/rfcs/attribute-model.md
TrueDoctor: Yes, this would require the use of a special dynamic attribute whisch is essentially a hashmap. We might hide the difference from the user, but this dynamic path would be a lot slower

node-graph/rfcs/attribute-model.md
TrueDoctor: I this case A is an identifier to associte input with output, so bsically just some token. And Y would be the type (though this syntax needs some more work it's currently more of a functional draft, but not pretty)

node-graph/rfcs/attribute-model.md
TrueDoctor: Because the loop is never materialized in the node itself, we instead poll the node 100 times each with a different index. This is basically the rust iterator approach, an explicit for loop would require an intermediate collect

node-graph/rfcs/attribute-model.md
TrueDoctor: The frame arena allocator so the allocations are cheap and you can share references to them as long as the graph is running

node-graph/rfcs/attribute-model.md
TrueDoctor: There are still some differences between the main type and the attributes (main type does not need to be `Copy`, can have a non trivial drop impl etc.) so there is at least some technical reason. The attributes are also ignored during the type analysis so they don't drive the monomorphization / implementation selection. On a ux side, I think the primary types would be needed to color the node graph wires right?
I'd suggest to use `()` as the carrier type if you want to basically have only attributes, but I don't have that strong of an opinion design wise

node-graph/rfcs/attribute-model.md
TrueDoctor: This all builds on the [async execution / iterator design](https://github.com/GraphiteEditor/Graphite/blob/pre-rank-poly/node-graph/rfcs/async-execution-model.md)

node-graph/rfcs/attribute-model.md
0HyperCube: > The attributes are also ignored during the type analysis

I'm very confused by this statement. You have said that during the graphene compilation, "each wire's layout is constructed from its upstream write set." If the types of the attributes are unknown, it would surely not be possible to do such an analysis. 

> I think the primary types would be needed to colour the node graph wires

I'm not convinced the colouring is very helpful. I always end up looking at the tooltip anyway. There would be no obstacle in having the editor decide based on some other mechanism what colour to make a wire (e.g. are any of the attributes containing raster data).


If I have a `List<Raster>` and attach an `f64` attribute to each item (e.g. average brightness), then I still have a `List<Raster>`. This is difficult if I want to pass it into a node that takes `List<f64>` (for example sorting). I guess you could add a node that would swap the element and one attribute, eating the quadratic number of implementations based on the number of types available?

node-graph/rfcs/attribute-model.md
0HyperCube: I'm not sure if it is ever useful to have a dynamically named attribute. You could ban it to reduce the complexity of the proposal (it seems already complex enough).

TrueDoctor: Yeah, I agree I've already deffered it mentally / in the implementation this is more of an eventual escape hatch which can be added later on

node-graph/rfcs/attribute-model.md
TrueDoctor: > If I have a List<Raster> and attach an f64 attribute to each item (e.g. average brightness), then I still have a List<Raster>. This is difficult if I want to pass it into a node that takes List<f64> (for example sorting). I guess you could add a node that would swap the element and one attribute, eating the quadratic number of implementations based on the number of types available?

I think in this case you could use a read attribute node (with implementations for `List<f64>`) which can take anything with an `List<f64>` attribute and return just `List<f64>` so you need one extra node, but no combinatorial explosion. (If we only read an attribute and never read the main type, we don't need monomorphizations for the main type at all)

PR #4417, opened 2026-08-07
Fix Text tool resize locking both max width and max height on a single-edge drag

### LINKED ISSUE #4396: Resizing a text box by one edge handle incorrectly locks both Max Width and Max Height, silently clipping text
Dragging a single edge handle to resize a text layer's bounding box incorrectly enables and locks both the Max Width and Max Height constraints together, even when only one axis was actually resized. Since Max Height causes any text beyond it to not be drawn, this can silently clip text that should still be auto-growing.

**Steps to reproduce:**
1. Select the Text tool and click (don't drag) to create point text, then type several lines — this creates auto-sized text with neither Max Width nor Max Height enabled
2. Switch to resizing and drag only the right-edge handle of the bounding box to set a width (so the text wraps)
3. Expected: only Max Width becomes active; height keeps auto-growing to fit the rewrapped text
4. Actual: Max Height also gets switched on and frozen to whatever the box's height happened to be before the rewrap — any lines that end up pushed past that frozen height are silently not drawn

**Cause:**
In `editor/src/messages/tool/tool_messages/text_tool.rs`, the `ResizingBounds` → `PointerMove` handler (around lines 851-920) unconditionally sets both `HasMaxWidthInput`/`MaxWidthInput` and `HasMaxHeightInput`/`MaxHeightInput`, regardless of which edge is being dragged. There's actually a TODO comment already sitting right above this code:

    // TODO: Don't set both max_width and max_height to true at the same time, only do one based on which edge is being dragged (or both if a corner is being dragged)

The information needed to fix it is already available and just unused: `SelectedEdges { top, bottom, left, right }` in `transformation_cage.rs` records exactly which edge(s) are being dragged, and `new_size()` already leaves `size.y` untouched (equal to the pre-drag height) when only a left/right edge is dragged — that stale value is what's getting wrongly frozen into `MaxHeightInput`.

**Suggested fix:**
Gate the two pairs of `SetInput` calls on which axis was actually touched, per the existing TODO:

    let (touches_width, touches_height) = (movement.left || movement.right, movement.top || movement.bottom);

    if touches_width {
        // set HasMaxWidthInput / MaxWidthInput
    }
    if touches_height {
        // set HasMaxHeightInput / MaxHeightInput
    }

A corner-handle drag naturally sets both pairs of edges, so both branches fire together there, matching the "(or both if a corner is being dragged)" note in the TODO.

GuTS805:
Fixes #4396

## Problem

While resizing a text box with the Text tool, dragging a **single edge** unconditionally set *both* `max_width` and `max_height` on the Text node.

So dragging just the right edge to set a width constraint also froze the height at whatever it happened to be before the drag. As the text rewrapped to more lines, everything past that frozen height was silently clipped — the user never asked for a height constraint at all.

This was a known gap; the code carried a TODO for exactly this case:

```rust
// TODO: Don't set both max_width and max_height to true at the same time, only do one
// based on which edge is being dragged (or both if a corner is being dragged)
```

## Fix

Gate the two pairs of `SetInput` calls on which edges the drag actually touched, using the `SelectedEdges` state that the resize already tracks:

```rust
let (touches_width, touches_height) = (movement.left || movement.right, movement.top || movement.bottom);
```

- Left/right edge drag → only `has_max_width` + `max_width`
- Top/bottom edge drag → only `has_max_height` + `max_height`
- Corner drag → both, as before

The untouched axis keeps whatever it had (usually unconstrained), so text continues to auto-grow in that direction.

`editor/src/messages/tool/tool_messages/text_tool.rs` — 1 file, resolves the TODO.

## Testing

Built successfully and manually verified in the editor:

- Dragging the right edge of a text box now constrains width only; the box grows downward as text rewraps instead of clipping. ✅
- Dragging the bottom edge constrains height only; width is untouched. ✅
- Dragging a corner still constrains both axes (unchanged behavior). ✅
- Re-dragging the other edge afterwards adds the second constraint as expected. ✅

### DIFF
--- a/editor/src/messages/tool/tool_messages/text_tool.rs
+++ b/editor/src/messages/tool/tool_messages/text_tool.rs
@@ -884,23 +884,29 @@ impl Fsm for TextToolFsmState {
 					// Find the translation necessary from the original position in viewport space
 					let translation_viewport = bounds.original_bound_transform.transform_vector2(translation_bounds_space);
 
-					// TODO: Don't set both max_width and max_height to true at the same time, only do one based on which edge is being dragged (or both if a corner is being dragged)
-					responses.add(NodeGraphMessage::SetInput {
-						input_connector: InputConnector::node(node_id, graphene_std::text::text::HasMaxWidthInput),
-						input: NodeInput::value(TaggedValue::Bool(true), false),
-					});
-					responses.add(NodeGraphMessage::SetInput {
-						input_connector: InputConnector::node(node_id, graphene_std::text::text::MaxWidthInput),
-						input: NodeInput::value(TaggedValue::F64(size_layer.x), false),
-					});
-					responses.add(NodeGraphMessage::SetInput {
-						input_connector: InputConnector::node(node_id, graphene_std::text::text::HasMaxHeightInput),
-						input: NodeInput::value(TaggedValue::Bool(true), false),
-					});
-					responses.add(NodeGraphMessage::SetInput {
-						input_connector: InputConnector::node(node_id, graphene_std::text::text::MaxHeightInput),
-						input: NodeInput::value(TaggedValue::F64(size_layer.y), false),
-					});
+					// Only set the max width/height for the axis (axes) whose edge(s) are actually being dragged, so a single-edge drag doesn't lock the other axis too
+					let (touches_width, touches_height) = (movement.left || movement.right, movement.top || movement.bottom);
+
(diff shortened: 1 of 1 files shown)


Keavon:
This has the appearance of being LLM-authored, which violates our contributor AI policy.

GuTS805:
ohk, sorry i improvised the description using LLM my bad

PR #4426, opened 2026-08-10
Only set max width/height for the axis being dragged in Text tool resize

### LINKED ISSUE #4396: Resizing a text box by one edge handle incorrectly locks both Max Width and Max Height, silently clipping text

GuTS805:
Fixes #4396

The problem
So while resizing a text box with the Text tool, if you drag just one edge it was setting both max_width and max_height on the Text node, no matter which edge you grabbed.

Which means dragging only the right edge to make the box narrower also locked the height to whatever it was at that moment. Then when the text rewrapped onto more lines, anything past that height just got clipped. And the user never asked for a height constraint in the first place, so it looks like the text randomly dissapears.

Turns out this was already known, there's a TODO sitting right there for this exact thing:

// TODO: Don't set both max_width and max_height to true at the same time, only do one
// based on which edge is being dragged (or both if a corner is being dragged)
The fix
I just gated the two pairs of SetInput calls on which edges the drag actually touched. The resize already keeps track of this in SelectedEdges, so it was mostly a matter of using what was already there:

let (touches_width, touches_height) = (movement.left || movement.right, movement.top || movement.bottom);
left/right edge → only has_max_width + max_width
top/bottom edge → only has_max_height + max_height
corner → both, same as before
The axis you didn't touch keeps whatever it had (normally unconstrained), so the text can still auto-grow that way.

One file: editor/src/messages/tool/tool_messages/text_tool.rs. TODO is gone now.

Testing
Builds fine, and I checked all of these by hand in the editor:

Dragging the right edge only constrains the width now — box grows downwards as the text rewraps instead of clipping ✅
Dragging the bottom edge only constrains height, width is left alone ✅
Corners still constrain both, no change there ✅
If you drag one edge and then the other afterwards, you get both constraints like you'd expect ✅


0HyperCube (approved):
Works well; thanks for the fix.

GuTS805:
Thanks for the review..

0HyperCube:
Consider merging @timon-schelling.

PR #4436, opened 2026-08-15
Fix async command sequence failure propagation

tomhargrave:
## Summary

- return the worker result from `Sequence::wait()`
- propagate child spawn failures, nonzero exits, and worker panics to callers
- make WASM package generation fail when a generation step fails
- add focused sequence success and failure tests

This prevents a failed `wasm-bindgen` or `wasm-opt` step from being reported as a successful WASM build.

## Testing

- `cargo fmt --package cargo-run -- --check`
- `cargo test -p cargo-run`


Keavon:
This appears to be AI authored, in violation of our policy.

PR #4443, opened 2026-08-17
Add time since pointerdown to input event handling

timon-schelling:
Will later be used for brushes that change over time


TrueDoctor:
Would it potentially make more sense to store a unix epoch here? or animation time?

timon-schelling:
> Would it potentially make more sense to store a unix epoch here? or animation time?

Anything that can capture the notion of this sample was captured x seconds after the stroke began.

This small change is s separate PR for this exact discussion.

TrueDoctor (approved):

PR #4462, opened 2026-08-19
Fix obvious spelling issues

oesayan:
Fixed obvious spelling issues in the error log messages and panic message.


Keavon:
Hi, thanks for your efforts, but we don't usually accept PRs of this nature.

PR #4464, opened 2026-08-19
Reintroduce improved logic for mesh interior region fill determination

Keavon:
As removed in #4463, now reintroduced by acting based on the topology of the mesh geometry itself instead of an inconsistently used boolean flag taking the form of the RegionDomain data.


Keavon:
Performance regression is reasonable in that actual work is now occurring that was previously skipped at the cost of correctness about path winding, which only happened to work in practice in some (but not all) cases.

PR #4467, opened 2026-08-21
Introduce the new brush stroke data types

timon-schelling:


Keavon (approved):
Approved, pending the discussed refactor that I'll do to unify it with the Freehand tool in the upcoming weeks.

PR #4468, opened 2026-08-21
Add caching infrastructure for the GPU brush nodes

timon-schelling:


node-graph/graph-craft/src/document/value.rs
```diff
@@ -300,13 +304,14 @@ macro_rules! tagged_value {
 					// MANUAL VARIANTS
 					// ===============
 					Self::None => concrete!(()),
-					Self::TypeDefault(td) => td.clone(),
+				Self::TypeDefault(td) => td.clone(),
```
Keavon: Whitespace issue

Keavon (approved):

PR #4369, opened 2026-07-24
Add the Heart's radius gizmo via the generic registry system

Ayush2k02:
EVERYTHING FROM WEEK 1 TO WEEK 4

https://github.com/user-attachments/assets/b14f07ad-a6e0-478d-afc2-2fdef6ad3486

PR #4365, opened 2026-07-24
Add Heart drawing mode to the Shape tool with gizmo registration

Ayush2k02:
Add Heart drawing mode to the Shape tool with gizmo registration

https://github.com/user-attachments/assets/8455b528-24ba-4071-8459-ab33dfee3f6f

PR #4471, opened 2026-08-23
Migrate all shape gizmos to the registry-driven generic system

Ayush2k02:
Migrates all parametric shape gizmos from hand-written handlers to the registry-driven generic system, and adds the Heart shape.

Supersedes [#4365](https://github.com/GraphiteEditor/Graphite/issues/4365)–#4369 — same work, rebased onto master and carried further. Happy to close those, or split this back up if you'd rather review in parts.

What changes

- gizmo_registry.rs — which parameters get gizmos, as data
- generic_gizmos/ — the shared mechanics
- gizmo_behaviors.rs — the shape-specific half, and the only place node geometry appears
- shape_gizmos deleted; ShapeGizmoHandlers down from 7 variants across 8 match arms to 2
- 7 shapes, 14 gizmos. Circle and Heart need no behaviour at all — five registry lines each
- Hooks documented in gizmos/README.md

Scope

- Heart is the only new generator node
- Polygon's radius handles are restored — master had them, an earlier commit in this stack dropped them
- Not carried over: red snap-alignment guides on star/arc (both still snap), and three declarations that were never reachable (spiral's two radii, grid spacing)


Ayush2k02:
Superseded by newer pull requests carrying this work rebased onto current master.

PR #4499, opened 2026-09-03
Clipboard: Cross-application copy-paste support

### LINKED ISSUE #2373: Graphite should be able to copy/paste vector and raster content between other apps
*This is a good first issue but not for beginners.*

We should be able to copy vector and raster content, including hierarchies of layers and groups, between creative programs. Both into Graphite and out of Graphite. This shouldn't break existing copy-pasting within Graphite, which needs to have actual graph information rather than just vector or bitmap layer info.

This will require research into how the clipboard APIs let us encode different types of data, and if perhaps we need a degraded experience on web due to API limitations compared to a full separate implementation for the Tuari desktop version. Testing other creative apps and using a utility for inspecting clipboard contents at a technical level will be a useful way to approach this. It's also worth checking if this differs between OSs.

VimYoung:
# Description

This PR aims to resolve #2373. Solution's approach has been thoroughly discussed in discord's development channel. The PR will fix this issue by introducing copy of both text and svg+xml  mime types when copying a selection to the clipboard. Essentially making it possible for other apps to take the svg representation while graphite picks up the internal json representation for copy pasting.

## Notes
1. As mentioned by hypercube, `image/svg+xml` is not supported by firefox and safari Stable so that is handled.
2. The intersecting by desktop still doesn't copy the svg and simply ignores it.

Following example showcase the same copy paste not working across apps (`krita` and `inkscape` shown as the other apps) in current build vs it working in this branch while still being compatible across graphite tabs.

Before(Only across graphite tabs, not across apps):

https://github.com/user-attachments/assets/6b35d6dd-25aa-4162-b369-81005502e368

After(Across tabs and other apps):

https://github.com/user-attachments/assets/64b181a1-ee4a-4f53-8bb3-9fd9e4947402


desktop/wrapper/src/intercept_frontend_message.rs
```diff
@@ -119,6 +119,9 @@ pub(super) fn intercept_frontend_message(dispatcher: &mut DesktopWrapperMessageD
 		FrontendMessage::TriggerClipboardWrite { content } => {
 			dispatcher.respond(DesktopFrontendMessage::ClipboardWrite { content });
 		}
+		FrontendMessage::TriggerClipboardSvgWrite { graphite_json, .. } => {
+			dispatcher.respond(DesktopFrontendMessage::ClipboardWrite { content: graphite_json });
```
0HyperCube, line 121: It seems the current desktop windowing library can only support putting strings onto the clipboard. There is some work by timon-schelling and myself to add full support for other data types directly to winit https://github.com/rust-windowing/winit/pull/4658.

It should be relatively easy to add this when completed. Please leave a todo here.

editor/src/messages/clipboard/clipboard_message_handler.rs
```diff
@@ -80,19 +80,22 @@ impl MessageHandler<ClipboardMessage, ClipboardMessageContext<'_>> for Clipboard
 				}
 			}
 			ClipboardMessage::Write { content } => {
-				let text = match content {
+				match content {
 					ClipboardContent::Svg(_) => {
 						log::error!("SVG copying is not yet supported");
-						return;
+						// Need to fix this.
```
0HyperCube: Are you going to fix this?

I guess you could also remove `ClipboardContent::Svg` since it is unused in this implementation of SVG copying.

editor/src/messages/frontend/frontend_message.rs
```diff
@@ -153,6 +153,10 @@ pub enum FrontendMessage {
 	TriggerClipboardWrite {
 		content: String,
 	},
+	TriggerClipboardSvgWrite {
```
0HyperCube: It is perhaps unclear that this contains an SVG and some nodes JSON?

editor/src/node_graph_executor.rs
```diff
@@ -52,6 +52,7 @@ pub enum NodeGraphUpdate {
 	CompilationResponse(CompilationResponse),
 	EyedropperPreview(Raster<CPU>),
 	NodeGraphUpdateMessage(NodeGraphUpdateMessage),
+	SvgTextCopyClipboard(String, String),
```
0HyperCube: Consider putting into struct like the other variants (since it isn't clear what the strings represent).

editor/src/messages/frontend/frontend_message.rs
```diff
@@ -153,6 +153,10 @@ pub enum FrontendMessage {
...
+		svg_string: String,
+		graphite_json: String,
```
0HyperCube: Perhaps consolidate these into one message with optional svg and text values.

editor/src/messages/clipboard/clipboard_message_handler.rs
```diff
@@ -80,19 +80,22 @@ impl MessageHandler<ClipboardMessage, ClipboardMessageContext<'_>> for Clipboard
...
 						log::error!("Image copying is not yet supported");
-						return;
 					}
-					ClipboardContent::Graphite(graphite) => format!("{CLIPBOARD_PREFIX}{graphite}"),
-					ClipboardContent::Text(text) => text,
-				};
-				responses.add(FrontendMessage::TriggerClipboardWrite { content: text });
+					ClipboardContent::Graphite(graphite) => {
+						let graphite_json = format!("{CLIPBOARD_PREFIX}{graphite}");
+						responses.add(PortfolioMessage::RequestSvgTextCopy { graphite_json });
```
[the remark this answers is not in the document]
0HyperCube: Nearly correct. The messages are mistakenly discarded by the `EditorTestUtils::eval_graph`. A simple fix would be to preserve them instead:

[fix_test.patch](https://github.com/user-attachments/files/31889227/fix_test.patch)

editor/src/node_graph_executor.rs
```diff
@@ -812,6 +816,12 @@ impl NodeGraphExecutor {
 
 		Ok(())
 	}
+
+	pub fn copy_svg_clipboard(&self, graphite_json: String, selected_nodes: Vec<NodeId>) {
+		self.runtime_io
+			.send(GraphRuntimeRequest::CopySvgTextClipboard(graphite_json, selected_nodes))
+			.expect("Failed to send runtime request");
```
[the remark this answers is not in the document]
0HyperCube: The send fails only if the node graph executor disappears. In this case it isn't very useful to continue.

Avoiding a panic here is also completely useless given that `queue_execution` will panic.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
 					});
 					return texture;
 				}
+				GraphRuntimeRequest::CopySvgTextClipboard(text_string_clipboard, selected_node_ids) => {
+					let mut combined_graphics = List::<Graphic>::new();
+
+					for monitor_node_path in &self.monitor_nodes {
+						// Skip inspect monitor node if active
+						if self.inspect_state.as_ref().is_some_and(|state| monitor_node_path.last().copied() == Some(state.monitor_node)) {
```
0HyperCube: Please make into function e.g. `self.is_inspect_monitor_node(&monitor_node_path)`.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
...
+						};
+
+						if selected_node_ids.contains(&parent_network_node_id) {
+							// Introspect using the full monitor node path
+							if let Ok(introspected_data) = self.executor.introspect(monitor_node_path) {
+								if let Some(io) = introspected_data.downcast_ref::<IORecord<Context, List<Graphic>>>() {
+									combined_graphics.extend(io.output.clone());
+								} else if let Some(io) = introspected_data.downcast_ref::<IORecord<Context, Item<Graphic>>>() {
+									combined_graphics.push(io.output.clone());
+								}
```
0HyperCube: There is no case where this should be called?

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
...
+							continue;
+						}
+
+						let Some(parent_network_node_id) = monitor_node_path.len().checked_sub(2).and_then(|index| monitor_node_path.get(index)).copied() else {
+							continue;
+						};
+
+						if selected_node_ids.contains(&parent_network_node_id) {
+							// Introspect using the full monitor node path
+							if let Ok(introspected_data) = self.executor.introspect(monitor_node_path) {
```
0HyperCube: Consider adding a warning if no type is matched.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
 					});
 					return texture;
 				}
+				GraphRuntimeRequest::CopySvgTextClipboard(text_string_clipboard, selected_node_ids) => {
+					let mut combined_graphics = List::<Graphic>::new();
+
+					for monitor_node_path in &self.monitor_nodes {
```
0HyperCube: Could be function e.g. `self.collect_graphics(&selected_node_ids)`

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
...
+					let raw_bounds = match bounds {
+						RenderBoundingBox::Rectangle(bounds) if (bounds[1] - bounds[0]) != DVec2::ZERO => bounds,
+						_ => [DVec2::ZERO, DVec2::ONE],
+					};
+
+					let footprint = Footprint {
+						transform: DAffine2::from_translation(DVec2::new(raw_bounds[0].x, raw_bounds[0].y)),
+						resolution: UVec2::new((raw_bounds[1].x - raw_bounds[0].x).abs().ceil() as u32, (raw_bounds[1].y - raw_bounds[0].y).abs().ceil() as u32).max(UVec2::ONE),
+						quality: RenderQuality::Full,
+					};
```
0HyperCube: Consider a function such as `Footprint::from_bounds(bounds, RenderQuality::Full)`.

It is also not clear here why these are called `raw_bounds` (presumably copypasta from the thumbnail rendering where the bounds are adjusted?)

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
...
+
+					let render_params = RenderParams {
+						footprint,
+						thumbnail: false,
```
0HyperCube: False is already default, you can omit this line if if you wish.

frontend/src/managers/clipboard.ts
```diff
@@ -23,13 +23,28 @@ export function createClipboardManager(subscriptions: SubscriptionsRouter, edito
 	subscriptions.subscribeFrontendMessage("TriggerSelectionWrite", async (data) => {
 		insertAtCaret(data.content);
 	});
+
+	subscriptions.subscribeFrontendMessage("TriggerClipboardSvgWrite", (data) => {
+		// Adopted from https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem#browser_compatibility
+		if (ClipboardItem.supports("image/svg+xml")) {
```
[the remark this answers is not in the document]
0HyperCube: I do not care about supporting deprecated browser versions.

0HyperCube (commented):
Works well; thanks for your work so far.

frontend/src/managers/clipboard.ts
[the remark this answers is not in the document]
VimYoung: I guess this can be resolved then.

timon-schelling:
I would like to wait with this until winit update is merged. And we might want to make this API lazy.

desktop/wrapper/src/intercept_frontend_message.rs
VimYoung: I have added a todo for now, as the mentioned PR will get merged, I will do the desktop relevant changes here (as timon said) if required.

editor/src/messages/clipboard/clipboard_message_handler.rs
VimYoung: Yes, I have removed the SVG variant as it was no longer required, I was hesitant to remove it before this PR getting reviewed. I have removed it now.

editor/src/messages/clipboard/clipboard_message_handler.rs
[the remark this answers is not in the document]
VimYoung: Thanks, I have applied the patch, I will wait to see if the test passes now.

editor/src/messages/frontend/frontend_message.rs
VimYoung: Renamed it from `TriggerClipboardSvgWrite` to `TriggerClipboardSvgAndJsonWrite`, It is a bit lengthy but explains itself better ig.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -193,7 +200,7 @@ impl NodeRuntime {
 			eyedropper.render_config.pointer = execution.render_config.pointer;
 		}
 
-		let requests = [preferences, graph, eyedropper, execution].into_iter().flatten();
+		let requests = [preferences, graph, eyedropper, execution, svg_clipboard].into_iter().flatten();
```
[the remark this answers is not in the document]
VimYoung: This is fixed by the fixed_test.patch shared by hypercube.

editor/src/node_graph_executor.rs
VimYoung: Changed it's definition from tuple variant to struct enum variant.

editor/src/node_graph_executor.rs
```diff
@@ -466,6 +467,9 @@ impl NodeGraphExecutor {
 					responses.add(EyedropperToolMessage::PreviewImage { data, width, height });
 				}
 				NodeGraphUpdate::NodeGraphUpdateMessage(_) => {}
+				NodeGraphUpdate::SvgTextCopyClipboard { svg_string, graphite_json } => {
+					responses.add(FrontendMessage::TriggerClipboardSvgAndJsonWrite { svg_string, graphite_json });
```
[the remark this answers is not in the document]
VimYoung: Forgot to fix `clipboard.ts` fixed the name change there also.

VimYoung:
Hey @timon-schelling , we can wait for the API to be merged and then do the changes, what do you mean for the API to be lazy here?
Also can you assign the issue to me as this PR is opened and in review?

timon-schelling:
> Hey @timon-schelling , we can wait for the API to be merged and then do the changes, what do you mean for the API to be lazy here?

See the top level API definitions in https://github.com/rust-windowing/winit/pull/4658

Both the DnD and Clipboard APIs are lazy in winit, meaning data only needs to be constructed once the other application accepts (based on data type).

I think we should model it in a similar way.
I'm not completely decided on how to model this with messages though.


> Also can you assign the issue to me as this PR is opened and in review?

We usually do that when something is merged.

editor/src/messages/frontend/frontend_message.rs
VimYoung: This is fixed in commit c54b027e .

editor/src/node_graph_executor/runtime.rs
VimYoung: done in f85130e2e.

editor/src/node_graph_executor/runtime.rs
VimYoung: done in f85130e2e.

editor/src/node_graph_executor/runtime.rs
VimYoung: I added the warning in the helper function created in f85130e2e.

editor/src/node_graph_executor/runtime.rs
VimYoung: Right, that was my mistake of copy pasting without thinking, fixed in f85130e2e. selected_node_ids only contain layer nodes so it is dead code. Moreover, active inspect nodes are already skipped.

editor/src/node_graph_executor/runtime.rs
VimYoung: Done in f85130e2e, also fixed it in render_thumbnail to prevent code duplication.

editor/src/node_graph_executor/runtime.rs
VimYoung: Fixed in f85130e2e

editor/src/messages/clipboard/clipboard_message_handler.rs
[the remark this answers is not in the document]
VimYoung: `cargo test --all-features` passes all tests, so closing this.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,59 @@ impl NodeRuntime {
 					});
 					return texture;
 				}
+				GraphRuntimeRequest::CopySvgTextClipboard(text_string_clipboard, selected_node_ids) => {
+					let mut combined_graphics = List::<Graphic>::new();
```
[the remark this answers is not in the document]
VimYoung: I don't think one can copy the artboard as is, even if they copy every content in the artboard, the copying type would still be graphic (of all selected nodes/layers) ig.

node-graph/libraries/core-types/src/transform.rs
```diff
@@ -189,6 +189,14 @@ impl Footprint {
 		quality: RenderQuality::Full,
 	};
 
+	pub fn from_bounds(bounds: [DVec2; 2], quality: RenderQuality) -> Self {
+		Footprint {
+			transform: DAffine2::from_translation(DVec2::new(bounds[0].x, bounds[0].y)),
+			resolution: UVec2::new((bounds[1].x - bounds[0].x).abs().ceil() as u32, (bounds[1].y - bounds[0].y).abs().ceil() as u32).max(UVec2::ONE),
```
0HyperCube, line 195:
```suggestion
			transform: DAffine2::from_translation(bounds[0].min(bounds[1])),
			resolution: (bounds[1] - bounds[0]).abs().ceil().as_uvec2().max(UVec2::ONE),
```

editor/src/node_graph_executor/runtime.rs
```diff
@@ -529,6 +550,36 @@ impl NodeRuntime {
...
+
+			if selected_node_ids.contains(&parent_network_node_id) {
+				// Introspect using the full monitor node path
+				if let Ok(introspected_data) = self.executor.introspect(monitor_node_path)
+					&& let Some(io) = introspected_data.downcast_ref::<IORecord<Context, List<Graphic>>>()
+				{
+					combined_graphics.extend(io.output.clone());
+				} else {
+					warn!("No graphic type is matched while extracting svg");
+				}
```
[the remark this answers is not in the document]
0HyperCube: This doesn't happen as it will always be made into a group.

editor/src/messages/clipboard/clipboard_message_handler.rs
```diff
@@ -79,21 +79,18 @@ impl MessageHandler<ClipboardMessage, ClipboardMessageContext<'_>> for Clipboard
...
+			ClipboardMessage::Write { content } => match content {
+				ClipboardContent::Image { .. } => {
+					log::error!("Image copying is not yet supported");
+				}
+				ClipboardContent::Graphite(graphite) => {
+					let graphite_json = format!("{CLIPBOARD_PREFIX}{graphite}");
+					responses.add(PortfolioMessage::RequestSvgTextCopy { graphite_json });
+				}
+				ClipboardContent::Text(graphite_json) => {
+					responses.add(FrontendMessage::TriggerClipboardSvgAndJsonWrite { svg_string: None, graphite_json });
```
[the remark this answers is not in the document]
0HyperCube: The `None` does become an `undefined` in js.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -340,6 +347,28 @@ impl NodeRuntime {
 					});
 					return texture;
 				}
+				GraphRuntimeRequest::CopySvgTextClipboard(text_string_clipboard, selected_node_ids) => {
+					let combined_graphics = self.collect_graphics(&selected_node_ids);
+
+					if combined_graphics.is_empty() {
+						self.sender.send_svg_text_clipboard(String::new(), text_string_clipboard);
+						return None;
```
0HyperCube, line 355: `continue` don't `return`.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -193,7 +200,7 @@ impl NodeRuntime {
 			eyedropper.render_config.pointer = execution.render_config.pointer;
 		}
 
-		let requests = [preferences, graph, eyedropper, execution].into_iter().flatten();
+		let requests = [preferences, graph, eyedropper, svg_clipboard, execution].into_iter().flatten();
 
 		for request in requests {
 			match request {
```
0HyperCube, line 206: I think the control flow would be much easier if these were made into functions:

```rs
match request {
	GraphRuntimeRequest::EditorPreferencesUpdate(preferences) => self.editor_preferences_update(preferences),
	GraphRuntimeRequest::GraphUpdate(graph_update) => self.graph_update(graph_update),
	...
}

File: editor/src/node_graph_executor/runtime.rs [hunk 23]
```diff
@@ -193,7 +200,7 @@ impl NodeRuntime {
 			eyedropper.render_config.pointer = execution.render_config.pointer;
 		}
 
-		let requests = [preferences, graph, eyedropper, execution].into_iter().flatten();
+		let requests = [preferences, graph, eyedropper, svg_clipboard, execution].into_iter().flatten();
```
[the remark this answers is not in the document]
0HyperCube: This is technically true however it probably isn't very relevant since the user would be copying an SVG that is not yet rendered into the viewport.

You could fix this by having a result variable instead of an early return:
```
let mut result = None;
for request in requests {
	match request {
		// ...
		GraphRuntimeRequest::ExecutionRequest(ExecutionRequest { execution_id, mut render_config, .. }) => {
			// ...
			result = Some(texture);
		}
	}
}
result

File: desktop/wrapper/src/intercept_frontend_message.rs (hunk 1 above)
Comment (0HyperCube) [reply desktop/wrapper/src/intercept_frontend_message.rs]: @VimYoung the winit clipboard support is added by #4519 (graphite now uses a winit fork). It should now be possible to implement the desktop support when you rebase.

File: editor/src/messages/clipboard/clipboard_message_handler.rs (hunk 20 above)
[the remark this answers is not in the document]
Comment (VimYoung) [reply editor/src/messages/clipboard/clipboard_message_handler.rs:91]: Yup, that's was my initial though. Closing this.

File: editor/src/messages/clipboard/clipboard_message_handler.rs [hunk 24]
```diff
@@ -79,21 +79,18 @@ impl MessageHandler<ClipboardMessage, ClipboardMessageContext<'_>> for Clipboard
...
+				}
```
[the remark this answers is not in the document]
VimYoung: I think this is an unnecessary change. I think in general we should use the default naming pickups that rust offers without having to explicitly mention the arg is both the arg and the value identifiers are the same.

node-graph/libraries/core-types/src/transform.rs
```diff
@@ -189,6 +189,14 @@ impl Footprint {
...
+			quality,
+		}
```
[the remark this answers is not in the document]
VimYoung: commit 72cbd1f resolves this,

editor/src/node_graph_executor/runtime.rs
VimYoung: fixed in 8cc582bb1.

editor/src/node_graph_executor/runtime.rs
```diff
@@ -529,6 +550,36 @@ impl NodeRuntime {
...
+			};
+
+			if selected_node_ids.contains(&parent_network_node_id) {
+				// Introspect using the full monitor node path
+				if let Ok(introspected_data) = self.executor.introspect(monitor_node_path)
+					&& let Some(io) = introspected_data.downcast_ref::<IORecord<Context, List<Graphic>>>()
+				{
+					combined_graphics.extend(io.output.clone());
+				} else {
+					warn!("No graphic type is matched while extracting svg");
```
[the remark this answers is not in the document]
VimYoung: This is not relevant as the generation of warning was prompted by the maintainer.

editor/src/node_graph_executor.rs
```diff
@@ -466,6 +467,12 @@ impl NodeGraphExecutor {
 					responses.add(EyedropperToolMessage::PreviewImage { data, width, height });
 				}
 				NodeGraphUpdate::NodeGraphUpdateMessage(_) => {}
+				NodeGraphUpdate::SvgTextCopyClipboard { svg_string, graphite_json } => {
+					responses.add(FrontendMessage::TriggerClipboardSvgAndJsonWrite {
+						svg_string: Some(svg_string),
```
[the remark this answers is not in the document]
VimYoung: I think this is also not relevant. If there are some selected nodes coming in, I think that there would be some content in the artboard that is renderable and is selected .

VimYoung:
Hey @0HyperCube, I have done the changes that you proposed after the second review, please take a look if this PR is still relevant (as there were some ongoing discussions on scrapping it it favor of something that would follow winit's new API changes).

0HyperCube:
Hi @VimYoung, thanks for your continued work on this. In my view this is the most viable approach for the moment.

The winit clipboard support with other types was added in #4519 (graphite now uses a winit fork). It should now be possible to implement the desktop support to this PR. See the [clipboard example](https://github.com/timon-schelling/winit/blob/739828c05de222d0851899e4441343eb14aaf922/winit/examples/clipboard.rs#L59-L69).

desktop/wrapper/src/intercept_frontend_message.rs
```diff
@@ -116,8 +116,8 @@ pub(super) fn intercept_frontend_message(dispatcher: &mut DesktopWrapperMessageD
 		FrontendMessage::TriggerClipboardRead => {
```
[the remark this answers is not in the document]
VimYoung: Closing as not relevant.

VimYoung:
I have pushed a base commit to propagate the `svg_string` over to `app.rs` where it is being part of the clipboard's "single source of truth" and provided if an image of extension svg is asked.
The `TypeHint` enum doesn't have an xml type variant so I thought image with svg would be the best alternative.

Also I don't have tests and a desktop version of graphite in dev to ensure if this is the case and the final copy paste is even working in desktop mode or not(I would need reviewer's guidance).

From the dnd.rs example I figured `Image` data needs to be passed as vector of bytes([ref](https://github.com/timon-schelling/winit/blob/739828c05de222d0851899e4441343eb14aaf922/winit/examples/dnd.rs#L129-L134)).

editor/src/node_graph_executor/runtime.rs
```diff
@@ -210,139 +217,169 @@ impl NodeRuntime {
...
+			inspect_result,
+		});
+		*return_texture = texture;
+	}
+
+	fn copy_svg_text_clipboard(&self, text_string_clipboard: String, selected_node_ids: Vec<NodeId>) {
+		let combined_graphics = self.collect_graphics(&selected_node_ids);
+
+		if combined_graphics.is_empty() {
+			self.sender.send_svg_text_clipboard(String::new(), text_string_clipboard);
```
0HyperCube: Consider sending `None` for the SVG rather than an empty string?

editor/src/node_graph_executor/runtime.rs
```diff
@@ -193,8 +200,8 @@ impl NodeRuntime {
 			eyedropper.render_config.pointer = execution.render_config.pointer;
 		}
 
-		let requests = [preferences, graph, eyedropper, execution].into_iter().flatten();
-
+		let requests = [preferences, graph, eyedropper, svg_clipboard, execution].into_iter().flatten();
+		let mut return_texture = None;
 		for request in requests {
 			match request {
```
0HyperCube, line 206: The control flow is much easier to reason about here now. Thanks.

0HyperCube (approved):
Thanks for adding the desktop support. It seems to be working well from my testing.

> Also I don't have […] a desktop version of graphite in dev

To run the desktop, use `cargo run -- desktop` from the `graphite/` directory. To see all options for the `cargo-run` tool, use `cargo run -- --help`.

> I don't have tests

To add an automated testing, which would also be desirable, you can add to `copy_layers_to_clipboard` by asserting that a suitable SVG is also copied. You can run this using `cargo test -p graphite-editor copy_layers_to_clipboard`.

> I figured Image data needs to be passed as vector of bytes

For PNG data which cannot necessarily be represented as an UTF8 string, bytes are required. However since the SVGs will always be UTF8 you can just use the `SendData::String` if you prefer (however both approaches work fine).

> The TypeHint enum doesn't have an xml type variant so I thought image with svg would be the best alternative.

The `TypeHint::Image { extension_hint: Some("svg") }` is the right choice since that is recognised by x11 and wayland.

editor/src/node_graph_executor/runtime.rs
VimYoung: Glad I could help !!

editor/src/node_graph_executor/runtime.rs
VimYoung: This was indeed a needed improvement, fixed in ade31cce0. Thanks for pointing it out.

VimYoung:
> To add an automated testing, which would also be desirable, you can add to copy_layers_to_clipboard by asserting that a suitable SVG is also copied.

Since the signature of this method only returns String (graphite json) , Should I change it to return a combo of both svg and json and then add a new test function using it??

But technically the svg is never used internally by graphite as it sticks to it's own internal structure , so I will just try to assert it for now.

0HyperCube:
> Since the signature of this method only returns String (graphite json) , Should I change it to return a combo of both svg and json and then add a new test function using it??

If you mean the `copy_layers_to_clipboard` utility function in the test module, then it would be reasonable to adjust to send back all copied data.

VimYoung:
I just had time to adjust the function signature of `copy_layers_to_clipboard` to send svg back. Now the clipboard tests check the sanity(being valid xml start and end) and presence (being not `None`) of SVG. @timon-schelling @TrueDoctor @0HyperCube . Please see the changes, I guess the PR is ready.

PR #4511, opened 2026-09-08
Fix snapping edge cases

### LINKED ISSUE #2451: Dragging handle back into its anchor makes its opposite handle move around randomly

devviktoria:
This is just a minimal fix for the issue without any snapping weighting.
If the snapping snaps directly to the anchor point the angle calculation is going to be wrong and the snapping could return an infinite point and in that case the snapping point will be 0,0 point. This fixes this case.
Here is the video of this behavior: 

https://github.com/user-attachments/assets/6e515a89-740d-457f-9053-c892f58e6533

Closes https://github.com/GraphiteEditor/Graphite/issues/2451

### DIFF
--- a/editor/src/messages/tool/tool_messages/path_tool.rs
+++ b/editor/src/messages/tool/tool_messages/path_tool.rs
@@ -21,7 +21,7 @@ use crate::messages::tool::common_functionality::pivot::{PivotGizmo, PivotGizmoT
 use crate::messages::tool::common_functionality::shape_editor::{
 	ClosestSegment, ManipulatorAngle, OpposingHandleLengths, SelectedLayerState, SelectedPointsInfo, SelectionChange, SelectionShape, SelectionShapeType, ShapeState,
 };
-use crate::messages::tool::common_functionality::snapping::{SnapCache, SnapCandidatePoint, SnapConstraint, SnapData, SnapManager};
+use crate::messages::tool::common_functionality::snapping::{SnapCache, SnapCandidatePoint, SnapConstraint, SnapData, SnapManager, SnappedPoint};
 use crate::messages::tool::common_functionality::utility_functions::{calculate_segment_angle, find_two_param_best_approximate, make_path_editable_is_allowed};
 use graphene_std::Color;
 use graphene_std::renderer::Quad;
@@ -1066,6 +1066,10 @@ impl PathToolData {
 		snap_angle: bool,
 		tangent_to_neighboring_tangents: bool,
 	) -> f64 {
+		if handle_vector.length_squared() < f64::EPSILON {
+			return self.angle;
+		}
+
 		let current_angle = -handle_vector.angle_to(DVec2::X);
 
 		if let Some((vector, layer)) = shape_editor
@@ -1142,7 +1146,7 @@ impl PathToolData {
 		let snap_data = SnapData::new(document, input, viewport);
 		let snap_point = SnapCandidatePoint::handle_neighbors(new_handle_position, [anchor_position]);
 
-		let snap_result = match using_angle_constraints {
+		let mut snap_result = match using_angle_constraints {
 			true => {
 				let snap_constraint = SnapConstraint::Line {
 					origin: anchor_position,
@@ -1153,6 +1157,14 @@ impl PathToolData {
 			}
 			false => self.snap_manager.free_snap(&snap_data, &snap_point, Default::default()),
 		};
+
+		//If the snapping result is a non-finite position
+		if snap_result.distance == f64::INFINITY {
+			snap_result = SnappedPoint {
+				snapped_point_document: new_handle_position,
+				..Default::default()
+			};
+		}
 
 		self.snap_manager.update_indicator(snap_result.clone());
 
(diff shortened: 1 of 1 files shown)


editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
 	hint_data.send_layout(responses);
 	responses.add(ToolMessage::UpdateHints);
 }
+
```
devviktoria: This whole test considered to be created with AI help.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
...
+mod test_path {
+	use crate::messages::input_mapper::utility_types::keyboard::ModifierKeys;
+	use crate::messages::input_mapper::utility_types::pointer::MouseKeys;
+	use crate::messages::portfolio::document::utility_types::misc::{SNAP_FUNCTIONS_FOR_BOUNDING_BOXES, SNAP_FUNCTIONS_FOR_PATHS};
+	pub use crate::test_utils::test_prelude::*;
+	use glam::DAffine2;
+	use graphene_std::vector::Vector;
+	use graphene_std::vector::misc::{BezierHandles, segment_to_handles};
+
+	async fn prepare_document_for_path_snap_weight(anchor_positions: &[DVec2]) -> EditorTestUtils {
```
devviktoria: This prepares the document for testing: creates the document, sets up the viewport, creates an artboard so that the snapping is working. the sets up the snapping to be along path only (this was the root cause of the issue so this is what the snapping is set up.), then creates a Bezier path with the pen tool and check if the document is what is expected.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
...
+				handle_start: DVec2::new(370.0, 420.0),
+				handle_end: DVec2::new(80.0, 70.0),
+			},
+		];
+		assert_handle_positions(&modified_path, &expected_handles, layer_to_viewport, 1e-10);
+
+		editor
+	}
+
+	fn get_path_data(editor: &EditorTestUtils) -> (Vector, DAffine2) {
```
devviktoria: This one is returns the modified path data from the path layer (which should be the second layer in this case) and also returns the document transform data.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
...
+		let path_layer = document.metadata().all_layers().nth(1).expect("Expected path layer");
+
+		let modified_path = document.network_interface.compute_modified_vector(path_layer).expect("Vector not found in the path layer");
+
+		let layer_to_document = document.metadata().transform_to_document(path_layer);
+
+		(modified_path, layer_to_document)
+	}
+
+	fn assert_anchor_positions(vector: &Vector, transform: DAffine2, expected_anchors: &[DVec2], epsilon: f64) {
```
devviktoria: This function checks the position of the anchors.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
...
+
+		for (i, expected) in expected_anchors.iter().enumerate() {
+			let actual = anchors_in_viewport[i];
+			let distance = (actual - *expected).length();
+
+			assert!(distance < epsilon, "Anchor {i} mismatch: expected {expected:?}, got {actual:?}, distance {distance}");
+		}
+	}
+
+	fn assert_handle_positions(vector: &Vector, expected_handles: &[BezierHandles], transform: DAffine2, epsilon: f64) {
```
devviktoria: This function checks the handles. Since handles are stored for segments it goes through the segments and checks type of the handles and their positions.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -3661,3 +3673,202 @@ fn update_dynamic_hints(
...
+				// Mismatch case
+				(actual, expected) => {
+					panic!("Segment {i} handle type mismatch: actual = {:?}, expected = {:?}", actual, expected);
+				}
+			}
+		}
+	}
+
+	#[tokio::test]
+	async fn path_move_handle_to_anchor() {
```
devviktoria: This test moves the path's middle anchor's handle to the anchor point and check if the handles are still cubic handles and their position is what it supposed to be. First it sets up the document by calling prepare_document_for_path_snap_weight then drags the path's middle anchor's handle to the anchor point. And finally it check both the anchor positions and handle type and positions. This is important to check because moving a handle to the anchor point is an important edge case.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1154,6 +1158,14 @@ impl PathToolData {
 			false => self.snap_manager.free_snap(&snap_data, &snap_point, Default::default()),
 		};
 
+		//If the snapping result is a non-finite position
+		if snap_result.distance == f64::INFINITY {
```
devviktoria: AI helped code. This is just a fail safe just in case the snapping system returns with an invalid snap we are returning the new handle.

editor/src/messages/tool/tool_messages/path_tool.rs
```diff
@@ -1066,6 +1066,10 @@ impl PathToolData {
 		snap_angle: bool,
 		tangent_to_neighboring_tangents: bool,
 	) -> f64 {
+		if handle_vector.length_squared() < f64::EPSILON {
```
devviktoria: AI helped code: The drag function calls this and the result is used to calculate the new_handle_position, but in case the handle is at the anchor this returned NaN, which made the new_handle_position NaN. That caused problems. Not it is returning with self.angle wich is better.

devviktoria:
Please note: Since the code are just copied and adopted from my previous code I copied my codereview comment from the precious one instead of rewriting them. I hope that is OK.

PR #4529, opened 2026-09-14
Replace the 'Luminance' node with a 'Desaturate' node with a better selection of desaturation methods

Keavon:


node-graph/nodes/raster/src/adjustments.rs
```diff
@@ -61,18 +89,36 @@ fn luminance<T: Adjust<Color>>(
...
+				srgb_to_linear(0.2126 * r + 0.7152 * g + 0.0722 * b)
+			}
+			DesaturateMethod::LumaRec601 => {
+				let [r, g, b, _] = gamma();
+				srgb_to_linear(0.299 * r + 0.587 * g + 0.114 * b)
+			}
+			DesaturateMethod::LightnessOkLab => {
+				// A gray's OkLab lightness is the cube root of its linear value, so cubing gives the gray of equal lightness
+				let lightness = color.lightness_oklab();
+				lightness * lightness * lightness
```
TrueDoctor, line 113: I'd prefer a `lightness.powi(3)`
```suggestion
				lightness.powi(3)
```

node-graph/nodes/raster/src/adjustments.rs
```diff
@@ -61,18 +89,36 @@ fn luminance<T: Adjust<Color>>(
...
+			}
+			DesaturateMethod::ChannelsAverage => color.average_rgb_channels(),
+			DesaturateMethod::ChannelsMinimum => color.minimum_rgb_channels(),
+			DesaturateMethod::ChannelsMaximum => color.maximum_rgb_channels(),
+			DesaturateMethod::LightnessHsl => {
+				let [r, g, b, _] = gamma();
+				srgb_to_linear((r.max(g).max(b) + r.min(g).min(b)) / 2.)
```
TrueDoctor: since the gamma transform is monotonic, computing the min / max (ideally via their dedicated `maximum_rgb_channels` function) would save one linear -> gamma computation (which matters since this is done per pixel so perf sensitive)

TrueDoctor (approved):

Issue #2309, opened 2025-02-21
Rendering the fills of vector meshes and intersecting paths


Keavon:
High-level goal:

Topological styling: we need to apply fill and stroke styles to the *topology* of the geometry in the scene (composed of subpaths), even when the geometry can move around *spatially* from frame to frame. On any frame, when rendering, we need to apply the topological styling to the current spatial arrangement of the geometry in a stable manner. We need to uniquely, stably address the **sub-segments** and **sub-regions** based on where intersections occur given the current spatial arrangement, to apply a stroke and fill style to each as given by the artist's intent.

---

Vector meshes generalize paths to support arbitrary topology. Loops of segments need a way of being rendered with a fill. This needs to support certain requirements, such as reasonable robustness during editing and animation.

The plan according to the current design is to define a filled region as the area bounded within a list of segments, taking into account each segment's winding order (flow direction or half-edge of a segment) to bound the curved area, akin to CSG. If a flow direction is backwards, it represents a back-face which isn't considered filled.

For example, the dots in the inside of this edge flow would indicate that the left lobe of the infinity shape should be filled as the front-face, while the right lobe's outer-facing dots would be considered a back-face and not filled.

![Image](https://github.com/user-attachments/assets/ebf427cc-543b-4927-be82-6ba3e42535b6)

Uniquely identifying cubic Beziers isn't possible because there can be up to 9 intersections with 8 enclosed regions between a pair of them, compared to quadratic Beziers with 4 intersections and 3 enclosed regions. To address the boundaries of quadratics, cubics need to be considered as quadratics. This can likely be done (without actually subdividing the geometry) by addressing the starting or ending halves of a half-edge, splitting a segment up into four pieces. The midpoint of a half-edge is where the number of intersections is equal to half the total. The starting half reaches from the start to that middle intersection and the ending reaches from the end to the middle intersection, overlapping by a shared segment in cases with even numbers of intersections. A full explainer is at https://discord.com/channels/731730685944922173/731738914812854303/1341703079770193951 (and for @Keavon and @otdavies, [this private Discord chat](https://discord.com/channels/@me/98632974106378240/1341623418055036949)).

This drawing shows some unique addressing cases of each individual region. Red circles indicate midpoint intersections (from one or both of either curve).

![Image](https://github.com/user-attachments/assets/57801a41-1f20-4985-92bf-ccbf96a625bd)

[...]

PR #4542, opened 2026-09-16
fix boolean operations: use topological merge instead of spatial merge

### LINKED ISSUE #4513: Boolean Difference/Exclude operation got broken
If I create a boolean Difference/Exclude operation on 2 shapes the intersecting part is still visible, however it should not be.

https://github.com/user-attachments/assets/9286ad3e-cb2c-4e1d-94ba-341747c44efa

(f I observed correctly the point and segment data is generated correctly for the result, but the display is still wrong.)

Zadfar:
This PR aims to fix the broken boolean operations, mainly difference(exclude) and subtract.

image

The other issues with subtract mentioned by @0HyperCube have also been fixed.
The topological merge by distance function is more suitable to remove the duplicated points from the result of boolean operations.
Fixes #4513


0HyperCube (changes requested):
I would consider this a hacky workaround rather than a proper fix.

You can still get the buggy behaviour by inserting a « Merge by Distance » node after the « Boolean Operation ».

A proper fix would address the faulty rendering algorithm. Specifically `Vector::construct_face` is incredibly baffling. It takes the next `FaceSide` (aka segment) from a list of adjacent edges ordered by the angle of the tangent:

https://github.com/GraphiteEditor/Graphite/blob/eb4a12232219955d6adaae7a63e35686bc6615a4/node-graph/libraries/vector-types/src/vector/vector_attributes.rs#L1144-L1154

This leads it to sometimes never produce a face because it goes into a cycle e.g.

rectangle with circle sharing a vertex at the bottom

It will explore the following segments:
```
Visiting FaceSide { segment_index: 0, reversed: false }
Visiting FaceSide { segment_index: 1, reversed: false }
Visiting FaceSide { segment_index: 2, reversed: false }
Visiting FaceSide { segment_index: 3, reversed: false }
Visiting FaceSide { segment_index: 5, reversed: false }
Visiting FaceSide { segment_index: 6, reversed: false }
Visiting FaceSide { segment_index: 7, reversed: false }
Visiting FaceSide { segment_index: 8, reversed: false }
Visiting FaceSide { segment_index: 3, reversed: true }
Visiting FaceSide { segment_index: 2, reversed: true }
Visiting FaceSide { segment_index: 1, reversed: true }
Visiting FaceSide { segment_index: 0, reversed: true }
Visiting FaceSide { segment_index: 4, reversed: true }
Visiting FaceSide { segment_index: 8, reversed: true }
Visiting FaceSide { segment_index: 7, reversed: true }
Visiting FaceSide { segment_index: 6, reversed: true }
Visiting FaceSide { segment_index: 5, reversed: true }
```
Before giving up because of the iteration limit.

This yields no faces although there clearly should be.

Please mark as draft.

Zadfar:
Thanks for the feedback. I'll spent some time getting to know the vector rendering code better.

> A proper fix would address the faulty rendering algorithm. Specifically `Vector::construct_face` is incredibly baffling. It takes the next `FaceSide` (aka segment) from a list of adjacent edges ordered by the angle of the tangent:

This confused me too. I'll experiment a bit and find where it breaks.

0HyperCube:
Hi, thanks for the response.

I've done some sketching and I'm even more convinced that the `Vector::construct_face` is nonsensical.

It doesn't really make sense to look at the tangents when they can collide or overlap. For example this rectangle with a bézier through (with one handle snapped to the edge) occasionally has a broken fill [rectangle_with_bézier.graphite](https://github.com/user-attachments/files/32310346/rectangle_with_bezier.graphite.json):
rectangle with bézier through


Some things I've observed:
- Probably ignore bézier handles
- Probably want faces with as few anchors/segments as possible
- One face should contain a segment at most once (currently can contain the segment in both directions but this is bad).
- A segment is part of at most 2 faces (however it is possible to get stuck if you do some bad faces and follow this rule).
- A relatively simple graph has very many combinations for faces

I'm happy to discuss or collaborate on this here or on discord.

Zadfar:
> It doesn't really make sense to look at the tangents when they can collide or overlap. For example this rectangle with a bézier through (with one handle snapped to the edge) occasionally has a broken fill

I agree, the ordering should be based on something else. 
While reading up on how inkscape does its vector rendering i came across [this](https://inkscape.gitlab.io/inkscape/doxygen/classShape.html). Boolean operations and such are performed on the planar eulerian graph generated using [this](https://inkscape.gitlab.io/inkscape/doxygen/classShape.html#ac66a70e4e25b859d9b762c6a6b7712f3), instead of directly on the original graph. Would a similar approach make sense here?

I havent gotten around to how exactly it should be ordered. The bezier handles should probably be ignored as you said. 
I am fairly new to this. so forgive me if this deviates from the intended design of graphite.

0HyperCube:
> While reading up on how inkscape does its vector rendering i came across [this](https://inkscape.gitlab.io/inkscape/doxygen/classShape.html)

In graphite, the boolean operations are carried out with the [linesweeper crate](https://docs.rs/crate/linesweeper) which does a similar thing by converting intersections to vertices and then stitching the path back together.

> the planar eulerian graph

Currently it seems that the `Vector` type models arbitrary graphs (since users can create or delete segments/points using the « Path » node via the path tool). The graphs passed to the renderer therefore do not necessarily have vertices at intersections or any other nice properties. I'm not sure if it is desirable to keep such functionality since it seems to make processing rather difficult. It is desirable to keep rendering simple and fast.

>  forgive me if this deviates from the intended design of graphite.

I am not a maintainer and have no clue about the intended design. I think there was some idea that the user could choose to fill or not fill individual regions like in figma. However I'm not really sure if that is still being considered. You'd probably have to consult with Keavon on this matter.
filling in a region in figma in a complex self intersecting shape

Zadfar:
There seems to be a plan to completely redesign how vector meshes are handled in #2309. Until that happens, This temporary workaround is an option (similar to #3884).
While going through how linesweeper was added to handle boolean operations in graphite, i saw [this](https://github.com/GraphiteEditor/Graphite/pull/2670#issuecomment-3797719762). which seems to indicate that merge by distance was removed (but it wasnt). I tested this without the merge by distance and it seems to work as well.

@Keavon What do you think? I can close the pr if this workaround isnt really needed for the moment.

PR #4543, opened 2026-09-16
Add a dropdown for resource selection and adopt it in the 'Image' node

Keavon:
image


editor/src/messages/frontend/frontend_message.rs
```diff
@@ -93,6 +93,15 @@ pub enum FrontendMessage {
 	TriggerImport {
 		filters: Vec<FileFilter>,
 	},
+	/// Asks the frontend for a file whose bytes come back through the `upload_resource` command, aimed at this node input.
```
timon-schelling:
```suggestion
```

timon-schelling:
1. desktop is completely missing
2. uploading an image should be one unified path. not two separate ones. the pasting an image branch should also use this.
3. should be separated out into a resource dialog/management handler, especially if you are planning to add other file types soon. (currently feels like duct tape together into the document handler).
4. could you make this a stack with other things that depend on it (other File types mostly)?
5. resources get garbage collected when reloading the page (because the are removed from the history)

